PR #5, opened 2015-08-17
refactor the locking

MaximilianMeister:
moved from https://github.com/crowbar/barclamp-crowbar/pull/1346


tboerger:
Some offenses from hound and failing gating but code looks good.

tboerger:
@aspiers please stop this commit commenting, I don't get any notification about that.

MaximilianMeister:
@aspiers i adressed the problem with the exclusivity of the remote lock. please have a look again

ovisan:
:+1:  LGTM

tboerger:
I like that!

crowbar_framework/app/models/service_object.rb
```diff
@@ -822,8 +805,17 @@ def apply_role(role, inst, in_queue)
     # We also check that all nodes we'll require are in the ready state.
     #
 
-    # Initialize variables used in ensure at the end of the method
-    chef_daemon_nodes = []
+    nodes_without_admin = []
+    NodeObject.find_all_nodes.each do |node|
```
dirkmueller: can we use .select do |node| here instead ? would be slightly simpler I guess

crowbar_framework/app/models/service_object.rb
```diff
@@ -1285,8 +1273,7 @@ def apply_role(role, inst, in_queue)
     process_queue unless in_queue
     [200, {}]
   ensure
-    # start chef daemon on all nodes
-    chef_daemon(:start, chef_daemon_nodes)
+    apply_lock.release
```
dirkmueller: what if apply_lock is (not yet) defined here?

tboerger: How should that happen? It's defined above within this method.

crowbar_framework/app/models/service_object.rb
tboerger: Indeed, select makes sense for this simple loop

MaximilianMeister: how much simpler would it be?

```ruby
NodeObject.find_all_nodes.select { node[:platform] != "windows" || !node.admin? }.map(&:name)
```
the `!admin?` part does not work somehow. so if you have a simpler version please tell me how and i will change it.

tboerger: What about this?

```ruby
NodeObject.find_all_nodes.reject do |node|
  node[:platform] == "windows" || node.admin?
end.map(&:name)
```

MaximilianMeister: thanks

crowbar_framework/lib/crowbar/lock.rb
```diff
@@ -0,0 +1,110 @@
...
+    end
+
+    protected
+
+    def remote_lock(options = {})
+      action = options.fetch :action
+      remote.each do |node|
+        if action == "acquire"
+          logger.info("Acquire remote temporary lock on #{path} on #{node}")
+          ssh_cmd(node, "echo '#{Process.pid}' >")
```
aspiers: This would steal the lock from any existing owner.

crowbar_framework/app/models/service_object.rb
aspiers: Yeah I think @tboerger is right, it looks safe to me.

crowbar_framework/app/models/network_service.rb
```diff
@@ -33,14 +33,6 @@ def role_constraints
     end
   end
 
-  def acquire_ip_lock
```
aspiers: I think you should have kept these to avoid duplicated code below.

crowbar_framework/app/models/network_service.rb
```diff
@@ -179,9 +171,8 @@ def deallocate_ip_by_type(bc_instance, network, object, type)
     end
 
     save = false
-    begin # Rescue block
-      f = acquire_ip_lock
-
+    file_lock = Crowbar::Lock.new(logger: @logger, path: Rails.root.join("tmp", "ip.lock")).acquire
```
aspiers: This code exactly duplicates line 58 above.  They should be in a helper method.

crowbar_framework/app/models/role_object.rb
```diff
@@ -286,8 +286,7 @@ def increment_crowbar_revision!
 
   def save
     Rails.logger.debug("Saving role: #{@role.name} - #{crowbar_revision}")
-    role_lock = FileLock.acquire "role:#{@role.name}"
-    begin
+    Crowbar::Lock.new(path: Rails.root.join("tmp", "role:#{@role.name}.lock")).with_lock do
```
aspiers: More duplicated code.

crowbar_framework/app/models/role_object.rb
```diff
@@ -307,19 +306,14 @@ def save
...
-      FileLock.release role_lock
     end
     Rails.logger.debug("Done saving role: #{@role.name} - #{crowbar_revision}")
   end
 
   def destroy
     Rails.logger.debug("Destroying role: #{@role.name} - #{crowbar_revision}")
-    begin
-      role_lock = FileLock.acquire "role:#{@role.name}"
+    Crowbar::Lock.new(path: Rails.root.join("tmp", "role:#{@role.name}.lock")).with_lock do
```
aspiers: More duplicated code.

crowbar_framework/app/models/service_object.rb
```diff
@@ -140,23 +140,11 @@ def random_password(size = 12)
...
-  end
-
-#
 # Helper routines for queuing
 #
 
   def set_to_applying(nodes, inst)
-    f = acquire_lock "BA-LOCK"
-    begin
+    Crowbar::Lock.new(logger: @logger, path: Rails.root.join("tmp", "BA-LOCK.lock")).with_lock do
```
aspiers: More duplicated code.

crowbar_framework/app/models/service_object.rb
```diff
@@ -165,14 +153,11 @@ def set_to_applying(nodes, inst)
...
       end
-    ensure
-      release_lock f
     end
   end
 
   def restore_to_ready(nodes)
-    f = acquire_lock "BA-LOCK"
-    begin
+    Crowbar::Lock.new(logger: @logger, path: Rails.root.join("tmp", "BA-LOCK.lock")).with_lock do
```
aspiers: More duplicated code.

crowbar_framework/app/models/service_object.rb
```diff
@@ -140,23 +140,11 @@ def random_password(size = 12)
   end
 
 #
-# Locking Routines
-#
-  def acquire_lock(name)
```
aspiers: I think you should have kept this to avoid duplicated code.

MaximilianMeister:
@aspiers sorry for not addressing all the comments that you posted, but i am sitting between 2 chairs here. for this PR, as it is part of the core rails app, i chose to do it the @tboerger way, because first of all I simply trust in his experience and Ruby and Rails skills. on top of that we are sitting together in a room and I get immediate and concrete feedback about what i am writing. I personally find this solution here good for now, the reasoning behind your comments I understand, not the importance though to discuss so much about it. so please if you have another solution, it would just be better if you submit it directly. or if you think this is a ship stopper, then let the PO decide what to do here.

aspiers:
> I personally find this solution here good for now, the reasoning behind your comments I understand

Sorry, somehow we have a disconnect in communications here.  I was trying to explain that I believe the locking code in this PR is open to race conditions.  So if you still think this solution is good and don't understand why I'm still discussing it, then I guess I must be doing a terrible job of explaining.

So let's discuss tomorrow on mumble - hopefully we can eliminate the confusion a lot quicker there.  I'm also reworking this PR and will resubmit a new version.  It won't be a rewrite because there is lots of good stuff already in this PR which I want to keep.  But it will fix the race conditions and other issues I already mentioned.

PR #6, opened 2015-08-18
Changes for SLES_12 based sleshammer

mjura:
This pull request adopt barclamp-provisioner for SLES_12 based images

### DIFF
--- a/chef/cookbooks/provisioner/recipes/bootdisk.rb
+++ b/chef/cookbooks/provisioner/recipes/bootdisk.rb
@@ -31,34 +31,51 @@ ruby_block "Find the fallback boot device" do
     basedir = "/dev/disk/by-path"
     dev = nil
     disk_by_path = nil
-    ["-usb-", nil].each do |deviceignore|
-      ::Dir.entries(basedir).sort.each do |path|
-        # Not a symlink?  Not interested.
-        next unless File.symlink?(File.join(basedir, path))
-        # Symlink does not point at a disk?  Also not interested.
-        dev = File.readlink("#{basedir}/#{path}").split("/")[-1]
-        # Prefer devices in a specific order
-        next if (deviceignore and path.include?(deviceignore))
-        disk_by_path = "disk/by-path/#{path}"
-        break if dev =~ /^[hsv]d[a-z]+$/
-        # pci-0000:0b:08.0-cciss-disk0 -> ../../cciss/c0d0
-        break if dev =~ /^c[0-9]+d[0-9]+$/
-        # xen-vbd-51712-part1 -> ../../xvda1
-        break if dev =~ /^xvd[a-z]+$/
-        dev = nil
-        disk_by_path = nil
+    if File.directory?(basedir)
+      ["-usb-", nil].each do |deviceignore|
+        ::Dir.entries(basedir).sort.each do |path|
+          # Not a symlink?  Not interested.
+          next unless File.symlink?(File.join(basedir, path))
+          # Symlink does not point at a disk?  Also not interested.
+          dev = File.readlink("#{basedir}/#{path}").split("/")[-1]
+          # Prefer devices in a specific order
+          next if deviceignore && path.include?(deviceignore)
+          disk_by_path = "disk/by-path/#{path}"
+          break if dev =~ /^[hsv]d[a-z]+$/
+          # pci-0000:0b:08.0-cciss-disk0 -> ../../cciss/c0d0
+          break if dev =~ /^c[0-9]+d[0-9]+$/
+          # xen-vbd-51712-part1 -> ../../xvda1
+          break if dev =~ /^xvd[a-z]+$/
+          dev = nil
+          disk_by_path = nil
+        end
+        break if dev
       end
-      break if dev
+    end
+    node[:crowbar_wall][:boot_device] = disk_by_path
+
+    # Searching boot disk with virtio driver if no other disks were found
+    # on SLES_12 sleshammer image
+    if dev.nil?
+      basedir = "/sys/block"
+      ::Dir.entries(basedir).sort.each do |path|
+        next unless File.symlink?(File.join(basedir, path))
+        dev = File.readlink("#{basedir}/#{path}")
+        if dev =~ /virtio/
+          dev = dev.split("/")[-1]
+          node[:crowbar_wall][:boot_device] = dev
+          break
+        end
+      end
     end
     raise "Cannot find a hard disk!" unless dev
-    node[:crowbar_wall][:boot_device] = disk_by_path
+
     # Turn the found device into its corresponding /dev/disk/by-id link.
     # This name should be more stable than the /dev/disk/by-path one.
-
     basedir = "/dev/disk/by-id"
     # /dev/disk/by-id is unstable under VirtualBox so don't rely on it.
     hardware = node[:dmi][:system][:product_name] rescue "unknown"
-    if hardware !~ /VirtualBox/i && File.exists?(basedir)
+    if hardware !~ /VirtualBox/i && File.directory?(basedir)
       bootdisks = ::Dir.entries(basedir).sort.select do |m|
         f = File.join(basedir, m)
         File.symlink?(f) && (File.readlink(f).split("/")[-1] == dev)
@@ -83,7 +100,9 @@ ruby_block "Find the fallback boot device" do
     # name for bootdisk instead of the by-path name. According to:
     #  https://www.mail-archive.com/systemd-devel@lists.freedesktop.org/msg20607.html
     # the by-path names for virtio device are not stable and don't even exist
-    # anymore with newer udev releases, e.g. on SLES12
+    # anymore with newer udev releases
+    # This is only for SLE_11_SP3 sleshammer. On SLE_12 based image
+    # directory "disk/by-path" with virtio disks driver doesn't exist
     if File.dirname(node[:crowbar_wall][:boot_device]) == "disk/by-path" and
         File.basename(node[:crowbar_wall][:boot_device]) =~ /-virtio-/
       Chef::Log.debug("#{node[:crowbar_wall][:boot_device]} appears to be a virtio device, falling back to simple device name #{dev}")
--- a/updates/control.sh
+++ b/updates/control.sh
@@ -27,8 +27,36 @@ fi
 #set -x
 export PS4='${BASH_SOURCE}@${LINENO}(${FUNCNAME[0]}): '
 
-function is_suse {
-    [ -f /etc/SuSE-release ]
+# File /etc/SuSE-release is deprecated and will be removed
+# in a future service pack or release. Therefore we would
+# like to use /etc/os-release about details release,
+# but for SLES 11 SP3 /etc/os-release doesn't exist.
+function is_suse() {
+    # This check will work on SLE 11 SP3 and SLE 12
+    [ -f /etc/SuSE-release ] && return
+
+    # This will work only on SLE 12 and above, after
+    # /etc/SuSE-release will be deprecated
+    if [ -f /etc/os-release ]; then
+        . /etc/os-release
+        [ "$NAME" == "SLES" ] && return
+    fi
+
+    return 1
+}
+
+function suse_ver() {
+    local ver=$1
+    local suse_ver=0
+
+    if [ -f /etc/os-release ]; then
+        . /etc/os-release
+        [ "$NAME" == "SLES" ] && suse_ver=$VERSION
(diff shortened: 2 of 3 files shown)


mjura:
It was tested with standard SLE 11 SP3 sleshammer and new version SLE 12 sleshammer

tboerger:
:+1:

chef/cookbooks/provisioner/recipes/bootdisk.rb
```diff
@@ -78,17 +79,42 @@
...
+          File.open("/sys/block/#{dev}/device/uevent").each_line do |line|
+            device_driver = line.strip.split("=")[1] if line =~ /DRIVER/
+          end
+          if device_driver =~ /^virtio/
+            node[:crowbar_wall][:boot_device] = dev
+            break
+          end
+        end
+      end
+      raise "Cannot find a hard disk!"
```
MaximilianMeister: not sure if i am missing something in the logic, but doesn't this raise always fire? because it is after the `each` block

chef/cookbooks/provisioner/recipes/bootdisk.rb
```diff
@@ -31,34 +31,35 @@
...
-        next if (deviceignore and path.include?(deviceignore))
-        disk_by_path = "disk/by-path/#{path}"
-        break if dev =~ /^[hsv]d[a-z]+$/
-        # pci-0000:0b:08.0-cciss-disk0 -> ../../cciss/c0d0
-        break if dev =~ /^c[0-9]+d[0-9]+$/
-        # xen-vbd-51712-part1 -> ../../xvda1
-        break if dev =~ /^xvd[a-z]+$/
-        dev = nil
-        disk_by_path = nil
+    if File.directory?(basedir)
```
MaximilianMeister: (MINOR) I would put this test on the bottom of the loop, as it is small and we dont have an `else`. so you would not have to indent deeper.

rsalevsky:
After fixing the small nitpicks from @MaximilianMeister I'm fine with it.

chef/cookbooks/provisioner/recipes/bootdisk.rb
mjura: Ok, I've added `unlsess dev` to it

chef/cookbooks/provisioner/recipes/bootdisk.rb
mjura: This has to stay like this, because on SLE12 with virtio disks `basedir = "/dev/disk/by-path"` doesn't exist, and then 
```
basedir = nil
::Dir.entries(basedir).sort.each
TypeError: no implicit conversion of nil into String
```

updates/control.sh
```diff
@@ -110,13 +138,14 @@ until $ntp || [[ $DHCP_STATE = 'debug' ]]; do
...
-#
-killall dhclient
-killall dhclient3
-
+if ! suse_ver 12; then
+  #
+  # rely on the DHCP server to do the right thing
+  # Stick with this address until we get finished.
+  #
+  killall dhclient
```
dirkmueller: should be fine to killall a  process that doesn't exist, right? why is the check for suse_ver 12 needed?

chef/cookbooks/provisioner/recipes/bootdisk.rb
```diff
@@ -78,17 +79,42 @@
...
+    device_driver = nil
+    if dev.nil?
+      ::Dir.entries(basedir).sort.each do |path|
+        next unless File.symlink?(File.join(basedir, path))
+        dev = File.readlink("#{basedir}/#{path}").split("/")[-1]
+        if dev =~ /^vd[a-z]+$/ && File.exists?("/sys/block/#{dev}/device/uevent")
+          File.open("/sys/block/#{dev}/device/uevent").each_line do |line|
+            device_driver = line.strip.split("=")[1] if line =~ /DRIVER/
+          end
+          if device_driver =~ /^virtio/
```
dirkmueller: I'm not sure why we need this code.. if the device name starts with /^vd[a-z]*/ it should be virtio, there is no need to double check that it that much..

updates/control.sh
mjura: Yeap, you are right. I just wanted to keep it clear or just mark section which are not needed for SLES 12

chef/cookbooks/provisioner/recipes/bootdisk.rb
mjura: Problem is that issue is only with VirtIO disks on SLE12 nodes. I'm not sure if /^vd[a-z] disks can also from other disk drivers for example for Virtualbox

mjura: Btw. this name /^vd[a-z] can be changed by using target name option in libvrit template
```
<disk type='block' device='disk'>
<driver name='qemu' type='raw'/>
<source dev='/dev/sda1/>
<target dev='vda' bus='virtio'/>
<address type='pci' domain='0x0000' bus='0x00' slot='0x05' function='0x0'/>
</disk>
```

mjura: Only one way yo check this is read /sys/block/dev/device/uevent file with kernel driver

MaximilianMeister:
:+1:

mjura:
Yupi!!! Gating tests are passing again!!!

rsalevsky:
LGTM

chef/cookbooks/provisioner/recipes/bootdisk.rb
```diff
@@ -78,17 +79,42 @@
...
     if File.dirname(node[:crowbar_wall][:boot_device]) == "disk/by-path" and
         File.basename(node[:crowbar_wall][:boot_device]) =~ /-virtio-/
       Chef::Log.debug("#{node[:crowbar_wall][:boot_device]} appears to be a virtio device, falling back to simple device name #{dev}")
       node[:crowbar_wall][:boot_device] = dev
     end
+
+    # Searching boot disk with virtio driver if no other disks were found
+    # on SLES_12 sleshammer image
+    basedir = "/sys/block"
+    device_driver = nil
```
vuntz: Why is this not inside the `if` block?

chef/cookbooks/provisioner/recipes/bootdisk.rb
```diff
@@ -78,17 +79,42 @@
...
+    if dev.nil?
```
vuntz: Also, I'm wondering if this `if` block shouldn't come earlier (and everything that was above would be in the `else` block).

chef/cookbooks/provisioner/recipes/bootdisk.rb
dirkmueller: well, sure, but it doesn't really matter at all. the only point of the code at this point is: 

"we haven't been able to find the root disk neither via /dev/disk/by-id nor via by-path, so we could just use the "simple name" and be done with it.

chef/cookbooks/provisioner/recipes/bootdisk.rb
```diff
@@ -31,34 +31,35 @@
...
     # This name should be more stable than the /dev/disk/by-path one.
-
     basedir = "/dev/disk/by-id"
     # /dev/disk/by-id is unstable under VirtualBox so don't rely on it.
     hardware = node[:dmi][:system][:product_name] rescue "unknown"
-    if hardware !~ /VirtualBox/i && File.exists?(basedir)
+    if hardware !~ /VirtualBox/i && File.directory?(basedir)
       bootdisks = ::Dir.entries(basedir).sort.select do |m|
         f = File.join(basedir, m)
         File.symlink?(f) && (File.readlink(f).split("/")[-1] == dev)
```
vuntz: So really, my issue is that we're comparing stuff with `dev` here and now it can be `nil`. Which means this entire block is useless in the `dev.nil?` case and should be in the `else` part I mentioned earlier.

chef/cookbooks/provisioner/recipes/bootdisk.rb
```diff
@@ -78,17 +79,36 @@
...
     # name for bootdisk instead of the by-path name. According to:
     #  https://www.mail-archive.com/systemd-devel@lists.freedesktop.org/msg20607.html
     # the by-path names for virtio device are not stable and don't even exist
-    # anymore with newer udev releases, e.g. on SLES12
-    if File.dirname(node[:crowbar_wall][:boot_device]) == "disk/by-path" and
-        File.basename(node[:crowbar_wall][:boot_device]) =~ /-virtio-/
+    # anymore with newer udev releases
+    # This is only for SLE_11_SP3 sleshammer. On SLE_12 based image
+    # directory "disk/by-path" with virtio disks driver doesn't exist
+    elsif ! node[:crowbar_wall][:boot_device].nil? and
```
vuntz: How can it be `nil` if `dev` is not `nil`?

chef/cookbooks/provisioner/recipes/bootdisk.rb
```diff
@@ -78,17 +79,36 @@
...
     end
+
+    # Searching boot disk with virtio driver if no other disks were found
+    # on SLES_12 sleshammer image
+    if dev.nil?
+      basedir = "/sys/block"
+      ::Dir.entries(basedir).sort.each do |path|
+        next unless File.symlink?(File.join(basedir, path))
+        dev = File.readlink("#{basedir}/#{path}").split("/")[-1]
+        if dev =~ /^vd[a-z]+$/
```
dirkmueller: thats kind of pointless. the symlink will always point to the same name. you're iterating sorted over /sys/block/..

how about you're doing the sorted iteration, but simply check if the link contains "/virtio" ? 

in a way it feels like you actually want to sort by the link, not by the name of the link..

chef/cookbooks/provisioner/recipes/bootdisk.rb
mjura: This `dev` variable could be also `nil` before, we have only once specific scenario on SLES)12 with virtio disks only

vuntz: No, it could not before because of the `raise "Cannot find a hard disk!" unless dev` line (that you dropped).

chef/cookbooks/provisioner/recipes/bootdisk.rb
mjura: Ok, new proposal would be 
```
      basedir = "/sys/block"
      ::Dir.entries(basedir).sort.each do |path|
        next unless File.symlink?(File.join(basedir, path))
        dev = File.readlink("#{basedir}/#{path}")
        if dev =~ /virtio/
          dev = dev.split("/")[-1]
          node[:crowbar_wall][:boot_device] = dev
          break
        end
      end
```
This part was moved earlier before `raise` condition

chef/cookbooks/provisioner/recipes/bootdisk.rb
mjura: Ok, so as we discussed I moved condition before `raise "Cannot find a hard disk!" unless dev`. Right now I'm testing changes

chef/cookbooks/provisioner/recipes/bootdisk.rb
```diff
@@ -31,34 +31,52 @@
...
+          dev = dev.split("/")[-1]
+          node[:crowbar_wall][:boot_device] = dev
+          break
+        end
       end
-      break if dev
     end
+
     raise "Cannot find a hard disk!" unless dev
     node[:crowbar_wall][:boot_device] = disk_by_path
```
vuntz: I think this line actually breaks things? It should stay before your new block.

chef/cookbooks/provisioner/recipes/bootdisk.rb
```diff
@@ -78,13 +96,17 @@
...
     # devices looks like a virtio device it is safer to use the base device
     # name for bootdisk instead of the by-path name. According to:
     #  https://www.mail-archive.com/systemd-devel@lists.freedesktop.org/msg20607.html
     # the by-path names for virtio device are not stable and don't even exist
-    # anymore with newer udev releases, e.g. on SLES12
-    if File.dirname(node[:crowbar_wall][:boot_device]) == "disk/by-path" and
+    # anymore with newer udev releases
+    # This is only for SLE_11_SP3 sleshammer. On SLE_12 based image
+    # directory "disk/by-path" with virtio disks driver doesn't exist
+    if !node[:crowbar_wall][:boot_device].nil? &&
```
vuntz: I don't think you need to update that test now, since the original `raise` is still there.

dirkmueller:
+1

vuntz:
+1

PR #9, opened 2015-08-21
Support MTU size setting for nodes other than admin and storage.

mmnelemane:
Alllows setting mtu size for nodes using "want_mtu_size" variable. Set want_mtu_size="jumbo" to enable jumbo frame testing.


MaximilianMeister:
schema update, revision bump and migrations are missing...
have a look here how it needs to be done https://github.com/crowbar/barclamp-nova/pull/476

rsalevsky:
@MaximilianMeister It's not needed here, was already done in a previous PR.

MaximilianMeister:
@rsalevsky @mmnelemane which one? that should be mentioned at least in the commit message. so this is only a fixup for an old PR?

tboerger:
LGTM

rsalevsky:
+1 but let's wait for the gating test

aspiers:
Agree with @MaximilianMeister - please can you mention the previous PR URL in the commit message? Thanks!

PR #11, opened 2015-08-25
Revert "Revert "Stop writing dev file for rainbows, bin/rails server …

MaximilianMeister:
…is the thruth""

This reverts commit c2112fe9a24a2c914f78a95e6d6031bff8fe5ec5.

This revert in c2112fe9a24a2c914f78a95e6d6031bff8fe5ec5 wouldn't have been necessary in the first revert, that's why we revert the revert here.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/default.rb
+++ b/chef/cookbooks/crowbar/recipes/default.rb
@@ -287,21 +287,6 @@ template "/opt/dell/crowbar_framework/rainbows.cfg" do
             app_location: "/opt/dell/crowbar_framework")
 end
 
-template "/opt/dell/crowbar_framework/rainbows-dev.cfg" do
-  source "rainbows.cfg.erb"
-  owner "crowbar"
-  group "crowbar"
-  mode "0644"
-  variables(web_host: "0.0.0.0",
-            web_port: node["crowbar"]["web_port"] || 3000,
-            user: "crowbar",
-            concurrency_model: "EventMachine",
-            group: "crowbar",
-            logdir: logdir,
-            logname: "development",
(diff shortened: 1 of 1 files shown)


rsalevsky:
+1

tboerger:
:+1:

aspiers:
It seems this is reverting part of https://github.com/crowbar/barclamp-crowbar/pull/1338

tboerger:
> It seems this is reverting part of crowbar/barclamp-crowbar#1338

Really? And as @MaximilianMeister said, this was a useless revert.

aspiers:
@tboerger Really what?  I'm not disputing the change, merely adding a link for future reference.  This PR only referenced the individual commit being reverted (https://github.com/crowbar/crowbar-core/commit/c2112fe9a24a2c914f78a95e6d6031bff8fe5ec5), but https://github.com/crowbar/barclamp-crowbar/pull/1338 contains more useful context.

dirkmueller:
LGTM

tboerger:
> @tboerger Really what? I'm not disputing the change, merely adding a link for future reference. This PR only referenced the individual commit being reverted (c2112fe), but crowbar/barclamp-crowbar#1338 contains more useful context.

Because the revert pull request doesn't really matter as this was simply a wrong revert. This commit has nothing to do with breaking anything as it just drops the creation of the rainbows-dev config which is anyway not used by somebody these days.

aspiers:
@tboerger Sounds like you still think I was questioning the change.  I am not!  I just found the context hard to understand without a link to https://github.com/crowbar/barclamp-crowbar/pull/1338, so I added it in case anyone else had the same problem.

PR #12, opened 2015-08-27
Do not add conf file entries for virtual interfaces (bsc#816746)

mmnelemane:
Solves bug: #816746.

Does not add .conf file in /etc/dhcp3/hosts.d/ for virtual interfaces. So no accumulation of conf files in crowbar node for each VM. Constant interfaces (like bridges, physical interfaces) will continue to hold .conf file in this path.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -22,6 +22,7 @@ admin_ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").
 web_port = node[:provisioner][:web_port]
 provisioner_web = "http://#{admin_ip}:#{web_port}"
 dhcp_hosts_dir = node["provisioner"]["dhcp_hosts"]
+virtual_intfs = ["tap", "qbr", "qvo", "qvb", "brq", "ovs"]
 
 nodes = search(:node, "*:*")
 if not nodes.nil? and not nodes.empty?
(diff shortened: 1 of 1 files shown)


rsalevsky:
Please make the commit title shorter and explain the details in the commit message. Also add the full bug URL and add bsc#XXXXXX to the commit title.

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -40,11 +40,13 @@
     mac_list = []
     unless mnode["network"].nil? || mnode["network"]["interfaces"].nil?
       mnode["network"]["interfaces"].each do |net, net_data|
-        net_data.each do |field, field_data|
-          next if field != "addresses"
-          field_data.each do |addr, addr_data|
-            next if addr_data["family"] != "lladdr"
-            mac_list << addr unless mac_list.include? addr
+        if !net.start_with?("tap", "qvo", "qbr", "qvb", "ovs")
```
dirkmueller: what about

brq5ade6b9d-7f


bridges? does it ignore those already?

mmnelemane: Files for bridges like br-int, br-tunnel, br-fixed are retained as there is only one instance of these for each node. They dont get accumulated for each VM. 

brqXXXX gets added for each VM for networkingplugin=linuxbridge. 
Currently ongoing test on local machine. Will update the patch adding to the list.

dirkmueller:
LGTM

mjura:
+1

toabctl:
+1

PR #14, opened 2015-08-27
enhance locking and pause chef-client daemon when applying proposals

aspiers:
Reworking of #5 to address concurrency issues, make the interface between the two locking subclasses consistent, and DRY up the calling code.


aspiers:
The unit tests pass but I still need to test this in a real deployment.  Let's see what the gate thinks.

MaximilianMeister:
Looks really good at a first glance, I can see a now bit better what you meant during the previous discussion. especially the remote locking. thanks for the effort

aspiers:
Woohoo, this works!  The gating run is already well through tempest, so all the proposals applied fine.  I also ssh'd to the `c7` cloud and verified that the `flock` mechanism is working as expected, e.g.

```
D, [2015-09-03T19:07:42.166824 #26444] DEBUG -- : apply_role-dns-config-default-default-26444 acquiring remote lock /var/chef/cache/pause-file.lock on d52-54-01-77-77-01.c7.cloud.suse.de
D, [2015-09-03T19:07:42.166962 #26444] DEBUG -- : Will execute on node d52-54-01-77-77-01.c7.cloud.suse.de: flock /var/chef/cache/pause-file.lock.meta bash -esand feed the following to STDIN:
mkdir -p "/var/chef/cache/pause-file.lock.d"
        file="/var/chef/cache/pause-file.lock.d/apply_role-dns-config-default-default-26444"
        if [ -e "$file" ]; then
          echo "ERROR: apply_role-dns-config-default-default-26444 already had lock via $file" >&2
          exit 1
        fi
        echo "apply_role(dns-config-default, default, false) pid 26444" > "$file"
        echo "Last required by apply_role-dns-config-default-default-26444" > "/var/chef/cache/pause-file.lock"
```

and later:

```
D, [2015-09-03T19:08:31.907517 #26444] DEBUG -- : apply_role-dns-config-default-default-26444 releasing remote lock /var/chef/cache/pause-file.lock on d52-54-0
1-77-77-01.c7.cloud.suse.de
D, [2015-09-03T19:08:31.907568 #26444] DEBUG -- : Will execute on node d52-54-01-77-77-01.c7.cloud.suse.de: flock /var/chef/cache/pause-file.lock.meta bash -es
and feed the following to STDIN:
rm -f /var/chef/cache/pause-file.lock.d/apply_role-dns-config-default-default-26444
        if [ -d /var/chef/cache/pause-file.lock.d ] && ! ls /var/chef/cache/pause-file.lock.d | grep -q .; then
          rm -f /var/chef/cache/pause-file.lock
        fi
D, [2015-09-03T19:08:32.353531 #26444] DEBUG -- : apply_role-dns-config-default-default-26444 released remote lock /var/chef/cache/pause-file.lock on d52-54-01-77-77-01.c7.cloud.suse.de
```

So we only need this reviewed and merged now, and then we can drop the ugly workaround patches.

crowbar_framework/lib/crowbar/lock/local_blocking.rb
```diff
@@ -0,0 +1,57 @@
...
+    def release
+      logger.debug("Release #{name} lock enter: #{file.inspect}")
+      if @file
+        @file.flock(File::LOCK_UN) if locked?
+        @file.close unless @file.closed?
+      else
+        logger.warn("release called without valid file")
+      end
+      logger.debug("Release #{name} lock exit")
+      @locked = false
```
vuntz: I'm wondering: does acquiring work after the lock as been released? Sounds to me like it won't work since the file was closed.

rsalevsky:
@aspiers Please fix all the hound issues, after this I will do a complete review.

crowbar_framework/app/models/service_object.rb
```diff
@@ -142,12 +142,18 @@ def random_password(size = 12)
...
+    new_lock(name).acquire
   end
 
-  def release_lock(f)
-    FileLock.release(f, logger: @logger)
+  def with_lock(name)
+    new_lock(name).with_lock do
+      yield
+    end
   end
```
vuntz: We have these methods here and in deployment_queue.rb. Should they somehow be shared?

crowbar_framework/app/models/service_object.rb
```diff
@@ -822,8 +822,19 @@ def apply_role(role, inst, in_queue)
     # We also check that all nodes we'll require are in the ready state.
     #
 
-    # Initialize variables used in ensure at the end of the method
-    chef_daemon_nodes = []
+    nodes_without_admin = NodeObject.find_all_nodes.reject do |node|
```
vuntz: Since we're also excluding windows nodes, would it make sense to rename this variable to something like `nodes_to_lock` or something closer to what it really is?

crowbar_framework/app/models/service_object.rb
```diff
@@ -822,8 +822,19 @@ def apply_role(role, inst, in_queue)
...
+      node[:platform] == "windows" || node.admin?
+    end.map(&:name)
+
+    apply_locks = nodes_without_admin.map do |node|
```
vuntz: Hrm. Are we really locking absolutely all nodes, even if they're not related to the proposal being applied?

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
```diff
@@ -0,0 +1,139 @@
...
+      logger.debug("#{owner} acquired remote lock #{path} on #{node}")
+      @locked = true
+      self
+    end
+
+    def release
+      logger.debug("#{owner} releasing remote lock #{path} on #{node}")
+      shell = <<-EOSHELL.strip
+        rm -f #{@readers_dir}/#{owner}
+        if [ -d #{@readers_dir} ] && ! ls #{@readers_dir} | grep -q .; then
```
vuntz: Just to clarify (which might mean a comment would be nice?): the `ls | grep` bit is to check that the directory is not empty?

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
```diff
@@ -0,0 +1,139 @@
...
+          rm -f #{path}
```
vuntz: Wouldn't it make sense to also remove the directory?

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
```diff
@@ -0,0 +1,139 @@
...
+        fi
```
vuntz: There's also one thing I'm not sure I like here: if the lock is shared, and we were the last owner, then by releasing it, the content of the lock file will reference us while we will have released the lock. Would it make sense to append to the lock file instead of recreating the lock file to say who last requested it?

crowbar_framework/app/models/service_object.rb
```diff
@@ -822,8 +822,19 @@ def apply_role(role, inst, in_queue)
...
+      Crowbar::Lock::SharedNonBlocking.new(
+        logger: @logger,
+        path: "/var/chef/cache/pause-file.lock",
```
vuntz: Now the stupid question: is chef-client looking for this lock?

vuntz: (I mean, the daemon)

aspiers: Yes :) https://build.opensuse.org/request/show/323220

aspiers:
@vuntz just noticed that https://build.opensuse.org/request/show/323220 was [not ported to `Devel:Cloud:Shared:Rubygem`](https://build.suse.de/package/revisions/Devel:Cloud:Shared:Rubygem/rubygem-chef) so technically the gating test was invalid :-(

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
aspiers: > There's also one thing I'm not sure I like here: if the lock is shared, and we were the last owner, then by releasing it, the content of the lock file will reference us while we will have released the lock.

No, once the last owner releases it, the lock file is removed.

> Would it make sense to append to the lock file instead of recreating the lock file to say who last requested it?

Yes, that could be useful for debugging.  I will also make releasing append to it in the cases where there is still at least one owner remaining after the release.

dirkmueller:
@aspiers fixed - (I guess you realize that you're maintainer of that package ?)

aspiers:
@dirkmueller Thanks! I can't remember but I suspect I thought that somehow these changes got automatically propagated from OBS to IBS ... or maybe I just forgot to do it :)

crowbar_framework/spec/lib/crowbar/lock/local_blocking_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+        expect(lock.locked?).to be true
+      end
+      expect(lock.locked?).to be false
+    end
+
+    it "is an exclusive lock" do
+      # Check that one of two attempts racing to obtain lock will win
+      lock2 = subject.class.new
+      lock.acquire
+      expect {
```
aspiers: In this case I disagree with the hound dog, because introducing `do` and `end` would disrupt the natural flow of the expectation statement.  It's supposed to read close to natural English language:

    expect [something] to [match some expectation]

If it's changed to

    expect do [something] end to [match some expectation]

then that really spoils the natural flow of the "sentence".

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
vuntz: >> There's also one thing I'm not sure I like here: if the lock is shared, and we were the last owner, then by releasing it, the content of the lock file will reference us while we will have released the lock.
> 
> No, once the last owner releases it, the lock file is removed.

I really meant "we were the last to acquire the lock" when I said the "last owner". That probably makes my original comment clearer ;-)

crowbar_framework/lib/crowbar/lock/local_blocking.rb
aspiers: You mean calling `#acquire` on a lock which was already acquired and released?  Why wouldn't it work?  `#acquire` will simply reopen it.

crowbar_framework/app/models/service_object.rb
aspiers: I did wonder about that - actually I spent quite a bit of agonizing over it, and even reduced the amount of duplication (previously there were 3 methods).  But in the end I decided it was more trouble than it was worth.  I'm open to changing my mind if you disagree, but then I'd require details on how to avoid duplication, e.g. what would the shared module/class be called and where would it live?

crowbar_framework/app/models/service_object.rb
aspiers: Very good point, we shouldn't do that.  I'll fix.

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
aspiers: Yes, isn't that reasonably obvious? I could add a comment but then that's an extra line of spam appearing in `production.log` every time a lock is released, since the shell code is logged at DEBUG level ...

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
aspiers: I decided against it because 1) it will get created again later anyway, and 2) it's a nice indicator that the locking mechanism has been used at least once on this node.

aspiers: Well, I suppose the `flock` file indicates that too.  I don't really mind.  If you want me to add removal I can do that.

aspiers:
@rsalevsky I fixed all the hound issues I agree with.  I deliberately didn't fix the quoting and `{ }` ones because I think hound is wrong.  I intend to submit changes to the SUSE style guide repo for that.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1412,26 +1418,6 @@ def run_remote_chef_client(node, command, logfile_name)
...
-
-      @logger.debug "apply_role: #{action.to_s} chef service on #{node_name}"
-      node.run_service :chef, action
-      wait_nodes << node_name
-    end
-
-    # wait for chef clients on all nodes
-    wait_nodes.each do |node_name|
-      wait_for_chef_clients(node_name, logger: true)
-    end if action == :stop
```
vuntz: Note that we do want to keep this. It's actually pretty important.

aspiers: Oops, yes.

aspiers:
@rsalevsky Based on my improved understanding of the SUSE style, there is only one hound violation yet and that is deliberate for reasons explained in the comment.

crowbar_framework/app/models/service_object.rb
aspiers: Fixed - now waits for intervallic chef-client runs to finish.

crowbar_framework/app/models/service_object.rb
aspiers: Agreed - renamed to `nodes_to_lock`.

crowbar_framework/app/models/service_object.rb
aspiers: Ugh, this is horrible.  I've had to move the locking code later, because at this point we don't yet know which nodes are related to the proposal being applied.  And that means also moving the code which waits for any currently running chef-client runs to finish so that it's later too.  Due to the tangled monstrosity of `apply_role`, it's hard to know if I got it right - however I think I chose the best spot which is basically as early as possible and before anything significant gets changed.  The only thing which worries me is an inexplicable `role.save` very early on in `apply_role` - any ideas what that's for?

@vuntz I'm relying heavily on you for sanity checks here :-/

crowbar_framework/spec/lib/crowbar/lock/local_blocking_spec.rb
rsalevsky: No strong opinion on that, maybe we should discussed this in the ruby SUSE style repo.

rsalevsky:
I know it's not your fault but when we touch the code anyway can you move the `self.*` methods into a `class << self` block? This makes our code more consistent.

crowbar_framework/spec/lib/crowbar/lock/local_blocking_spec.rb
aspiers: Could do, yes.  In general I agree with this guideline though - `{}` for single-line blocks and `do ... end` for multi-line.  I think this is just a rare exception to that guideline - but that's why it's called "style guide" not "style rules" ;-)

aspiers:
Sure, I agree that `class << self` is generally a nicer approach (although I wish Ruby had a less cryptic way to do it ...)

aspiers:
@rsalevsky Actually are you talking about the `self.*` methods in `remote_cmd.rb`?  If so, that's a module not a class, so I think [`extend self` or `module_function`](http://idiosyncratic-ruby.com/8-self-improvement.html) would be better.  Let me know what you think.

crowbar_framework/app/models/service_object.rb
vuntz: I'd probably just create a `LockHelper` module. But I won't fight over it :-)

crowbar_framework/app/models/service_object.rb
```diff
@@ -1043,12 +1036,43 @@ def apply_role(role, inst, in_queue)
...
+        logger: @logger,
+        path: "/var/chef/cache/pause-file.lock",
+        node: node,
+        owner: "apply_role-#{role.name}-#{inst}-#{Process.pid}",
+        reason: "apply_role(#{role.name}, #{inst}, #{in_queue}) pid #{Process.pid}"
+      ).acquire
+    end
+
+    # Now that we've ensured no new intervallic runs can be started,
+    # wait for any which started before we paused the daemons:
```
vuntz: Question on comment style, but usually, I think we just end the comment with `.`, not with `:` (since it's pretty clear the comment applies to what's below). Or am I the one doing this wrong? :-)

crowbar_framework/app/models/service_object.rb
```diff
@@ -1043,12 +1036,43 @@ def apply_role(role, inst, in_queue)
...
+    wait_for_chef_daemons(new_elements.values.flatten.uniq)
+
+    # By this point, no intervallic runs should be running, and no
+    # more will be able to start running until we release the locks
+    # after the proposal has finished applying.
+
     # save databag with the role removal intention
     proposal.save if save_proposal
```
vuntz: I'd probably save the proposal before the code you moved here, so that this isn't blocked by `wait_for_chef_daemons`.

crowbar_framework/app/models/service_object.rb
vuntz: See https://github.com/crowbar/crowbar-core/pull/14/files#r38840207

crowbar_framework/app/models/service_object.rb
```diff
@@ -1043,12 +1036,43 @@ def apply_role(role, inst, in_queue)
...
+        path: "/var/chef/cache/pause-file.lock",
+        node: node,
+        owner: "apply_role-#{role.name}-#{inst}-#{Process.pid}",
+        reason: "apply_role(#{role.name}, #{inst}, #{in_queue}) pid #{Process.pid}"
+      ).acquire
+    end
+
+    # Now that we've ensured no new intervallic runs can be started,
+    # wait for any which started before we paused the daemons:
+    wait_for_chef_daemons(new_elements.values.flatten.uniq)
```
vuntz: You should do it on `applying_nodes`, I think. Or actually, on `nodes_to_lock`. That is: we want to wait for nodes where we also lock. Waiting where we don't lock is kind of useless.

crowbar_framework/lib/crowbar/lock/local_blocking.rb
vuntz: Because `@file` is not `nil`?

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
vuntz: Well actually, I'm not sure it's good to be that spammy in the log in the first place ;-) Maybe something to fix (I'll comment on this in the right place)

Back to the original topic: it's not totally obvious as in "I had to think for more than one second to make sure I got it right". Having a comment outside of the shell code (in the ruby code) would be good enough.

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
```diff
@@ -0,0 +1,141 @@
...
+    end
+
+    private
+
+    # Run a critical region with an exclusive lock so that no
+    # concurrent thread can race with it.
+    def run_with_exclusive_lock(shell)
+      cmd = "flock #{@meta_lock} bash -es"
+      logger.debug("Will execute on node #{node}: #{cmd}" + \
+                   "and feed the following to STDIN:\n#{shell}")
```
vuntz: Do we really want to put the shell command there? Don't we know what it will be anyway?

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
vuntz: Why would it be created again later? It's not likely that a proposal will be re-applied. On the other hand, I would say it makes things a bit confusing because it might look like there's a lock (if you look too quickly).

FWIW, in general, I prefer such things to be automatically cleaned since they're just "helpers".

crowbar_framework/app/models/service_object.rb
aspiers: I'd never thought about it before, and I use either.  I don't think either is wrong, but I'll change this to keep (y)our inconsistent style OCD demons at bay ;-)

crowbar_framework/app/models/service_object.rb
aspiers: Whoops, yes - thanks.

crowbar_framework/app/models/service_object.rb
vuntz: Oh, I don't mind, tbh. It was really just a question :-)

crowbar_framework/app/models/service_object.rb
aspiers: I wondered about this.  Why would it be bad if it was blocked by `wait_for_chef_daemons`?  And are you sure there's no risk of any races with already running `chef-client` daemons?  I was fairly sure that no recipes save to proposal databags (I can't think of any good reason they ever would), and in any case proposals are stored via ActiveRecord now, not Chef, but even so I wasn't 100% sure I wanted to risk it.

vuntz: Oh, it's not an issue with the chef-client daemons. I'm more worried about the user saving the proposal again in the webui. By not saving early, we would save and override the user changes (I assume).

aspiers: OK, you persuaded me :)

crowbar_framework/lib/crowbar/lock/local_blocking.rb
aspiers: Ah ISWYM - you can't flock a closed filehandle.  OK, I've fixed this and added another test case.

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
aspiers: OK then.

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
aspiers: We re-apply proposals all the time, so I'm not sure where that comment is coming from.  But OK I'll make it clean up the directory.

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
aspiers: Yeah I guess so, it was more during debugging.  I'll get rid of that.

aspiers:
OK, all comments should now be addressed.  Just need to test this properly by

[ ] 1. starting a chef-client daemon before applying a proposal, and
[ ] 2. by doing `rcchef-client restart` whilst a proposal is already applying - this will make the daemon attempt a run almost immediately

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
```diff
@@ -0,0 +1,151 @@
...
+        end
+        instance_variable_set "@#{option}", options[option.to_sym]
+      end
+
+      # We manage execution of the command via dependency injection,
+      # so that the test suite can execute the commands locally.
+      default_executor = lambda do |node, cmd, shell|
+        out, err, status = RemoteNode.pipe_ssh_cmd(node, cmd, shell)
+        unless $?.success?
+          raise self.exec_failure_msg(node, cmd, shell, out, err, status)
```
aspiers: Bad dog! I don't think this `self` is redundant.

crowbar_framework/lib/crowbar/lock/local_blocking.rb
```diff
@@ -0,0 +1,58 @@
...
+        logger.error("Couldn't open #{path} for locking: #$!")
+        logger.error("cwd was #{Dir.getwd})")
+        raise "Couldn't open #{path} for locking: #$!"
+      end
+      logger.debug("Acquiring #{name} lock")
+      count = 0
+      loop do
+        count += 1
+        logger.debug("Lock #{path} attempt #{count}")
+        if file.flock(File::LOCK_EX | File::LOCK_NB)
```
vuntz: Hrm, s/file/@file/ ? Or am I missing something?

crowbar_framework/lib/remote_node.rb
```diff
@@ -18,43 +18,87 @@
...
+    ip = resolve_host(host)
     msg = "Waiting for already running chef-client on #{host} (#{ip}); next attempt in #{sleep_time} seconds until #{timeout_at}"
-    self.wait_until(msg, timeout_at.to_i, sleep_time, options) do
-      self.port_open?(ip, 22) && !self.chef_clients_running?(ip)
+    wait_until(msg, timeout_at.to_i, sleep_time, options) do
+      port_open?(ip, 22) && !chef_clients_running?(ip)
     end
   end
 
+  def reboot_done?(host_or_ip, reboot_requesttime, options = {})
```
vuntz: Why is this not private anymore?

crowbar_framework/lib/crowbar/lock/local_blocking.rb
aspiers: You're missing the `attr_reader :file` in `lock.rb`.

crowbar_framework/lib/remote_node.rb
aspiers: Because `remote_node_spec.rb` tests it.  This is explained in the commit message.

aspiers:
@vuntz This is ready for final review now.

rsalevsky:
Do we really need the `crowbar_framework/lib/crowbar/error/locking_failure.rb` file? Because then we need for every new Error Class a new file with only a view lines. I prefer a general file where we can add all custom error classes.

crowbar_framework/lib/remote_node.rb
rsalevsky: Can you refactor this method a little bit? For me there are to many `if else`.

crowbar_framework/lib/remote_node.rb
```diff
@@ -18,43 +18,87 @@
...
+      true
+    end
+  end
+
+  # get boot time of the given host or 0 if command can not be executed
+  def ssh_cmd_get_boottime(host_or_ip, options = {})
+    logger = options.fetch(:logger, nil)
+
+    ssh_cmd = ssh_cmd_base(host_or_ip)
+    ssh_cmd << "'echo $(($(date +%s) - $(cat /proc/uptime|cut -d \" \" -f 1|cut -d \".\" -f 1)))'"
```
rsalevsky: I prefer using the build in ruby functions so in this case `ssh_cmd.push`.

crowbar_framework/lib/remote_node.rb
```diff
@@ -18,43 +18,87 @@
...
+
+  def ssh_cmd(host_or_ip, cmd)
+    ssh = ssh_cmd_base(host_or_ip)
+    ssh << cmd
+    system(*ssh)
+  end
+
+  def pipe_ssh_cmd(host_or_ip, cmd, stdin)
+    ssh = ssh_cmd_base(host_or_ip)
+    ssh << cmd
```
rsalevsky: Same here.

crowbar_framework/lib/remote_node.rb
aspiers: @rsalevsky I agree about the `if else` but sorry but this code is totally unrelated to this PR, so refactoring it is outside the scope of the story (which has already taken far longer than we hoped and got a wider scope than was necessary).  We can easily submit a separate PR to refactor it after this one is merged, but I have to draw the line somewhere otherwise this story will last forever.

crowbar_framework/lib/remote_node.rb
aspiers: Above comment applies here too.  I'm more than happy to see `<<` changed but not in this PR.  Also since '<<` vs. `push` is a purely stylistic choice, I recommend you first send a PR to https://github.com/SUSE/style-guides, otherwise there's not much point changing it in a single file.

aspiers:
Woohoo, the gate finally passed!

tboerger:
> Do we really need the crowbar_framework/lib/crowbar/error/locking_failure.rb file? Because then we need for every new Error Class a new file with only a few lines of code. I prefer a general file where we can add all custom error classes.

Yes, we want that. It's more or less a convention for ruby to have each class in a separate file and I can not see any issue in doing it like I started it :)

crowbar_framework/lib/remote_node.rb
tboerger: I prefer push as well and used it for all parts I have refactored/written.

aspiers:
> > Do we really need the crowbar_framework/lib/crowbar/error/locking_failure.rb file? Because then we need for every new Error Class a new file with only a few lines of code. I prefer a general file where we can add all custom error classes.
>
> Yes, we want that.

One small advantage of separate files is that it reduces the risk of merge conflicts if multiple PRs are adding exception classes at the same time.  Although this risk is already very small.

> It's more or less a convention for ruby to have each class in a separate file and I can not see any issue in doing it like I started it :)

While I agree that the "one file per class" convention is very common, it is quite often broken, and multiple exception classes per file is one of the times where it's most commonly broken, e.g.

- https://github.com/sstephenson/sprockets/blob/master/lib/sprockets/errors.rb
- https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/errors.rb
- https://github.com/erikhuda/thor/blob/master/lib/thor/error.rb

Even the MRI Ruby source itself does this in several places, e.g.

- https://github.com/ruby/ruby/blob/trunk/lib/rubygems/errors.rb
- https://github.com/ruby/ruby/blob/trunk/lib/uri/common.rb#L140

If you do `egrep -r 'class .* < (Error|Exception)'` in `/usr/lib64/ruby` and/or `~/.rvm` you will find many more examples.

My personal opinion is that for a small number of custom exception classes I don't think it really matters either way.  If we had 20 or 50 then maybe one per file would be overkill :)

aspiers:
@vuntz shall I rebase this, to allow for another gate test on top of current `master` in case it doesn't get another +1 and merge in the next 2 hours?

vuntz:
@aspiers no need to rebase, imho. I'll just merge it but we need someone else to review :-)

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -46,9 +46,10 @@ def transition(inst, name, state)
 
     @logger.info("Crowbar transition enter: #{name} to #{state}")
 
-    f = acquire_lock "BA-LOCK"
-    begin
-      node = NodeObject.find_node_by_name name
+    pop_it = false
+    node = NodeObject.find_node_by_name name
```
vuntz: Hrm, I just noticed that in the old code, we were fetching the node after acquiring the lock, and now this is different.

I actually believe the old code is more correct here since the lock is there to protect the node (I think?)

rsalevsky:
No objections with merging this PR from my side. :+1:

crowbar_framework/lib/crowbar/lock/local_blocking.rb
vuntz: For the record, it's a bit confusing that `release` is actually using `@file` ;-)

crowbar_framework/app/models/crowbar_service.rb
aspiers: Fixed.

aspiers:
Tweaked with @vuntz's `node = nil` suggestion and rebased.

vuntz:
+1

crowbar_framework/lib/crowbar/lock/local_blocking.rb
aspiers: Yeah I see what you mean now - if another excuse emerges for a force-push, I'll change this.

aspiers:
Oops, I have no idea how this didn't break Swift and Tempest, but https://github.com/crowbar/crowbar-openstack/pull/37 should be required to unbreak.

vuntz:
> Oops, I have no idea how this didn't break Swift and Tempest, but crowbar/crowbar-openstack#37 should be required to unbreak.

Because these bits are in a webui that no one uses :-)

aspiers:
Ah right :-)

PR #15, opened 2015-08-28
Clean MD Raid markers from partitions when wiping disks (bsc#939684)

dirkmueller:
Those are hard to get rid of otherwise, and they will
screw up a following autoyast install.

### DIFF
--- a/updates/control.sh
+++ b/updates/control.sh
@@ -179,6 +179,8 @@ nuke_everything() {
         [[ -b /dev/$name && -w /dev/$name && $name != name ]] || continue
         [[ $name = loop* ]] && continue
         [[ $name = dm* ]] && continue
+        # Remove Software RAID markers from the disk
+        wipefs -a /dev/$name
         if (( blocks >= 2048)); then
             dd "if=/dev/zero" "of=/dev/$name" "bs=512" "count=2048"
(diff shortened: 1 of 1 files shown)


jsuchome:
Although this is meant to fix bnc#939684 , it actually helped me install RAID node when I was trying to reproduce https://bugzilla.suse.com/show_bug.cgi?id=943330. Without this fix, I was experiencing some problems reported by installer, that was not able to proper create RAID because it found something unexpected on the disks (http://imagebin.suse.de/1888/img)

tboerger:
:+1:

jsuchome:
There should be probably also STDERR redirect, otherwise it reports

"mdadm: Unrecognised md component device - /dev/vda"

for some partitions.

jsuchome:
@dirkmueller +1 from me (maybe add that  2>&1redirect ?). 

Could you create tex PR as well?

jsuchome:
Hm, retracting last comment: sometimes it still does not seem to be enough, I've just met http://imagebin.suse.de/1888/img again.

Would it help to add ``wipefs -a /dev/$name`` as well? (Suggestion by @bmwiedemann )

jsuchome:
``wipefs -a`` alone seems to be enough

jsuchome:
+1

vuntz:
+1

PR #16, opened 2015-08-28
Cleaner rspec output

aspiers:


tboerger:
Is there a card for that? LGTM codewise.

aspiers:
No card, just noticed it when running the test suite and realised would not take long to fix.

rsalevsky:
+1

PR #17, opened 2015-08-29
Start crowbar_notify_shutdown service after register

rsalevsky:
At the first shutdown or reboot systemd assumes that the crowbar_join
service and the coherently crowbar_notify_shutdown service was never
started. Due to this assumption the service will not stopped at the
first shutdown or reboot.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/base.rb
+++ b/chef/cookbooks/provisioner/recipes/base.rb
@@ -283,7 +283,7 @@ if node["platform"] == "suse" && !node.roles.include?("provisioner-server")
     end
 
     service "crowbar_notify_shutdown" do
-      action :enable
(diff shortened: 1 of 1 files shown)


rsalevsky:
Tested after the gating run, works as expected.

toabctl:
What about starting the service where it is created and enabled? See https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/provisioner/recipes/base.rb#L287 ?

rsalevsky:
@toabctl Done, but needs to be tested after the gating run.

tboerger:
:+1:

rsalevsky:
Tested the new patch while gating run, works as expected.

toabctl:
+1

PR #19, opened 2015-08-31
stop ntp server before checking for ntp availability

bmwiedemann:
otherwise, we would add ourselves to the list of valid ntp servers
and timeout later

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
+++ b/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
@@ -457,9 +457,14 @@ if [ "$MODE" == "setup" -o "$MODE" == "start" ]; then
     HOSTNAME=$(hostname -f)
 
     # stop ntpd before we run ntpdate, and start it again afterwards.
+<% if @target_platform_version.to_f < 12.0 then %>
+    NTP_SERVICE=ntp
+<% else %>
+    NTP_SERVICE=ntpd
+<% end %>
     NTP_START=0
-    if service ntp status > /dev/null; then
-        log_to time service ntp stop
+    if service $NTP_SERVICE status > /dev/null; then
+        log_to time service $NTP_SERVICE stop
         NTP_START=1
     fi
 
@@ -498,7 +503,7 @@ if [ "$MODE" == "setup" -o "$MODE" == "start" ]; then
     # make sure to keep hostname
     [ "$MODE" == "setup" ] && echo $HOSTNAME > /etc/HOSTNAME
 
-    [ $NTP_START -eq 1 ] && log_to time service ntp start
+    [ $NTP_START -eq 1 ] && log_to time service $NTP_SERVICE start
     [ "$MODE" == "setup" ] && log_to chef service chef-client start
 
     [ $final_state == "ready" ] || EXVAL=1
--- a/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
+++ b/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
@@ -456,6 +456,13 @@ if [ "$MODE" == "setup" -o "$MODE" == "start" ]; then
     # Get our hostname
     HOSTNAME=$(hostname -f)
 
+    # stop ntpd before we run ntpdate, and start it again afterwards.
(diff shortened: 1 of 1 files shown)


dguitarbite:
+1

bmwiedemann:
just noticed, that service ntp status does not work on 13.2 (and likely SLE12, because it was renamed to ntpd there)

dguitarbite:
@bmwiedemann I can confirm that its 'ntpd.service' in SLE12. Not sure if you need a systemd command for the same.

bmwiedemann:
2nd commit to allow for ntpd is not yet tested.

tboerger:
But the switch is only required for cloud 5

dguitarbite:
@tboerger Yes, for SLE11 (Cloud 5).

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
```diff
@@ -456,6 +456,15 @@ if [ "$MODE" == "setup" -o "$MODE" == "start" ]; then
     # Get our hostname
     HOSTNAME=$(hostname -f)
 
+    # stop ntpd before we run ntpdate, and start it again afterwards.
+    ntp=ntp
+    service ntp status > /dev/null || ntp=ntpd
```
dguitarbite: Removing this variable should be enough to only allow for SLE11

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
```diff
@@ -456,6 +456,15 @@ if [ "$MODE" == "setup" -o "$MODE" == "start" ]; then
...
+    NTP_START=0
+    if service $ntp status > /dev/null; then
```
dguitarbite: I suggest "service ntp status ..."

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
tboerger: Sure, removing this variable, and then? Then it's not working for SLE 12, awesome!

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
tboerger: So what?

dguitarbite: Just extension of the previous comment to remove the $ntp variable which is the logic for setting the ntp/ntpd based of SLE version :).

But there could be an ideal "crowbar/chef" way of checking versions which may be given more preference.

tboerger: This current approach works for any distribution.

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
vuntz: There's also some `<% if @target_platform_version.to_f < 12.0 %>` magic in the template, so why not use this to define the name of the service?

toabctl: +1 for @vuntz suggestion

bmwiedemann:
updated according to @vuntz and @toabctl and slightly tested it

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
```diff
@@ -456,6 +456,18 @@ if [ "$MODE" == "setup" -o "$MODE" == "start" ]; then
     # Get our hostname
     HOSTNAME=$(hostname -f)
 
+    # stop ntpd before we run ntpdate, and start it again afterwards.
+<% if @target_platform_version.to_f < 12.0 then %>
+    ntp=ntp
+<% else %>
+    ntp=ntpd
```
vuntz: I'd probably call the variable `NTP_SERVICE`. Small update to change this would be welcome, but I give my +1 in advance since the rest is totally fine.

dguitarbite:
+1

toabctl:
+1

PR #20, opened 2015-09-01
Move crowbar behind Apache proxy

rsalevsky:
Superseeds #3


tboerger:
Just as a sidenote: This removes the authentication from the rails application, but as the rails application is anyway running on 127.0.0.1:3000 it should be no problem as an attacker can do anyway what he wants if he got local access to the admin node. Hopefully this will be replaced with a real login soonish.

:+1:

MaximilianMeister:
when the commit message is updated and the gating passes, I am fine with it

:+1:

rsalevsky:
Gating job before commit message update: https://ci.suse.de/job/openstack-mkcloud/7195/

rsalevsky:
Originally gating passes, let's merge.

PR #22, opened 2015-09-03
Create 'bios_boot' partition for installation (bsc#935283)

dguitarbite:
SLE11 nodes have issues while booting/installation for harddisk size > 2
TB. Create a boot partition of type gpt with embeded mbr to allow using
bigger disks for non-raid type.

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
@@ -96,8 +96,28 @@
           <device>/dev/<%= @boot_device %></device>
         <% end %>
         <initialize config:type="boolean">true</initialize>
-        <partitions config:type="list"/>
         <type config:type="symbol">CT_DISK</type>
+        <disklabel>gpt</disklabel>
+        <partitions config:type="list">
+          <partition>
+            <create config:type="boolean">true</create>
+            <partition_id config:type="integer">263</partition_id>
+            <size>10M</size>
+          </partition>
+          <partition>
+            <create config:type="boolean">true</create>
+            <filesystem config:type="symbol">swap</filesystem>
+            <format config:type="boolean">true</format>
+            <mount>swap</mount>
+            <size>auto</size>
+          </partition>
+          <partition>
+            <create config:type="boolean">true</create>
+            <format config:type="boolean">true</format>
+            <mount>/</mount>
+            <size>auto</size>
+          </partition>
+        </partitions>
         <use>all</use>
       </drive>
     <% else -%>


jsuchome:
If it is for SLE11 only (I'm not saying it is, just based on commit message), shouldn't it be restricted for sle11sp3 nodes in the template?

dguitarbite:
@jsuchome Yes, I have to discuss this once before this is clarified but this should not affect installation for SLE12 based nodes anyways.

mjura:
@dguitarbite : I would like to keep simple setup as much as we can. Separate `/boot` partition comes from ancient times. I agree with @jsuchome that we should limit this change only for SLE_11_SP3 which will be deprecated soon.

dguitarbite:
WIP: Creating a fix for SLE12 nodes.

dirkmueller:
@mjura The change is the same for SLE11 and SLE12. The inability to boot from partitions > 2.2GB has nothing to do with the OS, its a firmware/BIOS limitation.

dguitarbite:
This should work with both SLE12 and SLE11 nodes. Tested it with different combinations of harddisk sizes 3TB HDD size max. default HDD size. min.

tboerger:
LGTM

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
```diff
@@ -96,8 +104,38 @@
...
         <% end %>
         <initialize config:type="boolean">true</initialize>
-        <partitions config:type="list"/>
         <type config:type="symbol">CT_DISK</type>
+        <disklabel>gpt</disklabel>
+        <partitions config:type="list">
+          <partition>
+            <create config:type="boolean">true</create>
+            <format config:type="boolean">true</format>
+            <fstopt>acl,user_xattr</fstopt>
```
dirkmueller: why is fstopt specified for /boot but not for the others?

dguitarbite: Since we are creating a boot partition manually which is not the defaults, I think its a good idea to explicitly mention fstopt. Without this, grub was not getting installed which was due to permissions issue (most likely reason).

For the other partitions, not specific specific fstopt would make yast decide these attrs. and would be set to defaults which work properly. We could set them if we have some specific requirements. If not, keeping them to be decided by yast (not setting them) should be preferred since in future if some defaults change, we need not worry about custom flags creating regressions :).

jsuchome:
@dguitarbite don't you think we should restrict it for large disks only?

dguitarbite:
@jsuchome Adding the checks for disk size would just add more code IMHO and may not be worth it to save ~150MBs of disk space.

dguitarbite:
-1, I need to check out @jsuchome's patch to create a bios_boot partition instead of /boot partition for SLE11. It seems it works for him.

jsuchome:
@dguitarbite btw, for the RAID case, I didn't change the configuration that offers creating /boot partition. So actually both BIOS boot and /boot partition are created.

dguitarbite:
@jsuchome Ok, So for crowbar-core its a WIP anyways since they are all SLE12 nodes :).

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
```diff
@@ -96,8 +104,34 @@
...
         <% end %>
         <initialize config:type="boolean">true</initialize>
-        <partitions config:type="list"/>
         <type config:type="symbol">CT_DISK</type>
+        <disklabel>gpt</disklabel>
+        <partitions config:type="list">
+          <partition>
+            <create config:type="boolean">true</create>
+            <partition_id config:type="integer">263</partition_id>
+            <partition_nr config:type="integer">1</partition_nr>
```
jsuchome: I think partition_nr is not needed explicitely

jsuchome:
+1 based on the code (but not tested it myself)

dirkmueller:
-1, mkcloud gating runs reliably fail with that code..

dguitarbite:
I'm not sure if the reason for mkcloud gating to fail in the last run is something to do with this change. IMHO, this patch should be ready to be merged.

toabctl:
@dguitarbite please remove the merge commit and rebase.

dguitarbite:
IMO, CI mkcloud gating failure is not because of this patch.

dirkmueller:
+1 (after removing the merge commit)

dguitarbite:
Is this patch ready to be merged?

toabctl:
+1 (but let's wait with merging until M6 is out)

PR #23, opened 2015-09-03
Revert "batch build: abort on unknown aliases" (bsc#944404)

toabctl:
This reverts commit 2b351ca757aac468497198baf97f98d5138bd8cb.

### DIFF
--- a/bin/crowbar_batch
+++ b/bin/crowbar_batch
@@ -97,11 +97,6 @@ def build(input_file)
     input_data.gsub!(ALIAS_TEMPLATE % aliaz, hostname)
   end
 
-  unknown_aliases = input_data.to_enum(:scan, ALIAS_REGEXP).map { Regexp.last_match.to_s }
-  unless unknown_aliases.empty?
-    abort "Some of the aliases in the YAML file are not known to Crowbar:\n- #{unknown_aliases.uniq.join("\n- ")}"
-  end
-
   data = YAML.load(input_data)
   abort "#{input_file} is empty" unless data
 


aspiers:
Why?

toabctl:
> Why?

See bug report.

toabctl:
@thutterer maybe you want to have a look?

aspiers:
-1, we can fix this properly simply by reintroducing `ALIAS_REGEXP`.  It was dropped by 5849ca5917621436 since we didn't reuse it, and I guess 2b351ca was written before that but merged after it appeared, causing the breakage.

BTW best practice is to include the full bugzilla URL inside the commit message as explained at https://etherpad.nue.suse.com/p/cloud-pull-requests

dirkmueller:
I disagree with the requirement to include full bugzilla urls btw..

aspiers:
I didn't say it was a requirement, and nor does the etherpad.  However the advantages of giving the full URL are very obvious (and also explained in the etherpad), which is why I said it's best practice.  If you have arguments against this, please provide them.  And BTW "pasting a URL into a commit message is an extra 3 seconds of work" is not a valid argument ;-)

aspiers:
Closing as it's just as easy to reintroduce the missing constant, and that will avoid having to ditch the valuable sanity check introduced by 2b351ca.

toabctl:
@aspiers are you going to fix this then? ```crowbar_batch``` is unusable for now.

aspiers:
It's a one-line fix to add the constant, but OK I can do it if you want.

PR #25, opened 2015-09-04
provisioner: Rework dependencies of crowbar_join/crowbar_notify_shutdown

vuntz:
I assume the After dependencies were moved to crowbar_notify_shutdown
to avoid duplication, but it's actually a bit confusing. And
crowbar_notify_shutdown doesn't actually depend on services like
libvirtd or xend.


toabctl:
Yes. Was done to avoid duplication. But fine for me to make it more explicit. So +1

rsalevsky:
Any idea why gating fails?

vuntz:
@rsalevsky it fails during install-suse-cloud, so unrelated to this... Not sure if it's a temporary error or a bug that got introduced, though.

mjura:
LGTM +1

PR #27, opened 2015-09-04
Add missing description to barclamp YAMLs

rsalevsky:


toabctl:
+1

nfs_client.yml
```diff
@@ -17,6 +17,7 @@
 barclamp:
   name: 'nfs_client'
   display: 'NFS Client'
+  description: 'NFS Mount Configuration'
```
vuntz: I currently see "Setup of NFS mounts" in the crowbar webui, which I slightly prefer to the above.

ntp.yml
```diff
@@ -18,6 +18,7 @@
 barclamp:
   name: 'ntp'
   display: 'NTP'
+  description: 'Common NTP service for the cluster'
```
vuntz: Can we take this as an opportunity to get a better string? :-) Either "NTP Management" like for many other barclamps or something else.

tboerger:
:+1: Can you update my registry pull request as well? :)

rsalevsky:
@tboerger Will do this when we work an the card.

PR #28, opened 2015-09-04
batch: reintroduce ALIAS_REGEXP

aspiers:
2b351ca introduced a nice new sanity check for invalid aliases, but it relies on `ALIAS_REGEXP` and was written before 5849ca5 which dropped the definition of `ALIAS_REGEXP`.  So we need to bring back `ALIAS_REGEXP` to make this check work again.

Supercedes https://github.com/crowbar/crowbar-core/pull/23

### DIFF
--- a/bin/crowbar_batch
+++ b/bin/crowbar_batch
@@ -44,6 +44,7 @@ require "easy_diff"
 require "chef/mixin/deep_merge"
 require "pp"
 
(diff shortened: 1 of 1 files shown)


tboerger:
:+1:

rsalevsky:
+1

toabctl:
thanks @aspiers

PR #29, opened 2015-09-07
refactor provisioner::base and autogenerate /etc/profile.d/crowbar-vars.sh

aspiers:
Split up the enormous `base.rb` into manageable chunks, and add a new provisioner::shell-vars recipe which auto-generates `/etc/profile.d/crowbar-vars.sh`.  For each machine alias, it defines shell variables mapping to that machine's node and role.  For example, if you have a machine with alias `controller1`, it means that you can type stuff like:

    knife node show $controller1
    knife role show $controller1r
    less +F /var/log/crowbar/chef-client/$controller1.log

rather than having to endure the pain of tedious messing around with MAC addresses.


chef/cookbooks/provisioner/recipes/ssh.rb
```diff
@@ -0,0 +1,89 @@
...
+
+# Also put authorized_keys in tftpboot path on the admin node so that discovered
+# nodes can use the same.
+if node.roles.include? "crowbar"
+  case node[:platform]
+  when "suse"
+    tftpboot_path = "/srv/tftpboot"
+  else
+    tftpboot_path = "/tftpboot"
+  end
```
vuntz: Please drop this and use `node[:provisioner][:root]` instead.

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -214,170 +75,11 @@
...
-          .gsub("SUFFIX", "${prompt_suffix}") \
-          .gsub("ALIAS", "\\[\\e[01;35m\\]#{aliaz}\\[\\e[0m\\]") \
-          .gsub("HOST", "\\[\\e[01;35m\\]#{node["hostname"]}\\[\\e[0m\\]") \
-          .gsub("FQDN", "\\[\\e[01;35m\\]#{node["fqdn"]}\\[\\e[0m\\]")
-      }
-    )
-  end
-end
+include_recipe "provisioner::shell-prompt"
+include_recipe "provisioner::shell-vars"
```
vuntz: Shouldn't we have the variables only on the admin server? Or is it really useful everywhere?

chef/cookbooks/provisioner/recipes/shell-vars.rb
```diff
@@ -0,0 +1,41 @@
...
+
+env = node[:provisioner][:config][:environment]
+
+node_vars = []
+role_vars = []
+
+search(:node, "provisioner_config_environment:#{env}") do |n|
+  aliaz = n["crowbar"]["display"]["alias"] rescue nil
+  next unless aliaz && !aliaz.empty?
+  short = aliaz.gsub("node", "n")
```
vuntz: Not sure about this `gsub`; this is a bit too magic for people. If they really wanted to use a shorter version, then they would surely have used `n` directly?

chef/cookbooks/provisioner/recipes/shell-vars.rb
```diff
@@ -0,0 +1,41 @@
...
+  node_vars.push [short, n.fqdn]
+
+  role = "crowbar-" + n.fqdn.gsub(".", "_")
+  role_vars.push [short + "r", role]
```
vuntz: I don't think I would add this bit, or only in a developer mode. This is really useful just for us.
(btw, this bit is definitely admin-server-only)

chef/cookbooks/provisioner/recipes/ssh.rb
aspiers: I didn't touch that code - only move it to a new file.  But sure, I can fix that :)

tboerger:
You know that it is bad behavior to use dashes instead of underscores for filenames in ruby?

chef/cookbooks/provisioner/recipes/shell-vars.rb
aspiers: OK I'll drop it.

chef/cookbooks/provisioner/recipes/shell-vars.rb
aspiers: Sorry, I disagree.  Even SEs would find this useful.  And it doesn't do any harm to have it everywhere (see updated version), so why not keep it?  If anything, I'd prefer to drop the whole file from all non-admin nodes.

aspiers:
@tboerger Excellent point - I forgot about that since I was thinking of the files as recipes not Ruby files. Will fix!

aspiers:
@tboerger BTW I found [another reason to prefer underscores](https://github.com/sethvargo/chefspec):

> WARNING Cookbooks with dashes (hyphens) are difficult to test with ChefSpec because of how Chef classifies objects. We recommend naming cookbooks with underscores (_) instead of dashes (-).

aspiers:
Oh, that's talking about cookbooks not recipes.  Never mind :-)

chef/cookbooks/provisioner/recipes/base.rb
aspiers: I thought it could be useful in other places and I didn't see any harm in defining them everywhere.  But I have no problem dropping them from non-admin nodes if you prefer.

aspiers:
For the record, I found a way to automatically change all `.rb` files affected by this PR to have underscores instead of hyphens:

```shell
base_commit=`git upstream` # commit on which this PR is based

# Find and store all files affected by this PR
git diff --name-only $base_commit..HEAD > ~/touched-files.txt

git filter-branch --index-filter '
  # get the current index
  git ls-files -s > index.txt

  # Replace any hyphens in the affected files with underscores
  grep "\.rb$" ~/touched-files.txt | while read file; do
    perl -pi -e "m%\t\Q$file\E\$% && s/-/_/g" index.txt
  done

  # Convert index.txt to a new (binary) index and then replace the old one.
  # We cannot use git-update-index directly on .git/index because then the
  # files pre-rename would persist.
  cat index.txt | \
    GIT_INDEX_FILE=$GIT_INDEX_FILE.new git update-index --index-info && \
  mv "$GIT_INDEX_FILE.new" "$GIT_INDEX_FILE"
  ' $base_commit..HEAD
```

This was based off [this blog post](http://www.fussylogic.co.uk/blog/?p=1250).

chef/cookbooks/provisioner/recipes/shell_prompt.rb
```diff
@@ -0,0 +1,66 @@
...
+          .gsub("ALIAS", "%{\\e[0;35m%}#{aliaz}%{\\e[0m%}") \
+          .gsub("HOST", "%{\\e[0;35m%}#{node["hostname"]}%{\\e[0m%}") \
+          .gsub("FQDN", "%{\\e[0;35m%}#{node["fqdn"]}%{\\e[0m%}")
+      },
+
+      bash_prompt_from_template: proc {
+        node["provisioner"]["shell_prompt"].to_s \
+          .gsub("USER", "\\[\\e[01;31m\\]\\u\\[\\e[0m\\]") \
+          .gsub("CWD", "\\[\\e[01;31m\\]\\w\\[\\e[0m\\]") \
+          .gsub("SUFFIX", "${prompt_suffix}") \
```
aspiers: I cannot reproduce this style error when running Rubocop locally, so I can't experiment to try to fix it.  Any ideas @tboerger or @rsalevsky?

vuntz:
+1

aspiers:
Rebased because it looks like the gate failed due to the PR being based on `master` from before the change regarding digest authentication.

aspiers:
And now the gate fails for some weird reason which looks unrelated :-(

vuntz:
@aspiers not sure since when, but this needs a rebase :/

vuntz:
@aspiers ping for the rebase

aspiers:
@vuntz Rebased.  I also locally backported to 3.0 and tested on my c19 which is currently running SOC6 - seemed to work fine, but I didn't test on SOC7 yet.

aspiers:
@tboerger I ran `rubocop -c .hound.ruby.yaml` against the files changed by this PR, and it didn't generate any of the `Style/DotPosition` violations Hound is complaining about - any ideas why?

aspiers:
That was with `rubocop` 0.33.  I upgraded to 0.38 and now it complains:

```
Error: obsolete parameter RunRailsCops (for AllCops) found in /home/adam/SUSE/cloud/crowbar/git/crowbar/barclamps/crowbar-core/.hound.ruby.yml
Use the following configuration instead:
Rails:
  Enabled: true
```

but if I fix that, I still can't reproduce the errors Hound is generating.

tboerger:
Hound is not complaining anymore :)

aspiers:
@tboerger That's because of the last commit, but the `Style/DotPosition` question still needs to be answered.

aspiers:
Rebased.

chef/cookbooks/provisioner/libraries/find_server.rb
```diff
@@ -0,0 +1,32 @@
...
+        begin
+          env = node[:provisioner][:config][:environment]
+          server = search(
+            :node,
+            "roles:provisioner-server AND provisioner_config_environment:#{env}"
+          ).first
+          Chef::Log.info("Provisioner server is #{server[:hostname]}")
+          server
+        end
+    end
```
vuntz: I'm a bit reluctant about this, because you're caching a full node. As we've seen, the cache will stay alive forever in the chef-client daemon, and I'm guessing (although not sure) that the node attributes will slowly be outdated after a while, but the cache never invalidated.

At the minimum please add the hack using ohai_time to invalidate the cache.

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -207,186 +74,13 @@
...
-aliaz = begin
-  display_alias = node["crowbar"]["display"]["alias"]
-  if display_alias && !display_alias.empty?
-    display_alias
-  else
-    node["hostname"]
-  end
-rescue
-  node["hostname"]
+  include_recipe "provisioner::client"
```
vuntz: Can you move the `if platform == suse` bit inside this new recipe? As the recipe should be run regardless of the platform.

chef/cookbooks/provisioner/recipes/shell_vars.rb
```diff
@@ -0,0 +1,38 @@
...
+role_vars = []
+
+search(:node, "provisioner_config_environment:#{env}") do |n|
+  next unless n["crowbar"]["display"]
+  aliaz = n["crowbar"]["display"]["alias"]
+  next unless aliaz && !aliaz.empty?
+  node_vars.push [aliaz, n.fqdn]
+
+  role = "crowbar-" + n.fqdn.tr(".", "_")
+  role_vars.push [aliaz + "r", role]
```
vuntz: You should probably do this before the alias, since right now this might not be run (due to the `next`).

aspiers: No that's intentional, because the role variable is also named based on the alias, so you can do stuff like

    knife role show $controller1r

chef/cookbooks/provisioner/libraries/find_server.rb
aspiers: Good point, will do.

chef/cookbooks/provisioner/recipes/base.rb
aspiers: Yep, good catch.

aspiers:
Rebased again.

chef/cookbooks/provisioner/recipes/shell_vars.rb
vuntz: Doh, of course.

vuntz:
+1

rsalevsky:
+1

PR #30, opened 2015-09-08
For large disks, we're gonna need to create BIOS boot partition

jsuchome:
Just testing first solution. What is missing:

- partition_nr ? [EDIT: not needed]
- more details about swap (likely not needed) [EDIT: true, not needed]
- size is just a guess EDIT: 1MB should be enough based on https://en.wikipedia.org/wiki/BIOS_boot_partition
- limit this for big disks only (maybe not needed)

Based on real example: https://bugzilla.suse.com/attachment.cgi?id=646431 . Should fix https://bugzilla.suse.com/show_bug.cgi?id=944467

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
@@ -108,6 +108,11 @@
           <initialize config:type="boolean">true</initialize>
           <partitions config:type="list">
             <partition>
+              <create config:type="boolean">true</create>
+              <partition_id config:type="integer">263</partition_id>
+              <size>10M</size>
+            </partition>
+            <partition>
               <mount>swap</mount>
               <size>auto</size>
             </partition>


dguitarbite:
@jsuchome I could not make a 1mb partition using autoyast, the default minimum partition size was 10mbs ... does the 1mb work for you with SLE12?

jsuchome:
@dguitarbite Hm, I didn't test it acutally, the tested version was with 10MB. Let me ask Olli.

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
```diff
@@ -108,6 +108,11 @@
           <initialize config:type="boolean">true</initialize>
           <partitions config:type="list">
             <partition>
+              <create config:type="boolean">true</create>
```
dguitarbite: I suggest adding partition table type as gpt.

jsuchome: Not needed, I assume YaST will do it automatically.

dguitarbite: @jsuchome Yes, I think for SLE12 it should do it automatically.

jsuchome:
@dguitarbite about the size: looks like YaST will automatically change 1MB value to something it can use (https://bugzilla.suse.com/show_bug.cgi?id=944467#c31), so I think it's fine we leave 1MB here

dguitarbite:
I would suggest using 10 MB, since its going to update it anyways ... better to do it in the XML file.

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
jsuchome: Yes, it's tested, see the bug report

dguitarbite: Then we could live without it.

dguitarbite:
@jsuchome Small nitpick please add (#BSC944467)

dguitarbite:
:+1:  Looking at the autoyast.xml.erb changes. I did not test this with the cloud product. If you need me to test it, please request since its a tricky to get RAID setup with VMs.

jsuchome:
@dguitarbite no, thanks, I think it's enough tested by Olli, and I will also test with normal size disks to make sure nothing is broken

vuntz:
Can't really say if that's correct or not, but nothing obviously wrong. So +0.5 from me because it fixes the issue according to the bug.

rsalevsky:
LGTM

jsuchome:
I've tested it with normal size disks, Cloud6 and SLE12 nodes. Everything seems to work fine.

PR #31, opened 2015-09-08
Rename YAMLs with hyphen to underscore

rsalevsky:
To avoid using workarounds for testing if the name has a hyphen or a
underscore we should always use underscores.


tboerger:
:+1: but merge request conflicts.

MaximilianMeister:
+1

PR #33, opened 2015-09-08
Let the network barclamp create OVS bridges

rhafer:
Initial work for moving the ovs creation for neutron to the network barclamp as an alternative approach to https://github.com/crowbar/crowbar-openstack/pull/3

It's not quite ready yet (but it works already on my test system), but I'd like to get some feedback.

This should allow to make the neutron::common_agent recipe a lot simpler. I am still somewhat undecided on the new proposal attributes. An alternative approach to the new ```add_ovs_brigde``` attribute would be to use the existing ```add_bridge``` flag and add an (optional?) ```bridge_type``` attribute. I am open to suggestion. 

Also it might make sense to create a common base class for ```Nic::OvsBridge``` and ```Nic::Bridge``` in barclamp/libraries/nic.rb since they share some code. 

The related change to the neutron barclamp will follow soon.

### DIFF
--- a/chef/cookbooks/barclamp/libraries/nic.rb
+++ b/chef/cookbooks/barclamp/libraries/nic.rb
@@ -648,17 +648,18 @@ class ::Nic
     end
 
     def add_slave(slave)
-      slave = self.class.coerce(slave)
       unless ::Nic.exists?(slave)
         raise ::ArgumentError.new("#{slave} does not exist, cannot add to bridge#{@nic}")
       end
-      return if slaves.member?(slave)
+      slave = self.class.coerce(slave)
+      return slave if slaves.member?(slave)
       if current_master = slave.master
         current_master.remove_slave(slave)
       end
       slave.up
       usurp(slave)
       ::Kernel.system("brctl addif #{@nic} #{slave}")
+      slave
     end
 
     def remove_slave(slave)
@@ -668,6 +669,7 @@ class ::Nic
       end
       ::Kernel.system("brctl delif #{@nic} #{slave}")
       slave.down
+      slave
     end
 
     def stp
@@ -745,17 +747,18 @@ class ::Nic
     end
 
     def add_slave(slave)
-      slave = self.class.coerce(slave)
       unless ::Nic.exists?(slave)
         raise ::ArgumentError.new("#{slave} does not exist, cannot add to bridge#{@nic}")
       end
-      return if slaves.member?(slave)
+      slave = self.class.coerce(slave)
+      return slave if slaves.member?(slave)
       if current_master = slave.master
         current_master.remove_slave(slave)
       end
       slave.up
       usurp(slave)
       ::Kernel.system("ovs-vsctl add-port #{@nic} #{slave}")
+      slave
     end
 
     def remove_slave(slave)
@@ -765,6 +768,7 @@ class ::Nic
       end
       ::Kernel.system("ovs-vsctl del-port #{@nic} #{slave}")
       slave.down
+      slave
     end
 
     def up
--- a/chef/cookbooks/barclamp/libraries/nic.rb
+++ b/chef/cookbooks/barclamp/libraries/nic.rb
@@ -374,7 +374,9 @@ class ::Nic
   # Return the bond we are enslaved to, or nil if we are not in a bond.
   def bond_master
     return nil unless File.exists?("#{@nicdir}/master")
-    Nic.new(File.readlink("#{@nicdir}/master").split("/")[-1])
+    master = File.readlink("#{@nicdir}/master").split("/")[-1]
+    return nil unless Nic.bond?(master)
+    Nic.new(master)
   end
 
   # Return the bridge we are enslaved to, or nil if we are not in a bridge.
--- a/chef/cookbooks/barclamp/libraries/nic.rb
+++ b/chef/cookbooks/barclamp/libraries/nic.rb
@@ -650,8 +650,8 @@ class ::Nic
       unless ::Nic.exists?(slave)
         raise ::ArgumentError.new("#{slave} does not exist, cannot add to bridge#{@nic}")
       end
-      return if self.slaves.member?(slave)
-      if current_master = slave.master()
+      return if slaves.member?(slave)
+      if current_master = slave.master
         current_master.remove_slave(slave)
       end
       slave.up
@@ -692,7 +692,7 @@ class ::Nic
     end
 
     def up
-      slaves.each{ |s|s.up }
+      slaves.each(&:up)
       super
     end
 
@@ -705,7 +705,7 @@ class ::Nic
       nil
     end
 
-    def self.create(nic,slaves=[])
+    def self.create(nic, slaves = [])
       Chef::Log.info("Creating new bridge #{nic}")
       if self.exists?(nic)
         raise ::ArgumentError.new("#{nic} already exists.")
@@ -747,8 +747,8 @@ class ::Nic
       unless ::Nic.exists?(slave)
         raise ::ArgumentError.new("#{slave} does not exist, cannot add to bridge#{@nic}")
       end
-      return if self.slaves.member?(slave)
-      if current_master = slave.master()
+      return if slaves.member?(slave)
+      if current_master = slave.master
         current_master.remove_slave(slave)
       end
       slave.up
@@ -766,7 +766,7 @@ class ::Nic
     end
 
     def up
-      slaves.each{ |s|s.up }
+      slaves.each(&:up)
       super
     end
 
--- a/chef/cookbooks/barclamp/libraries/nic.rb
+++ b/chef/cookbooks/barclamp/libraries/nic.rb
@@ -729,8 +729,7 @@ class ::Nic
     end
   end
 
-  # Base class for an ovs-bridge. This is just enough to be able to
-  # implement the remove_slave call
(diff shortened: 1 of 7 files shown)


rhafer:
FYI. The mkcloud gating for this is currently failing because of: https://github.com/openSUSE/obs-build/pull/198

chef/cookbooks/barclamp/libraries/nic.rb
```diff
@@ -751,6 +766,51 @@ def remove_slave(slave)
...
+        remove_slave(slave)
+      end
+      super
+      ::Kernel.system("ovs-vsctl del-br #{@nic}")
+      nil
+    end
+
+    def replug(slave)
+      slave = self.class.coerce(slave)
+      unless self.slaves.member?(slave)
```
rsalevsky: I think we can ignore this.

chef/data_bags/crowbar/bc-template-network.schema
```diff
@@ -82,6 +82,8 @@
                     "vlan": { "type": "int", "required": true },
                     "use_vlan": { "type": "bool", "required": true },
                     "add_bridge": { "type": "bool", "required": true },
+                    "add_ovs_bridge": { "type": "bool", "required": false },
+                    "bridge_name": { "type": "str", "required": false },
```
rhafer: Note: I defined those to attribute to be optional because they should only ever be needed for the "nova_floating" and "nova_fixed" networks (and any addtional external network that might get added). I'd like to avoid them being added to all the other networks ("admin", "os_sdn", ...). IMO that would just cause confusion.

chef/data_bags/crowbar/migrate/network/011_add_ovs_attributes.rb
```diff
@@ -0,0 +1,43 @@
+def upgrade(ta, td, a, d)
+  a["network"]["nova_floating"]["add_ovs_bridge"] = ta["network"]["nova_floating"]["add_ovs_bridge"]
+  a["network"]["nova_floating"]["bridge_name"] = ta["network"]["nova_floating"]["bridge_name"]
+  a["network"]["nova_fixed"]["add_ovs_bridge"] = ta["network"]["nova_fixed"]["add_ovs_bridge"]
+  a["network"]["nova_fixed"]["bridge_name"] = ta["network"]["nova_fixed"]["bridge_name"]
+
+  # We need to switch 'add_ovs_bridge' to true on nodes that have the
+  # "neutron-network" role and use the 'openvswitch' mechanism driver
+  nodes = NodeObject.find("roles:neutron-network")
+  nodes.each do |node|
```
rhafer: Actually it might be better to move this part of the migration to the neutron barclamp. If only to avoid the duplication with https://github.com/crowbar/crowbar-openstack/pull/23/files#diff-1cab7f3b6842279ae4d6bd875e073cacR324

rhafer:
I just removed the [WIP] tag from this. I think this is basically complete now.

@vuntz, @aplanas I would really like you to have a closer look at this since you worked on the first iteration of it (https://github.com/crowbar/crowbar-openstack/pull/3).
See https://github.com/crowbar/crowbar-openstack/pull/23 on how this is supposed to be used.

chef/cookbooks/barclamp/libraries/nic.rb
tboerger: Why? Just remove `self.`

chef/cookbooks/barclamp/libraries/nic.rb
```diff
@@ -743,6 +744,20 @@ def slaves
       ports.map{ |i| ::Nic.new(i) }
     end
 
+    def add_slave(slave)
+      slave = self.class.coerce(slave)
+      unless ::Nic.exists?(slave)
```
vuntz: If we manage to coerce, then the slave does exist. I think it makes more sense to check for existence first, and then coerce (that's what we do in the Bond class)

chef/cookbooks/barclamp/libraries/nic.rb
```diff
@@ -743,6 +744,20 @@ def slaves
...
+        raise ::ArgumentError.new("#{slave} does not exist, cannot add to bridge#{@nic}")
+      end
+      return if slaves.member?(slave)
+      if current_master = slave.master
+        current_master.remove_slave(slave)
+      end
+      slave.up
+      usurp(slave)
+      ::Kernel.system("ovs-vsctl add-port #{@nic} #{slave}")
```
vuntz: It's worth mentioning that the Bond class actually returns the slave from this method (also in the `return if` bit a few lines above).

chef/cookbooks/barclamp/libraries/nic.rb
```diff
@@ -751,6 +766,50 @@ def remove_slave(slave)
...
+    end
+
+    def self.create(nic, slaves = [])
+      Chef::Log.info("Creating new OVS bridge #{nic}")
+      if self.exists?(nic)
+        raise ::ArgumentError.new("#{nic} already exists.")
+      end
+      ::Kernel.system("ovs-vsctl add-br #{nic}")
+      5.times do
```
vuntz: Can we maybe have a comment explaining why we try several times?

chef/cookbooks/network/recipes/default.rb
```diff
@@ -249,6 +239,63 @@ def kill_nic(nic)
...
+      ::Kernel.system("modprobe #{node[:network][:ovs_module]}")
+    end
+
+    s = service node[:network][:ovs_service] do
+      service_name node[:network][:ovs_service]
+      supports status: true, restart: true
+      action [:nothing]
+    end
+    s.run_action :enable
+    s.run_action :start
```
vuntz: Won't chef complain that we have resources with the same name (since we might run this for br-fixed and br-public, so we define the same package and same service resources twice)?

chef/cookbooks/network/recipes/default.rb
```diff
@@ -249,6 +239,63 @@ def kill_nic(nic)
...
+    our_kids.each do |k|
+      next unless Nic.vlan?(k)
+      ovs_master = k.ovs_master
+      unless ovs_master.nil?
+        Chef::Log.warn("Replugging #{k.name} to #{ovs_master.name} (workaround bnc#945219)")
+        ovs_master.replug(k.name)
+      end
+    end
+    ifs[br.name]["slaves"] = [our_iface.name]
+    ifs[br.name]["type"] = "bridge"
```
vuntz: Can you move these two lines with the other `ifs[br.name]` lines? Or was there a good reason to have them here? (Possibly waiting for the add_slave, but if it fails, then the fact that we had these lines earlier won't matter due to the exception that will be raised)

vuntz: Btw, why `bridge` and not `ovs_bridge`?

vuntz: > Can you move these two lines with the other ifs[br.name] lines? Or was there a good reason to have them here? (Possibly waiting for the add_slave, but if it fails, then the fact that we had these lines earlier won't matter due to the exception that will be raised)

Ah, you did the same as what was done for bridge. Fair enough.

vuntz:
Good stuff. I have a few comments, but nothing critical.

chef/cookbooks/barclamp/libraries/nic.rb
rhafer: True. It seems the Bridge class does this the wrong way around as well.

chef/cookbooks/barclamp/libraries/nic.rb
rhafer: Also true. I'll align these methods (Bridge.add_slave also does it different to Bond) so that they behave similar. It doesn't seem to matter much currently as the return value is not used anywhere at the moment.

chef/cookbooks/barclamp/libraries/nic.rb
rhafer: I am not entirely sure (I just copied this from the Bridge class), but my assumption was that it is just waiting for the sysfs file for the bridge to show up. It's just the ```self.exists?(nic) && self.ovs_bridge?(nic)``` that is executed multiple times. The code would be racey otherwise. 

I'll add a comment.

chef/cookbooks/network/recipes/default.rb
rhafer: Yeah, right. And I actually had code already to only call it once. Most have lost it during a rebase or something. :(

chef/cookbooks/network/recipes/default.rb
rhafer: > Btw, why bridge and not ovs_bridge?

It seems ```type``` is only used in the config template for debian/ubuntu. And that doesn't currently handle ovs bridges at all.

rhafer:
@vuntz thanks for the feedback. I think I've address all the issues you raised. Though I didn't test the last changes yet.

chef/cookbooks/network/recipes/default.rb
vuntz: Hrm, I think I'd still use `ovs_bridge` because I don't think the settings used in the debian/ubuntu template all apply for an ovs bridge. This can eventually be fixed when someone goes on and fix the debian/ubuntu support.

vuntz:
+1

aplanas:
+1

PR #35, opened 2015-09-14
prevent crowbar batch from locking itself and the user out

jsuchome:
Originally https://github.com/crowbar/barclamp-crowbar/pull/1336


aspiers:
+1 and this already got two +1s as https://github.com/crowbar/crowbar-core/pull/35

vuntz:
+1

PR #38, opened 2015-09-15
fix error responses from REST API

aspiers:
[N.B. There could well be a good reason why `ProxyErrorOverride On` was included.  I can't think of any, but if there is, this PR will need some more discussion.]

Sometimes Crowbar returns JSON-formatted errors from the REST API.  One easily reproducible example is to attempt to allocate a node which is already allocated.  This is what should happen:

    root@crowbar:~ # crowbar machines -d allocate d52-54-01-77-77-01.c19.cloud.suse.de

    [... snipped authentication stuff ...]

    DEBUG: (post) hostname: 127.0.0.1:80
    DEBUG: (post) request: /crowbar/machines/1.0/allocate/d52-54-01-77-77-01.c19.cloud.suse.de
    DEBUG: (post) data: {}
    DEBUG: (post) return code: 422
    DEBUG: (post) return body: {"error":"Node already allocated"}
    Failed to talk to service allocate: Node already allocated (422)

However https://github.com/crowbar/crowbar-core/pull/20 put Crowbar behind an Apache proxy, and one of the directives introduced was `ProxyErrorOverride On` which intercepts all errors and replaces the response body with its own documents as directed by ErrorDocument directives.  This caused the responses to be a) HTML, which confused the client-side JSON parser, and b) uninformative, because the original error (`Node already allocated`) never reaches the user.

Removing the `ProxyErrorOverride On` directive fixes this.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/apache.conf.erb
+++ b/chef/cookbooks/crowbar/templates/default/apache.conf.erb
@@ -11,7 +11,6 @@
 
     ProxyPass / http://127.0.0.1:<%= @port %>/ timeout=600 Keepalive=On
     ProxyPassReverse / http://127.0.0.1:<%= @port %>/
-    ProxyErrorOverride On
 
     <Location />
         <%- unless @realm.nil? %>


aspiers:
Gate failed due to some signing issue with the cloud media.  I've emailed `cloud-devel` for suggestions why.

tboerger:
@rsalevsky why have you added `ProxyErrorOverride`?

rsalevsky:
@tboerger this was based on your commit https://github.com/crowbar/crowbar-core/pull/3/files#diff-d8e60db934822ee5f78b3bce37f7614aR14

vuntz:
+1

aspiers:
Discussed gating failure with @dirkmueller and @toabctl and we can't explain the failure, so I've forced a rebuild to see if we can reproduce.

aspiers:
Another unrelated gate failure.

```
00:40:17 + curl -L -m 40 -s -S -k http://d52-54-01-77-77-01.c1.cloud.suse.de
00:40:17 + grep -q -e csrfmiddlewaretoken -e '<title>302 Found</title>'
00:40:57 curl: (28) Operation timed out after 29395 milliseconds with 0 bytes received
00:40:57 + complain 101 'simple horizon test failed'
```

aspiers:
This time it failed with

    ERROR tempest.api.compute.base [-] Waiting for deletion of server 83c15ee4-1911-4fa5-a832-67cbaf2686c0 failed

which also seems unrelated.  Sheesh!  Have added to [cloud-gating etherpad](https://etherpad.nue.suse.com/p/cloud-gating).  Forcing **another** rebuild :-/

aspiers:
@jsuchome Once you have tested this please review and give a +1 :-)

jsuchome:
+1 This indeed seems to help with my crowbar_batch deployment

tboerger:
I can't remember the reason why I have added it. If it works without that I'm fine with dropping the line.

vuntz:
Gating is finally green.

PR #41, opened 2015-09-17
fix windows install by using IPv4 tftp symlinks

bmwiedemann:
In the Hyper-V case we use name-remapping for the bootmgr.exe
    using the \x expanding to the hexadecimal version of the IP-address.
    With IPv6 now being used (even when connecting from a IPv4 addr),
    this address and thus the remapped name is different than in Cloud5
    and thus did not find our symlinks.

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/tftp.erb
+++ b/chef/cookbooks/provisioner/templates/default/tftp.erb
@@ -8,7 +8,7 @@ service tftp
        socket_type             = dgram
        protocol                = udp
        wait                    = yes
-       flags                   = IPv6 IPv4
+       flags                   = IPv4
        user                    = root
        server                  = /usr/sbin/in.tftpd
        server_args             = -m /etc/tftpd.conf -s <%=@tftproot%>


ovisan:
:+1:

vuntz:
See https://github.com/crowbar/barclamp-provisioner/pull/446 and https://github.com/crowbar/barclamp-provisioner/pull/447 (still, I'm fine with the user change)

chef/cookbooks/provisioner/templates/default/tftp.erb
```diff
@@ -8,8 +8,8 @@ service tftp
        socket_type             = dgram
        protocol                = udp
        wait                    = yes
-       flags                   = IPv6 IPv4
+       flags                   = IPv4
```
vuntz: How does it avoid the memory corruption? It's not really explained in the bug.

dirkmueller:
that breaks it if the change gets backed out..

chef/cookbooks/provisioner/templates/default/tftp.erb
bmwiedemann: Bug now has more info. There is a memory overflow in tftpd because IPv4 addrs are mapped to IPv6-addrs like ::ffff:123.123.123.123 which are much longer when our \x expansion is applied for hyper-v booting

bmwiedemann:
dropped the user change and added more info in the commit message

vuntz:
Oh, so I guess this also fixes things where the symlinks are actually wrong.

+1

mjura:
LGTM

chef/cookbooks/provisioner/templates/default/tftp.erb
dirkmueller: looks like the bug is fixed already: https://build.opensuse.org/request/show/332002

dirkmueller:
fine by me, but I'd rather take the fix in tftp (which is already fixed apparently by the package maintainer)

bmwiedemann:
updated the commit message, as the change did not actually avoid the memory corruption, but we still need it

dirkmueller:
:+1:

bmwiedemann:
btw: this is not needed on Cloud5 / SLE11 - maybe xinetd does not map IPv4 to IPv6 addrs.

PR #44, opened 2015-09-18
Minor cleaning in the locking code

vuntz:
Use the reader instead of instance variable whenever possible.

cc @aspiers

### DIFF
--- a/crowbar_framework/lib/crowbar/lock/local_blocking.rb
+++ b/crowbar_framework/lib/crowbar/lock/local_blocking.rb
@@ -31,18 +31,18 @@ module Crowbar
       loop do
         count += 1
         logger.debug("Lock #{path} attempt #{count}")
-        if file.flock(File::LOCK_EX | File::LOCK_NB)
+        if @file.flock(File::LOCK_EX | File::LOCK_NB)
           break
         end
         sleep 1
       end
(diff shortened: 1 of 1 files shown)


aspiers:
I'm in favour of the increased consistency, although I would have maybe gone the other way and switched everything to `@file`.  The accessor is more for external usage.  What do you think?

vuntz:
@aspiers: done. As long as it's consistent one way or the other, I'm happy :-)

aspiers:
+1 for the change but the gate is failing.  I have suggested a reason in the new https://etherpad.nue.suse.com/p/cloud-gating (and already I am annoyed that we didn't try this experiment with Trello which would have allowed deeplinking ...)

vuntz:
@aspiers it's failing because the code it's patching is not yet in the package...

aspiers:
@vuntz Damn, that sounds like a flaw in our gating process.

vuntz:
Anyone with another +1?

toabctl:
+1

tboerger:
I usually prefer the reader overall.

aspiers:
@tboerger That was my first instinct too, but then I thought that assigning to `@file` and reading from it via `#file` within the same method was a bit unnecessary obfuscated so I changed my mind.  Although if I'm missing something I could easily be persuaded back :-)

PR #47, opened 2015-09-23
Raise gem version to new package versions

rsalevsky:


tboerger:
You should raise the `Gemfile` versions as well.

dirkmueller:
also chef/cookbooks/crowbar/recipes/default.rb needs to be adjusted

tboerger:
LGTM

dirkmueller:
+1

vuntz:
Do you intend to update also other gems in this PR, or is this  a later step? (hashie, for instance)

rsalevsky:
I think we should not mix the rails 4.2 upgrade and the gem version upgrade to SP1. So I prefer to open another PR for this.

vuntz:
+1

vuntz:
@rsalevsky we can merge (as, well, we have the patch in our package anyway now); were you planning anything label, or can we drop the "do not merge yet" label and merge?

dirkmueller:
finished testing, merging.

PR #48, opened 2015-09-23
Move home of crowbar user to /var/lib/crowbar

vuntz:


rhafer:
Makes sense. +1

chef/cookbooks/crowbar/files/default/knife.rb
```diff
@@ -1,2 +0,0 @@
-node_name "chef-webui"
```
aspiers: -1, this is still used by `cookbook_file "/root/.chef/knife.rb"` on line 178 of the `crowbar::default` recipe, which caused the gate to break.

jsuchome:
+1

tboerger:
And what happens to the knife config now?

vuntz:
> And what happens to the knife config now?

Does the crowbar user use knife?

tboerger:
> Does the crowbar user use knife?

Maybe just the administrator that signs in as crowbar user on the shell

vuntz:
> Maybe just the administrator that signs in as crowbar user on the shell

The crowbar user has /sbin/nologin as shell, so no :-)

tboerger:
Than it's fine :)
:+1:

PR #52, opened 2015-09-30
Fix/workaround node-role revision races

rhafer:
This will reduce the race window that exists when multiple nodes transition to
"ready" at the same time and trigger the processing for the proposal queue.
This will NOT really fix the underlying issue for the race.

The following actions lead to the possible race:

Preconditions:
 - 2 Nodes are allocated at roughly the same time (e.g. using bulk edit)
 - While the nodes are being installed proposal for some barclamps are created
   and at least two independent are assigned to one of the above nodes. (e.g
   database and rabbitmq). The proposals are queued for deployment.

Now:
 - Node1 transitions to "ready"
 - that triggers processing of the proposal queue
 - one of the queued proposals is unqueued
 - that causes apply_role to be called
 - apply_role tries to lock the node (which takes some time), the node state is
   not yet updated to "applying"
 - during that Node2 transitions to "ready"
 - that triggers processing of the proposal queue again
 - the queued proposals is unqueued (for Node1(!) because it's state is still
   "ready"
 - apply_role is called again (in a different crowbar process). <- this should
   NEVER happen, even while it won't cause chef-client to be started twice on
   the same node it will lead to various kinds for revisions races on the node
   role.

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1044,6 +1044,10 @@ class ServiceObject
 
     applying_nodes = run_order.flatten.uniq.sort
 
+    # Mark nodes as applying; beware that all_nodes do not contain nodes that
+    # are actually removed.
+    set_to_applying(applying_nodes, inst, pre_cached_nodes)
+
     # Prevent any intervallic runs from running whilst we apply the
     # proposal, in order to avoid the orchestration problems described
     # in https://bugzilla.suse.com/show_bug.cgi?id=857375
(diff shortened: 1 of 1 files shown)


vuntz:
+1

toabctl:
Commit message has 2x should (```should should```)

rhafer:
> Commit message has 2x should (should should)

Fixed.

dirkmueller:
+1

vuntz:
Still +1

vuntz:
@dirkmueller is your +1 still valid with the additional patch?

aspiers:
> apply_role is called again (in a different crowbar process). <- this should NEVER happen, even while it won't cause chef-client to be started twice on the same node it will lead to various kinds for revisions races on the node role.

FWIW the new shared locking I added specifically allowed for (and even expected) `apply_role` to be called from different threads concurrently.  So at least I didn't add to the problem, but only made existing race windows easier to find ;-)

dirkmueller:
yep, still +1

PR #55, opened 2015-10-02
Bump requirements to be in line with SLE12 SP1

dirkmueller:


Gemfile
```diff
@@ -19,9 +19,9 @@ source "https://rubygems.org"
 
 group :development do
   gem "closure-compiler", "~> 1.1.10"
-  gem "sass", "~> 3.2.19"
+  gem "sass", "~> 3.4.13"
   gem "sprockets-standalone", "~> 1.2.1"
-  gem "sprockets", "~> 2.11.0"
+  gem "sprockets", "~> 3.1.0"
```
tboerger: As I can see you have not checked if it still works. The dependencies can't get resolved.
Reference: https://rubygems.org/gems/sprockets-standalone

rsalevsky: I think we should drop `sprockets-standalone` completely.

dirkmueller: right, standalone requires 2.11. I'm happy if we could drop that

toabctl:
+1

crowbar_framework/config/boot.rb
```diff
@@ -55,7 +55,7 @@
   gem "dotenv", version: "~> 1.0"
   require "dotenv"
 
-  gem "hashie", version: "~> 2.1"
+  gem "hashie", version: "~> 3.4"
```
aplanas: In another PR you suggested to use patch level versioning. Still makes sense here?

rsalevsky:
+1

vuntz:
What about js-routes?

crowbar_framework/config/boot.rb
dirkmueller: yeah, I still think, but this patch is more urgent than the other one. I expect the other one to be discussed for a couple of months..

dirkmueller:
@vuntz fixed

aplanas:
+1

vuntz:
+1

Gemfile
tboerger: This can get dropped on crowbar-core, but not from any other component. So we will still have this issue.

PR #56, opened 2015-10-02
Sync config/boot.rb version specifications with Gemfile requirements

dirkmueller:
Assuming that rubygems follow semantic versioning guidelines,
we want to lock on patch level releases. It seems useful to me
to be consistent with the versioning in the Gemfile and this
part of crowbar to ensure that production and dev environments
always have exactly the same rubygem dependency restrictions.

Rebase of https://github.com/crowbar/barclamp-crowbar/pull/1324

### DIFF
--- a/crowbar_framework/config/boot.rb
+++ b/crowbar_framework/config/boot.rb
@@ -27,62 +27,62 @@ if File.exists? ENV["BUNDLE_GEMFILE"]
   Bundler.require(:default, Rails.env)
 else
   # rails related
-  gem "rails", version: "~> 4.2"
+  gem "rails", version: "~> 4.2.2"
   require "rails/all"
 
-  gem "haml-rails", version: "~> 0.9"
+  gem "haml-rails", version: "~> 0.9.0"
   require "haml-rails"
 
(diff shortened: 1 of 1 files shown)


vuntz:
Revisiting this: I'm +1 (this needs to be rebased, of course).

aspiers:
Actually this is syncing `config/boot.rb` with `Gemfile`, not the other way around.  Can you change the commit/PR titles accordingly?

aspiers:
And +1 with that change.

aplanas:
This is going to be fun to maintain : )

+1

tboerger:
> This is going to be fun to maintain : )
> 
> +1

If we switch to rails engines and gem based distribution it will be only a single place to change ;)

tboerger:
Gating failed, just to say it...

aspiers:
Looks like it hung at trove, which means the failure was almost certainly nothing to do with this PR.

tboerger:
> Looks like it hung at trove, which means the failure was almost certainly nothing to do with this PR.

But still the convention no merge of red gating ;)

PR #57, opened 2015-10-02
Replace old status LED's

rsalevsky:
This change is based on the changes in our SUSE design branch.


aplanas:
+1

Are there more changes in the queue? If so maybe we can add all together

vuntz:
-1, some of the images are not updated, which will result in very weird behavior.

vuntz:
Missing one is yellow-green-spin.png

rsalevsky:
@vuntz Thanks, missed that one.

@aplanas Yep, but I prefer to put every change in one PR because I lost the overview completely.

aplanas:
> @aplanas Yep, but I prefer to put every change in one PR because I lost the overview in the design branch completely.

But this will put UI in a very funny stages if we merge it out of order.

rsalevsky:
Nop, I started already with grouping them so they are independently.

vuntz:
+1

PR #58, opened 2015-10-02
Change global link color

rsalevsky:
This change is based on the SUSE branding branch where we changed the
link color to blue. The blue is nearly the same as the default crowbar
blue and the changes also matches the default crowbar design.


vuntz:
+1

aplanas:
+1

Is the queue very long? I really do no feel comfortable with one PR for every CSS change, when a single PR with multiple commits can do the job

PR #60, opened 2015-10-04
Only install suse-openstack-cloud-release package on SLES 12.1

dirkmueller:
In case we need to install a SLES 12.0 node (for SES2 :-(), don't
try to install additional software on it.

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
@@ -278,10 +278,10 @@ sync
 <% else -%>
   <% if @target_platform_version == "11.3" -%>
       <package>suse-sle11-openstack-cloud-release</package>
-  <% else -%>
+  <% elsif @target_platform_version == "12.1" -%>
       <package>suse-openstack-cloud-release</package>
-  <% end -%>
       <package>supportutils-plugin-suse-openstack-cloud</package>
+  <% end -%>
 <% end -%>
 <% @packages.each do |package| -%>
       <package><%= package %></package>
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -299,7 +299,7 @@ PACKAGES_INSTALL="$PACKAGES_INSTALL openssh"
 
 # From autoyast profile
 PATTERNS_INSTALL="$PATTERNS_INSTALL Minimal base"
-PACKAGES_INSTALL="$PACKAGES_INSTALL netcat ruby2.1-rubygem-chef supportutils-plugin-suse-openstack-cloud"
+PACKAGES_INSTALL="$PACKAGES_INSTALL netcat ruby2.1-rubygem-chef"
 
 case $ARCH in
     x86_64) PACKAGES_INSTALL+=" biosdevname";;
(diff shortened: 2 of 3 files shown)


toabctl:
+1

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
```diff
@@ -278,10 +278,11 @@ sync
 <% else -%>
   <% if @target_platform_version == "11.3" -%>
       <package>suse-sle11-openstack-cloud-release</package>
-  <% else -%>
-      <package>suse-openstack-cloud-release</package>
   <% end -%>
+  <% if @target_platform_version == "12.1" -%>
```
rsalevsky: `elsif` should be better in this case

vuntz:
Hrm, just realized: don't we have the issue of ruby2.1-rubygem-chef? (which would be a deal breaker)

dirkmueller:
@vuntz SES2 contains chef right now (I don't know why)

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
```diff
@@ -278,10 +278,10 @@ sync
 <% else -%>
   <% if @target_platform_version == "11.3" -%>
       <package>suse-sle11-openstack-cloud-release</package>
```
vuntz: Strictly speaking, we'd need the supportutils package here too (but I'd be equally fine removing 11.3 from there)

vuntz:
Can you also update the crowbar_register template, since I expect we will have the same issue for packages (and maybe ntp)?

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
dirkmueller: depends on if we'll ever have a 11.3 compute nodes..

dirkmueller:
@vuntz done

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
vuntz: Well, there's the idea of potentially getting 11.4 compute nodes later on, and having the code here would be a reminder to update it. But again, I'm fine removing it too :-)

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -309,8 +309,8 @@ zypper --non-interactive install -t pattern $PATTERNS_INSTALL
 # Auto-agree with the license since it was already agreed on for the admin server
 <% if @target_platform_version.to_f < 12.0 -%>
 zypper --non-interactive install --auto-agree-with-licenses suse-sle11-openstack-cloud-release
-<% else -%>
-zypper --non-interactive install --auto-agree-with-licenses suse-openstack-cloud-release
+<% elsif @target_platform_version == "12.1" -%>
+zypper --non-interactive install --auto-agree-with-licenses suse-openstack-cloud-release supportutils-plugin-suse-openstack-cloud
```
vuntz: Hrm, you forgot to remove the package from line 302?

dirkmueller: yes, fixed.

rhafer:
+1

PR #62, opened 2015-10-05
Fix crowbar production.log ownership

toabctl:
When install-chef-suse.sh fails for some reason and the script
is reexecuted, /var/log/crowbar/production.log can be owned by
root:root which prevents crowbar from executing the db:migrate rake
task:

rake aborted!
Permission denied @ rb_sysopen - /var/log/crowbar/production.log
[snipped backtrace]
Tasks: TOP => db:migrate => environment

chef-client runs before the rake task execution so setting the owner
in the crowbar recipe fixes the problem.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/default.rb
+++ b/chef/cookbooks/crowbar/recipes/default.rb
@@ -226,6 +226,13 @@ directory logdir do
   action :create
 end
 
+file "#{logdir}/production.log" do
+  owner "crowbar"
+  group "crowbar"
+  mode "0644"
+  action :create
+end
+
 directory "#{logdir}/chef-client" do
   owner "crowbar"
   group "crowbar"


toabctl:
FYI: This is reproducable when hitting https://bugzilla.suse.com/show_bug.cgi?id=948717 and then executing on the admin node ```screen  /opt/dell/bin/install-chef-suse.sh  --verbose```

dirkmueller:
codewise LGTM. I wonder though if that indicates that we're running something as root that should be run by the crowbar user instead. perhaps it is this ?

```
# Create the secrets env
cat<<EOF > /opt/dell/crowbar_framework/config/secrets.env
export SECRET_KEY_BASE=$(cd /opt/dell/crowbar_framework && RAILS_ENV=production ./bin/rake secret)
EOF
```

rsalevsky:
I think we should find the real bug instead of fixing a symptom.

toabctl:
> codewise LGTM. I wonder though if that indicates that we're running something as root that should be run by the crowbar user instead. perhaps it is this ?

That code doesn't create production.log or changes ownership. Must be something different.

toabctl:
So the production.log file is created somehow during ```/opt/dell/bin/barclamp_install.rb $BARCLAMP_INSTALL_OPTS $required_components```

toabctl:
It's also not the code in the migration (```migrate_cmd_su = "su -s /bin/sh - crowbar sh -c \"#{migrate_cmd}\" >> #{log}"```). This code is not executed. My guess is that it's something with the ```active_support``` usage?! But I already spend to much time on this so my suggestion is to merge it as it is. It's a workaround but doesn't hurt and isn't ugly I think.

dirkmueller:
I was wondering whether we could use chattr +i production.log to figure out where it crashes and trace the user of that log file that way.. 

I'm not against merging this, but I'd rather find the real problem ..

vuntz:
I tried to reproduce, but was not able to do so.

What's inside /var/log/crowbar/production.log when this happens?

toabctl:
> What's inside /var/log/crowbar/production.log when this happens?

Just one line:
```
# Logfile created on 2015-10-09 09:56:57 +0000 by logger.rb/44203
```

dirkmueller:
didn't you say already it would be barclamp_install.rb ? can you try running that as crowbar user ? does that work?

toabctl:
> didn't you say already it would be barclamp_install.rb ? can you try running that as crowbar user ? does that work?

No. Running into multiple permission problems then.

toabctl:
> I tried to reproduce, but was not able to do so.

What works for me:
* ./mkcloud cleanup prepare setupadmin prepareinstcrowbar
* ssh to the admin node and change ```ProxyPass``` to ```Proxy123Pass``` in /opt/dell/chef/cookbooks/crowbar/templates/default/apache.conf.erb (to trigger the apache a2enmod bug for sure)
*  ./mkcloud instcrowbar
* now ```production.log``` has root:root as u/g
* then after the failure rerun on the admin node: ```screen /opt/dell/bin/install-chef-suse.sh --verbose```

This works for me 100% to reproduce the bug.

vuntz:
https://github.com/SUSE-Cloud/supportutils-plugin-suse-openstack-cloud/pull/50 fixes the issue for me

toabctl:
See https://github.com/crowbar/crowbar-core/pull/62#issuecomment-146903728

PR #63, opened 2015-10-06
Update bootstrap to 3.3.5

rsalevsky:


rsalevsky:
Tested crowbar after the upgrade, looks good so far didn't seen any problem in the UI.

aplanas:
Only for curiosity, why bootstraps, jquery plugins, and the rest are not installed as packages?

Also, why we include the full source of boostrap?

rsalevsky:
But it in packages is not possible because we want to run crowbar out of git. This is required for the development setup.

We can use the gems but there was also a reason why we don't use it. @tboerger Can explain why.

tboerger:
> Only for curiosity, why bootstraps, jquery plugins, and the rest are not installed as packages?
> 
> Also, why we include the full source of boostrap?

Sure, we can do that if you wants to maintain these gems. I think it's quite difficult to get somebody who wants to takeover the maintenance if it can get done that easy within the source.

tboerger:
> Also, why we include the full source of boostrap?

I don't know what you mean with that.

tboerger:
`crowbar_framework/app/assets/javascripts/bootstrap.js` should be an empty file, as it's already done similar by `bootstrap-sprockets`. I would take this file only for bootstrap extensions that will be used for sure soon.

crowbar_framework/app/assets/javascripts/jquery.js
```diff
@@ -16,3 +16,4 @@
 //= require jquery/dynamicTable
 //= require jquery/changeDetection
 //= require jquery/hideShowText
+//= require jquery/migrate
```
tboerger: I think that should be the first loaded plugin.

crowbar_framework/app/assets/stylesheets/bootstrap.scss
```diff
@@ -1,7 +1,19 @@
+/*!
```
tboerger: Move all this content into `bootstrap-sprockets` and leave this file for bootstrap extensions.

tboerger:
I don't know if i have missed something. As you have put everything into a single commit and github got a limitation for the diff view I can more or less just guess if that is fine.

rsalevsky:
@tboerger Fixed all you comments.

tboerger:
I think that looks good so far!

rsalevsky:
@vuntz Can we merge this PR?

dirkmueller:
+1

PR #65, opened 2015-10-07
add netwait=60 (bsc#927997)

bmwiedemann:
this is now also needed for SLES12-based sleshammer images
https://bugzilla.suse.com/show_bug.cgi?id=927997

(and maybe it can help avoid an observed race where node1 comes up without network and thus cannot be discovered)

### DIFF
--- a/chef/cookbooks/provisioner/recipes/setup_base_images.rb
+++ b/chef/cookbooks/provisioner/recipes/setup_base_images.rb
@@ -44,6 +44,8 @@ end
 if crowbar_key != ""
   append_line += " crowbar.install.key=#{crowbar_key}"
 end
+# workaround broken IPMI/BMC firmwares not transmitting traffic for 35s (bsc#927997)
+append_line += " netwait=60"
 append_line = append_line.split.join(" ")
 node.set[:provisioner][:sledgehammer_append_line] = append_line
 


aspiers:
Isn't this going to significantly slow down every single autoyast-based deployment?  If so, -1 unless there really is no other way to solve this.

vuntz:
Can you add a comment before the line to explain why we need it?

bmwiedemann:
This change is about sleshammer. autoyast already uses the same netwait=60 since cloud5 and it is not going to slow things down, because it is just the maximum wait at which point it gives up and you have a broken cloud.

dirkmueller:
@aspiers nonsense, thats just the maximum timeout it waits, not the minimum timeout (default is 1s or so which is far too low).

dirkmueller:
but sleshammer has this code:

```bash

if suse_ver 12; then
    /usr/lib/wicked/bin/wickedd-dhcp4 --test --test-output /tmp/wicked-dhcp-eth0 --test-timeout 60 $BOOTDEV

```

it doesn't check for netwait being set on /proc/cmdline. it probably could though.

aspiers:
@dirkmueller It's not nonsense, it was a valid question not a statement of fact.

bmwiedemann:
added a comment to explain the line

aspiers:
Thanks, +1

tboerger:
:+1:

rsalevsky:
+1 Maybe change bnc to bsc.

bmwiedemann:
nay, bnc is the traditional canonical form and using different variants of it makes life harder

aspiers:
Not relevant to this PR, but +1 for use of bsc over bnc - it's really not hard to adapt our tools and workflows to support it and there are useful benefits since https://bugzilla.suse.com actually behaves differently to https://bugzilla.novell.com.  Plus it's not nice to stick with obsolete references to Novell - not just an aesthetic preference but it can be confusing for newcomers.

vuntz:
As a sidenote: everyone started using bsc, and most tools (if not all) have been updated to support that.

bmwiedemann:
updated to bsc

chef/cookbooks/provisioner/recipes/setup_base_images.rb
```diff
@@ -59,6 +59,8 @@
 if crowbar_key != ""
   append_line += " crowbar.install.key=#{crowbar_key}"
 end
+# workaround broken IPMI/BMC firmwares not transmitting traffic for 35s (bsc#927997)
+append_line += " netwait=60"
```
bmwiedemann: :-1: there is still some problem here, as in testing, it had many netwait entries in the resulting file - one from each chef-client run.

tboerger:
Needs rebase

bmwiedemann:
rebased

PR #53, opened 2015-10-01
replace checks for platform with checks for platform_family

jsuchome:

PR #66, opened 2015-10-07
Fixes for recent ovs bridge related changes

rhafer:


chef/cookbooks/network/recipes/default.rb
```diff
@@ -509,4 +511,18 @@ def kill_nic(nic)
...
+  # should be correctly configure at this point.
+  template "/etc/wicked/local.conf" do
+    source "wicked-local.conf.erb"
+    owner "root"
+    group "root"
+    mode "0644"
+    variables(
+      use_nanny: ovs_setup_once
+    )
+  end
```
rhafer: Hm, I think I am missing an ```only_if``` condition here. And using ```ovs_setup_once``` once for ```use_nanny``` might also be wrong. I'll reconsider.

chef/cookbooks/network/recipes/default.rb
```diff
@@ -509,4 +511,18 @@ def kill_nic(nic)
...
+  # for the next boot.
+  # Note: There's no need to restart wicked here as all interfaces
+  # should be correctly configure at this point.
+  template "/etc/wicked/local.conf" do
+    source "wicked-local.conf.erb"
+    owner "root"
+    group "root"
+    mode "0644"
+    variables(
+      use_nanny: ovs_setup_once
```
vuntz: Just checking: there's no plan to restart wicked when that file changes, or is there? That was breaking HA when we had that.

rhafer: No, see the comment above the template resource. :)

vuntz: Heh, sorry, was reading too quickly :-)

rhafer:
Reworked. Should be ready now. Removed the wip label.

vuntz:
+1

rsalevsky:
+1

rhafer:
rebased

chef/cookbooks/network/recipes/default.rb
```diff
@@ -509,4 +511,26 @@ def kill_nic(nic)
...
+    # Note: There's no need to restart wicked here as all interfaces
+    # should be correctly configure at this point.
+    template "/etc/wicked/local.conf" do
+      source "wicked-local.conf.erb"
+      owner "root"
+      group "root"
+      mode "0644"
+      variables(
+        use_nanny: true
+      )
```
vuntz: Hrm, can you put the whole code inside a big if `node[:platform_family] == "suse" && !(node[:platform] == "suse" && node[:platform_version] < 12.0)` ?

rhafer: @vuntz Hm, we are already in the  ```when "suse"``` branch of ```case node["platform"]``` here. So I guess only the ```if !node[:platform_version] < 12.0``` would make sense here, right? Or probably better: ```node[:platform_version] >= 12.0```

vuntz: Ah, right. So the `case` will change to `case node[:platform_family]` soon (see https://github.com/crowbar/crowbar-core/pull/53), so the test should become `unless node[:platform] == "suse" && node[:platform_version] < 12.0` (to make it really SLE11-specific, and not considering openSUSE and others). Technically, just looking at the version would work, but I prefer to be slightly more explicit.

rhafer: Hm, that code is not SLES11 specific. It's for everything that has wicked. (SLE12 and newer, including openSUSE if we supported that)

vuntz: Hence the `unless` and not `if` :-)

vuntz:
Gating doesn't pass because the test package can't be found anymore; need to rebase or manually trigger a package rebuild.

chef/cookbooks/network/recipes/default.rb
rhafer: Addressed.

vuntz:
+1

mjura:
+1 looks OK

I'm wondering if this time gating will finish successfully

rhafer:
> I'm wondering if this time gating will finish successfully

Me too ;). Though none of the failures was related to this PR itself :/

vuntz:
You're reaching cct, so tempest is passing => good enough for merging.

PR #67, opened 2015-10-08
avoid case where bind9 was started without named.conf

aspiers:
Also fix outdated Chef resource syntax.

### DIFF
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -313,7 +313,7 @@ end
 
 service "bind9" do
   case node[:platform]
-  when "centos","redhat","suse"
+  when "centos", "redhat", "suse"
     service_name "named"
   end
   supports restart: true, status: true, reload: true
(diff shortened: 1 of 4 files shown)


aspiers:
Did this 2-3 weeks ago but forgot to submit it.

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -322,5 +311,14 @@ def make_zone(zone)
   notifies :restart, "service[bind9]", :immediately
```
vuntz: Just checking: this restart won't fail, right?

aspiers: You mean because the `bind9` service declaration appears below it?  Surely Chef's two-phase approach will prevent that?  I think I also tested but TBH I can't remember now.

vuntz: It should work fine (because you used the string way to reference the service), but wanted to double-check :-)

vuntz:
+1

tboerger:
LGTM

PR #68, opened 2015-10-08
[UNTESTED] only run ovs-vsctl if openvswitch service is running

aspiers:
(N.B. Here's a real world example of [a PR which needs an `untested` label but not `wip`](https://github.com/SUSE/cloud/pull/87#issuecomment-142867716).  This is a consequence of a debugging session with a real customer this morning, and I'm submitting this now before I forget it, before I have time to test it, and before the gate has had time to run.  It's cosmetic only so I can't prioritise testing it for quite a while.  But it's simple enough that I'm pretty sure it's finished.)

If openvswitch is installed, running `ovs-vsctl` when the openvswitch service is not running will result in an ugly error in syslog about not being able to connect to a socket in `/var/run`.  This can be misleading
and cause confusion during debugging, so let's avoid it.

### DIFF
--- a/chef/cookbooks/barclamp/libraries/nic.rb
+++ b/chef/cookbooks/barclamp/libraries/nic.rb
@@ -91,7 +91,8 @@ class ::Nic
   def self.ovs_bridge?(nic)
     nic.kind_of?(::Nic::OvsBridge) or
       # If no ovs tools are installed there's no ovs switch
-      if ::Kernel.system("which ovs-vsctl > /dev/null 2>&1")
+      if ::Kernel.system("which ovs-vsctl > /dev/null 2>&1 && " \
+                         "systemctl -q is-active openvswitch.service")
         ::Kernel.system("ovs-vsctl br-exists #{nic}")
       end
   end


chef/cookbooks/barclamp/libraries/nic.rb
```diff
@@ -91,7 +91,8 @@ def self.bridge?(nic)
   def self.ovs_bridge?(nic)
     nic.kind_of?(::Nic::OvsBridge) or
       # If no ovs tools are installed there's no ovs switch
-      if ::Kernel.system("which ovs-vsctl > /dev/null 2>&1")
+      if ::Kernel.system("which ovs-vsctl > /dev/null 2>&1 && " \
+                         "systemctl -q is-active openvswitch.service")
```
vuntz: What about systems where systemd is not used, or the service is called differently? :-)

aspiers: I already thought about that and decided that since this is the `master` branch, it is safe enough to assume systemd is running, and better to avoid making assumptions about implementation details (which something like `ps -ef | grep -q ovsdb-server` would do).

rhafer: BTW, there is an option to just stop ovs-vsctl from writing to syslog (which could be ok in this case):

```ovs-vsctl --verbose=vsctl:syslog:emer --verbose=reconnect:syslog:emer br-exists #{nic}```

Though I am undecided about which approach to prefer.

Comment (aspiers) [inline chef/cookbooks/barclamp/libraries/nic.rb]: Well, if the service isn't active, surely it's totally pointless to run `ovs-vsctl` and cleaner to skip it rather than rely on it to return the exit code we want when the expected behaviour in that case might be undefined.

File: chef/cookbooks/barclamp/libraries/nic.rb [hunk 2]
```diff
@@ -91,7 +91,8 @@ def self.bridge?(nic)
   def self.ovs_bridge?(nic)
     nic.kind_of?(::Nic::OvsBridge) or
       # If no ovs tools are installed there's no ovs switch
-      if ::Kernel.system("which ovs-vsctl > /dev/null 2>&1")
+      if ::Kernel.system("which ovs-vsctl > /dev/null 2>&1 && " \
```
tboerger: Is `which` still installed per default on SLE 12? If I remember correctly you should use `hash` instead of `which` as you are touching that code anyway.

aspiers: Wow.  Which genius decided that SLES 12 didn't need `which` by default?  It's used all over the place.  Do you have a URL of e.g. a mailing list thread or bug about this change?

aspiers: Hmm, looking closer into this, actually it's not used as often as I expected.  But it is still used in enough places that not installing by default is asking for trouble IMHO.

tboerger: I'm asking if it gets installed because I already had issues on minimalistic openSUSE and so on.

PR #69, opened 2015-10-08
Fix apache syntax error for crowbar vhost (bsc#948717)

toabctl:
Crowbar needs now mod_proxy (and some other related modules) but
using these modules via the chef cookbook only triggers a a2enmod
which writes the module to /etc/sysconfig/apache2.
A reload is not enough because the listed modules from the sysconfig
file are added as parameters to the command line when starting apache.

https://bugzilla.suse.com/show_bug.cgi?id=948717

### DIFF
--- a/chef/cookbooks/apache2/recipes/default.rb
+++ b/chef/cookbooks/apache2/recipes/default.rb
@@ -222,8 +222,3 @@ end
 
 # We don't need this.
 #apache_site "default" if platform?("centos", "redhat", "fedora")
-
-service "apache2" do
-  action :start
-  ignore_failure true
-end
--- a/chef/cookbooks/crowbar/recipes/default.rb
+++ b/chef/cookbooks/crowbar/recipes/default.rb
@@ -393,6 +393,10 @@ template "#{node[:apache][:dir]}/vhosts.d/crowbar.conf" do
   notifies :reload, resources(service: "apache2")
 end
 
+service "apache2" do
+  action [:enable, :start]
+end
+
 # The below code swiped from:
 # https://github.com/opscode-cookbooks/chef-server/blob/chef10/recipes/default.rb
 # It will automaticaly compact the couchdb database when it gets too large.


dirkmueller:
related bug report https://bugzilla.suse.com/show_bug.cgi?id=949434

chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -390,7 +390,7 @@
     realm: realm
   )
 
-  notifies :reload, resources(service: "apache2")
+  notifies :restart, resources(service: "apache2"), :immediately
```
vuntz: Do we need the `:immediately`?

Also wondering how this would impact things when applying the crowbar proposal later on (if it ever rewrites that config file). Restarting apache means killing the connection that is used for the "apply" process, and I'm not sure it's very good.

What about just including the `mod_*` recipes before the `apache2` one?

toabctl: > What about just including the mod_* recipes before the apache2 one?

Tried that but doesn't work. The mod recipe includes the apache recipe which starts apache.

toabctl: @vuntz We could also just add the needed modules to https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/apache2/recipes/default.rb#L205 (there's already a comment for suse). Downside is that we than have the proxy_* modules (and all others needed by crowbar) also for the dashboard.

toabctl:
I think with the fix from https://github.com/crowbar/crowbar-core/commit/6eb20f9ace8ada856589a551d47b9040796af9e4 we would also need to adjust the dashboard cookbook to start apache...

chef/cookbooks/apache2/recipes/default.rb
```diff
@@ -222,8 +222,3 @@
 
 # We don't need this.
 #apache_site "default" if platform?("centos", "redhat", "fedora")
-
-service "apache2" do
-  action :start
-  ignore_failure true
-end
```
vuntz: That sounds dangerous and could break horizon (which is using this cookbook)

toabctl: That's why I added https://github.com/crowbar/crowbar-core/pull/69#issuecomment-146556700 . The think is that I don't find a nice way to fix it as long as ```a2enmod``` is broken for a graceful apache restart

vuntz:
> I think with the fix from 6eb20f9 we would also need to adjust the dashboard cookbook to start apache...

Ah, sorry, hadn't seen your comment. Indeed, and I feel uncomfortable about it, tbh.

What about just restarting apache on the first run? You could do that by restarting it when the systemd service file is installed, for instance.

vuntz:
Or, well, what about fixing the mod_proxy recipe to restart apache?

toabctl:
> Or, well, what about fixing the mod_proxy recipe to restart apache?

We could just do a restart for every new module. that would be a clean fix. but results in a lot of restart during the first run.

vuntz:
I'm just surprised because most modules already work without a restart, and we do reload apache after enabling a module -- which should be enough.

vuntz:
Btw, this worked in SLE12 GA, so wouldn't that be a change in SP1 that breaks this?

toabctl:
> I'm just surprised because most modules already work without a restart, and we do reload apache after enabling a module -- which should be enough.

The modules from the default list work fine. And there is a race so it sometimes work. But I'm currently not sure about the race and can't explain it...

toabctl:
> Btw, this worked in SLE12 GA, so wouldn't that be a change in SP1 that breaks this?

Are we sure it always worked with SLE12 GA ?

vuntz:
> Are we sure it always worked with SLE12 GA ?

We moved to apache before M4. I never saw the issue since then until the switch to SP1. So I'm assuming it always worked, yes.

Might be worth trying to manually reproduce (without crowbar) on GA and SP1...

toabctl:
> Might be worth trying to manually reproduce (without crowbar) on GA and SP1...

Done. Seems to be a SP1 bug. See https://bugzilla.suse.com/show_bug.cgi?id=949434#c1

toabctl:
But is in the apache2 package. let's wait for the fix there.

PR #70, opened 2015-10-09
adjust to new sntp (bnc#949513)

bmwiedemann:
https://bugzilla.suse.com/show_bug.cgi?id=949513
we need to set the system time when we start booting.
But according to man sntp it changed meaning of params

```
12-GA:
       -s, --settod
              Set (step) the time with settimeofday().  This option  must  not
              appear  in  combination  with any of the following options: adj-
              time.

12-SP1:
       -S, --step
              OK to 'step' the time with settimeofday(2).

       -s, --slew
              OK to 'slew' the time with adjtime(2).
```

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
+++ b/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
@@ -134,8 +134,10 @@ sync_time() {
 
 <% if @target_platform_version.to_f < 12.0 %>
     SNTP_OPTS="-P no -r"
-<% else %>
+<% elsif @target_platform_version.to_f < 12.09 %>
     SNTP_OPTS="-s"
+<% else %>
+    SNTP_OPTS="-S"
 <% end %>
     while [[ $tries_left > 0 ]] ; do
         for ntpserver in $VALID_NTP_SERVERS ; do


vuntz:
+1

That's only in crowbar_register, though. Do we have any other code using this?

vuntz:
Oh, I see that we're using ntpdate in updates/control.sh, maybe we should be consistent?

bmwiedemann:
updated to use ntpdate

bmwiedemann:
updated to also fix the correct place - using sntp for simplicity there

rsalevsky:
+1

vuntz:
+1

bmwiedemann:
rebased

tboerger:
:+1:

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
```diff
@@ -134,8 +134,10 @@ sync_time() {
 
 <% if @target_platform_version.to_f < 12.0 %>
     SNTP_OPTS="-P no -r"
-<% else %>
+<% elsif @target_platform_version.to_f == 12.0 %>
```
dirkmueller: I'd prefer 

```ruby

elsif @target_platform_version.to_f < 12.1 

```

to avoid float comparison precision issues.

bmwiedemann: then I'd prefer < 12.09 exactly to avoid float comparison issues. 12.0 can be represented exactly as a float

bmwiedemann:
updated to avoid floating comparison == 12.0

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
dirkmueller: I wasn't sure whether 12 can be represented or not. in any case "12.0 < 12.1" should be true no matter which IEEE rounding mode or precision error is in effect, and thats all that matters here. we have not and will not release SLES 12.0999999999999999 for that matter. And Float::EPSILON is pretty small (and in IEEE rounding, 12.1-3*Float::EPSILON is still 12.1)

12.09 is slightly less readable though.

PR #72, opened 2015-10-09
fix suse_ver

bmwiedemann:
on SLES12-SP1, the suse_ver string contained "12-SP1"
which was not an integer

### DIFF
--- a/updates/control.sh
+++ b/updates/control.sh
@@ -51,7 +51,7 @@ function suse_ver() {
 
     if [ -f /etc/os-release ]; then
         . /etc/os-release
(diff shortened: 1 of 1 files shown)


vuntz:
I think https://build.suse.de/package/rdiff/Devel:Cloud:6/sleshammer?linkrev=base&rev=38 would be a better fix (but this wouldn't be needed with https://github.com/crowbar/crowbar-core/pull/60)

bmwiedemann:
updated it to use the same approach

vuntz:
+1 (in case other PR doesn't get merged)

vuntz:
Obsoleted by https://github.com/crowbar/crowbar-core/pull/60

PR #73, opened 2015-10-09
webui: Display node architecture along with the CPU info

vuntz:


jsuchome:
+1

crowbar_framework/app/models/node_object.rb
```diff
@@ -553,10 +553,26 @@ def memory
...
   end
 
+  def cpu_arch
+    if !cpu.blank? && !architecture.blank?
+      "#{cpu} (#{architecture})"
+    elsif !cpu.blank?
+      cpu
+    elsif !architecture.blank?
+      architecture
+    else
```
tboerger: This is indeed redundant :)

tboerger:
Drop the `else` than I'm fine.

vuntz:
@tboerger done

crowbar_framework/app/models/node_object.rb
```diff
@@ -553,10 +553,26 @@ def memory
     @node["memory"]["total"] rescue nil
   end
 
+  def architecture
+    @node["kernel"]["machine"] rescue nil
```
tboerger: Let's ignore that for now.

tboerger:
:+1:

PR #75, opened 2015-10-12
Revert jquery and bootstrap bump

vuntz:
This reverts https://github.com/crowbar/crowbar-core/pull/63 as we seem to have a broken webui with the new jquery.


tboerger:
It should be pretty bad to revert this as the new wizard screen and so on relies on the new bootstrap version.

rsalevsky:
It seams this is just an issue with Firefox with Chrome it the migrate plugin works as expected.

I'm against the revert because the fix should be easy for an jquery expert and Ken started already with rewriting stuff to use the new bootstrap. It also blocks current cards in our sprint.

vuntz:
@tboerger @rsalevsky really, the proposal is to revert it because noone is fixing it right now. That doesn't mean that we won't add it back later, but this is blocking the milestone... If someone fixes the bug, then of course it's better.

vuntz:
> It seams this is just an issue with Firefox with Chrome it the migrate plugin works as expected.

The UI is also broken in chromium, btw.

rsalevsky:
@tboerger As far as I know @tboerger works already on this.

dirkmueller:
Whats the timeline for a fix ? we ought to release the milestone 9 days ago and this is something that got merged and is breaking stuff even though I asked to delay the merge until after the milestone was cut..

toabctl:
I added this PR to the package for now to be able to continue with M6

dirkmueller:
@toabctl the problem with having the revert only in the package is that it most likely breaks all other PR's gating

toabctl:
@dirkmueller in this case it's just UI changes which shouldn't break gating I think

tboerger:
Closed in favor of https://github.com/crowbar/crowbar-core/pull/77 and https://github.com/crowbar/crowbar-ha/pull/31

PR #78, opened 2015-10-14
Allocate node IP from correct (node) range only after the  node itself was already allocated and not when it is only discovered.

jsuchome:
https://bugzilla.suse.com/show_bug.cgi?id=941528

Note for backporting to Cloud5: get_network_by_type needs to return nil if network is not there

### DIFF
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -136,7 +136,7 @@ if not nodes.nil? and not nodes.empty?
           hostname mnode.name
           macaddress mac_list[i]
           if mnode.macaddress == mac_list[i]
-            ipaddress admin_data_net.address
+            ipaddress admin_data_net.address unless admin_data_net.nil?
             options [
      '      if option arch = 00:06 {
         filename = "discovery/bootia32.efi";
--- a/crowbar_framework/app/models/deployer_service.rb
+++ b/crowbar_framework/app/models/deployer_service.rb
@@ -134,49 +134,6 @@ class DeployerService < ServiceObject
       roles << node.run_list_to_roles
       roles.flatten!
 
-      # Allocate required addresses
-      range = node.admin? ? "admin" : "host"
-      @logger.debug("Deployer transition: Allocate admin address for #{name}")
-      ns = NetworkService.new @logger
-      result = ns.allocate_ip("default", "admin", range, name)
-      @logger.error("Failed to allocate admin address for: #{node.name}: #{result[0]}") if result[0] != 200
-      if result[0] == 200
-        address = result[1]["address"]
-        boot_ip_hex = sprintf("%08X", address.split(".").inject(0) { |acc, i| (acc << 8) + i.to_i })
-      end
-
-      @logger.debug("Deployer transition: Done Allocate admin address for #{name} boot file:#{boot_ip_hex}")
-
-      if node.admin?
-        # If we are the admin node, we may need to add a vlan bmc address.
-        # Add the vlan bmc if the bmc network and the admin network are not the same.
-        # not great to do it this way, but hey.
-        admin_net = Chef::DataBag.load "crowbar/admin_network" rescue nil
-        bmc_net = Chef::DataBag.load "crowbar/bmc_network" rescue nil
-        if admin_net["network"]["subnet"] != bmc_net["network"]["subnet"]
-          @logger.debug("Deployer transition: Allocate bmc_vlan address for #{name}")
-          result = ns.allocate_ip("default", "bmc_vlan", "host", name)
-          @logger.error("Failed to allocate bmc_vlan address for: #{node.name}: #{result[0]}") if result[0] != 200
-          @logger.debug("Deployer transition: Done Allocate bmc_vlan address for #{name}")
-        end
-
-        # Allocate the bastion network ip for the admin node if a bastion
-        # network is defined in the network proposal
-        bastion_net = Chef::DataBag.load "crowbar/bastion_network" rescue nil
-        unless bastion_net.nil?
-          result = ns.allocate_ip("default", "bastion", range, name)
-          if result[0] != 200
-            @logger.error("Failed to allocate bastion address for: #{node.name}: #{result[0]}")
-          else
-            @logger.debug("Allocated bastion address: #{result[1]["address"]} for the admin node.")
-          end
-        end
-      end
-
-      node = NodeObject.find_node_by_name node.name
-      # save this on the node after it's been refreshed with the network info.
-      node.crowbar["crowbar"]["boot_ip_hex"] = boot_ip_hex if boot_ip_hex
-
       # Walk map to categorize the node.  Choose first one from the bios map that matches.
       role = RoleObject.find_role_by_name "deployer-config-#{inst}"
       done = false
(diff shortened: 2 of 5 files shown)


crowbar_framework/app/models/deployer_service.rb
```diff
@@ -181,6 +162,27 @@ def transition(inst, name, state)
...
+      if result[0] == 200
+        address = result[1]["address"]
+        boot_ip_hex = sprintf("%08X", address.split(".").inject(0){ |acc,i| (acc << 8) + i.to_i })
+      end
+
+      @logger.debug("Deployer transition: Done Allocate admin address for #{name} boot file:#{boot_ip_hex}")
+
+      # Let it fly to the provisioner. Reload to get the address.
+      node = NodeObject.find_node_by_name node.name
+      node.crowbar["crowbar"]["usedhcp"] = true
```
jsuchome: Looks like this isn't used anywhere... or is it?

crowbar_framework/app/models/deployer_service.rb
```diff
@@ -112,19 +112,6 @@ def transition(inst, name, state)
...
-      result = ns.allocate_ip("default", "admin", range, name)
-      @logger.error("Failed to allocate admin address for: #{node.name}: #{result[0]}") if result[0] != 200
-      if result[0] == 200
-        address = result[1]["address"]
-        boot_ip_hex  = sprintf("%08X",address.split(".").inject(0){ |acc,i|(acc << 8)+i.to_i })
-      end
-
-      @logger.debug("Deployer transition: Done Allocate admin address for #{name} boot file:#{boot_ip_hex}")
-
       if node.admin?
```
jsuchome: Not sure about this bmc section, could it stay here?

crowbar_framework/app/models/deployer_service.rb
vuntz: It seems you're right, try killing it? :-)

crowbar_framework/app/models/deployer_service.rb
```diff
@@ -181,6 +162,27 @@ def transition(inst, name, state)
...
+      end
+
+      @logger.debug("Deployer transition: Done Allocate admin address for #{name} boot file:#{boot_ip_hex}")
+
+      # Let it fly to the provisioner. Reload to get the address.
+      node = NodeObject.find_node_by_name node.name
+      node.crowbar["crowbar"]["usedhcp"] = true
+
+      # save this on the node after it's been refreshed with the network info.
+      node.crowbar["crowbar"]["boot_ip_hex"] = boot_ip_hex if boot_ip_hex
```
vuntz: Maybe move this inside `if result[0] == 200`?

crowbar_framework/app/models/deployer_service.rb
```diff
@@ -181,6 +162,27 @@ def transition(inst, name, state)
...
+      node.save
```
vuntz: Don't save it, just set `save_it` to `true`.

crowbar_framework/app/models/deployer_service.rb
```diff
@@ -151,18 +138,12 @@ def transition(inst, name, state)
...
-
       role = RoleObject.find_role_by_name "deployer-config-#{inst}"
       unless role.default_attributes["deployer"]["use_allocate"] and !node.admin?
         @logger.debug("Automatically allocating node #{node.name}")
         node.allocate!
       end
 
-      # save this on the node after it's been refreshed with the network info.
-      node.crowbar["crowbar"]["boot_ip_hex"] = boot_ip_hex  if boot_ip_hex
       node.save
```
vuntz: Do we still need to save here?

crowbar_framework/app/models/deployer_service.rb
vuntz: It doesn't really matter, tbh, but to be fully correct, I'd also move it.

crowbar_framework/app/models/deployer_service.rb
```diff
@@ -181,6 +162,27 @@ def transition(inst, name, state)
       roles << node.run_list_to_roles
       roles.flatten!
 
+      # Allocate required addresses
```
vuntz: Should we check if an address has already been allocated and not do anything if that's the case?

crowbar_framework/app/models/deployer_service.rb
jsuchome: i was not sure, it was 'just in case'...

crowbar_framework/app/models/deployer_service.rb
jsuchome: ah, true (left from testing)

crowbar_framework/app/models/deployer_service.rb
vuntz: Also: I'm wondering whether this should move to the transition method of the network service? (since `delete` / deallocating IP address is handled there)

jsuchome:
This config (adapted to Cloud5) currently works for me

jsuchome:
@vuntz modified for network service. 
Do you think we should prepare PTF (Cloud5-based) or wait until this gets reviewed and merged?

aplanas:
@jsuchome are we going to address the DNS part?

crowbar_framework/app/models/deployer_service.rb
```diff
@@ -199,10 +152,9 @@ def transition(inst, name, state)
 
       os_map = role.default_attributes["deployer"]["os_map"]
       node.crowbar["crowbar"]["hardware"]["os"] = os_map[0]["install_os"]
-      save_it = true
     end
 
-    node.save if save_it
+    node.save
```
vuntz: Why always save? It should only happen if we have to save.

jsuchome: ah, leftover from previous version with IP allocation

crowbar_framework/app/models/network_service.rb
```diff
@@ -302,6 +302,54 @@ def transition(inst, name, state)
...
+          result = allocate_ip("default", "bastion", range, name)
+          if result[0] != 200
+            @logger.error("Failed to allocate bastion address for: #{node.name}: #{result[0]}")
+          else
+            @logger.debug("Allocated bastion address: #{result[1]["address"]} for the admin node.")
+          end
+        end
+      end
+
+      node = NodeObject.find_node_by_name node.name
```
vuntz: We already have `node`, why do you reload it? (maybe there's a good reason, btw)

jsuchome: I am not sure, but without that reload it didn't work, AFAIK. There's some kind of explanation on  next line:
``# save this on the node after it's been refreshed with the network info.``

vuntz: Ah, that makes sense. Can you just move that comment above, and clarify that it is why we reload?

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -136,7 +136,7 @@
           hostname mnode.name
           macaddress mac_list[i]
           if mnode.macaddress == mac_list[i]
-            ipaddress admin_data_net.address
+            ipaddress admin_data_net.address unless admin_data_net.nil?
```
vuntz: Just to confirm: when that condition is not met, we don't put an IP address in the config file for the node; does it still get an IP address from the right pool (dhcp) of the admin network? (if you reboot, for instance)

vuntz:
+1 if my question has the expected answer :-)

chef/cookbooks/provisioner/recipes/update_nodes.rb
jsuchome: It has 192.168.126.24 in my test, which is in the correct range, installed nodes have .81+

aplanas: I have the same results. Worked for me.

bmwiedemann:
I did not review or test, but I would welcome such a fix very much.

PR #80, opened 2015-10-15
Avoid endlessly looping in the deployment queue in some cases

vuntz:
It can happen that we end up stuck in the deployment queue because a
proposal stays in the queue can't be applied. It happens, for instance,
if the proposal is not valid (or being applied).

So ignore proposals that can't be committed and that results in 4xx or
5xx errors.


rsalevsky:
+1

tboerger:
:+1:

PR #81, opened 2015-10-16
Fix UEFI support when using grub

vuntz:


vuntz:
Anyone up for a review? This blocks the multi-architecture patches (as they're built on top of that, since EFI support was also updated for multiple architectures).

chef/cookbooks/provisioner/recipes/setup_base_images.rb
```diff
@@ -579,12 +596,12 @@
...
   node.set[:provisioner][:available_oses][os] ||= Mash.new
   if /^(hyperv|windows)/ =~ os
-    node.set[:provisioner][:available_oses][os][:kernel] = "../#{os}/#{kernel}"
+    node.set[:provisioner][:available_oses][os][:kernel] = "../../#{os}/#{kernel}"
     node.set[:provisioner][:available_oses][os][:initrd] = " "
     node.set[:provisioner][:available_oses][os][:append_line] = " "
   else
-    node.set[:provisioner][:available_oses][os][:kernel] = "../#{os}/install/#{kernel}"
-    node.set[:provisioner][:available_oses][os][:initrd] = "../#{os}/install/#{initrd}"
+    node.set[:provisioner][:available_oses][os][:kernel] = "../../#{os}/install/#{kernel}"
```
aplanas: Looks like that `#{kernel}` and `#{initrd}` is going to be replaced with `../something`, isn't better to avoid this `install` part of the path directly?

vuntz: I'm not sure I understand, where do you see it replaced?

(I do plan to get rid of the `../..` bits eventually, in a later PR, btw)

aplanas: Sorry, I was wrong.  I mixed the parameter to the template with this later code.

aplanas:
+1 for me. Not tested.

nkrinner:
+1

PR #82, opened 2015-10-16
Minor fixes in provisioner

vuntz:
These are just some bits I fixed during multi-architecture, but that aren't related to this.


tboerger:
LGTM

rsalevsky:
+1

PR #84, opened 2015-10-16
provisioner: Better support for multiple architectures for discovery

vuntz:
We also use a per-architecture directory for discovery now.

~~Note that this contains https://github.com/crowbar/crowbar-core/pull/83, and gating will fail because sleshammer needs to be updated accordingly.~~

~~Note that this contains https://github.com/crowbar/crowbar-core/pull/81~~

PR #83, opened 2015-10-16
Support multiple architecture for repo layout and OS attributes

vuntz:
~~Note that this contains https://github.com/crowbar/crowbar-core/pull/81 and https://github.com/crowbar/crowbar-core/pull/82 https://github.com/crowbar/crowbar-core/pull/84, so please review these first.~~

Also, this depends on https://github.com/SUSE-Cloud/automation/pull/613.

PR #85, opened 2015-10-16
provisioner: Support for discovery/allocation of ppc64le nodes

vuntz:
~~Note that this contains https://github.com/crowbar/crowbar-core/pull/84 https://github.com/crowbar/crowbar-core/pull/83.~~

### DIFF
--- a/chef/cookbooks/provisioner/recipes/setup_base_images.rb
+++ b/chef/cookbooks/provisioner/recipes/setup_base_images.rb
@@ -26,8 +26,8 @@ append_line = ""
 tftproot = node[:provisioner][:root]
 
 # This is what we could support, but this requires validation
-#discovery_arches = ["x86_64", "ia32"]
-discovery_arches = ["x86_64"]
+#discovery_arches = ["x86_64", "ppc64le", "ia32"]
+discovery_arches = ["x86_64", "ppc64le"]
 
 discovery_dir = "#{tftproot}/discovery"
 pxe_subdir = "bios"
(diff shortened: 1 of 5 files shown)


abel-navarro:
For review simplicity, I'd wait until https://github.com/crowbar/crowbar-core/pull/83 is merged, then I would rebase this without those changes.

vuntz:
Rebased

chef/cookbooks/provisioner/recipes/setup_base_images.rb
```diff
@@ -60,8 +63,9 @@
 end
 
 # PXE config
-%w(x86_64).each do |arch|
-  # Make it easy to totally disable/enable ppc64le
+# ppc64le bootloader can parse pxelinux config files
+%w(x86_64 ppc64le).each do |arch|
```
jsuchome: Why not use filtered `discovery_arches` here?

vuntz: Because this bit is only for PXE boot, we have another loop later on for EFI boot. So we would need to do the filter twice.

jsuchome: Maybe I'm just don't seeing something, but it looks like you filter `discovery_arches` at line 36, and than you basically iterate over same values at lines 67 and 106 (exiting at lines 69 and 108 means you really care about current value of `discovery_arches`)

vuntz: The values I'm iterating over are the architectures for which PXE boot is used or EFI boot is used; it doesn't mean that it's the architectures for which we support discovery, that's why we have the `next` at the beginning of the loop.

jsuchome: Oh, and of course, I missed the second time the values are different. :-(

jsuchome:
+1

tboerger:
LGTM

PR #87, opened 2015-10-19
provisioner: Fix use of wrong attribute to get platform version

vuntz:
This got introduced in https://github.com/crowbar/crowbar-core/pull/53


jsuchome:
+

vuntz:
Merging as CI seems broken without this (unsure why CI passed without this?).

PR #92, opened 2015-10-19
Remove unused cookbooks

vuntz:


rsalevsky:
Travis fails...

tboerger:
@vuntz you need to remove the rails part from the rspec fixtures.

vuntz:
Just to make it clear: gating can't be green because the patch fails when building the package (due to the removal of binary files).

rsalevsky:
+1

tboerger:
:+1:

PR #96, opened 2015-10-22
Adding the functionality to search and replace a line

ovisan:
The search is done with regex expressions and the replacement is string

### DIFF
--- a/chef/cookbooks/utils/providers/line.rb
+++ b/chef/cookbooks/utils/providers/line.rb
@@ -87,6 +87,20 @@ ensure
   unlock(lock) unless lock.nil?
 end
 
+def search_and_replace(filename, search_regex, replaced_text)
+  lock = nil
+  lock = lock(@new_resource)
+  rc = Chef::Util::FileEdit.new(filename)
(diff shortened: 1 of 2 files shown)


aspiers:
Nice idea!  I wonder why Chef doesn't provide this by default.  Maybe [the recommended way is to always use templates](http://stackoverflow.com/questions/14848110/how-i-can-change-a-file-with-chef), but I think there are genuine scenarios where search+replace is a valid approach.

Having said that, I wonder if this really belongs in the existing `line` LWRP, since it will potentially modify multiple lines.  And also I have to give -1 for now, since you just added a merge commit to the PR.  PRs should not have merge commits.

ovisan:
@aspiers fixed the merge commit. Right, i was also looking at the recommended way using templates and i agree that should be done for a broader scenario. In this case we just want to modify a line or a specific text matched by a regular expression.

tboerger:
:+1:

ovisan:
Seems search_file_replace_line exists and it's working, closing the PR.

PR #98, opened 2015-10-23
Make DNS resolve discovered nodes (+ some other fixes)

vuntz:


chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -182,19 +187,48 @@ def make_zone(zone)
...
+      base_name = "#{net_name}.#{base_name_no_net}"
+      alias_name = "#{net_name}.#{alias_name_no_net}" if alias_name_no_net
     end
     cluster_zone[:hosts][base_name] = Mash.new
     cluster_zone[:hosts][base_name][:ip4addr] = network.address
     cluster_zone[:hosts][base_name][:alias] = alias_name if alias_name
   end
+
+  # Also set DNS name with temporary DHCP address for discovered nodes
+  if n[:state] == "discovered" &&
```
vuntz: I disagree with you here, sorry :-)

crowbar_framework/app/controllers/nodes_controller.rb
```diff
@@ -427,7 +427,7 @@ def get_node_and_network(node_name)
     @node = NodeObject.find_node_by_name(node_name) if @node.nil?
     if @node
       # If we're in discovery mode, then we have a temporary DHCP IP address.
-      if not ["discovering", "discovered", "hardware-installing", "hardware-installed"].include? @node.state
+      if !["discovering", "discovered"].include?(@node.state)
```
MaximilianMeister: MINOR: i would just use `unless` here

vuntz: That'd be against the style guidelines because there's a `else` :-)

abel-navarro:
I'm not really qualified to review this so bear with me if the question is not relevant: I can see that now the discovered nodes will get a dns entry, but what about the transition from discovered to allocated? will the record be deleted/updated in that case?

vuntz:
> what about the transition from discovered to allocated? will the record be deleted/updated in that case?

Yes, it will.

vuntz:
CI is red, but failing in cct. That means tempest passed, and we're far beyond what this change could have broken, so I'll just pretend CI is green.

Still missing some +1, though.

rsalevsky:
+1

crowbar_framework/app/controllers/nodes_controller.rb
rhafer: @vuntz Does the IP the node is currently using really already match the one we allocated from the "host" range. Withouth having tested this, I'd expect the node to still use the IP address from the "dhcp" range. Displaying the allocated address in that case would be confusing IMO

vuntz: Hrm, good question. And the answer is: "it depends":

 - first, when doing hardware-installing/hardware-installed outside of the discovery process (that can happen with the bios/raid barclamps), then yes, it just works fine. But arguable, not really a case that we're seeing for all practical purposes ;-)

 - in discovery, we have two things:

  - the network role is not used, so it won't set the IP address from the host range. So at first, it would seem that you are correct, but...

  - the dhcp configuration is actually updated, and on lease renewal, the node will get the allocated IP address.

All in all, we are in a gray area here, and it pretty much depends on the timing. That being said, I would argue that people generally won't want to connect to the discovered node at this point in time (they would connect before allocating it) and what we display in the UI here is consistent with what is put in the DNS config. Which is why I lean towards keeping this.

(question: should I put all this in the commit message?)

rhafer: > Hrm, good question. And the answer is: "it depends":
[..]
> Which is why I lean towards keeping this.

Thanks for the details. And yes, I agree with you (now) that it is fine to keep it.

> question: should I put all this in the commit message?

Yes please!

aplanas:
+1!

crowbar_framework/app/controllers/nodes_controller.rb
vuntz: done

vuntz:
Rebased and improved commit message, but no code change.

rhafer:
+1

vuntz:
Please don't merge yet, it seems that dnsmasq still has the wrong IP in its cache (I suspect something fills it in before bind restart).

vuntz:
Changing the notification for dnsmasq to a delayed one (see updated commit message of first commit)

vuntz:
Ok, fixed the issue that was blocking the PR, removing wip.

vuntz:
Btw, I apologize, but I'm resetting the +1 count because I changed the PR.

@aplanas @rhafer @rsalevsky if you have time to review this again...

rsalevsky:
Still +1

aplanas:
+1

PR #99, opened 2015-10-25
Add "role recipes" with the goal of simplifying the roles

vuntz:
We want the roles to only reference one recipe, so that we can put some
intelligence in that recipe:

 - in a first step, we will stop changing the run list of a node
   depending on its state, and instead have the role recipe "decide"
   whether something should be done or not, depending on the state. This
   will solve the issue that the search results for nodes are not
   reliable when a node is in some state that is not "readying",
   "applying" or "ready".

 - in a later step, we might want to have a new attribute that will tell
   us which role to apply instead of applying everything.


aspiers:
Sounds good but I don't understand why 

> the search results for nodes are not reliable when a node is in some state that is not "readying", "applying" or "ready"

Please can you clarify this in the commit messages for each PR?

vuntz:
> Please can you clarify this in the commit messages for each PR?

Is the following good enough:

> When a node is not in one of these states, the run list is updated by the crowbar engine and removes most of the roles (except roles from some core barclamps). This makes the search results unreliable when a node is, for instance, rebooting.

tboerger:
Makes it harder to see the actual runlist but LGTM

aspiers:
Vincent Untz <notifications@github.com> wrote:
> Is the following good enough:
> 
> > When a node is not in one of these states, the run list is updated by the crowbar engine and removes most of the roles (except roles from some core barclamps). This makes the search results unreliable when a node is, for instance, rebooting.

It's definitely helpful, but I still don't get it.  Presumably you are
talking about a specific type of search for nodes which have a certain
role?  But I'm wondering why it makes sense to use that kind of search
in this situation.  Would help a lot if you could give an exact
example of such a search?

vuntz:
@aspiers it's really any search in the cookbook, like https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/tempest/recipes/config.rb#L137

vuntz:
Not really sure how to make this clearer, btw :/

aspiers:
Erm, is that a good example?  Because that `glance_node` variable isn't used anywhere :)

aspiers:
But I think I get it now ...

aspiers:
Maybe

    use_docker = !search(:node, "roles:nova-multi-compute-docker").empty?

is a better example.

I think the key point you are not stating *completely* explicitly here is that we sometimes need to make decisions or take actions depending on which nodes "normally" have certain roles, rather than which nodes have certain roles at a given instant.  So if there is only one docker compute node, we want `use_docker` to be `true` even if it happens to be rebooting at the time the search is done.

Perhaps you could add that example (or similar) to the commit message?  I didn't get it until I looked at one.  Or maybe it's just jetlag ... if it's a big hassle to update then don't bother.

However @tboerger has a very good point - I do worry that this will make debugging harder.  If the recipes aren't included, there will be a silent gap in the `chef-client` logs, and if a particular recipe was already quiet, it could be hard to be sure whether it was included or not.  Or does something always get logged by `include_recipe`?

vuntz:
> However @tboerger has a very good point - I do worry that this will make debugging harder.

I can add a log in a `else` branch for all role recipes, so we see when a role is skipped.

vuntz:
@aspiers done

vuntz:
Gating reached cct (with tempest passing), and failed there. I'll consider this as green.

Still missing reviews, though.

vuntz:
Removed wip tag.

tboerger:
:+1:

aspiers:
Looks a lot better now.  There is however now quite a lot of code duplication - is it possible to reduce that?

mjura:
+1

toabctl:
@vuntz are you going to address @aspiers comments to remove the duplicated code? i.e. create a function todo the role check and logging?

vuntz:
@toabctl we'll revisit this in the next sprint of team grapes. The first step will be a bit simplified, as I think I went a bit too far by doing the next step at the same time (which triggers some issues).

vuntz:
Closing; this is partly done, will create a new PR for the remaining.

PR #101, opened 2015-10-26
Adding the functionality to search and replace a line

ovisan:
The search is done with regex expressions and the replace is string
https://github.com/crowbar/crowbar-core/pull/96
reopening for idiomatic chef

### DIFF
--- a/chef/cookbooks/utils/providers/line.rb
+++ b/chef/cookbooks/utils/providers/line.rb
@@ -87,14 +87,38 @@ ensure
   unlock(lock) unless lock.nil?
 end
 
+def search_and_replace(filename, search_regex, replaced_text)
+  rc = Chef::Util::FileEdit.new(filename)
+  rc.search_file_replace_line(
+    search_regex,
+    replaced_text
+  )
+  rc.write_file
+end
+
 action :add do
-  updated = add_and_filter(@new_resource.file,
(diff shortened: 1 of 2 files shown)


tboerger:
:+1:

MaximilianMeister:
would have made the formatting an extra commit but it doesn't really matter
:+1:

chef/cookbooks/utils/providers/line.rb
```diff
@@ -87,14 +87,40 @@ def add_and_filter(file, add, filter_re)
...
+    nil,
+    "^#{@new_resource.name}$"
+  )
+  @new_resource.updated_by_last_action(true) if updated
+end
+
+action :replace do
+  updated = search_and_replace(
+    @new_resource.file,
+    @new_resource.regexp,
```
vuntz: There's no `regexp` attribute in https://github.com/ovisan/crowbar-core/blob/searchandreplace/chef/cookbooks/utils/resources/line.rb

ovisan: @vuntz correct, i don't know what happened, but it's fixed now. Also `:replace` was not defined

tboerger:
Still :+1:

rsalevsky:
+1

PR #18, opened 2015-08-31
add repositories page

MaximilianMeister:
Part of https://trello.com/c/CRJYcP59/466-5-p6-move-repo-checks-out-of-install-suse-cloud-also-ux

This still needs some work, the views are just a very basic table right now, so depending on where we want to go with the whole thing, (adding new repos, first time landing page...) we need to involve @kwwii in the design.
please don't judge the views too much here, as they are just a dummy right now to have something.

repo_page

So for now shall we just implement a simple table like that? or do we need sth. more designed? any concrete suggestions are more than welcome. same goes for the lib, I have put this to WIP to get some input from you.

Bits that are still unclear/TODO:
- [ ] decide on a good UI representation (view)
- [ ] define YAML file
- [ ] where do we store the yaml file(s) ?
- [ ] fit the lib to [repositories.rb](https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/provisioner/libraries/repositories.rb) **or** fit [repositories.rb](https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/provisioner/libraries/repositories.rb) to the lib?
- [ ] unittests for the lib

PR #102, opened 2015-10-26
Followup PR to fix issues in new repo magic

vuntz:
This is a follow up to https://github.com/crowbar/crowbar-core/pull/18.

Comments not addressed yet:
- [x] [missing API to enable all available repos](https://github.com/crowbar/crowbar-core/pull/18/files#r42978260)
- [x] [renaming md5 attribute in yaml file](https://github.com/crowbar/crowbar-core/pull/18/files#r42975805)
- [x] [making SOC repos mandatory for cloud](https://github.com/crowbar/crowbar-core/pull/18/files#r42981978)
- [x] [removing hack for SES](https://github.com/crowbar/crowbar-core/pull/18/files#r42982071)
- [x] [ability to see which repo is active, but with stale data](https://github.com/crowbar/crowbar-core/pull/18/files#r42990584)
- [x] nothing forbidding you to allocate a node if mandatory repos are missing
- [x] probably need a link to repo page at the right of the platform select when allocating a node

### DIFF
--- a/chef/cookbooks/provisioner/libraries/repositories.rb
+++ b/chef/cookbooks/provisioner/libraries/repositories.rb
@@ -74,12 +74,15 @@ class Provisioner
       # they can be used.
       def get_repos(platform, version)
         repositories = {}
-        repos_from_databag = Chef::DataBag.load("repositories").keys rescue []
-        repos_from_databag.each do |reponame|
-          dbi = Chef::DataBagItem.load("repositories", reponame)
-          if dbi["platform"] == "#{platform}-#{version}"
-            repositories[dbi["name"]] = { url: dbi["url"], ask_on_error: dbi["ask_on_error"] }
-          end
+
+        repos_db = begin
+          Chef::DataBag.load("crowbar/repositories")
+        rescue Net::HTTPServerException
+          {}
+        end
+
+        repos_db.fetch("#{platform}-#{version}", {}).each do |id, config|
+          repositories[config["name"]] = { url: config["url"], ask_on_error: config["ask_on_error"] }
         end
         repositories
       end
--- a/chef/cookbooks/provisioner/libraries/repositories.rb
+++ b/chef/cookbooks/provisioner/libraries/repositories.rb
@@ -78,7 +78,7 @@ class Provisioner
         repos_from_databag.each do |reponame|
           dbi = Chef::DataBagItem.load("repositories", reponame)
           if dbi["platform"] == "#{platform}-#{version}"
-            repositories[dbi["name"]] = { url: dbi["url"] }
+            repositories[dbi["name"]] = { url: dbi["url"], ask_on_error: dbi["ask_on_error"] }
           end
         end
         repositories
--- a/chef/cookbooks/provisioner/libraries/repositories.rb
+++ b/chef/cookbooks/provisioner/libraries/repositories.rb
@@ -22,7 +22,7 @@ class Provisioner
           raise "Internal error: inspect_repos method should only be called on provisioner-server node."
         end
 
-        common_available, cloud_available, hae_available, storage_available = nil
+        common_available, cloud_available, hae_available, storage_available = false
         repos_from_databag = Chef::DataBag.load("repositories").keys rescue []
 
         case node[:platform]
--- a/chef/cookbooks/provisioner/libraries/repositories.rb
+++ b/chef/cookbooks/provisioner/libraries/repositories.rb
@@ -17,58 +17,6 @@
 class Provisioner
   class Repositories
     class << self
-      def inspect_repos(node)
-        unless node.roles.include? "provisioner-server"
-          raise "Internal error: inspect_repos method should only be called on provisioner-server node."
-        end
-
-        common_available, cloud_available, hae_available, storage_available = false
-        repos_from_databag = Chef::DataBag.load("repositories").keys rescue []
-
-        case node[:platform]
-        when "suse"
-          repos_from_databag.each do |repo_from_databag|
-            dbi = Chef::DataBagItem.load("repositories", repo_from_databag)
-            case dbi["product_name"]
-            when "common"
-              common_available = true
-            when "cloud"
-              cloud_available = true
-            when "hae"
-              hae_available = true
-            when "storage"
-              storage_available = true
-            end
-          end
-
-          # set an attribute about available repos so that cookbooks and crowbar
-          # know that HA can be used
-          # know that SUSE_Storage can be used
-          # know that OpenStack can be used
-          node_set = false
-          node.set[:provisioner][:suse] ||= {}
-          if node[:provisioner][:suse][:common_available] != common_available
-            node.set[:provisioner][:suse][:common_available] = common_available
-            node_set = true
-          end
-          if node[:provisioner][:suse][:cloud_available] != cloud_available
-            node.set[:provisioner][:suse][:cloud_available] = cloud_available
-            node_set = true
-          end
-          if node[:provisioner][:suse][:hae_available] != hae_available
-            node.set[:provisioner][:suse][:hae_available] = hae_available
-            node_set = true
-          end
-          if node[:provisioner][:suse][:storage_available] != storage_available
-            node.set[:provisioner][:suse][:storage_available] = storage_available
-            node_set = true
-          end
-          if node_set
-            node.save
-          end
-        end
-      end
-
       # This returns a hash containing the data about the repos that must be
       # used on nodes; optional repos (such as HA) will only be returned if
       # they can be used.
--- a/crowbar_framework/app/controllers/repositories_controller.rb
+++ b/crowbar_framework/app/controllers/repositories_controller.rb
@@ -50,11 +50,13 @@ class RepositoriesController < ApplicationController
   def activate
     return render_not_found if params[:platform].nil? || params[:repo].nil?
     respond_to do |format|
-      repo_object = Crowbar::Repository.where(platform: params[:platform], repo: params[:repo]).first
-      ret = ProvisionerService.new(logger).create_repository_item(repo_object) unless repo_object.nil?
-      if ret
+      ret = ProvisionerService.new(logger).enable_repository(params[:platform], params[:repo])
+      case ret
+      when 200
         format.json { head :ok }
         format.html { redirect_to repositories_url }
+      when 404
+        render_not_found
       else
         format.json do
           render json: { error: I18n.t("cannot_activate_repo", scope: "error", id: params[:repo]) },
@@ -77,11 +79,13 @@ class RepositoriesController < ApplicationController
   def deactivate
     return render_not_found if params[:platform].nil? || params[:repo].nil?
     respond_to do |format|
-      repo_object = Crowbar::Repository.where(platform: params[:platform], repo: params[:repo]).first
-      ret = ProvisionerService.new(logger).destroy_repository_item(repo_object) unless repo_object.nil?
-      if ret
+      ret = ProvisionerService.new(logger).disable_repository(params[:platform], params[:repo])
+      case ret
+      when 200
         format.json { head :ok }
         format.html { redirect_to repositories_url }
+      when 404
+        render_not_found
       else
         format.json do
           render json: { error: I18n.t("cannot_deactivate_repo", scope: "error", id: params[:repo]) },
--- a/crowbar_framework/app/controllers/repositories_controller.rb
+++ b/crowbar_framework/app/controllers/repositories_controller.rb
@@ -15,7 +15,7 @@
 #
 
 class RepositoriesController < ApplicationController
-  before_filter :load_registry
+  before_filter :reload_registry
   #
   # Repository Check
   #
@@ -101,7 +101,7 @@ class RepositoriesController < ApplicationController
 
   protected
 
-  def load_registry
+  def reload_registry
     Crowbar::Repository.load!
   end
 end
--- a/crowbar_framework/app/controllers/repositories_controller.rb
+++ b/crowbar_framework/app/controllers/repositories_controller.rb
@@ -24,6 +24,11 @@ class RepositoriesController < ApplicationController
   # Renders an HTML view in the UI
   def index
     @repocheck = Crowbar::Repository.check_all_repos
+    @repocheck.sort! do |a, b|
+      required_a = RepositoriesHelper.repository_required_to_i(a.required)
+      required_b = RepositoriesHelper.repository_required_to_i(b.required)
+      [required_a, a.name] <=> [required_b, b.name]
+    end
     respond_to do |format|
       format.html { @repocheck }
       format.xml { render xml: @repocheck }
--- a/crowbar_framework/app/controllers/repositories_controller.rb
+++ b/crowbar_framework/app/controllers/repositories_controller.rb
@@ -49,8 +49,8 @@ class RepositoriesController < ApplicationController
   # required parameters: platform, repo
   def activate
     return render_not_found if params[:platform].nil? || params[:repo].nil?
+    ret, _message = ProvisionerService.new(logger).enable_repository(params[:platform], params[:repo])
     respond_to do |format|
-      ret = ProvisionerService.new(logger).enable_repository(params[:platform], params[:repo])
       case ret
       when 200
         format.json { head :ok }
@@ -78,8 +78,8 @@ class RepositoriesController < ApplicationController
   # required parameters: platform, repo
   def deactivate
     return render_not_found if params[:platform].nil? || params[:repo].nil?
+    ret, _message = ProvisionerService.new(logger).disable_repository(params[:platform], params[:repo])
     respond_to do |format|
-      ret = ProvisionerService.new(logger).disable_repository(params[:platform], params[:repo])
       case ret
       when 200
         format.json { head :ok }
--- a/crowbar_framework/app/controllers/repositories_controller.rb
+++ b/crowbar_framework/app/controllers/repositories_controller.rb
@@ -99,6 +99,32 @@ class RepositoriesController < ApplicationController
     end
   end
 
+  #
+  # Activate all Repositories
+  #
+  # Provides the restful api call for
+  # Activate all Repositories   /utils/repositories/activate_all   POST  Creates Repository DataBagItem
+  def activate_all
+    ProvisionerService.new(logger).enable_all_repositories
+    respond_to do |format|
+      format.json { head :ok }
+      format.html { redirect_to repositories_url }
+    end
+  end
+
+  #
+  # Deactivate all Repositories
+  #
+  # Provides the restful api call for
+  # Deactivate all Repositories   /utils/repositories/deactivate_all   POST   Destroys Repository DataBagItem
+  def deactivate_all
+    ProvisionerService.new(logger).disable_all_repositories
+    respond_to do |format|
+      format.json { head :ok }
+      format.html { redirect_to repositories_url }
+    end
+  end
+
   protected
 
   def reload_registry
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -68,7 +68,7 @@ class NodeObject < ChefObject
       unless provisioner.nil? || provisioner["provisioner"]["default_os"].nil?
          provisioner["provisioner"]["default_os"]
       else
-        admin = NodeObject.find("role:crowbar").find { |n| n.admin? }
+        admin = admin_node
         if admin.nil?
           ""
         else
@@ -171,7 +171,7 @@ class NodeObject < ChefObject
   end
 
   def self.admin_node
-    all.detect(&:admin?)
+    find("role:crowbar").detect(&:admin?)
   end
 
   def self.make_role_name(name)
--- a/crowbar_framework/app/models/provisioner_service.rb
+++ b/crowbar_framework/app/models/provisioner_service.rb
@@ -210,7 +210,7 @@ class ProvisionerService < ServiceObject
     # create DataBag if it doesn't exist yet
     load_repository_bag
     # create DataBagItem with the repo metadata used by provisioner recipes
-    if repo_object.available? && repo_object.valid_key_file?
+    if repo_object.available?
       @logger.debug("Setting #{repo_object.registry['name']} repository for #{repo_object.platform} as active.")
       repo_object.to_databag!.save
       saved = true
--- a/crowbar_framework/app/models/provisioner_service.rb
+++ b/crowbar_framework/app/models/provisioner_service.rb
@@ -179,76 +179,84 @@ class ProvisionerService < ServiceObject
   end
 
   def load_repository_bag
-    bag = Chef::DataBag.load("repositories") rescue nil
+    bag = Chef::DataBag.load("crowbar/repositories") rescue nil
     if bag.nil?
-      create_repository_bag
+      @logger.debug("Creating repositories data bag")
+      bag = Chef::DataBagItem.new
+      bag.data_bag "crowbar"
+      bag["id"] = "repositories"
+      bag.save
     end
     bag
   end
 
   def synchronize_repositories(platforms)
-    # create DataBag if it doesn't exist yet
-    load_repository_bag
+    bag = load_repository_bag
+
     platforms.each do |platform, repos|
-      repos.each do |reponame, active|
-        repo_object = Crowbar::Repository.where(platform: platform, repo: reponame).first
+      repos.each do |repo, active|
         case active.to_i
         when 0
-          db = Chef::DataBagItem.load("repositories", repo_object.id) rescue nil
-          # skip repo if it hasn't been active before
-          next if db.nil?
-          destroy_repository_item(repo_object)
+          disable_repository(platform, repo, bag)
         when 1
-          create_repository_item(repo_object)
+          enable_repository(platform, repo, bag)
         end
       end
     end
   end
 
-  def create_repository_item(repo_object)
-    saved = false
-    # create DataBag if it doesn't exist yet
-    load_repository_bag
-    # create DataBagItem with the repo metadata used by provisioner recipes
+  def enable_repository(platform, repo, bag = nil)
+    bag = load_repository_bag if bag.nil?
+
+    repo_object = Crowbar::Repository.where(platform: platform, repo: repo).first
+    if repo_object.nil?
+      @logger.debug("#{repo} repository for #{platform} does not exist.")
+      return 404
+    end
+
+    code = 200
+
     if repo_object.available?
-      @logger.debug("Setting #{repo_object.registry['name']} repository for #{repo_object.platform} as active.")
-      repo_object.to_databag!.save
-      saved = true
+      bag[platform] ||= {}
+      repo_hash = repo_object.to_databag_hash
+      if bag[platform][repo] != repo_object.to_databag_hash
+        @logger.debug("Setting #{repo} repository for #{platform} as active.")
+        bag[platform][repo] = repo_hash
+        bag.save
+      else
+        @logger.debug("#{repo} repository for #{platform} is already active.")
+      end
     else
-      @logger.debug("Cannot set #{repo_object.registry['name']} repository for #{repo_object.platform} as active.")
+      @logger.debug("Cannot set #{repo} repository for #{platform} as active.")
+      if bag.fetch(platform, {}).key? repo
+        @logger.debug("Forcefully disabling #{repo} repository for #{platform}.")
+        disable_repository(platform, repo, bag)
+      end
+
+      code = 403
     end
-    saved
+
+    code
   end
 
-  def destroy_repository_item(repo_object)
-    destroyed = false
-    # create DataBag if it doesn't exist yet
-    load_repository_bag
-    # otherwise destroy DataBagItem
-    db_item = repo_object.bag_item
-    ret = db_item.destroy("repositories", db_item.id) rescue nil
-    if ret
-      @logger.debug("Deactivating #{repo_object.registry['name']} repository.")
-      destroyed = true
+  def disable_repository(platform, repo, bag = nil)
+    bag = load_repository_bag if bag.nil?
+
+    repo_object = Crowbar::Repository.where(platform: platform, repo: repo).first
+    if repo_object.nil?
+      @logger.debug("#{repo} repository for #{platform} does not exist.")
+      return 404
+    end
+
+    if bag.fetch(platform, {}).key? repo
+      @logger.debug("Setting #{repo} repository for #{platform} as inactive.")
+      bag[platform].delete(repo)
+      bag.delete(platform) if bag[platform].empty?
+      bag.save
     else
-      @logger.debug("Cannot deactivate #{repo_object.registry['name']} repository.")
+      @logger.debug("#{repo} repository for #{platform} is already inactive.")
     end
-    # destroy the DataBag if it is empty
-    destroy_repository_bag if load_repository_bag.empty?
-    destroyed
-  end
 
-  private
-
-  def create_repository_bag
-    bag = Chef::DataBag.new
-    bag.name "repositories"
-    @logger.debug("Creating DataBag '#{bag.name}'.")
-    bag.save
-  end
-
-  def destroy_repository_bag
-    @logger.debug("No more active repositories. Destroying repositories DataBag.")
-    Chef::DataBag.chef_server_rest.delete_rest("data/repositories")
+    200
   end
 end
--- a/crowbar_framework/app/models/provisioner_service.rb
+++ b/crowbar_framework/app/models/provisioner_service.rb
@@ -205,6 +205,24 @@ class ProvisionerService < ServiceObject
     end
   end
 
+  def enable_all_repositories
+    @logger.debug("Enabling all repositories.")
+    bag = load_repository_bag
+    Crowbar::Repository.check_all_repos.each do |repo|
+      enable_repository(repo.platform, repo.id, bag)
+    end
+  end
+
+  def disable_all_repositories
+    @logger.debug("Disabling all repositories.")
+    bag = load_repository_bag
+    bag.keys.each do |platform|
+      next if platform == "id"
+      bag.delete(platform)
+    end
+    bag.save
+  end
+
   def enable_repository(platform, repo, bag = nil)
     bag = load_repository_bag if bag.nil?
 
--- a/crowbar_framework/app/models/provisioner_service.rb
+++ b/crowbar_framework/app/models/provisioner_service.rb
@@ -228,11 +228,13 @@ class ProvisionerService < ServiceObject
 
     repo_object = Crowbar::Repository.where(platform: platform, repo: repo).first
     if repo_object.nil?
-      @logger.debug("#{repo} repository for #{platform} does not exist.")
-      return 404
+      message = "#{repo} repository for #{platform} does not exist."
+      @logger.debug(message)
+      return [404, message]
     end
 
     code = 200
+    message = ""
 
     if repo_object.available?
       bag[platform] ||= {}
@@ -245,7 +247,8 @@ class ProvisionerService < ServiceObject
         @logger.debug("#{repo} repository for #{platform} is already active.")
       end
     else
-      @logger.debug("Cannot set #{repo} repository for #{platform} as active.")
+      message = "Cannot set #{repo} repository for #{platform} as active."
+      @logger.debug(message)
       if bag.fetch(platform, {}).key? repo
         @logger.debug("Forcefully disabling #{repo} repository for #{platform}.")
         disable_repository(platform, repo, bag)
@@ -254,7 +257,7 @@ class ProvisionerService < ServiceObject
       code = 403
     end
 
-    code
+    [code, message]
   end
 
   def disable_repository(platform, repo, bag = nil)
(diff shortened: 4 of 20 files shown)


crowbar_framework/app/models/node_object.rb
```diff
@@ -171,7 +171,7 @@ def self.all
   end
 
   def self.admin_node
-    all.detect(&:admin?)
+    NodeObject.find("role:crowbar").find { |n| n.admin? }
```
tboerger: `find("role:crowbar").detect(&:admin?)` is what you want.

crowbar_framework/app/views/repositories/index.html.haml
```diff
@@ -33,7 +33,7 @@
           - @repocheck.sort_by{ |check| check.registry["name"] }.each do |repo|
             - if repo.platform == platform
               %tr
-              - if repo.available?
+              - if repo.exists?
```
tboerger: `exist?` is more ruby conventional.

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -531,9 +531,9 @@ en:
       status: 'Status'
       available: 'Available'
       unavailable: 'Not present'
-      md5_check: 'MD5 Check'
-      md5_ok: 'OK'
-      md5_failed: 'Failed'
+      repo_check: 'Repository Check'
```
tboerger: What about just `Check`? It's obvious that it's repo related as you are anyway in the repositories list.

crowbar_framework/config/repos.yml
```diff
@@ -5,169 +5,121 @@ suse-12.0:
       name: "SLES12-Pool"
       required: "mandatory"
       product_name: "common"
-      content_key:
```
tboerger: These changes are totally different to the strongly pushed schema defined within https://etherpad.nue.suse.com/p/cloud-repo-checks at "Repository schema"

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -78,43 +74,42 @@ def repositories(platform)
...
       @platform = platform
       @id = repo
-      @registry = Repository.registry[@platform]["repos"][@id]
+      @config = Repository.registry[@platform]["repos"][@id]
       @url = url
     end
 
-    def available?
-      all_repo_dirs.include?(@registry["name"]) && check_repo_tag
+    def is_remote?
```
tboerger: Just `remote?` is more ruby conventional, like `available?`.

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -78,43 +74,42 @@ def repositories(platform)
...
-      else
-        false
-      end
+    def available?
+      is_remote? || (check_directory && check_repo_tag && check_key_file)
     end
 
-    def repodata_path
-      "/srv/tftpboot/#{@platform}/repos/#{@registry['name']}/repodata"
+    def exists?
```
tboerger: `exist?` is a more current ruby standard.

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -20,12 +20,12 @@ class Repository
```
MaximilianMeister: s/registry/config

crowbar_framework/app/views/repositories/index.html.haml
```diff
@@ -33,7 +33,7 @@
           - @repocheck.sort_by{ |check| check.registry["name"] }.each do |repo|
```
MaximilianMeister: s/registry/config

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -138,16 +134,25 @@ def bag_item
 
     private
 
+    def repos_dir
```
tboerger: To be in same naming... `repos_path` or change the method below to `repodata_dir`.

crowbar_framework/app/views/repositories/index.html.haml
MaximilianMeister: ...and below a couple of times

crowbar_framework/lib/crowbar/repository.rb
tboerger: And just using `Pathname` instead of this string makes other parts a lot easier as well.

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -138,16 +134,25 @@ def bag_item
...
+    end
+
     #
     # validation helpers
     #
-    def all_repo_dirs
-      path = File.join("/srv/tftpboot", @platform, "repos")
-      Dir["#{path}/*"].map { |p| p.split("/").last }
+    def check_directory
+      Dir.exist? File.join(repos_dir, @config["name"])
```
tboerger: Related to above comments: `repos_path.join(@config["name"]).directory?`

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -138,16 +134,25 @@ def bag_item
...
     end
 
     def check_repo_tag
-      expected = @registry["repomd"]["tag"]
+      expected = @config.fetch("repomd", {})["tag"]
```
tboerger: This looks totally weird

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -155,5 +160,18 @@ def check_repo_tag
         false
       end
     end
+
+    def check_key_file
+      expected = @config.fetch("repomd", {})["md5"]
```
tboerger: Same as above

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -155,5 +160,18 @@ def check_repo_tag
...
+      return true if expected.blank?
+
+      key_path = "#{repodata_path}/repomd.xml.key"
```
tboerger: `repodata_path.join("repomd.xml.key")` and `key_path.file?` ;)

crowbar_framework/app/controllers/repositories_controller.rb
```diff
@@ -50,11 +50,12 @@ def sync
   def activate
     return render_not_found if params[:platform].nil? || params[:repo].nil?
     respond_to do |format|
-      repo_object = Crowbar::Repository.where(platform: params[:platform], repo: params[:repo]).first
-      ret = ProvisionerService.new(logger).create_repository_item(repo_object) unless repo_object.nil?
-      if ret
+      ret = ProvisionerService.new(logger).enable_repository(params[:platform], params[:repo])
+      if ret == 200
```
tboerger: Please take an example on other already refactored controllers and do it like:

```ruby
case ret
when 200
when 404
else
end
```

tboerger:
In general I'm really against a single data bag for all repos. That way you can never be sure that the data stays correct, no way for atomic writes. If it's split into separate data bags it stays for sure in a more consistent state as in the other cases. That's why i pushed that implementation for @MaximilianMeister.

crowbar_framework/config/repos.yml
vuntz: First, nothing validating the value of these attributes exists, so it's useless. Second, the etherpad marks this as optional, so it's actually fine to define repos without this attribute.

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -20,12 +20,12 @@ class Repository
...
       end
 
       def registry
-        load! unless defined? @config
-        @config
+        load! unless defined? @all_repos
+        @all_repos
       end
 
       def where(options = {})
```
MaximilianMeister: here is another missing s/registry/config

crowbar_framework/lib/crowbar/repository.rb
vuntz: Why? This just ensures that we can get a value (even nil) if there's no repomd attribute.

crowbar_framework/lib/crowbar/repository.rb
vuntz: done, even though github doesn't show it :-)

crowbar_framework/app/models/provisioner_service.rb
```diff
@@ -179,76 +179,85 @@ def transition(inst, name, state)
...
         case active.to_i
         when 0
-          db = Chef::DataBagItem.load("repositories", repo_object.id) rescue nil
-          # skip repo if it hasn't been active before
-          next if db.nil?
-          destroy_repository_item(repo_object)
+          disable_repository(platform, repo, bag)
         when 1
-          create_repository_item(repo_object)
+          enable_repository(platform, repo, bag)
```
MaximilianMeister: either rename reponame to repo, or change repo to reponame here

aspiers:
Thomas Boerger <notifications@github.com> wrote:
> In general I'm really against a single data bag for all repos. That way you can never be sure that the data stays correct, no way for atomic writes. If it's split into separate data bags it stays for sure in a more consistent state as in the other cases. That's why i pushed that implementation for @MaximilianMeister.

Definitely not disagreeing - it sounds like you have some solid reasons here, but I'm curious and would like to understand them better.  Any chance you could explain a little further? e.g. give an example scenario where a single data bag would lose consistency?

crowbar_framework/lib/crowbar/repository.rb
vuntz: done

crowbar_framework/app/models/provisioner_service.rb
vuntz: done

crowbar_framework/app/models/provisioner_service.rb
```diff
@@ -179,76 +179,85 @@ def transition(inst, name, state)
...
-    if repo_object.available? && repo_object.valid_key_file?
-      @logger.debug("Setting #{repo_object.registry['name']} repository for #{repo_object.platform} as active.")
-      repo_object.to_databag!.save
-      saved = true
-    else
-      @logger.debug("Cannot set #{repo_object.registry['name']} repository for #{repo_object.platform} as active.")
+  def enable_repository(platform, repo, bag = nil)
+    bag = load_repository_bag if bag.nil?
+
+    repo_object = Crowbar::Repository.where(platform: params[:platform], repo: params[:repo]).first
```
MaximilianMeister: the params Hash only works in a Controller, probably copy paste error. should be just platform and repo

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -78,79 +74,97 @@ def repositories(platform)
...
-        "http://#{Repository.admin_ip}:#{Repository.web_port}/#{@platform}/repos/#{@registry['name']}"
+      @config["url"] || \
+        "http://#{Repository.admin_ip}:#{Repository.web_port}/#{@platform}/repos/#{@config["name"]}"
     end
 
     def active?
-      !bag_item.nil?
-    end
+      active_repos = Chef::DataBag.load("crowbar/repositories") rescue {}
+      active_repos ||= {}
```
MaximilianMeister: i think you can drop this line, the above code does already that.

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -78,79 +74,97 @@ def repositories(platform)
...
-    def to_databag!
-      repository_item = Chef::DataBagItem.new
-      repository_item.data_bag "repositories"
-      repository_item["id"] = @id
-      repository_item["platform"] = @platform
-      repository_item["name"] = @registry["name"]
-      repository_item["url"] = url
-      repository_item["product_name"] = @registry["product_name"]
-      repository_item
+      active_repos.fetch(@platform, {}).key? @id
```
MaximilianMeister: that should be `@config["name"]` instead of `@id`

crowbar_framework/lib/crowbar/repository.rb
vuntz: Nope, if the `load` returns `nil`, the `rescue` won't help.

crowbar_framework/lib/crowbar/repository.rb
vuntz: Hrm, I need to run the code to check, but I think the code is correct.

vuntz:
> In general I'm really against a single data bag for all repos. That way you can never be sure that the data stays correct, no way for atomic writes. If it's split into separate data bags it stays for sure in a more consistent state as in the other cases.

Interesting. That being said, how is the way I've picked different from anything else we're currently doing with data bags (or, for that matter, with a database table)?

I'm happy to find yet another way of storing the data, but the old way was broken as soon as we add multiple architectures.

tboerger:
> Definitely not disagreeing - it sounds like you have some solid reasons here, but I'm curious and would like to understand them better.  Any chance you could explain a little further? e.g. give an example scenario where a single data bag would lose consistency?

The most simple example? Two processes try to update different repos at the same time?

crowbar_framework/lib/crowbar/repository.rb
tboerger: What about `active_repos = Chef::DataBag.load("crowbar/repositories") || {}`?

tboerger:
> I'm happy to find yet another way of storing the data, but the old way was broken as soon as we add multiple architectures.

Maybe for the ids of the databags, but it's always safe to write with multiple processes at the same time for different repos.

crowbar_framework/lib/crowbar/repository.rb
MaximilianMeister: Either it crashes with a HTTP exception or it gives you back a databag, AFAICS it never returns nil

crowbar_framework/lib/crowbar/repository.rb
MaximilianMeister: I have checked that already

crowbar_framework/lib/crowbar/repository.rb
tboerger:
```
> Chef::DataBag.load("crowbar/repositories")                                                                                                                                                
Net::HTTPServerException: 404 "Not Found"
```

chef/cookbooks/provisioner/libraries/repositories.rb
```diff
@@ -74,12 +74,9 @@ def inspect_repos(node)
       # they can be used.
       def get_repos(platform, version)
         repositories = {}
-        repos_from_databag = Chef::DataBag.load("repositories").keys rescue []
-        repos_from_databag.each do |reponame|
-          dbi = Chef::DataBagItem.load("repositories", reponame)
-          if dbi["platform"] == "#{platform}-#{version}"
-            repositories[dbi["name"]] = { url: dbi["url"] }
-          end
+        repos_db = Chef::DataBag.load("crowbar/repositories") || {}
```
tboerger: https://github.com/crowbar/crowbar-core/pull/102#discussion_r42999096 shows that you have to catch a `Net::HTTPServerException` if the data bag doesn't exist. If I look at the chef code I can not see a case where it returns `nil`.

crowbar_framework/lib/crowbar/repository.rb
vuntz: Thanks, done.

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -63,94 +97,134 @@ def check_all_repos
...
-    def repomd_key_md5
-      @registry["repomd"]["md5"]
+    def valid_key_file?
+      remote? || check_key_file
     end
 
-    def repomd_key_path
-      @registry["repomd"]["key"] || "#{repodata_path}/repomd.xml.key"
+    def name
+      @config["name"] || @id
```
MaximilianMeister: related to https://github.com/crowbar/crowbar-core/pull/102#discussion_r42997204:

`@id` is always the key from the yml file which is lower case, `@config["name"]` is the actual uppercase name, so that is not correct here

vuntz: Can you clarify what is not correct? The goal of the commit is to make the name attribute optional (so that we're a bit more solid when parsing yaml files written by non-developers). If the name is specified, then it will be used.

MaximilianMeister: I have only seen that you still didn't change the `active?` method, that's why I commented this part. but if it is not really related, and that change is wanted here then it's a different story.

crowbar_framework/lib/crowbar/repository.rb
vuntz: fixed

vuntz:
Ok, I think this is ready now. The only missing bit is the API to active all repos that are available.

@MaximilianMeister could you play with it a bit and see if things work as intended for you?

vuntz:
Btw, I'm not 100% sure the repo page behavior is 100% correct, but that's unrelated to the changes in this PR. For instance, failed repo checks for optional repo are not marked as "failed", which is not intuitive.

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -48,109 +82,197 @@ def all_platforms
...
+      repos_path.join(name).directory?
     end
 
     def check_repo_tag
-      expected = @registry["repomd"]["tag"]
-      repomd_path = "#{repodata_path}/repomd.xml"
-      if File.exist?(repomd_path)
-        REXML::Document.new(File.open(repomd_path)).root.elements["tags/repo"].text == expected
+      expected = @config.fetch("repomd", {})["tag"]
+      return true if expected.blank?
```
MaximilianMeister: maybe we should show "Skipped" instead of faking a successful check here

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -48,109 +82,197 @@ def all_platforms
...
+          !repo_tag.nil? && repo_tag.text == expected
+        end
+      else
+        false
+      end
+    end
+
+    def check_key_file
+      expected = @config.fetch("repomd", {})["key_md5"]
+      return true if expected.blank?
```
MaximilianMeister: same here with "Skipped"

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -48,109 +82,197 @@ def all_platforms
...
     def check_repo_tag
-      expected = @registry["repomd"]["tag"]
-      repomd_path = "#{repodata_path}/repomd.xml"
-      if File.exist?(repomd_path)
-        REXML::Document.new(File.open(repomd_path)).root.elements["tags/repo"].text == expected
+      expected = @config.fetch("repomd", {})["tag"]
+      return true if expected.blank?
+
+      repomd_path = repodata_path.join("repomd.xml")
+      repomd_path = repodata_media_path.join("repomd.xml") unless repomd_path.exist?
```
MaximilianMeister: maybe just `repomd_path ||= repodata_media_path.join("repomd.xml")`

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -48,109 +82,197 @@ def all_platforms
...
+        false
+      end
+    end
+
+    def check_key_file
+      expected = @config.fetch("repomd", {})["key_md5"]
+      return true if expected.blank?
+
+      key_path = repodata_path.join("repomd.xml.key")
+      key_path = repodata_media_path.join("repomd.xml.key") unless key_path.exist?
```
MaximilianMeister: same here with `key_path ||= repodata_media_path.join("repomd.xml")`

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -16,16 +16,50 @@
...
+          @all_repos = YAML.load_file(Rails.root.join("config/repos-cloud.yml"))
+        end
+
+        # merge data from etc config file
+        etc_repos = {}
+        if File.exist? etc_yml
+          begin
+            etc_repos = YAML.load_file(etc_yml)
+          rescue SyntaxError
+            # ok, let's live without it
```
MaximilianMeister: i would put at least a log message here. wouldn't it run it that rescue all over again without anyone noticing?

chef/cookbooks/provisioner/libraries/repositories.rb
```diff
@@ -17,69 +17,20 @@
...
-            repositories[dbi["name"]] = { url: dbi["url"] }
-          end
+
+        begin
+          repos_db = Chef::DataBag.load("crowbar/repositories")
+        rescue Net::HTTPServerException
+          repos_db = {}
+        end
+
+        repos_db.fetch("#{platform}-#{version}", {}).each do |id, config|
```
MaximilianMeister: why use a fetch with another `{}` "rescue" here?

chef/cookbooks/provisioner/libraries/repositories.rb
```diff
@@ -17,69 +17,20 @@
...
+          repositories[config["name"]] = { url: config["url"], ask_on_error: config["ask_on_error"] }
```
MaximilianMeister: for what do we need the `ask_on_error` exactly? it is not mentioned in https://github.com/vuntz/crowbar-core/commit/b6968b16071e556d8e017e40cfeadb1d9bfef960 why we need it

chef/cookbooks/provisioner/libraries/repositories.rb
```diff
@@ -17,69 +17,20 @@
 class Provisioner
   class Repositories
     class << self
-      def inspect_repos(node)
```
MaximilianMeister: I understand dropping this from the crowbar perspective, but the cookbooks don't need to know about the availability anymore?

chef/cookbooks/provisioner/libraries/repositories.rb
```diff
@@ -17,69 +17,20 @@
...
       def get_repos(platform, version)
         repositories = {}
-        repos_from_databag = Chef::DataBag.load("repositories").keys rescue []
-        repos_from_databag.each do |reponame|
-          dbi = Chef::DataBagItem.load("repositories", reponame)
-          if dbi["platform"] == "#{platform}-#{version}"
-            repositories[dbi["name"]] = { url: dbi["url"] }
-          end
+
+        begin
```
tboerger: I think you can directly write

```
        repos_db = begin
          Chef::DataBag.load("crowbar/repositories")
        rescue Net::HTTPServerException
          {}
        end
```

chef/cookbooks/provisioner/libraries/repositories.rb
tboerger: Implementing a customized Hash makes pretty sure sense to avoid all these fetch calls overall.

crowbar_framework/app/controllers/repositories_controller.rb
```diff
@@ -50,11 +50,13 @@ def sync
   def activate
     return render_not_found if params[:platform].nil? || params[:repo].nil?
     respond_to do |format|
-      repo_object = Crowbar::Repository.where(platform: params[:platform], repo: params[:repo]).first
-      ret = ProvisionerService.new(logger).create_repository_item(repo_object) unless repo_object.nil?
-      if ret
+      ret = ProvisionerService.new(logger).enable_repository(params[:platform], params[:repo])
```
tboerger: If you are taking this strange behavior of Crowbar you should at least return a code and a message.

crowbar_framework/app/controllers/repositories_controller.rb
```diff
@@ -95,9 +99,35 @@ def deactivate
...
   end
 
+  #
+  # Activate all Repositories
+  #
+  # Provides the restful api call for
+  # Activate all Repositories   /utils/repositories/activate_all   POST  Creates Repository DataBagItem
+  def activate_all
+    respond_to do |format|
+      ProvisionerService.new(logger).enable_all_repositories
```
tboerger: This should be outside of the respond_to block. Maybe even for the other actions as well.

MaximilianMeister:
I also tried to enable/disable the repos with some curl requests, but this doesn't seem to work anymore. I haven't checked too closely why though....

MaximilianMeister:
I tried to overall test everything, but I haven't tested the `provided_and_required` method and the `/etc/crobar/repos.yml` feature.

there are some really good new features like the more specific error displays (haven't tested them though) and I didn't look too closely on the haml changes, @tboerger can maybe have a closer look.

MaximilianMeister:
Ah and I am still not convinced why to make the name optional, especially because we are depending on the name in the url for example and in some other places in the `repository` lib. making it optional feels a bit confusing and maybe could lead to arbitrary behavior at some point (at least that's my feeling)

MaximilianMeister:
other than that it looks good to me, maybe you could squash some commits like the pathname, minor cleanup, trailing slash and some yml related changes but that is up to you of course. now i will close my 41 review tabs ;-)

but overall thanks for taking this over, it was definitely quicker and more efficient this way.

MaximilianMeister:
> I also tried to enable/disable the repos with some curl requests, but this doesn't seem to work anymore. I haven't checked too closely why though....

maybe we should change the routes that don't require any parameters, like `activate_all` to `GET`
I have tested that and it works as expected

crowbar_framework/lib/crowbar/repository.rb
vuntz: That won't work, `repomd_path` is not `nil`.

crowbar_framework/lib/crowbar/repository.rb
vuntz: That won't work, `key_path` is not `nil`.

chef/cookbooks/provisioner/libraries/repositories.rb
vuntz: No, they don't.

chef/cookbooks/provisioner/libraries/repositories.rb
vuntz: It's used in the autoyast template...

chef/cookbooks/provisioner/libraries/repositories.rb
vuntz: @MaximilianMeister to avoid `nil.each`, in case the platform doesn't exist in the db.

vuntz:
> I also tried to enable/disable the repos with some curl requests, but this doesn't seem to work anymore. I haven't checked too closely why though....

Hrm, weird. That works here. Would you have time to debug?

vuntz:
> maybe we should change the routes that don't require any parameters, like activate_all to GET
> I have tested that and it works as expected

So I don't know what's the "right" way here, I'm happy to change to `GET` if that's the correct way.

Btw, which `curl` command were you using?

vuntz:
> Ah and I am still not convinced why to make the name optional, especially because we are depending on the name in the url for example and in some other places in the repository lib. making it optional feels a bit confusing and maybe could lead to arbitrary behavior at some point (at least that's my feeling)

It just means that people can put the name directly as the id, really. The point of this is to make it easier for people to write the `/etc/crowbar/repos.yml` file.

MaximilianMeister:
> Btw, which curl command were you using?

you're right, the `activate_all` route works as it is, I had given the wrong port for my dev setup

but the single activate routes dont work for me. try e.g.

`curl -X POST -d "platform=suse-12.1" -d "repo=SLES12-SP1-Updates" http://localhost:3000/utils/repositories/activate`

vuntz:
@MaximilianMeister that's because the repo is an id now. I can actually change code to match on both id and name...

crowbar_framework/lib/crowbar/repository.rb
vuntz: I was about to implement that, but I'm afraid that "Skipped" sends the wrong message to the user: it will make them feel that something is broken, while that's really how the repos are configured. And the issue is that it would happen by default because the PTF repo has no check.

So I'm leaning towards not changing this for now.

chef/cookbooks/provisioner/libraries/repositories.rb
MaximilianMeister: but as we already found out, `Chef::DataBag.load` doesn't return nil in any case, that is why i was wondering

chef/cookbooks/provisioner/libraries/repositories.rb
```diff
@@ -17,69 +17,20 @@
...
-            repositories[dbi["name"]] = { url: dbi["url"] }
-          end
+
+        repos_db = begin
+          Chef::DataBag.load("crowbar/repositories")
+        rescue Net::HTTPServerException
+          {}
+        end
+
+        repos_db.fetch("#{platform}-#{version}", {}).each do |id, config|
```
vuntz: Replying to https://github.com/crowbar/crowbar-core/pull/102/files#r43094115

The issue is not `Chef::DataBag.load` returning `nil`, but `repos_db["foo"]` returning `nil`.

vuntz:
> @MaximilianMeister that's because the repo is an id now. I can actually change code to match on both id and name...

Done.

crowbar_framework/app/views/repositories/index.html.haml
```diff
@@ -30,49 +30,68 @@
                 %i.glyphicon.glyphicon-question-sign{ data: { tooltip: true, placement: "top" }, title: t(".active_repo_tooltip") }
 
           %tbody
-          - @repocheck.sort_by{ |check| check.registry["name"] }.each do |repo|
+          - @repocheck.sort{ |a, b| [repository_required_to_i(a.required), a.name] <=> [repository_required_to_i(b.required), b.name] }.each do |repo|
```
tboerger: This should be a helper method or part of the model as it's quite complex now.

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -530,10 +531,13 @@ en:
       required: 'Required'
       status: 'Status'
       available: 'Available'
+      stale: 'Changed'
       unavailable: 'Not present'
-      md5_check: 'MD5 Check'
-      md5_ok: 'OK'
-      md5_failed: 'Failed'
+      check: 'Repository Check'
```
tboerger: I were talking about the translation to `Check`and not about the key :)

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -63,94 +97,136 @@ def check_all_repos
...
 
-      def repositories(platform)
-        registry[platform]["repos"].keys
+          repositories(platform).each do |repo|
+            provided_features = registry[platform]["repos"][repo]["features"] || []
+
+            next unless provided_features.include? feature
+            found = true
+
+            r = self.new(platform, repo)
```
tboerger: That is indeed redundant. Just `new()´

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -48,109 +82,197 @@ def all_platforms
...
+        answer
       end
 
-      def web_port
-        Proposal.where(barclamp: "provisioner").first.raw_attributes["web_port"]
+      def platform_available?(platform)
+        available = true
+
+        repositories(platform).each do |repo|
+          r = self.new(platform, repo)
```
tboerger: Same as above.

crowbar_framework/lib/crowbar/repository.rb
tboerger: `repomd_path ||=` is totally different. It's setting it only if the file exists.

tboerger: Ah, but you want to check for `file?`instead of `exist?`.

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -48,109 +82,197 @@ def all_platforms
...
-      repomd_path = "#{repodata_path}/repomd.xml"
-      if File.exist?(repomd_path)
-        REXML::Document.new(File.open(repomd_path)).root.elements["tags/repo"].text == expected
+      expected = @config.fetch("repomd", {})["tag"]
+      return true if expected.blank?
+
+      repomd_path = repodata_path.join("repomd.xml")
+      repomd_path = repodata_media_path.join("repomd.xml") unless repomd_path.exist?
+
+      if repomd_path.exist?
```
tboerger: Same as above, check for `file?`

crowbar_framework/lib/crowbar/repository.rb
tboerger: Same as above, check for `file?`

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -48,109 +82,197 @@ def all_platforms
...
+    end
+
+    def check_key_file
+      expected = @config.fetch("repomd", {})["key_md5"]
+      return true if expected.blank?
+
+      key_path = repodata_path.join("repomd.xml.key")
+      key_path = repodata_media_path.join("repomd.xml.key") unless key_path.exist?
+
+      if key_path.exist?
```
tboerger: Same as above, check for `file?`

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -48,109 +82,197 @@ def all_platforms
...
+        md5 = Digest::MD5.hexdigest(File.read(key_path.to_s))
```
tboerger: `md5 = Digest::MD5.hexdigest(key_path.read)

crowbar_framework/config/routes.rb
```diff
@@ -75,6 +75,8 @@
   post "utils/repositories/sync", controller: "repositories", action: "sync", as: "sync_repositories"
   post "utils/repositories/activate", controller: "repositories", action: "activate", as: "activate_repository"
   post "utils/repositories/deactivate", controller: "repositories", action: "deactivate", as: "deactivate_repository"
+  post "utils/repositories/activate_all", controller: "repositories", action: "activate_all", as: "activate_all_repositories"
+  post "utils/repositories/deactivate_all", controller: "repositories", action: "deactivate_all", as: "deactivate_all_repositories"
```
tboerger: Maybe it's even better to update the routes to something like:

```ruby
get "utils/repositories/activate/all", controller: "repositories", action: "activate_all", as: "activate_all_repositories"
get "utils/repositories/activate/:platform/:repo", controller: "repositories", action: "activate", as: "activate_repository"
get "utils/repositories/deactivate/all", controller: "repositories", action: "deactivate_all", as: "deactivate_all_repositories"
get "utils/repositories/deactivate/:platform/:repo", controller: "repositories", action: "deactivate", as: "deactivate_repository"
```

Than you can even drop the check if the required params are set.

tboerger: Writing it in form of `resource`route definition would hurt much lesser in the eyes as well, but that can be something for a separate pr.

tboerger:
In general there are good improvements included, but from my side the kind of storing everything in a single data bag is :-1: from my side. As explained above we can run into race conditions because there is no way (other than working with even more locking files which sucks pretty hard because of poor design) to avoid it with a single data bag.

crowbar_framework/config/routes.rb
vuntz: @tboerger I don't have any feeling on this, I just thought that `POST` was more REST here (and I guess @MaximilianMeister felt the same for the initial code).

vuntz:
@tboerger I understand your concern about this, but I would really prefer we try to fix this after we merge this and then the multi-arch work. The old way is just not compatible with multi-arch, and this means a new solution has to be found.

I'm actually fine putting some locking for this in the mean time if you're really concerned that this is going to be hit by people.

For the record, my guess is that the real fix is actually to not use the data bag as the place where to store the data, but use the database for this instead (updating the database shouldn't have the same issues, as we can update only one row) and then mirror what's in the database in the databag (which shouldn't be subject to any concurrency issue) so that it's accessible from chef-client.

tboerger:
> For the record, my guess is that the real fix is actually to not use the data bag as the place where to store the data, but use the database for this instead (updating the database shouldn't have the same issues, as we can update only one row) and then mirror what's in the database in the databag (which shouldn't be subject to any concurrency issue) so that it's accessible from chef-client.

That's exactly what I suggested 2 months ago ;)

vuntz:
> That's exactly what I suggested 2 months ago ;)

No, you suggested replacing the yaml file with the database. I'm merely suggesting to store whether a repo is active or not in the database, no replacing the yaml file.

vuntz:
For the record, CI can't be green because the crowbar package is outdated, so we can't build this package. Also this depends on https://github.com/crowbar/crowbar/pull/2172

MaximilianMeister:
>>That's exactly what I suggested 2 months ago ;)

> No, you suggested replacing the yaml file with the database. I'm merely suggesting to store whether a repo is active or not in the database, no replacing the yaml file.

at some point, maybe when we make the repos configurable from the UI, the repos will end up in the database, that's my guess, even if it is only the dynamic overlay, and the static yml file stays the same. (I would prefer an API for that instead of yml, but that will be discussed again later anyways)

we only wanted to go that direction already with a real model, and now the lib is not used by chef directly, which was one reason why we ended up with a lib instead of a model, I think that is the issue here.

But apart from that the code works good as it is now, haven't tested all the internals though, some more specs would be cool, but for me it is fine and mergable :+1:

MaximilianMeister:
@vuntz one more thing, shall we add the debuginfo repos as optional? I remember this was suggested some time ago

vuntz:
> @vuntz one more thing, shall we add the debuginfo repos as optional? I remember this was suggested some time ago

Nearly everything we ship is python or ruby, so it's not really useful, imho. They can be added to the etc yaml file, when needed.

crowbar_framework/config/routes.rb
tboerger: I think initially more because of the parameters ;)

vuntz:
Let's bite the bullet and check this in.

tboerger:
I really dislike that this simply gets merged even with a :-1:.

vuntz:
Just for the record: I already have a patch that should fix @tboerger's concern, that sits on top of https://github.com/crowbar/crowbar-core/pull/83 (will open the PR after the multi-arch bit is merged, current commit is https://github.com/crowbar/crowbar-core/commit/bd39cc05fc133e5f225a6c99e3e780942f24c498).

PR #103, opened 2015-10-27
Cleanup leftovers from switch to puma

rhafer:
This removes some things that are not needed (or used) anymore since we
switched to the puma webserver.


vuntz:
+1

tboerger:
:+1:

vuntz:
Deployment worked, and tempest passed (even if later test failed) => good enough for merging this.

PR #107, opened 2015-10-29
fix locale after theming update

jsuchome:
This should fix recent installation failures (see https://trello.com/c/cXXNRuVe/64-actionview-template-error-hash-does-not-contain-any-of-parameters-text-icon).

@tboerger suggested this, I'm creating it so it can be merged ASAP


vuntz:
This was a bug in the branding patch, it's already fixed there.

jsuchome:
Good.

PR #109, opened 2015-10-30
Go back to one data bag item per repository

vuntz:
cc @tboerger @MaximilianMeister

### DIFF
--- a/crowbar_framework/app/models/provisioner_service.rb
+++ b/crowbar_framework/app/models/provisioner_service.rb
@@ -178,29 +178,15 @@ class ProvisionerService < ServiceObject
     [200, { name: name }]
   end
 
-  def load_repository_bag
-    bag = Chef::DataBag.load("crowbar/repositories") rescue nil
-    if bag.nil?
-      @logger.debug("Creating repositories data bag")
-      bag = Chef::DataBagItem.new
-      bag.data_bag "crowbar"
-      bag["id"] = "repositories"
-      bag.save
-    end
-    bag
-  end
-
   def synchronize_repositories(platforms)
-    bag = load_repository_bag
-
     platforms.each do |platform, arches|
       arches.each do |arch, repos|
         repos.each do |repo, active|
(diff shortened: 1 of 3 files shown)


aspiers:
The first commit is missing a justification for the change.  In particular I'm curious to understand why having multiple data bags prevents the race conditions @tboerger mentioned.

Also I'm wondering how the second commit is related to this PR?

vuntz:
> Also I'm wondering how the second commit is related to this PR?

It's not, I moved it to https://github.com/crowbar/crowbar-core/pull/114

vuntz:
> The first commit is missing a justification for the change.

I improved the commit message for this.

crowbar_framework/app/models/provisioner_service.rb
```diff
@@ -279,16 +267,25 @@ def disable_repository(platform, arch, repo, bag = nil)
...
       @logger.debug("#{repo_id} repository for #{platform} / #{arch} is already inactive.")
     end
 
     [200, ""]
   end
+
+  private
+
+  # workaround for Chef::DataBag.destroy not working
+  def chef_data_bag_destroy(name)
```
tboerger: Why do you need that? My tests showed that destroying a data bag works fine.

vuntz: Interesting, that doesn't work for me. Can you try the patch and use the proper destroy method to see if that works for you?

tboerger: You are right... Maybe i mixed it up with data_bag items. For reference: https://github.com/chef/chef/blob/10-stable/chef/lib/chef/knife/data_bag_delete.rb#L39

vuntz:
@tboerger reminder: you wanted this, so maybe a +1/-1 would be good ;-)

tboerger:
Can not fine real mistakes. :+1:

rsalevsky:
:+1:

PR #110, opened 2015-10-30
Update rspec conf to get better output

rsalevsky:
This change helps to get a better overview which checks are failing and
it's at the same time a kind of documentation.


vuntz:
+1, that's indeed an improvement.

tboerger:
:+1:

PR #111, opened 2015-10-30
provisioner: Cleanup path to kernel/initrd used in boot files

vuntz:
The current setup happened to work, but only because grub2 was nice
enough to understand "(tftp)/discovery/../../../" as "(tftp)/". Let's
play it a bit safer.


jsuchome:
+1 (tested forgetting + new discovery)

tboerger:
:+1: but have not tested it.

PR #112, opened 2015-11-02
provisioner: Fix possible crash in cookbook when node has no ohai data

vuntz:
If a node has no ohai data for some reason (getting discovered, but ohai
not running in sleshammer for one reason or another), we don't want the
provisioner-server chef-client to crash. So provide a fallback
architecture for that node.


aplanas:
+1

jsuchome:
+1

PR #113, opened 2015-11-02
Fixes to avoid issues with discovery image

vuntz:


vuntz:
It's unfortunately still not enough, but it does help a bit and this makes sense anyway imho, so removing the WIP tag.

rhafer:
+1

jsuchome:
+1

aplanas:
No idea about this command, but looks like that also there is this version: `ifcfg=*=dhcp4` or `ifcfg=eth*=dhcp4`. Not sure if one of the previous can be more accurate for this solution.

vuntz:
@aplanas I don't know either, I just copied what we use in the update_nodes recipe :-)

PR #114, opened 2015-11-02
provisioner: Totally ignore PTF repo in crowbar_register when disabled

vuntz:
crowbar_register tries to auto import the GPG key of the PTF repo, but
that doesn't make sense when the repo is disabled.


aplanas:
+1

jsuchome:
+1

tboerger:
:+1:

PR #116, opened 2015-11-03
Start crowbar server without initial chef run

rsalevsky:


MaximilianMeister:
:+1: assuming it works, havent tested

MaximilianMeister:
maybe add the "do not merge yet" tag, as we need to adapt `install-chef-suse`

rsalevsky:
@MaximilianMeister We can merge it, the migration is then just twice but this doesn't hurt.

tboerger:
:+1:

vuntz:
Does that really work? I mean, the webui will just display an error because it won't be able to connect to chef, won't it?

tboerger:
> Does that really work? I mean, the webui will just display an error because it won't be able to connect to chef, won't it?

Why should it do that? For node views that is clear, but you won't get there after merging the install wizard pull request.

rsalevsky:
This PR is also just preparation for the packing part until this is not done we can't start the server anyway.

crowbar_framework/config/initializers/migration.rb
```diff
@@ -0,0 +1,32 @@
...
+      )
+    )
+    if migrator.pending_migrations.any?
+      puts "Migrating database schema to #{migrator.pending_migrations.last.version}..."
+      ActiveRecord::Migrator.migrate(
+        ActiveRecord::Migrator.migrations_paths
+      )
+    end
+  end
+end
```
vuntz: Don't you need to also generate /opt/dell/crowbar_framework/config/secrets.env?

vuntz:
> until this is not done we can't start the server anyway.

So I guess I was confused by the PR title ;-)

crowbar_framework/config/initializers/migration.rb
rsalevsky: Yes, it's part of the package `%post` section because this needs to be done before the crowbar server start.

PR #119, opened 2015-11-04
Enable a first time installer for crowbar

MaximilianMeister:
This page should be displayed after the first boot of the crowbar app.
From there the user can trigger `install-chef-suse` and get an overview about the progress.

this depends on a few things:
* crowbar needs to be able to start without the initial chef client run https://github.com/crowbar/crowbar-core/pull/116
* install-chef-suse needs to be aware of a `crowbar` mode https://github.com/crowbar/crowbar/pull/2178


MaximilianMeister:
first_install

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,45 @@
+$(document).ready(function() {
+  function statusCheck() {
+    $.ajax({
+      type: "GET",
+      dataType: "json",
+      url: "/installer/status.json",
```
tboerger: We've got js-routes, so use the routes please.

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,45 @@
...
+      success: function(data)
+      {
+        var failed = data.failed;
+        data = data.steps;
+        var mostRecent = $("li." + data[data.length-1]);
+        var mostRecentIcon = mostRecent.children();
+
+        $.each(data, function (index, value) {
+          $("li." + value).removeClass("list-group-item-success");
+          $("li." + value).children("span").removeClass("fa-hourglass-o fa-circle-o-notch fa-spin");
```
tboerger: You know about chaining jquery methods?

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,45 @@
...
+          mostRecent.next("li").addClass("list-group-item-danger");
+          mostRecent.next("li").children("span").removeClass("fa-hourglass-o fa-circle-o-notch fa-spin");
+          mostRecent.next("li").children("span").addClass("fa-times");
+        } else {
+          mostRecent.next("li").addClass("list-group-item-success");
+          mostRecent.next("li").children("span").removeClass("fa-hourglass-o");
+          mostRecent.next("li").children("span").addClass("fa-circle-o-notch fa-spin");
+        }
+      },
+      error: function(data) {
```
tboerger: You need to define something on error, maybe some flash message?

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,45 @@
...
+      }
+    });
+  }
+  // call statusCheck to not wait for setInterval to kick in
+  statusCheck();
+  setInterval(statusCheck, 3000);
```
tboerger: Overall you need to match for some specific element, else the installer ajax will run forever, even after the installation.

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -24,6 +24,7 @@ class ApplicationController < ActionController::Base
   rescue_from Crowbar::Error::ChefOffline, with: :chef_is_offline
 
   before_filter :enable_profiler, if: Proc.new { ENV["ENABLE_PROFILER"] == "true" }
+  before_filter :crowbar_mode
```
tboerger: Just look one line above how we have done it for the profiler.

crowbar_framework/app/controllers/docs_controller.rb
```diff
@@ -17,6 +17,7 @@
 
 class DocsController < ApplicationController
   before_filter :generate_books
+  skip_before_filter :crowbar_mode
```
tboerger: I don't believe that you need to disable it for the docs controller.

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,135 @@
...
+        format.html { redirect_to installer_url }
+      end
+    end
+  end
+
+  #
+  # Set the status of an installation step
+  #
+  # Provides the restful api call for
+  # /installer/set_status 	POST 	set the status
```
tboerger: Why `/installer/set_status POST`? It's different than `/installer/status GET`, so just do `/installer/status POST` instead of `set_status`.

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  def steps_done
+    steps = all_steps.select do |step|
+      File.exist?(Rails.root.join("..", ".#{step}"))
+    end
+    { steps: steps, failed: failed? }
+  end
+
+  def failed?
+    File.exist?(Rails.root.join(".crowbar-install-failed"))
```
tboerger: Oh come on, please read the docs about Pathname... `Rails.root.join(".crowbar-install-failed").exist?`

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,135 @@
...
+  end
+
+  def successful?
+    File.exist?(Rails.root.join(".crowbar-installed-ok"))
```
tboerger: Same as above

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,135 @@
...
+      :apply_crowbar_config,
+      :transition_crowbar,
+      :chef_client_daemon,
+      :post_sanity_checks
+    ]
+  end
+
+  def steps_done
+    steps = all_steps.select do |step|
+      File.exist?(Rails.root.join("..", ".#{step}"))
```
tboerger: It's a `Pathname`...

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,135 @@
...
+  def failed?
+    File.exist?(Rails.root.join(".crowbar-install-failed"))
+  end
+
+  def successful?
+    File.exist?(Rails.root.join(".crowbar-installed-ok"))
+  end
+
+  def write_file(path, filename)
+    if FileUtils.touch("#{path}/#{filename}")
```
tboerger: `path` is a `Pathname` as well.

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,135 @@
...
+      return [200, ""]
+    else
+      return [404, ""]
```
tboerger: When should that happen?

crowbar_framework/config/initializers/assets.rb
```diff
@@ -22,6 +22,7 @@
     /application\.(css|js)$/,
     /ie\.(css|js)$/,
     /branding\.(css|js)$/,
-    /\.(jpg|png|gif|svg|ico|eot|woff|woff2|ttf)$/
+    /\.(jpg|png|gif|svg|ico|eot|woff|woff2|ttf)$/,
+    /installer\.(css|js)$/
```
tboerger: Why? You put it anyway into `application.(js|css)`

crowbar_framework/config/initializers/installer.rb
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+Rails.application.config.tap do |config|
+  unless File.exist?(Rails.root.join(".crowbar-installed-ok")) || ENV["RAILS_ENV"] == "test"
+    ENV["CROWBAR_MODE"] = "installer"
```
tboerger: So this env var is only used within the before filter, why do we need to define it than?

crowbar_framework/config/routes.rb
```diff
@@ -123,5 +123,10 @@
   get "crowbar/:barclamp", action: "versions", controller: "barclamp"
   post "crowbar/:barclamp/1.0/:action/:id", controller: "barclamp"
 
+  get "installer", controller: "installer", action: "index", as: "installer"
+  get "installer/status", controller: "installer", action: "status", as: "installer_status"
+  post "installer/set_status", controller: "installer", action: "set_status", as: "installer_set_status"
```
tboerger: `post "installer/status", controller: "installer", action: "update", as: "update_installer_status"`

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,45 @@
...
+        var mostRecentIcon = mostRecent.children();
+
+        $.each(data, function (index, value) {
+          $("li." + value).removeClass("list-group-item-success");
+          $("li." + value).children("span").removeClass("fa-hourglass-o fa-circle-o-notch fa-spin");
+          $("li." + value).children("span").addClass("fa-check");
+        });
+
+        mostRecentIcon.removeClass(function (index, css) {
+          match = css.match(/^fa-/g);
```
tboerger: `var match`

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,45 @@
...
+          if (match !== null) {
+            return match.join("");
+          }
+        });
+        mostRecentIcon.addClass("fa-check");
+
+        if (failed) {
+          mostRecent.next("li").addClass("list-group-item-danger");
```
tboerger: Something like this should work and there is no need to lookup the elements again and again:

```
mostRecent
  .next("li")
  .addClass("list-group-item-danger")
  .find('span')
  .removeClass("fa-hourglass-o fa-circle-o-notch fa-spin")
  .addClass("fa-times");
```

crowbar_framework/app/assets/stylesheets/installer.scss
```diff
@@ -0,0 +1,16 @@
+.installer {
```
tboerger: I would suggest to avoid this custom CSS, I don't think that we really need it with the proper usage of bootstrap elements.

MaximilianMeister:
after_install

crowbar_framework/config/initializers/installer.rb
rsalevsky: We need it also for the backup mode.

rsalevsky:
Can you rewrite the commit messages? They seams temporary to me, maybe you can also combine some commits.

crowbar_framework/app/assets/javascripts/application.js
```diff
@@ -8,6 +8,7 @@
 //= require fontawesome
 //= require_self
 //= require branding
+//= require installer
```
tboerger: Branding should always stay at the last place.

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,69 @@
+$(document).ready(function() {
+  $("body.installer").each(function() {
+    function statusCheck() {
+      $.ajax({
+        type: "GET",
+        dataType: "json",
+        url: Routes.installer_status_path(),
+        success: function(data)
+        {
```
tboerger: This should go to the line above

crowbar_framework/config/initializers/installer.rb
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+Rails.application.config.tap do |config|
+  unless Rails.root.join(".crowbar-installed-ok").exist? || ENV["RAILS_ENV"] == "test"
+    ENV["CROWBAR_MODE"] = "installer"
```
tboerger: I still can't see a real requirement for that.

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,142 @@
...
+    Rails.root.join(".crowbar-install-failed").exist?
+  end
+
+  def successful?
+    Rails.root.join(".crowbar-installed-ok").exist?
+  end
+
+  def write_file(path, filename)
+    if FileUtils.touch("#{path.to_path}/#{filename}")
+      return [200, ""]
```
tboerger: I think you can just remove this anyway weird return code stuff.

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,142 @@
...
+    ]
+  end
+
+  def crowbar_lib_dir
+    Pathname.new("/var/lib/crowbar")
+  end
+
+  def steps_done
+    steps = all_steps.select do |step|
+      File.exist?("#{crowbar_lib_dir.to_path}/#{step}")
```
tboerger: `crowbar_lib_dir.join(step).exist?`

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,142 @@
...
+  def failed?
+    Rails.root.join(".crowbar-install-failed").exist?
+  end
+
+  def successful?
+    Rails.root.join(".crowbar-installed-ok").exist?
+  end
+
+  def write_file(path, filename)
+    if FileUtils.touch("#{path.to_path}/#{filename}")
```
tboerger: Simply `FileUtils.touch path.join(filename)` and nothing else.

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,142 @@
...
+    crowbar_dir = Rails.root.join("..")
+    if installed_ok.exist?
+      msg = "If you want to reinstall please remove /opt/dell/crowbar_framework/crowbar-installed-ok"
+      respond_to do |format|
+        format.json { render text: msg and return }
+        format.html { redirect_to installer_url and return }
+      end
+    end
+
+    if !File.exist?("#{crowbar_lib_dir.to_path}/crowbar_installing")
```
tboerger: Swap the if and the else branch, and than use `if crowbar_lib_dir.join("crowbar_installing").exist?`

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,142 @@
...
+  #
+  # Provides the restful api call for
+  # /installer/install 	POST 	triggers install-chef-suse.sh
+  # the bash Process will be spawned in the background and therefore has not a direct return value which we can use here
+  def install
+    installed_ok = Rails.root.join(".crowbar-installed-ok")
+    crowbar_dir = Rails.root.join("..")
+    if installed_ok.exist?
+      msg = "If you want to reinstall please remove /opt/dell/crowbar_framework/crowbar-installed-ok"
+      respond_to do |format|
```
tboerger: You should avoid multiple respond_to blocks in a single controller action.

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -541,6 +541,25 @@ en:
       active: 'Active'
       active_repo_tooltip: 'Managed nodes will be automatically configured to use repositories only if they are marked as active'
 
+  installer:
+    status:
+      installation_failed: 'Installation failed prematurely. Please examine /var/log/crowbar/install.log'
+    index:
+      installation_header: 'Installing SUSE<span class="copyright-tag">©</span> OpenStack Cloud'
```
vuntz: No downstream branding here, please.

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,74 @@
...
+              .addClass("fa-circle-o-notch fa-spin");
+          }
+
+          if (!done) {
+            setTimeout(statusCheck, 3000);
+          } else {
+            $(".panel")
+              .parent()
+              .parent()
+              .prepend("<div class='col-lg-12 col-md-12 col-sm-12 col-xs-12'> \
```
tboerger: Just `<div class=\"col-lg-12\">` and nothing else.

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,74 @@
...
+                         <div class='alert alert-success'> \
+                           <span>Installation was successful. You will be redirected in a few seconds.</span> \
```
tboerger: Some i18n should be good here as well.

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,153 @@
...
+    end
+  end
+
+  #
+  # Set the status of an installation step
+  #
+  # Provides the restful api call for
+  # /installer/update 	POST 	set the status
+  # touch a file in /var/lib/crowbar named after the step
+  # not really needed yet, but maybe later
```
vuntz: If it's not really needed yet, I think we should drop it for now -- otherwise, we'll forget to remove it if it's really not needed. Feel free to keep it in a local branch of course.

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -19,6 +19,7 @@
 require "json"
 
 class BarclampController < ApplicationController
+  skip_before_filter :enforce_installer
```
vuntz: So we have to put this in all controllers? Isn't the bit from `application_controller.rb` enough?

rsalevsky: Nop, we just put this in the controllers which we need while installing.

vuntz: Oh, sorry, I misread and didn't see the `skip`.

Hrm. It's slightly more complicated, though: if chef is not started yet, then this controller will likely fail. We probably need to allow this controller only when we're at the step where it will really work.

crowbar_framework/app/controllers/installer_controller.rb
MaximilianMeister: okay. done

crowbar_framework/config/initializers/installer.rb
MaximilianMeister: okay dropped

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -23,7 +23,8 @@ class ApplicationController < ActionController::Base
   rescue_from Crowbar::Error::NotFound, with: :render_not_found
   rescue_from Crowbar::Error::ChefOffline, with: :chef_is_offline
 
-  before_filter :enable_profiler, if: Proc.new { ENV["ENABLE_PROFILER"] == "true" }
+  before_filter :enable_profiler, if: proc { ENV["ENABLE_PROFILER"] == "true" }
+  before_filter :enforce_installer, if: proc { !Rails.root.join(".crowbar-installed-ok").exist? }
```
tboerger: `before_filter :enforce_installer, unless: proc { Rails.root.join(".crowbar-installed-ok").exist? }`

:)

MaximilianMeister: right :)

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,129 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class InstallerController < ApplicationController
+  skip_before_filter :enforce_installer
```
rsalevsky: Maybe change enforce to just force, sounds better for me.

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,129 @@
...
+  #
+  # Perform Crowbar Installation
+  #
+  # Provides the restful api call for
+  # /installer/install 	POST 	triggers install-chef-suse.sh
+  # the bash Process will be spawned in the background and therefore has not a direct return value which we can use here
+  def install
+    installed_ok = Rails.root.join(".crowbar-installed-ok")
+    crowbar_dir = Rails.root.join("..")
+    if installed_ok.exist?
```
rsalevsky: Is this line to long if you drop the variable?

tboerger: Nope

tboerger:
IMHO we should wait for getting https://github.com/crowbar/crowbar-core/pull/121 merged, than we can display the documentation entries on the sidebar withtout the need to hardcode the values.

vuntz:
Stupid question, but is it intended that it doesn't match the design from @kwwii ? (colors, symbols)

crowbar_framework/app/assets/stylesheets/content/content.scss
```diff
@@ -65,6 +65,22 @@ h3.modal-title {
...
+    line-height: 2em;
+    padding-right: 5px;
+  }
+
+  .action > li > span {
+    margin-right: 5px;
+  }
+}
+
+.copyright-tag {
```
vuntz: Where is this used?

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,86 @@
...
+        type: "GET",
+        dataType: "json",
+        url: Routes.installer_status_path(),
+        success: function(data) {
+          var failed = data.failed;
+          var success = data.success;
+          var errorMsg = data.errorMsg
+          var successMsg = data.successMsg
+          var noticeMsg = data.noticeMsg
+          data = data.steps;
```
vuntz: Hrm, sounds weird to me to reassign this to `data`; why not a new variable?

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,86 @@
...
+          var mostRecent = $("li." + data[data.length-1]);
+          var mostRecentIcon = mostRecent.children();
+
+          $.each(data, function (index, value) {
+            $("li." + value)
+              .removeClass("list-group-item-success")
+              .children("span")
+              .removeClass("fa-hourglass-o fa-circle-o-notch fa-spin")
+              .addClass("fa-check");
+          });
```
vuntz: Do I understand right that we change all the previous steps all the time? If yes, this shouldn't be needed. Or maybe I'm missing something?

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,86 @@
...
+
+          mostRecentIcon.removeClass(function (index, css) {
+            var match = css.match(/^fa-/g);
+            if (match !== null) {
+              return match.join("");
+            }
+          });
```
vuntz: Why do we use a function here, while in other places we hardcode the classes to remove?

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,86 @@
...
+          mostRecentIcon.addClass("fa-check");
+
+          if (failed) {
+            mostRecent
+              .next("li")
```
vuntz: Just a suggestion, but maybe it'd make sense to have a `currentStep` variable for this. (And maybe rename `mostRecent` to `lastSuccessfulStep`, if I understand what `mostRecent` is)

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,86 @@
...
+              .addClass("list-group-item-success")
+              .children("span")
+              .removeClass("fa-hourglass-o")
+              .addClass("fa-circle-o-notch fa-spin");
+          }
+
+          if (failed) {
+            $(".panel")
+              .parent()
+              .parent()
```
vuntz: So I'm not a web developer, but I find it weird that we have to call `parent` twice. Don't we have a way to select this element?

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,86 @@
...
+                         </div> \
+                         <div class='alert alert-info'> \
+                           <span>" + noticeMsg + "</span> \
+                         </div> \
+                       </div>");
+            setTimeout(function(){
+              window.location.replace(Routes.root_path());
+            }, 10000);
+          }
+        },
```
vuntz: Generally speaking, I would just preprend the div if `errorMsg`/`successMsg`/`noticeMsg` are non-empty. I would not even check for `failed`/`success`. This makes the code a bit more generic, and allows sending more notifications if we need too.

Do you think it would make sense?

crowbar_framework/app/controllers/barclamp_controller.rb
vuntz: This comment is still valid, afaict.

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+      format.json { render json: Crowbar::Installer.status_hash }
+      format.html { redirect_to installer_url }
+    end
+  end
+
+  #
+  # Perform Crowbar Installation
+  #
+  # Provides the restful api call for
+  # /installer/install 	POST 	triggers install-chef-suse.sh
```
vuntz: I probably wouldn't mention the name of the script here since we'll eventually kill it; just say "triggers installation".

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+  def install
+    if Rails.root.join(".crowbar-installed-ok").exist?
+      respond_to do |format|
+        format.json { render json: Crowbar::Installer.status_hash.to_json and return }
```
vuntz: Why do we need `to_json` here and not a few lines earlier, and why do we need `and return`?

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+        format.html { redirect_to installer_url and return }
```
vuntz: Why do we need `and return`?

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+      end
+    else
+      # the shell Process will be spawned in the background and therefore has
+      # not a direct return value which we can use here
+      if File.exist?("#{Crowbar::Installer.crowbar_lib_dir.to_path}/crowbar_installing")
+        flash[:notice] = "Crowbar is already installing. Please wait."
```
vuntz: Should we localize the string? Also, maybe rephrase to: "Installation is already ongoing, please wait."

crowbar_framework/app/controllers/network_controller.rb
```diff
@@ -16,6 +16,8 @@
 #
 
 class NetworkController < BarclampController
+  skip_before_filter :enforce_installer
```
vuntz: Hrm, don't we already do that in the barclamp controller?

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -541,6 +541,27 @@ en:
       active: 'Active'
       active_repo_tooltip: 'Managed nodes will be automatically configured to use repositories only if they are marked as active'
 
+  installer:
+    status:
+      installation_failed: 'Installation failed prematurely. Please examine /var/log/crowbar/install.log'
```
vuntz: Can it fail non-prematurely? ;-) Just drop that word. And maybe s/examine/check/

vuntz: (+ end the sentence with a dot)

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -541,6 +541,27 @@ en:
...
+    index:
+      installation_header: 'Installing Crowbar'
+      installation_successful: 'Installation was successful. You will be redirected in a few seconds'
```
vuntz: Missing dot at the end.

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -541,6 +541,27 @@ en:
...
+      reinstall_notice: 'If you want to install again please remove /opt/dell/crowbar_framework/crowbar-installed-ok'
```
vuntz: I would rephrase to: "If you want to reinstall to get a fresh Crowbar setup, please remove the file ..."

(I know the string was probably already like this in the script, btw)

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -541,6 +541,27 @@ en:
...
+      install_button: 'Install'
```
vuntz: Start Installation? Not sure if it's better.

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -541,6 +541,27 @@ en:
...
+      pre_sanity_checks: 'Performing pre-installation sanity checks'
+      run_services: 'Starting required services'
+      initial_chef_client: 'Performing initial chef-client run'
+      barclamp_install: 'Installing Barclamps'
+      bootstrap_crowbar_setup: 'Bootstrapping Crowbar setup'
+      apply_crowbar_config: 'Applying Crowbar configuration for Admin Server'
```
vuntz: s/Admin/Administration/

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -541,6 +541,27 @@ en:
...
+      transition_crowbar: 'Transitioning Admin server to ready state'
```
vuntz: Same

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -541,6 +541,27 @@ en:
...
+      chef_client_daemon: 'Starting chef-client'
+      post_sanity_checks: 'Performing post-installation sanity checks'
+      documentation: 'Documentation'
+      release_notes: 'Release Notes'
+      deployment_guide: 'Crowbar Deployment Guide'
```
vuntz: Wondering if this should just be "Deployment Guide"?

vuntz:
Should we hide the install button if the install process is started, or if the .crowbar-installed-ok file exists?

crowbar_framework/lib/crowbar/installer.rb
```diff
@@ -0,0 +1,85 @@
...
+#
+
+module Crowbar
+  class Installer
+    class << self
+      def install
+        crowbar_dir = Rails.root.join("..")
+
+        pid = spawn("sudo #{crowbar_dir}/bin/install-chef-suse.sh --crowbar")
+        write_file(crowbar_lib_dir, "crowbar_installing")
```
vuntz: Shouldn't we create that file from the script?

crowbar_framework/lib/crowbar/installer.rb
```diff
@@ -0,0 +1,85 @@
...
+# limitations under the License.
+#
+
+module Crowbar
+  class Installer
+    class << self
+      def install
+        crowbar_dir = Rails.root.join("..")
+
+        pid = spawn("sudo #{crowbar_dir}/bin/install-chef-suse.sh --crowbar")
```
vuntz: Somehow we need to check that we run on a SUSE system or not, since in theory, it's not the same script ;-) (I know, the other script is broken, but still)

crowbar_framework/lib/crowbar/installer.rb
```diff
@@ -0,0 +1,85 @@
...
+        {
+          steps: steps,
+          failed: failed?,
+          success: successful?,
+          errorMsg: error_msg,
+          successMsg: success_msg,
+          noticeMsg: notice_msg
+        }
+      end
+
```
vuntz: Should all the methods after this be protected?

crowbar_framework/app/assets/javascripts/installer.js
vuntz: Ah, it's needed if we reload the page. Doesn't sound optimal, though.

crowbar_framework/app/assets/javascripts/installer.js
vuntz: Also, I'd feel more comfortable if we get the info about which step is the current one from the API call instead of simply taking the next one.

crowbar_framework/app/assets/stylesheets/content/content.scss
tboerger: It's used for the (c) after SUSE :)

crowbar_framework/app/assets/javascripts/installer.js
tboerger: In general it's not an issue to directly access `data.steps`.

crowbar_framework/app/controllers/installer_controller.rb
tboerger: `to_json`is useless. And he needs the return because he still relies on multiple respond_to's.

crowbar_framework/app/controllers/installer_controller.rb
tboerger: Bad style, multiple respond_to :)

crowbar_framework/app/controllers/installer_controller.rb
tboerger: :+1:

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+  #
+  # Perform Crowbar Installation
+  #
+  # Provides the restful api call for
+  # /installer/install 	POST 	triggers installation
+  def install
+    if Crowbar::Installer.crowbar_lib_dir.join("crowbar-installed-ok").exist?
+      respond_to do |format|
+        format.json { render json: Crowbar::Installer.status_hash.to_json and return }
+        format.html { redirect_to installer_url and return }
```
vuntz: (continuation of https://github.com/crowbar/crowbar-core/pull/119/files#r45450878)

@tboerger just to help me fully understand (I'm still learning rails ;-)): the `and return` here matters, even though the other `respond_to` will not be called (this one is in `if` while the other one is in `else`)?

crowbar_framework/app/assets/stylesheets/content/content.scss
vuntz: Ah, but this was gone in the PR because the string got changed to Crowbar. Now this makes sense :-)

crowbar_framework/app/controllers/installer_controller.rb
vuntz: (btw, `to_json` is still here.)

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -541,6 +541,29 @@ en:
       active: 'Active'
       active_repo_tooltip: 'Managed nodes will be automatically configured to use repositories only if they are marked as active'
 
+  installer:
+    status:
+      installation_failed: 'Installation failed. Please check the log at: /var/log/crowbar/install.log'
```
vuntz: s/://

crowbar_framework/app/assets/javascripts/installer.js
MaximilianMeister: I am not too familiar with it. maybe @tboerger has an idea?

crowbar_framework/app/controllers/network_controller.rb
MaximilianMeister: i will try it again without. i have added it because install-chef-suse failed at the admin ip check, and this fixed it somehow

crowbar_framework/app/assets/javascripts/installer.js
MaximilianMeister: not really sure

crowbar_framework/app/assets/javascripts/installer.js
MaximilianMeister: maybe @tboerger can help out here a bit. the javascript part could maybe be reworked a bit, i am not sure but i couldn't come up with sth better now.

crowbar_framework/app/assets/javascripts/installer.js
MaximilianMeister: we have to remove specific ones, not always **all**

MaximilianMeister:
> Stupid question, but is it intended that it doesn't match the design from @kwwii ? (colors, symbols)

he was fine with it until now

crowbar_framework/app/assets/javascripts/installer.js
tboerger: You can match a parent element with `.parents('good_selector')`

crowbar_framework/app/assets/javascripts/installer.js
```diff
@@ -0,0 +1,86 @@
+$(document).ready(function() {
+  $("body.installer").each(function() {
+    function statusCheck() {
+      $.ajax({
+        type: "GET",
+        dataType: "json",
+        url: Routes.installer_status_path(),
+        success: function(data) {
+          var failed = data.failed;
```
tboerger: I generally dislike to define all these variables. Just access the `data` object.

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class InstallerController < ApplicationController
+  skip_before_filter :enforce_installer
+  before_filter :hide_navigation
+
+  def index
+    @steps = Crowbar::Installer.all_steps
```
tboerger: Why all_steps? Just `steps` is enough :)

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+  #
+  # Perform Crowbar Installation
+  #
+  # Provides the restful api call for
+  # /installer/status 	GET 	return done steps, error and success
+  # returns a hash with an indicator if the installation failed/succeeded
+  # and the steps that are done
+  def status
+    respond_to do |format|
+      format.json { render json: Crowbar::Installer.status_hash }
```
tboerger: And why not simply `status`?

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+    end
+  end
+
+  #
+  # Perform Crowbar Installation
+  #
+  # Provides the restful api call for
+  # /installer/install 	POST 	triggers installation
+  def install
+    if Crowbar::Installer.crowbar_lib_dir.join("crowbar-installed-ok").exist?
```
tboerger: I would name that simply `lib_dir`, IMHO it's obvious enough that this should be the crowbar libdir.

tboerger: Beside that this check can be something like `Crowbar::Installer.installed?`

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+      respond_to do |format|
+        format.json { render json: Crowbar::Installer.status_hash.to_json }
+        format.html { redirect_to installer_url }
+      end
+    else
+      # the shell Process will be spawned in the background and therefore has
+      # not a direct return value which we can use here
+      if File.exist?("#{Crowbar::Installer.crowbar_lib_dir.to_path}/crowbar_installing")
```
tboerger: This is weired... `Crowbar::Installer.lib_dir.join("crowbar_installing").exist?` is much cleaner.

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+        flash[:notice] = I18n.t(".installation_ongoing", scope: "installer.index")
```
tboerger: The scope is wrong. It's `installer.install`, beside that to safe some chars you can do `I18n.t("installer.install.installation_ongoing")`as well

crowbar_framework/app/controllers/installer_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+      else
+        ret, msg = Crowbar::Installer.install
```
tboerger: I already said before that the current implementation of crowbar with the return code and the message is simply bad coding style. You should use exceptions or error array/hash attributes instead of just taking the Dell coding style.

crowbar_framework/app/views/installer/index.html.haml
```diff
@@ -0,0 +1,20 @@
+.col-lg-6.col-md-6.col-sm-6.col-xs-12
+  .panel.panel-default
+    .panel-heading
+      %h2
+        = t(".installation_header").html_safe
```
tboerger: Why do we need html_safe?

crowbar_framework/app/views/installer/index.html.haml
```diff
@@ -0,0 +1,20 @@
...
+        .btn-group.pull-right
+          = button_to t(".install_button"), install_path, class: "btn btn-success", method: :post
```
tboerger: As the action is called by a button/link without any form I would suggest to make it simply a GET route.

crowbar_framework/config/routes.rb
```diff
@@ -123,5 +123,9 @@
   get "crowbar/:barclamp", action: "versions", controller: "barclamp"
   post "crowbar/:barclamp/1.0/:action/:id", controller: "barclamp"
 
+  get "installer", controller: "installer", action: "index", as: "installer"
```
tboerger: Maybe it's a pretty bad habit to write all the routes manually instead of just using something like

```
resource :installer, only: [:show] do
  collection do
    get :status
    get :install
  end
end
```

(The resource definition is just written out of my head without testing)

crowbar_framework/lib/crowbar/installer.rb
```diff
@@ -0,0 +1,91 @@
...
+        end
+
+        # spawn the script asynchronously in the background
+        pid = spawn(cmd)
+        Process.detach(pid)
+
+        return [200, ""]
+      end
+
+      def all_steps
```
tboerger: s/all_steps/steps/

crowbar_framework/lib/crowbar/installer.rb
```diff
@@ -0,0 +1,91 @@
...
+          :barclamp_install,
+          :bootstrap_crowbar_setup,
+          :apply_crowbar_config,
+          :transition_crowbar,
+          :chef_client_daemon,
+          :post_sanity_checks
+        ]
+      end
+
+      def crowbar_lib_dir
```
tboerger: s/crowbar_lib_dir/lib_path/ or s/crowbar_lib_dir/lib_dir/. I would prefer _path as it's a pathname.

crowbar_framework/lib/crowbar/installer.rb
```diff
@@ -0,0 +1,91 @@
...
+        Pathname.new("/var/lib/crowbar/install")
+      end
+
+      def status_hash
```
tboerger: s/status_hash/status/

crowbar_framework/app/assets/javascripts/installer.js
vuntz: But why not do like what you did a few lines above: `.removeClass("fa-hourglass-o fa-circle-o-notch fa-spin")`? AFAICT, you're in the exact same case.

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -541,6 +541,29 @@ en:
       active: 'Active'
       active_repo_tooltip: 'Managed nodes will be automatically configured to use repositories only if they are marked as active'
 
+  installer:
+    status:
+      installation_failed: 'Installation failed. Please check the log at /var/log/crowbar/install.log'
+    index:
+      installation_header: 'Installing Crowbar'
```
vuntz: Can we replace "Crowbar" with some string that comes from /opt/dell/crowbar_framework/config/branding.yml? That will make branding easier.

tboerger: :+1:

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -541,6 +541,29 @@ en:
...
+      installation_successful: 'Installation was successful. You will be redirected in a few seconds.'
+      installation_ongoing: 'Installation is already ongoing, please wait.'
+      reinstall_notice: 'If you want to reinstall to get a fresh Crowbar setup, please remove the following file: /var/lib/crowbar/install/crowbar-installed-ok'
```
tboerger: s/Crowbar//

crowbar_framework/app/views/installer/index.html.haml
MaximilianMeister: makes sense, but this is a bit less risky. i did a `post` action not to accidentally trigger the installation codepath by opening the url.

crowbar_framework/config/routes.rb
rsalevsky: I agree that how we write our routes is not optimal but should we start now and have two different styles? Or should we do this once after the merge?

tboerger: I think we should do it piece by piece. Doing all at once is too error prone

crowbar_framework/app/views/installer/index.html.haml
vuntz: @MaximilianMeister thanks for the rationale, agree with you.

tboerger: Than please do it with a real form and not with that kind of akward code.

crowbar_framework/app/assets/javascripts/installer.js
MaximilianMeister: it is not the same case

crowbar_framework/app/controllers/barclamp_controller.rb
MaximilianMeister: https://github.com/crowbar/crowbar-core/commit/faf9bc1e79a357a942fef83da8d85d325751144c should hopefully fix that issue

crowbar_framework/app/assets/javascripts/installer.js
vuntz: What is different?

MaximilianMeister: well after debugging i found out that this is completely useless and can be dropped, it was maybe a remainder of the many try and errors

crowbar_framework/app/controllers/installers_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  #
+  # Perform Crowbar Installation
+  #
+  # Provides the restful api call for
+  # /installer/start 	POST 	triggers installation
+  def start
+    if Crowbar::Installer.successful?
+      respond_to do |format|
+        format.json { render json: Crowbar::Installer.status.to_json }
```
tboerger: Skip `.to_json`

crowbar_framework/app/controllers/installers_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+        format.html { redirect_to installer_url }
+      end
+    else
+      # the shell Process will be spawned in the background and therefore has
+      # not a direct return value which we can use here
+      if Crowbar::Installer.lib_dir_path.join("crowbar_installing").exist?
```
tboerger: `Crowbar::Installer.installing?`

crowbar_framework/app/views/installers/show.html.haml
```diff
@@ -0,0 +1,21 @@
+.col-lg-6.col-md-6.col-sm-6.col-xs-12
+  .panel.panel-default
+    .panel-heading
+      %h2
+        = t(".installation_header", product: Crowbar::Installer.branding_product)
```
tboerger: `product: branding.page_title`

crowbar_framework/lib/crowbar/installer.rb
```diff
@@ -0,0 +1,101 @@
...
+
+      def failed?
+        lib_dir_path.join("crowbar-install-failed").exist?
+      end
+
+      def successful?
+        lib_dir_path.join("crowbar-installed-ok").exist?
+      end
+
+      def branding_product
```
tboerger: Remove this

crowbar_framework/lib/crowbar/installer.rb
```diff
@@ -0,0 +1,101 @@
...
+          :barclamp_install,
+          :bootstrap_crowbar_setup,
+          :apply_crowbar_config,
+          :transition_crowbar,
+          :chef_client_daemon,
+          :post_sanity_checks
+        ]
+      end
+
+      def lib_dir_path
```
tboerger: Please name that just `lib_path`

tboerger:
Some NITPICK, beside that :+1:

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -23,7 +23,8 @@ class ApplicationController < ActionController::Base
   rescue_from Crowbar::Error::NotFound, with: :render_not_found
   rescue_from Crowbar::Error::ChefOffline, with: :chef_is_offline
 
-  before_filter :enable_profiler, if: Proc.new { ENV["ENABLE_PROFILER"] == "true" }
+  before_filter :enable_profiler, if: proc { ENV["ENABLE_PROFILER"] == "true" }
+  before_filter :enforce_installer, unless: proc { Crowbar::Installer.lib_dir_path.join("crowbar-installed-ok").exist? || ENV["RAILS_ENV"] == "test" }
```
rsalevsky: Can you break this line?

rsalevsky:
The long line hound issues should be fixed but otherwise. :+1:

PR #120, opened 2015-11-05
Proxy timeout is configurable and set to 1500

nkrinner:
Barclamp deployments can last significantly longer than the apache proxy
is configured to keep the connection, causing crowbar batch running into
timeouts and errors in the web view. Make the proxy timeout configurable
and set it to a reasonable value.


tboerger:
:-1: Travis fails related to this PR.

chef/cookbooks/crowbar/templates/default/apache.conf.erb
```diff
@@ -9,7 +9,7 @@
     ErrorDocument 422 /422.html
     ErrorDocument 500 /500.html
 
-    ProxyPass / http://127.0.0.1:<%= @port %>/ timeout=600 Keepalive=On
+    ProxyPass / http://127.0.0.1:<%= @port %>/ timeout=<%= @proxy_timeout %> Keepalive=On
     ProxyPassReverse / http://127.0.0.1:<%= @port %>/
     SetEnv proxy-nokeepalive 1
```
rsalevsky: We have still `Keepalive=On` and `SetEnv proxy-nokeepalive 1` which makes no sense.

nkrinner:
Updated fixtures so gating passes again. Updated apache configuration template as @rsalevsky pointed out.

rsalevsky:
Please also adapt https://github.com/crowbar/crowbar-core/blob/master/configs/crowbar.conf.

jsuchome:
Just deployed develcloud6 with this patch +1

vuntz:
Why do we need an attribute for this?

nkrinner:
We do not need it as an attribute as far as I can see. To me it looked like the 'cleanest' implementation though to have it that way.

vuntz:
I'm just worried that we're starting to add attributes for everything, which will end up being very messy. So I wouldn't add an attribute unless we know it's needed.

nkrinner:
Removed the attribute and added the wip label until i finished testing this change

configs/crowbar.conf
```diff
@@ -9,8 +9,7 @@
     ErrorDocument 422 /422.html
     ErrorDocument 500 /500.html
 
-    ProxyPass / http://127.0.0.1:3000/ timeout=600 Keepalive=On
+    ProxyPass / http://127.0.0.1:3000/ timeout=1500 Keepalive=On
```
dirkmueller: we would need to pass through @proxy_timeout ?

nkrinner: This file is only used during iniital crowbar setup from the rpm package. It is replaced with a chef-controlled file later. We can probably do without proxy_timeout here.

vuntz:
I would even kill the attribute on the node, but I can live with that.

+1

configs/crowbar.conf
rsalevsky: We need to do it without proxy_timeout because we have no chef running at this time. It's just for the time while installing the package and start install-chef-suse.

rsalevsky:
+1

nkrinner:
Tested successfully localy and rebased. Removing the wip label.

configs/crowbar.conf
dirkmueller: I understand. thats why I recommended using the same string ("timeout=1500") for both locations so that simple search&replace will find both occurrences

jsuchome:
Gating build still not finished, but as it did not fail during crowbar related deployment tasks, I would think we can go and merge this...

dirkmueller:
@jsuchome would still be good to remove the extra attribute..

jsuchome:
For the record, gating is green now (https://ci.suse.de/job/openstack-mkcloud/11067/)

nkrinner:
So if I understand this correctly, you prefer a simple value change in the template to a chef-controlled attribute? @dirkmueller @vuntz

dirkmueller:
@nkrinner yes, thats what I meant (to have a grep/search finding both places should we ever need to touch that code again)

nkrinner:
Commit stripped down and only changing the value, although I still see more value in having the attribute set  with chef. Also rebased.

PR #122, opened 2015-11-06
Start apache2.service while crowbar.service start

rsalevsky:
Starting crowbar without the apache service makes no sense since we use
the apache proxy to reach the crowbar web UI. This change makes sure
that the service is always started.


tboerger:
Is Wants correct? Or Requires?

rsalevsky:
Changed to `Requires=`, wants makes no real sense.

vuntz:
+1

MaximilianMeister:
:+1:

PR #123, opened 2015-11-09
provisioner: Remove useless bits from kernel boot line for sledgehammer

vuntz:
ifcfg and netwait are not used when booting in standard SLES (no
linuxrc), so they have no effect. Let's drop them.

Instead, these bits have been put in the sleshammer image.

See discussion in https://bugzilla.suse.com/show_bug.cgi?id=953107


tboerger:
After reading the bug report it looks good to me.

rsalevsky:
+1

mjura:
Yeap, seems to be OK +1

PR #125, opened 2015-11-10
Added partial path to location at SMT server

jsuchome:
We could probably just use the product name and version (like `SLE-SERVER/12`), but isn't this bit more readable? 

(But I'm fine with any way)

### DIFF
--- a/crowbar_framework/config/repos-cloud.yml
+++ b/crowbar_framework/config/repos-cloud.yml
@@ -9,6 +9,7 @@ suse-12.0:
         tag: "obsproduct://build.suse.de/SUSE:SLE-12:GA/SLES/12/POOL/x86_64"
         key_md5: "9a62177e1c6852d48453ed3909956d6b"
       url:
+      smt_path: "SUSE/Products/SLE-SERVER/12/x86_64/product"
       ask_on_error: false
     sles12-updates:
       name: "SLES12-Updates"
@@ -18,6 +19,7 @@ suse-12.0:
         tag: "obsrepository://build.suse.de/SUSE:Updates:SLE-SERVER:12:x86_64/update"
         key_md5: "9a62177e1c6852d48453ed3909956d6b"
       url:
+      smt_path: "SUSE/Updates/SLE-SERVER/12/x86_64/update"
       ask_on_error: false
     sle12-ha-pool:
       name: "SLE12-HA-Pool"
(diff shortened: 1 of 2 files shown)


tboerger:
And where gets it used?

jsuchome:
@tboerger commit message updated

crowbar_framework/config/repos-cloud.yml
```diff
@@ -18,6 +19,7 @@ suse-12.0:
         tag: "obsrepository://build.suse.de/SUSE:Updates:SLE-SERVER:12:x86_64/update"
         key_md5: "9a62177e1c6852d48453ed3909956d6b"
       url:
+      smt_path: "SUSE/Products/SLE-SERVER/12/x86_64/update"
```
vuntz: Looks wrong, I'd expect SUSE/Updates

vuntz: (same for the other update repos)

vuntz:
Can you also update repos-ses.yml?

crowbar_framework/config/repos-cloud.yml
jsuchome: Fixed

tboerger:
If yast2 is the only tool that uses it I'm fine with it.

crowbar_framework/config/repos-cloud.yml
```diff
@@ -45,6 +47,7 @@ suse-12.0:
         tag: "obsproduct://build.suse.de/SUSE:SLE-12:Update:Products:SES2/ses/2/POOL/x86_64"
         key_md5: "9a62177e1c6852d48453ed3909956d6b"
       url:
+      smt_path: "suse/products/storage/2/x86_64/product"
```
vuntz: s/suse/SUSE/

crowbar_framework/config/repos-ses.yml
```diff
@@ -45,6 +47,7 @@ suse-12.0:
         tag: "obsproduct://build.suse.de/SUSE:SLE-12:Update:Products:SES2/ses/2/POOL/x86_64"
         key_md5: "9a62177e1c6852d48453ed3909956d6b"
       url:
+      smt_path: "suse/products/storage/2/x86_64/product"
```
vuntz: same here

vuntz:
+1

toabctl:
+1

PR #126, opened 2015-11-10
Renew DHCP lease after hardware-installing transtion

rhafer:
During the "hardware-installing" transition a new admin IP Address (from the
"host" ranges) gets allocated to the node. Force a DHCP renewal to change to
the new IP Address directly after the transition. This avoids having the IP
Address changed at a later (unpredictable) time during the remainder of the
sledgehammer phase.

### DIFF
--- a/updates/control.sh
+++ b/updates/control.sh
@@ -234,6 +234,20 @@ run_hooks() {
     done
 }
 
+renew_dhcp_after_hwinstalling () {
+    # The Admin IP address of the node is allocated in the
+    # "hardware-installing" transition. Force a renewal here get the new IP
+    # address and avoid it changing at an unexpected time during the
+    # subsequent transitions
+    if [[ $1 = 'hardware-installing' ]]; then
+        echo "Forcing DHCP renewal after Admin IP allocation"
+        ifup $BOOTDEV > /dev/null
+        echo "New local IP Addresses:"
+        ip a | awk '/127.0.0./ { next; } /inet / { print }'
+    fi
+    return 0
+}
+
 walk_node_through () {
     # $1 = hostname for chef-client run
     # $@ = states to walk through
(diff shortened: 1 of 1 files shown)


updates/control.sh
```diff
@@ -234,6 +234,20 @@ run_hooks() {
     done
 }
 
+renew_dhcp_after_hwinstalling () {
+    # The Admin IP address of the node is allocated in the
+    # "hardware-installing" transition. Force a renewal here get the new IP
+    # address and avoid it changing and an unexpected time during the
+    # subsequent transitions
+    if [[ $1 = 'hardware-installing' ]]; then
```
jsuchome: `==` ?

rhafer: The ```=``` operator is equivalent to ```==``` here and AFAICS we're using ```=``` everywhere else in this script.

rhafer: Hm, actually we're using both forms throughout this.

updates/control.sh
```diff
@@ -234,6 +234,20 @@ run_hooks() {
     done
 }
 
+renew_dhcp_after_hwinstalling () {
+    # The Admin IP address of the node is allocated in the
+    # "hardware-installing" transition. Force a renewal here get the new IP
+    # address and avoid it changing and an unexpected time during the
```
vuntz: s/and/at/

updates/control.sh
```diff
@@ -234,6 +234,20 @@ run_hooks() {
...
+    # subsequent transitions
+    if [[ $1 = 'hardware-installing' ]]; then
+        echo "Forcing DHCP renewal after Admin IP allocation"
+        ifup $BOOTDEV > /dev/null
```
vuntz: Just wondering: does `ifup` work on all distros?

rhafer: You mean distros != SUSE? I have no idea. But it should work on all SUSE distros.

updates/control.sh
rhafer: fixed

rsalevsky:
LGTM

jsuchome:
+ 1 (not tested)

vuntz:
+1

dirkmueller:
+1

PR #127, opened 2015-11-12
Fix ifcfg template for openvswitch device

rhafer:
According to the SLES12-SP1 documentation STARTMODE can be "auto" for
OvsBridges as well.
We need the OVS_BRIDGE='yes' and OVS_BRIDGE_PORT_DEVICE_* settings as wicked
will remove the bridge during shutdown and recreate them at boottime. (Which
went unnoticed until recently because of a bug in the openvswitch service
dependencies)

### DIFF
--- a/chef/cookbooks/network/templates/default/suse-cfg.erb
+++ b/chef/cookbooks/network/templates/default/suse-cfg.erb
@@ -18,13 +18,9 @@ iface=@interfaces[@nic.name]
 -%>
 NAME=<%=quote(@nic.name)%>
 <% if iface["ovs_slave"] -%>
-STARTMODE=hotplug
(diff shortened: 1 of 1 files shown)


rhafer:
I put this here for some early review. Still needs some more testing. (Especially reboot)

vuntz:
Looks reasonable.

rhafer:
Did a manual reboot test now. Seemed to work. At least on the compute nodes. There is however still a problem with shutting down the controller (nothing that this PR introduces, it's also there without this change and probably even unrelated to networking)

vuntz:
Thanks for testing, changing my comment to +1

rsalevsky:
+1

PR #131, opened 2015-11-18
Fix repotag that is used for checking Cloud 6 media

vuntz:
We didn't have the right URL for the official build...


dirkmueller:
*sigh*.. why do we check the REPOID on a media? we should check the DISTRO :/

dirkmueller:
+1

kbaikov:
+1

tboerger:
:+1:

PR #132, opened 2015-11-19
Don't create ovs-system related pieces anymore

rhafer:
They are no longer needed with the latest wicked (>= 0.6.28)


toabctl:
+1

vuntz:
+1 if gating is green :-)

rhafer:
@vuntz it's not exactly green :), but survived tempest and AFAICS cct failed just because the chef-client run was so slow that crowbar considered the node state "unknown".

vuntz:
@rhafer does this survive a reboot?

rhafer:
> does this survive a reboot?

@vuntz yes. At least on my manual deployment. Hm, probably I'll just retrigger the CI  with reboot enabled.

rhafer:
@vuntz that last ci run was with reboot enable. YAY!

PR #133, opened 2015-11-20
webui: Automatically create symlinks to SMT repositories

vuntz:
If a repo doesn't exist but is mirrored by the local SMT, automatically
create a symlink to it.

### DIFF
--- a/crowbar_framework/lib/crowbar/repository.rb
+++ b/crowbar_framework/lib/crowbar/repository.rb
@@ -264,7 +264,7 @@ module Crowbar
     private
 
     def repos_path
-      Pathname.new(File.join("/srv/tftpboot", @platform, @arch, "repos"))
+      Pathname.new("/srv/tftpboot").join(@platform, @arch, "repos")
     end
 
     def repo_path
--- a/crowbar_framework/lib/crowbar/repository.rb
+++ b/crowbar_framework/lib/crowbar/repository.rb
@@ -266,19 +266,23 @@ module Crowbar
       Pathname.new(File.join("/srv/tftpboot", @platform, @arch, "repos"))
     end
 
+    def repo_path
+      repos_path.join(name)
+    end
+
     def repodata_path
-      repos_path.join(name, "repodata")
+      repo_path.join("repodata")
     end
 
     def repodata_media_path
-      repos_path.join(name, "suse", "repodata")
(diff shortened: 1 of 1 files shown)


crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -266,19 +267,35 @@ def repos_path
       Pathname.new(File.join("/srv/tftpboot", @platform, @arch, "repos"))
     end
 
+    def repo_path
+      repos_path.join(name)
+    end
+
+    def smt_path
+      unless @config["smt_path"].blank?
```
tboerger: I would suggest something like

```ruby
Pathname.new(
  "/srv/www/htdocs/repo"
).join(
  @config["smt_path"]
) unless @config["smt_path"].blank?
```

tboerger:
Some NITPIK otherwise :+1:

crowbar_framework/lib/crowbar/repository.rb
vuntz: I'll do the `Pathname.new().join()` bit, but I'm not a big fan of the multiline style. Is this somehow recommended by some guidelines?

tboerger: For singlelines you should pu the unless behind it :)

vuntz: Sure, but then I have a line that is too long. And your suggestion is 5 lines instead of 3 lines (for a readability that is equivalent :-)

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -263,22 +264,38 @@ def to_databag_hash
...
+      repo_path.join("repodata")
     end
 
     def repodata_media_path
-      repos_path.join(name, "suse", "repodata")
+      repo_path.join("suse", "repodata")
+    end
+
+    def ensure_link_smt_path
+      unless remote? || smt_path.nil? || repo_path.directory? || !smt_path.directory?
```
toabctl: Does this check if the link already exists?

vuntz: @toabctl yes, that's the `smt_path.directory?` bit. Although, this might not be enough if the symlink is dangling. Not sure if we should care about that, though?

toabctl:
+1

PR #134, opened 2015-11-20
Avoid HoundCI bitching about has_key?

dirkmueller:
According to our now master of style, Hash#has_key? is deprecated in
favor of Hash#key?.  (https://github.com/bbatsov/ruby-style-guide#hash-key)


tboerger:
There are still other offenses but :+1:

rsalevsky:
+1

toabctl:
+1

PR #135, opened 2015-11-20
webui: Add favicon option in branding to help downstreams

vuntz:


tboerger:
:+1: for now until we rebuild the branding with the possibility to build subpackages.

vuntz:
@tboerger that's actually to enable this that this change is done ;-)

tboerger:
> @tboerger that's actually to enable this that this change is done ;-)

Because of that I gave my :+1: :)

rsalevsky:
:+1:

vuntz:
Gating is red because it conflicts with the branding patch...

PR #136, opened 2015-11-20
Run iptables network check as root (bsc#955984)

rsalevsky:
The crowbar user is part of the sudoer's so we can just run iptables
as root.


tboerger:
I have not tested it but looks ok codewise.

vuntz:
+1

mjura:
Ok, I've checked this
+1

PR #137, opened 2015-11-20
AutoYaST Upgrade profile

rhafer:
This is the contains the base work for an autoyast profile to upgrade a SLE11 openstack controller to SLE12-SP1.

Currently this just creates autoyast-upgrade.xml just in parallel to the "normal" autoyast profile and hardcodes the autoupgrade parameter into the pxe config for all nodes that get allocated. This was just done to be able to test the profile. The final solution will have to handle it differently of course. This PR is just meant to be input for whoever is going to implement the "upgrade" state in crowbar.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -258,7 +258,8 @@ if not nodes.nil? and not nodes.empty?
 
         when os =~ /^(open)?suse/
           append << "install=#{install_url} autoyast=#{node_url}/autoyast.xml"
-          append << "ifcfg=dhcp4 netwait=60"
+          append << " ifcfg=dhcp4 netwait=60"
+          append << " autoupgrade=1" if mnode[:state] == "os_upgrading"
 
           target_platform_version = os.gsub(/^.*-/, "")
           repos = Provisioner::Repositories.get_repos("suse", target_platform_version, arch)
@@ -282,9 +283,10 @@ if not nodes.nil? and not nodes.empty?
             cloud_available = true if name.include? "Cloud"
           end
 
+          autoyast_template = mnode[:state] == "os_upgrading" ? "autoyast-upgrade" : "autoyast"
           template "#{node_cfg_dir}/autoyast.xml" do
             mode 0644
-            source "autoyast.xml.erb"
+            source "#{autoyast_template}.xml.erb"
             owner "root"
             group "root"
             variables(
--- a/chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
@@ -1,0 +1,103 @@
+<?xml version="1.0"?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <deploy_image>
+    <image_installation config:type="boolean">false</image_installation>
+  </deploy_image>
+  <general>
+    <ask-list config:type="list"/>
+    <mode>
+      <confirm config:type="boolean">false</confirm>
+      <final_halt config:type="boolean">false</final_halt>
+      <final_reboot config:type="boolean">false</final_reboot>
+      <halt config:type="boolean">false</halt>
+      <second_stage config:type="boolean">true</second_stage>
+    </mode>
+    <mouse>
+      <id>none</id>
+    </mouse>
+    <proposals config:type="list"/>
+    <signature-handling>
+      <accept_file_without_checksum config:type="boolean">true</accept_file_without_checksum>
+      <accept_non_trusted_gpg_key config:type="boolean">true</accept_non_trusted_gpg_key>
+      <accept_unknown_gpg_key config:type="boolean">true</accept_unknown_gpg_key>
+      <accept_unsigned_file config:type="boolean">true</accept_unsigned_file>
+      <accept_verification_failed config:type="boolean">true</accept_verification_failed>
+      <import_gpg_key config:type="boolean">true</import_gpg_key>
+    </signature-handling>
+    <storage/>
+  </general>
+  <add-on>
+    <add_on_products config:type="list">
+    <% @repos.keys.sort.each do |name| %>
+      <listentry>
+        <name><%= name %></name>
+        <product><%= name %></product>
+        <product_dir>/</product_dir>
+        <media_url><%= @repos[name][:url] %></media_url>
+        <!-- See bnc#787157 for discussion on ask_on_error -->
+        <ask_on_error config:type="boolean"><%= @repos[name][:ask_on_error] || 'false' %></ask_on_error>
+      </listentry>
+    <% end %>
+    </add_on_products>
+  </add-on>
+  <timezone>
+    <hwclock>UTC</hwclock>
+    <timezone><%=@timezone%></timezone>
+  </timezone>
+  <services-manager>
+    <services>
+      <enable config:type="list">
+        <service>sshd</service>
+      </enable>
+    </services>
+  </services-manager>
+  <software>
+    <packages config:type="list">
+      <package>autoyast2-installation</package>
+<% if @architecture == "x86_64" -%>
+      <package>biosdevname</package>
+<% end -%>
+      <package>netcat</package>
+      <package>ruby2.1-rubygem-chef</package>
+<% if @is_ses -%>
+      <package>ses-release</package>
+<% else -%>
+      <package>suse-openstack-cloud-release</package>
+      <package>supportutils-plugin-suse-openstack-cloud</package>
+<% end -%>
+<% @packages.each do |package| -%>
+      <package><%= package %></package>
+<% end -%>
+    </packages>
+    <patterns config:type="list">
+      <pattern>Minimal</pattern>
+      <pattern>base</pattern>
+    </patterns>
+    <!-- Cleanup leftover Cloud 5 packages -->
+    <remove-packages config:type="list">
+     <package>facter</package>
+     <package>gstreamer-0_10</package>
+     <package>gstreamer-0_10-lang</package>
+     <package>ifplugd</package>
+     <package>ksh</package>
+     <package>libgstreamer-0_10-0</package>
+     <package>libipset2</package>
+     <package>libjpeg-32bit</package>
+     <package>libopensc2</package>
+     <package>libopenssl0_9_8</package>
+     <package>libopenssl0_9_8-32bit</package>
+     <package>libstdc++33</package>
+     <package>libtiff3-32bit</package>
+     <package>libyajl1</package>
+     <package>netdate</package>
+     <package>pcre-32bit</package>
+     <package>puppet</package>
+     <package>python-ordereddict</package>
+     <package>ruby2.1-rubygem-treetop-1_4</package>
+     <package>rubygem-bundler</package>
+     <package>rubygems</package>
+     <package>yast2-wagon</package>
+    </remove-packages>
+  </software>
+</profile>
--- a/chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
@@ -70,6 +70,74 @@
       <package><%= package %></package>
 <% end -%>
     </packages>
+    <scripts>
+      <chroot-scripts config:type="list">
+        <script>
+          <chrooted config:type="boolean">false</chrooted>
+          <debug config:type="boolean">true</debug>
+          <filename>crowbar_post</filename>
+          <source>
+            <![CDATA[
+            export PS4='${BASH_SOURCE}@${LINENO}(${FUNCNAME[0]}): '
+            HTTP_SERVER="<%= @admin_node_ip %>:<%= @web_port %>"
+            IP="<%= @admin_node_ip %>"
+            HOSTNAME="<%= @node_fqdn %>"
+
+            key_re='crowbar\.install\.key=([^ ]+)'
+            if [[ $(cat /proc/cmdline) =~ $key_re ]]; then
+                export CROWBAR_KEY="${BASH_REMATCH[1]}"
+                echo "$CROWBAR_KEY" >/mnt/etc/crowbar.install.key
+            elif [[ -f /mnt/etc/crowbar.install.key ]]; then
+                export CROWBAR_KEY="$(cat /mnt/etc/crowbar.install.key)"
+            fi
+            post_state() {
+                mkdir -p /var/log/crowbar
+                local curlargs=(-o "/mnt/var/log/crowbar/$1-$2.json" --connect-timeout 60 -s \
+                  -L -X POST --data-binary "{ \"name\": \"$1\", \"state\": \"$2\" }" \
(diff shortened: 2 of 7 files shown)


chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
```diff
@@ -0,0 +1,107 @@
...
+      <id>none</id>
+    </mouse>
+    <proposals config:type="list"/>
+    <signature-handling>
+      <accept_file_without_checksum config:type="boolean">true</accept_file_without_checksum>
+      <accept_non_trusted_gpg_key config:type="boolean">true</accept_non_trusted_gpg_key>
+      <accept_unknown_gpg_key config:type="boolean">true</accept_unknown_gpg_key>
+      <accept_unsigned_file config:type="boolean">true</accept_unsigned_file>
+      <accept_verification_failed config:type="boolean">true</accept_verification_failed>
+      <import_gpg_key config:type="boolean">true</import_gpg_key>
```
dirkmueller: do we really need those settings to be enabled?

rhafer: Actually I don't really know. Those are just copied from the non-upgrade template. If we don't need them there anymore, we don't need them here.

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -257,7 +257,7 @@
           end
 
         when os =~ /^(open)?suse/
-          append << "install=#{install_url} autoyast=#{node_url}/autoyast.xml"
+          append << "install=#{install_url} autoyast=#{node_url}/autoyast-upgrade.xml autoupgrade=1"
```
mjura: Shouldn't we check somehow if this upgrade is needed otherwise we could use standard `autoyast.xml` template

rhafer: Yes of course. But the underlying mechanism required for that in crowbar are still being worked out (e.g. here: https://github.com/crowbar/barclamp-updater/pull/11) and implemented yet. This commit here contains just the bare minimum that we needed to create/test an AutoYaST profile for upgrading. It's not meant to be merged in this form (see my initial comment)

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
```diff
@@ -0,0 +1,107 @@
...
+        <product><%= name %></product>
+        <product_dir>/</product_dir>
+        <media_url><%= @repos[name][:url] %></media_url>
+        <!-- See bnc#787157 for discussion on ask_on_error -->
+        <ask_on_error config:type="boolean"><%= @repos[name][:ask_on_error] || 'false' %></ask_on_error>
+      </listentry>
+    <% end %>
+    </add_on_products>
+  </add-on>
+  <timezone>
```
mjura: I think this is not needed, if we want to do only upgrade.

chef/cookbooks/provisioner/recipes/update_nodes.rb
mjura: Ok, ack

mjura:
LGTM

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
```diff
@@ -0,0 +1,107 @@
...
+    <patterns config:type="list">
+      <pattern>Minimal</pattern>
+      <pattern>base</pattern>
+    </patterns>
+    <!-- Avoid default Cloud pattern "cloud_admin" to be installed on all nodes -->
+    <remove-patterns config:type="list">
+	<pattern>cloud_admin</pattern>
+    </remove-patterns>
+    <!-- Cleanup leftover Cloud 5 packages -->
+    <remove-packages config:type="list">
```
vuntz: Have you generated this list from a node where you had installed every package from Cloud 5?

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
```diff
@@ -0,0 +1,107 @@
...
+<% end -%>
+    </packages>
+    <patterns config:type="list">
+      <pattern>Minimal</pattern>
+      <pattern>base</pattern>
+    </patterns>
+    <!-- Avoid default Cloud pattern "cloud_admin" to be installed on all nodes -->
+    <remove-patterns config:type="list">
+	<pattern>cloud_admin</pattern>
+    </remove-patterns>
```
vuntz: Why is this needed? Is this just to be safe?

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
rhafer: The Cloud5 node was basically an  HA controller node, that had all barclamps deployed. We didn't install everything that's on the media though. The list was basically gathered by doing this search on the upgraded node:

```
zypper se -t package -s |grep \(System | cut -d \| -f 2
```


But looking at the package diff between a node upgraded from Cloud 5 and one cleanly installed from the latested Cloud6 image there seems to be a lot more potential for cleanup (maybe even before the upgrade is done).

rhafer:
I rebased this on top of #177 so please ignore those changes for now.

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -282,9 +283,11 @@
             cloud_available = true if name.include? "Cloud"
           end
 
+          autoyast_template="autoyast"
+          autoyast_template="autoyast-upgrade" if mnode[:state] == "os_upgrading"
```
jsuchome: Place to use ternary operator?

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -331,6 +331,7 @@ en:
         unknown: '%{count} Unknown'
         unready: '%{count} Not Ready'
         pending: '%{count} Pending'
+        crowbar_upgrade: '%{count} Upgrade'
```
mmnelemane: Question: Wont you need to add "Upgrade" in # Global state list too in this file ? (around Line 627)

tboerger:
Some offenses can be fixed but :+1:

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
rhafer: IIRC YaST  was selecting the default pattern for add-on when upgrading. But it seems we removed the default pattern options from the image meanwhile so I guess this can go away now.

rhafer:
> Some offenses can be fixed

Yeah, will try to address some of that. But hound is now complaining about lines > 80 that seems really wrong and is not what we configured.

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
rhafer: Yeah, I was unsure about this one. But it's an allowed autoyast section for the upgrade and doesn't really hurt. I can remove it though if you really insist :)

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
rhafer: @vuntz this one is now fixed

mjura:
Except that hound is bitching +1

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -176,6 +176,48 @@ def create_proposal
...
+
+    # wait for the pxe_config to be updated, then reboot the nodes
+    discovery_dir = "#{NodeObject.admin_node[:provisioner][:root]}/discovery/"
+    pxecfg_subdir = "bios/pxelinux.cfg"
+
+    all_nodes.each do |node|
+      next if node.admin? || node["platform"] == "windows"
+
+      boot_ip_hex = node["crowbar"]["boot_ip_hex"]
+      node_arch = node["kernel"]["machine"] rescue "x86_64"
```
rhafer: @tboerger Do you think this needs to be addressed. If yes, what's a good way of fixing it?

mmnelemane: @tboerger, @ralf: How about
    node_arch = node["kernel"]["machine"] || "x86_64"  ? Its similar to what is suggested in the link by hound. "No exception for control flow."

mmnelemane:
+1

crowbar_framework/app/models/crowbar_service.rb
rhafer: That still throws an exception when node["kernel"] in nil

vuntz: `node.fetch("kernel", {}).fetch("machine", "x86_64")`

vuntz: That being said, why would node["kernel"] be nil?

rhafer: @vuntz Good question, but why would node["kernel"]["machine"] else throw an excpetion? 

But this check might really be unneeded here. I admittedly copied it from somewhere else (provisioner cookbook)

rhafer:
Note, this is still not bring the nodes into the "ready" state. For that I need to deactivate all the non-core barclamps before (or as part of) calling ```prepare_nodes_for_os_upgrade```, but I'd prefer to create another PR for that.

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
jsuchome: I think that when we are upgrading, it's really best to minimize the profile to stuff that needs changing. As this should stay as it is, I'm also for removing.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -176,6 +176,47 @@ def create_proposal
...
+      node_arch = node["kernel"]["machine"]
+      pxe_conf = "#{discovery_dir}/#{node_arch}/#{pxecfg_subdir}/#{boot_ip_hex}"
+      ready_for_reboot = false
+
+      while Time.now.to_i < Time.now.to_i + 120 && !ready_for_reboot
+        @logger.debug("waiting for pxe configuration to be updated for #{node.name}")
+        if File.file?(pxe_conf)
+          File.open(pxe_conf).each_line do |line|
+            line.chomp!
+            @logger.debug("line: #{line}")
```
jsuchome: Forgotten debug line? (Or maybe you want to have it there...)

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -176,6 +176,47 @@ def create_proposal
...
+            if line =~ /^DEFAULT\s+.+_install$/
```
jsuchome: Wouldn't it be possible to grep the file instead?

jsuchome:
I do not see any problems here (but I haven't test it)

crowbar_framework/app/models/crowbar_service.rb
rhafer: No, that should not be there. I'll update.

crowbar_framework/app/models/crowbar_service.rb
rhafer: grep, like calling out the external tool? I think that's really ugly.

rhafer:
Addressed the last comments and squashed the first two commits into a single one.

jsuchome:
+1 (without testing)

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -258,7 +258,8 @@
 
         when os =~ /^(open)?suse/
           append << "install=#{install_url} autoyast=#{node_url}/autoyast.xml"
-          append << "ifcfg=dhcp4 netwait=60"
+          append << " ifcfg=dhcp4 netwait=60"
+          append << " autoupgrade=1" if mnode[:state] == "os_upgrading"
```
vuntz: I don't think you need the leading ` `: we are joining the items of `append` (`append.join(" ")`).

vuntz:
If possible, I'd like to not use `_` but `-` in the state names. We allowed that to move forward for `crowbar_upgrade` (due to the led / pie), but I don't see anything depending on this for these states.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -176,6 +176,46 @@ def create_proposal
     base
   end
 
+  def prepare_nodes_for_os_upgrade
+    all_nodes = NodeObject.all
```
vuntz: What about:
```
```

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -176,6 +176,46 @@ def create_proposal
...
+          end.close
+        end
+        sleep(5) unless ready_for_reboot
+      end
+      if ready_for_reboot
+        @logger.debug("Rebooting node #{node.name} for operating system upgrade")
+        node.ssh_cmd("/sbin/reboot")
+      end
+    end
+    true
```
vuntz: Should we return `false` if one node has not rebooted?

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
```diff
@@ -0,0 +1,171 @@
...
+
+            if ! curl -s -o /mnt/root/.ssh/authorized_keys.wget \
+                http://$HTTP_SERVER/authorized_keys ||\
+                grep -q "Error 404" /mnt/root/.ssh/authorized_keys.wget; then
+                rm -f /mnt/root/.ssh/authorized_keys.wget
+            else
+                chmod 644 /mnt/root/.ssh/authorized_keys
+                cat /mnt/root/.ssh/authorized_keys.wget >> /mnt/root/.ssh/authorized_keys
+                rm -f /mnt/root/.ssh/authorized_keys.wget
+            fi
```
vuntz: Doesn't really matter, but more out of curiosity: this bit (for authorized keys) should not be needed, since the old file is still there, right?

vuntz:
For the record, I can live with the PR being merged and my comments addressed in a followup PR.

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
rhafer: Ok, will remove that.

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
rhafer: Right. I think the same is true for autoyast_set_hostentries.sh, I'll remove those two here.

crowbar_framework/app/models/crowbar_service.rb
rhafer: Probably, but currently I have the problem that the ```node.ssh_cmd("/sbin/reboot")``` call on SLE12 systems will always return an error. I guess this is because it cause the ssh connection to be killed. Might need to figure something out for that.

rhafer:
> If possible, I'd like to not use _ but - in the state names. We allowed that to move forward for crowbar_upgrade (due to the led / pie), but I don't see anything depending on this for these states.

Ok, I'll switch that.

PR #142, opened 2015-11-24
Added option to select FS type for a node.

jsuchome:
Tested manually. 

The fs type values were took from /etc/sysconfig/storage

### DIFF
--- a/crowbar_framework/app/controllers/nodes_controller.rb
+++ b/crowbar_framework/app/controllers/nodes_controller.rb
@@ -400,6 +400,7 @@ class NodesController < ApplicationController
         description: :description,
         availability_zone: :availability_zone,
         intended_role: :intended_role,
+        default_fs: :default_fs,
         raid_type: :raid_type,
         raid_disks: :raid_disks
       }.each do |attr, param|
--- a/crowbar_framework/app/helpers/form_helper.rb
+++ b/crowbar_framework/app/helpers/form_helper.rb
@@ -41,6 +41,20 @@ module FormHelper
     )
   end
 
+  # Possible values for nodes default file system
+  def fs_for_select(selected)
+    fs_values = [
+      "ext3",
+      "ext4",
+      "xfs",
+      "btrfs"
+    ]
+    options_for_select(
+      fs_values.zip(fs_values),
+      selected.to_s
+    )
+  end
+
   def booleans_for_select(selected)
     options_for_select(
       [
--- a/crowbar_framework/config/locales/crowbar/en.yml
+++ b/crowbar_framework/config/locales/crowbar/en.yml
@@ -443,6 +443,7 @@ en:
       availability_zone: 'Availability Zone'
       availability_zone_hint: 'Availability zones allow to arrange sets of either OpenStack Compute or OpenStack Block Storage hosts into logical groups. If empty, the default availability zone will be used.'
       intended_role: 'Intended Role'
+      default_fs: 'Filesystem Type'
       controller: 'Controller'
       compute: 'Compute'
       storage: 'Storage'
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -295,7 +295,8 @@ if not nodes.nil? and not nodes.empty?
                       architecture: arch,
                       is_ses: node[:provisioner][:suse] &&
                         !node[:provisioner][:suse][:cloud_available] && node[:provisioner][:suse][:storage_available],
-                      crowbar_join: "#{os_url}/crowbar_join.sh")
+                      crowbar_join: "#{os_url}/crowbar_join.sh",
+                      default_fs: mnode[:crowbar_wall][:default_fs] || "ext4")
           end
 
         when os =~ /^(hyperv|windows)/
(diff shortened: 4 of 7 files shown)


crowbar_framework/app/controllers/nodes_controller.rb
```diff
@@ -139,6 +139,11 @@ def bulk
               dirty = true
             end
 
+            unless node.default_fs == node_attributes["default_fs"]
```
tboerger: I don't believe that we want to provide it on the bulk edit section as we already have not enough space.

jsuchome:
I should probably also adapt RAID disks FS...

crowbar_framework/app/controllers/nodes_controller.rb
jsuchome: It's actually _not_ in bulk UI (so this part should be unnecessary)

tboerger: Yeah, but this change is only related to processing of bulk editing :)

jsuchome:
I've adapted RAID's `/` partition as well, but not `/boot`. Does it matter which FS does boot partition have?

crowbar_framework/app/controllers/nodes_controller.rb
jsuchome: Hence unnecessary. Anyway, it doesn't hurt here, does it? I mean, it would be useful to have it also in bulk view, so once the UI manages to fit everything we want into the page, this part is ready

abel-navarro:
I'm no expert in this area, but I'd say we either provide a closed list of FS to choose from or we add a validation checking the input was correct. I think this is not provided right now.

jsuchome:
@abel-navarro ouch, I forgot to commit some parts :-)

crowbar_framework/app/controllers/nodes_controller.rb
vuntz: @jsuchome the issue is that if you set the default fs of a node, then go to the bulk edit page, then you'll trigger this code and you will lose the default fs that was set.

vuntz: (at least, that's what I'm guessing)

crowbar_framework/app/helpers/form_helper.rb
```diff
@@ -41,6 +41,19 @@ def roles_for_select(selected)
     )
   end
 
+  # Possible values for nodes default file system
+  def fs_for_select(selected)
+    options_for_select(
+      [
+        ["ext3", "ext3"],
```
vuntz: Do we really want ext3?

Should we put some words of caution about the non-ext4 fs (or at least, btrfs -- since it's totally untested for controllers or kvm compute nodes)

crowbar_framework/app/controllers/nodes_controller.rb
jsuchome: Ah, true

crowbar_framework/app/helpers/form_helper.rb
jsuchome: Well, as for now, we still use ext3 for RAID. But it's just an option.

ext4 is still default, so you user has to change it explicitely. My idea was that further explanation when to use btrfs and possibly why not should go to Deployment guide.

But if we would consider anything else than ext4 as unsupported, than hint in UI definitely makes sense

jsuchome: @vuntz  the values are from /etc/sysconfig/storage , there's even ext2 as allowed value.

jsuchome: @vuntz do you think we should say about some filesystem that is/is not supported?

vuntz: I think they're all supported by SLES12, so my only concern is really performance impacts that btrfs could have if we don't tweak anything for it in our product. Not really sure how to phrase it, though :/

jsuchome: Hm, something like "Think carefully before you change it" looks too discouraging, I'd say...

I think we should leave it this way (without any additional comment) and rather add some more text to Deployment guide... although I'm not sure what else could we write there besides the fact that we recommend btrfs for nova-docker nodes...

crowbar_framework/app/helpers/form_helper.rb
```diff
@@ -41,6 +41,19 @@ def roles_for_select(selected)
     )
   end
 
+  # Possible values for nodes default file system
+  def fs_for_select(selected)
+    options_for_select(
```
tboerger:
```ruby
    values = [
      "ext3",
      "ext4",
      "xfs",
      "btrfs"
    ]

    options_for_select(
      values.zip(values),
      selected.to_s
    )
```

:)

abel-navarro:
+1 for me provided that customer documentation mentions the btrfs for docker nodes.

crowbar_framework/app/helpers/form_helper.rb
vuntz: @jsuchome works for me.

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -296,6 +296,7 @@ en:
         target_platform_not_set: 'Not set'
         availability_zone: 'Availability Zone'
         intended_role: 'Intended Role'
+        default_fs: 'Filesystem Type'
```
vuntz: I think this string is not used and can be dropped.

vuntz:
+1

PR #145, opened 2015-11-27
refactoring the installers_controller responses

MaximilianMeister:
make it usable for crowbarctl

### DIFF
--- a/crowbar_framework/app/controllers/installers_controller.rb
+++ b/crowbar_framework/app/controllers/installers_controller.rb
@@ -46,47 +46,72 @@ class InstallersController < ApplicationController
   # Provides the restful api call for
   # /installer/start 	POST 	triggers installation
   def start
-    if Crowbar::Installer.successful?
+    if params[:force]
+      ret = Crowbar::Installer.install!
+      case ret[:status]
+      when 501
+        respond_to do |format|
+          format.json do
+            head :not_implemented
+          end
+        end
+        format.html do
+          flash[:alert] = ret[:msg]
+          redirect_to installer_url
+        end
+      end
+
       respond_to do |format|
         format.json do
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/controllers/installers_controller.rb
```diff
@@ -51,18 +51,38 @@ def start
       # the shell Process will be spawned in the background and therefore has
       # not a direct return value which we can use here
       if Crowbar::Installer.installing?
-        flash[:notice] = I18n.t(".installers.start.installation_ongoing")
+        respond_to do |format|
+          format.json do
+            head :im_used
```
MaximilianMeister: just for reference: http://httpstatusdogs.com/226-im-used

not sure if there is some more suitable status-code... 226 refers to GET and we do a POST. any suggestions?

MaximilianMeister:
previous gating was successful https://ci.suse.de/job/openstack-mkcloud/12092/

just added a `410` in case crowbar is already installed. so lets wait for the next one ...

crowbar_framework/app/controllers/installers_controller.rb
```diff
@@ -42,27 +46,74 @@ def status
   # Provides the restful api call for
   # /installer/start 	POST 	triggers installation
   def start
```
tboerger: A lot of code duplication, maybe you should create some variables and do the respond_to block only at the bottom of the method?

tboerger:
Beside the code duplication :+1:

MaximilianMeister:
> A lot of code duplication, maybe you should create some variables and do the respond_to block only at the bottom of the method?

@tboerger good idea. got rid of code duplication

rsalevsky:
+1

PR #146, opened 2015-11-27
Add suse-12.1/s390x

dirkmueller:
We need this entry so that the crowbar_register.sh scripts
get generated properly.


vuntz:
You probably need a migration for existing setups to get this.

dirkmueller:
migration 32 takes care of that still :)

vuntz:
Fair enough, +1

tboerger:
:+1:

PR #148, opened 2015-11-30
S390x bits

dirkmueller:


vuntz:
I would probably have removed the sleshammer 11SP3 bit from boot device detection in a separate commit (this confused me at first), but no big deal.

+1

dirkmueller:
@vuntz splitted the two commits

vuntz:
still +1

MaximilianMeister:
:+1:

PR #150, opened 2015-12-02
provisioner: fix is_ses check for autoyast profile

tserong:
Commit bf4e98b removed the cloud_available and storage_available
attributes, which broke the is_ses check when setting up the autoyast
profile for storage-only deploys.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -273,6 +273,15 @@ if not nodes.nil? and not nodes.empty?
 
           packages = node[:provisioner][:packages][os] || []
 
+          # Need to know if we're doing a storage-only deploy so we can tweak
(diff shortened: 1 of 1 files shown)


tserong:
I'm not sure if this is the most correct way to do this check (@vuntz?) but it seems to do the trick...

vuntz:
Didn't have time to think about it since you've posted the comments in the other PR. This feels hacky, though... I'm happy to merge that if this is blocking storage, but I feel we could export in the data bags that represent the repos whether they provide openstack / ceph features. Which could be enough for this.

tserong:
Yeah, I agree it'd be better to expose it by feature. Still, if you don't mind merging this in the meantime, until we can improve it, that'd be helpful :-)

tboerger:
I think on the short term I'm fine with that. :+1:

tserong:
(rebased)

rsalevsky:
+1

PR #151, opened 2015-12-02
webui: Do not have branding js/css compiled in application js/css

vuntz:
This is an issue when compiling the assets as this means that we can't
simply replace the compiled branding bits, but we have to re-compile the
application assets.

cc @tboerger


tboerger:
Perfect, +2
Can be merged after successful gating.

rsalevsky:
+1

PR #152, opened 2015-12-03
webui: Hide some steps in the installer

vuntz:
Some steps make no sense to display, so just group them together.

cc @MaximilianMeister @kwwii


vuntz:
Ken was mentioning that we could possibly merge the remaining first two as "Starting up" and change the last one as "Finishing...".

rsalevsky:
I was thinking about making the steps anyway more dynamic and sending them in the json instead of define them in the view. This can be useful for the backup-restore case as well as for other distros

tboerger:
I would prefer more steps to show more progress if it can be updated frequently enough.

vuntz:
> I would prefer more steps to show more progress if it can be updated frequently enough.

For more steps, the user can just look at the log directly. I don't think it makes the UI clearer to have more steps.

tboerger:
As the SE on our last meeting said... It makes the users happier to see progress ;)

vuntz:
Won't be used...

PR #153, opened 2015-12-03
Distinguish between cases when nodes are not ready and when

jsuchome:
some proposals were not applied yet (bsc#954406)


jsuchome:
Gating failed in tempest run, so it's unrelated

jsuchome:
@vwallfahrer as you've reported the original buggy behavior, would you mind testing if this PR helps? Thanks

vwallfahrer:
+1

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -393,14 +393,24 @@ def proposal_update
...
             missing_nodes = answer[1].map { |node_dns| NodeObject.find_node_by_name(node_dns) }
 
-            unready_nodes = missing_nodes.map(&:alias)
+            unready_nodes = missing_nodes.select { |n| n.state != "ready" }.map(&:alias)
             unallocated_nodes = missing_nodes.reject(&:allocated?).map(&:alias)
 
-            flash[:notice] = "#{t('barclamp.proposal_show.commit_proposal_queued')}: #{(unready_nodes - unallocated_nodes).join(", ")}"
-
+            unless unready_nodes.empty?
+              flash[:notice] = "#{t('barclamp.proposal_show.commit_proposal_queued')}: #{(unready_nodes - unallocated_nodes).join(", ")}"
```
tboerger: PLEASE use the power of rails i18n:

```ruby
t(
  "barclamp.proposal_show.commit_proposal_queued",
  nodes: (unready_nodes - unallocated_nodes).join(", ")
)
```

```yaml
commit_proposal_queued_dependency: 'Successfully queued the proposal until %{nodes} are deployed'
```

That should be applied for all the translation strings. Do you know what happens if we add more languages that require the list on a different place?

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -393,14 +393,24 @@ def proposal_update
...
+            end
             unless unallocated_nodes.empty?
               flash[:alert]  = "#{t('barclamp.proposal_show.commit_proposal_queued_unallocated')}: #{unallocated_nodes.join(", ")}"
             end
+            if unready_nodes.empty? && unallocated_nodes.empty?
+              # find out which proposals were not applied yet
+              deps = @service_object.proposal_dependencies(ServiceObject.proposal_to_role(@proposal, params[:barclamp]))
```
tboerger:
```ruby
deps = @service_object.proposal_dependencies(
  ServiceObject.proposal_to_role(
    @proposal,
    params[:barclamp]
  )
)
```

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -393,14 +393,24 @@ def proposal_update
...
+              missing_barclamps = deps.select.map do |dep|
+                depprop = Proposal.where(barclamp: dep["barclamp"], name: dep["inst"]).first
+                dep["barclamp"] unless depprop["deployment"][dep["barclamp"]]["crowbar-status"] == "success"
+              end.compact
+              flash[:notice] = "#{t('barclamp.proposal_show.commit_proposal_queued_dependency')}: #{missing_barclamps.join(", ")}"
```
tboerger: Same translation problem as above.

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -393,14 +393,24 @@ def proposal_update
           if answer[0] == 202
             missing_nodes = answer[1].map { |node_dns| NodeObject.find_node_by_name(node_dns) }
 
-            unready_nodes = missing_nodes.map(&:alias)
+            unready_nodes = missing_nodes.select { |n| n.state != "ready" }.map(&:alias)
             unallocated_nodes = missing_nodes.reject(&:allocated?).map(&:alias)
 
-            flash[:notice] = "#{t('barclamp.proposal_show.commit_proposal_queued')}: #{(unready_nodes - unallocated_nodes).join(", ")}"
-
+            unless unready_nodes.empty?
```
tboerger: For the condition you take only `unready_nodes` but for the flash message you remove `unallocated_nodes`?

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -393,14 +393,24 @@ def proposal_update
...
             unless unallocated_nodes.empty?
               flash[:alert]  = "#{t('barclamp.proposal_show.commit_proposal_queued_unallocated')}: #{unallocated_nodes.join(", ")}"
             end
+            if unready_nodes.empty? && unallocated_nodes.empty?
+              # find out which proposals were not applied yet
+              deps = @service_object.proposal_dependencies(ServiceObject.proposal_to_role(@proposal, params[:barclamp]))
+              missing_barclamps = deps.select.map do |dep|
+                depprop = Proposal.where(barclamp: dep["barclamp"], name: dep["inst"]).first
+                dep["barclamp"] unless depprop["deployment"][dep["barclamp"]]["crowbar-status"] == "success"
+              end.compact
```
vuntz: This is actually not enough: you nee to also check if the proposal is queued or not. See https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/deployment_queue.rb#L228

vuntz: Btw, it might be good to actually rework things a bit so we can get this information directly from the deployment queue. This would involve not just returning the nodes that are not ready, but a structure with the deps that are not ready and the nodes that are not ready.

crowbar_framework/app/controllers/barclamp_controller.rb
jsuchome: For `unallocated_nodes` there is a different flash message

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -584,13 +584,36 @@ def proposal_update
...
+                "barclamp.proposal_show.commit_proposal_queued_unallocated",
+                nodes: unallocated_nodes.join(", ")
+              )
+            end
+            if unready_nodes.empty? && unallocated_nodes.empty?
+              # find out which proposals were not applied yet
+              deps = @service_object.proposal_dependencies(
+                ServiceObject.proposal_to_role(@proposal, params[:barclamp])
+              )
+              missing_barclamps = deps.select.map do |dep|
```
tboerger: Why `select.map`?

tboerger:
Beside my single question (I'm not sure if that correctly works) you get my :+1:

jsuchome:
Tested again and it worked. Two +1's -> merging

PR #154, opened 2015-12-04
add network.json validation

MaximilianMeister:
this will replace the check in the installation script
As long as the network.json is not valid, the user is not
allowed to start the installation

cc @vuntz

### DIFF
--- a/crowbar_framework/lib/crowbar/installer.rb
+++ b/crowbar_framework/lib/crowbar/installer.rb
@@ -61,7 +61,7 @@ module Crowbar
           failed: failed?,
           success: successful?,
           installing: installing?,
-          validNetwork: valid_network_json?,
+          network: network_status,
           errorMsg: error_msg,
           successMsg: success_msg,
           noticeMsg: notice_msg
@@ -84,8 +84,10 @@ module Crowbar
         steps_done.include?(:initial_chef_client)
       end
 
-      def valid_network_json?
-        Crowbar::Validator::NetworkValidator.new.validate
+      def network_status
+        validator = Crowbar::Validator::NetworkValidator.new
+        validator.cache if validator.network_changed?
+        validator.status
       end
 
       def steps_done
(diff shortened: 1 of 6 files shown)


MaximilianMeister:
not sure yet where or how to cache the last modification time of `network.json`, right now the validation script gets called on each request. any ideas?

tboerger:
> not sure yet where or how to cache the last modification time of network.json, right now the validation script gets called on each request. any ideas?

Rails.cache?

crowbar_framework/lib/crowbar/installer.rb
```diff
@@ -77,6 +78,10 @@ def initial_chef_client?
         steps_done.include?(:initial_chef_client)
       end
 
+      def valid_network_json?
+        Crowbar::Validator::NetworkJsonValidator.new.validate
```
tboerger: Please call it just `NetworkValidator`

MaximilianMeister:
@tboerger @vuntz I have used `Rails.cache` now. please have another look

MaximilianMeister:
I just realized that I need to adapt the `/installer/start` API in the back end as well, to not allow installation... will fix tomorrow.

MaximilianMeister:
related PR for the crowbar-client is https://github.com/crowbar/crowbar-client/pull/43

rsalevsky:
Can you drop the code from `crowbar_framework/lib/crowbar/checks/network.rb`? Otherwise we have some duplicated code.

MaximilianMeister:
> Can you drop the code from crowbar_framework/lib/crowbar/checks/network.rb? Otherwise we have some duplicated code.

right, thx I forgot about it. it is not used by anything right now right?

rsalevsky:
@MaximilianMeister Yes, it's not used at all.

rsalevsky:
+1

aplanas:
I do not like that `bin/network-json-validator` is called to validate the JSON document. IMHO this logic needs to be in a library used both in the CLI and here.

But I understand that this is not in the scope of the PR.

+1

PR #157, opened 2015-12-08
Make network cookbook compatible with Leap

vuntz:


chef/cookbooks/network/attributes/default.rb
```diff
@@ -18,8 +18,11 @@
   default[:network][:base_pkgs] = ["bridge-utils",
                                    "vlan"]
   default[:network][:ovs_pkgs] = ["openvswitch",
```
rsalevsky: Makes it sense to remove the line break? The intention is anyway wrong so I prefer removing the line break or
```ruby
default[:network][:ovs_pkgs] = [
  "openvswitch",
  "openvswitch-switch"
]
```

vuntz: @rsalevsky added a commit for this

rsalevsky:
+1

aplanas:
I was confused because the link : )

+1

chef/cookbooks/network/attributes/default.rb
```diff
@@ -15,11 +15,18 @@
...
+  default[:network][:ovs_pkgs] = [
+    "openvswitch",
+    "openvswitch-switch"
+  ]
+  # openSUSE uses the module shipped with upstream kernel
+  if node[:platform] == "suse"
+    default[:network][:ovs_pkgs].push("openvswitch-kmp-default")
+  end
   # SLES11 uses a different service name for openvswitch
   if node[:platform] == "suse" && node[:platform_version].to_f < 12.0
```
tboerger: I think we can drop the SLE 11 features?

chef/cookbooks/network/attributes/default.rb
```diff
@@ -27,18 +34,26 @@
...
-                                   "vlan"]
-  default[:network][:ovs_pkgs] = ["linux-headers-#{`uname -r`.strip}",
-                                  "openvswitch-datapath-dkms",
-                                  "openvswitch-switch"]
+  default[:network][:base_pkgs] = [
+    "bridge-utils",
+    "vlan"
+  ]
+  default[:network][:ovs_pkgs] = [
+    "linux-headers-#{`uname -r`.strip}",
```
tboerger: To be really correct I would replace this with `"linux-headers-#{node["kernel"]["release"]}",`

tboerger:
Small NITPIK but otherwise :+1:

chef/cookbooks/network/attributes/default.rb
dirkmueller: Unrelated to the purpose of this PR, cleanup can be done later

dirkmueller:
added "do not merge" until M8 is released.

aplanas:
Time to merge?

PR #159, opened 2015-12-08
Enable use of Leap for nodes

vuntz:
We probably can't merge that as is, if only because I'm referencing a personal OBS project.

### DIFF
--- a/crowbar_framework/config/repos-cloud.yml
+++ b/crowbar_framework/config/repos-cloud.yml
@@ -134,3 +134,75 @@ suse-12.1:
       required: "recommended"
       url:
       ask_on_error: false
+opensuse-42.1:
+  x86_64:
+    leap-42_1:
+      name: "openSUSE-Leap"
+      required: "mandatory"
+      features: ["os", "ha", "ceph"]
+      repomd:
+        tag: ["obsproduct://build.opensuse.org/openSUSE:Leap:42.1/openSUSE/42.1/ftp/x86_64",
+              "obsproduct://build.opensuse.org/openSUSE:Leap:42.1/openSUSE/42.1/cd/x86_64"]
+        key_md5: "d8039df72c407bc7dd7c1d9c66d58627"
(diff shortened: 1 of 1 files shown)


dirkmueller:
the other parts could be merged.. if it would be in a separate PR..

vuntz:
@dirkmueller I've put the other parts in https://github.com/crowbar/crowbar-core/pull/163 in case we want to merge that earlier.

dirkmueller:
adding "do not merge" before M8 is released

crowbar_framework/config/repos-cloud.yml
```diff
@@ -97,3 +97,75 @@ suse-12.1:
...
+      name: "Chef-10"
+      required: "mandatory"
+      #repomd:
+      #  tag: "obsrepository://build.opensuse.org/systemsmanagement:chef:10/openSUSE_Leap_42.1"
+      #  key_md5: "52a2a3f69993789a5362844ec81442dd"
+      #url: "http://download.opensuse.org/repositories/systemsmanagement:/chef:/10/openSUSE_Leap_42.1"
+      repomd:
+        tag: "obsrepository://build.opensuse.org/home:vuntz:crowbar:3.0/openSUSE_Leap_42.1"
+        key_md5: "aef42abc9d4311b5f72f46db225f2575"
+      url: "http://download.opensuse.org/repositories/home:/vuntz:/crowbar:/3.0/openSUSE_Leap_42.1"
```
tboerger: Why do you need this home project here?

vuntz: Because it was hackweek :-) We need to move that to some other location, as mentioned in the PR.

tboerger: But the crowbar repo got a higher build priority and you should be a maintainer of it :)

vuntz:
Let's close for now, no interest in pushing this further.

PR #160, opened 2015-12-08
logging: Prefer rsyslog to systemd-logger on Leap

vuntz:
An alternative is to remove the systemd-logger in the cookbook. But this
way seems cleaner (for now).

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
@@ -305,6 +305,9 @@ sync
       <package>supportutils-plugin-suse-openstack-cloud</package>
   <% end -%>
 <% end -%>
+<% if @platform == "opensuse" -%>
+      <package>rsyslog</package>
+<% end -%>
 <% @packages.each do |package| -%>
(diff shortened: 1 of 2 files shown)


jsuchome:
+1

rsalevsky:
+1

mjura:
+1

dirkmueller:
+1

tboerger:
Just to know why, why do we need rsyslog instead of systemd-logger?

vuntz:
> Just to know why, why do we need rsyslog instead of systemd-logger?

For the logging barclamp.

tboerger:
> For the logging barclamp.

Does it make sense to switch to systemd-logger? Is remote logging not supported by it?

vuntz:
> > For the logging barclamp.
> 
> Does it make sense to switch to systemd-logger? Is remote logging not supported by it?

journald has some features for this, but this requires rewriting the logging barclamp. And it's unclear if it's actually as convenient.

tboerger:
> journald has some features for this, but this requires rewriting the logging barclamp. And it's unclear if it's actually as convenient.

Thanks for explaining it.

PR #161, opened 2015-12-09
Allow to define roles for disks owned by nodes

jsuchome:
When such role (ceph,swift, sbd etc.) for each device exists,
other barclamps should use it when claiming devices for themselves.

This is based on https://github.com/crowbar/barclamp-crowbar/pull/1306 and https://github.com/crowbar/barclamp-deployer/pull/248

Some screenshots:
- Selecting new role (can be changed even after Save)
  crowbar-node-disk-roles-selecting
- All roles already selected AND claimed (no further change possible, but we can see the info)
  crowbar-node-disk-roles-swift


chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -270,6 +270,24 @@ def self.claimed(node,owner)
...
+        #   - first with the disks that have the given role set
+        #   - second with the disks with no assigned role
+        def self.prepared_for(node, role)
+          disk_roles = node["crowbar_wall"]["disk_roles"] || {}
+          prepared_role = []
+          without_role = []
+          unclaimed(node).each do |d|
+            disk_role = disk_roles.fetch(d.unique_name, "")
+            if disk_role == role
+              prepared_role << d
```
rsalevsky: prepared_role.push(d)

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -270,6 +270,24 @@ def self.claimed(node,owner)
...
+            elsif disk_role.empty?
+              without_role << d
```
rsalevsky: without_role.push(d)

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -270,6 +270,24 @@ def self.claimed(node,owner)
           end
         end
 
+        # return 2 lists of unclaimed disks:
+        #   - first with the disks that have the given role set
+        #   - second with the disks with no assigned role
+        def self.prepared_for(node, role)
```
rsalevsky: Where is this function called?

jsuchome: See related PR's, like https://github.com/crowbar/crowbar-ceph/pull/28

jsuchome:
cc @tboerger @MaximilianMeister (btw, I think it was already reviewed before repo merge)

crowbar_framework/app/helpers/form_helper.rb
```diff
@@ -55,6 +55,31 @@ def fs_for_select(selected)
     )
   end
 
+  def beautify_disk_owner(owner)
+    @owner2label ||= {
+      "Ceph" => t(".disk_role_ceph"),
+      "Cinder" => t(".disk_role_cinder"),
+      "Swift" => t(".disk_role_swift"),
+      "LVM_DRBD" => t(".disk_role_drbd"),
+      "sbd" => t(".disk_role_sbd")
```
mjura: Don't we need also role for software RAID?
Btw. it is bad that we have different format names for different disk roles. Can we standardize this too?

crowbar_framework/app/models/node_object.rb
```diff
@@ -307,6 +307,42 @@ def default_fs=(value)
...
+      end
+    end
+  end
+
+  def add_unclaimed_disks_to_roles
+    # Check for current unclaimed disks, and add unclaimed disk to
+    # disk_roles hash if it is not there yet.
+    # (New disk could have been added to the node)
+    unclaimed_physical_drives.each do |d|
+      unique_name = unique_device_for(d[0])
```
MaximilianMeister: `d.first`

crowbar_framework/app/models/node_object.rb
```diff
@@ -307,6 +307,42 @@ def default_fs=(value)
...
+      unless crowbar_wall["disk_roles"][unique_name]
+        crowbar_wall["disk_roles"][unique_name] = ""
+      end
+    end
+  end
+
+  def disk_roles(refresh = false)
```
MaximilianMeister: i would use an options hash and fetch from it. that gives the calling function more sense.
```ruby
def disk_roles(options = {})
  refresh = options.fetch :refresh, false
  ...
end
```

crowbar_framework/app/views/nodes/_form.html.haml
```diff
@@ -108,6 +108,43 @@
...
+          %tbody
+            - all_disks = all_disks_info
+            - (@node.crowbar_wall["claimed_disks"] || {}).each do |unique_name, disk|
+              %tr
+                %td
+                  = all_disks.fetch(unique_name, "-")
+                %td
+                  = unique_name
+                %td
+                  %span.form-control{ :disabled => "disabled" }
```
MaximilianMeister: can we use the new ruby hash syntax here? `disabled: "disabled"`

crowbar_framework/app/helpers/form_helper.rb
jsuchome: About RAID - those are set at different time, right? You have to prepare RAID before node installation, and than there's no way (and reason) to change it. But these disk roles are for non-system disks.

About names - I guess you are right, but wouldn't it be better in some other PR?

crowbar_framework/app/views/nodes/_form.html.haml
jsuchome: Well, rest of the file keeps the same syntaxc...

crowbar_framework/app/helpers/form_helper.rb
jsuchome: Also, those are current strings used for claiming the disks; isn't it better, just to be on safe side, to leave them as they are, so we don't run into confusion after upgrade?

crowbar_framework/app/views/nodes/_form.html.haml
```diff
@@ -108,6 +108,43 @@
...
+                    = beautify_disk_owner(disk["owner"])
+
+            - @node.disk_roles({:refresh => true}).each do |unique_name, disk_role|
```
tboerger: `- @node.disk_roles(refresh: true).each do |unique_name, disk_role|`

mjura:
There is issue with `Software raid`, I've tested this PR and installed nova with soft RAID 10 on disks /dev/vdc and /dev/vdd. UI doesn't show up that those disks were allocated and showing that /dev/vda is Boot disk.

jsuchome:
@mjura Do you have the output of knife node show from your scenario?

dirkmueller:
do not merge before M8

mjura:
> @mjura Do you have the output of knife node show from your scenario?

Yes, I've generated this for you
```
"crowbar_wall": {
      "ipmi": {
      },
      "status": {
        "ipmi": {
          "user_set": false,
          "address_set": false
        }
      },
      "openstack": {
        "availability_zone": ""
      },
      "intended_role": "no_role",
      "default_fs": "ext4",
      "raid_type": "raid1",
      "raid_disks": [
        "/dev/vdc",
        "/dev/vdd"
      ],
      "boot_device": "vda",
      "claimed_disks": {
        "/dev/vda": {
          "owner": "Boot"
        }
      }
```

jsuchome:
This could be the solution for the RAID problem mentioned by @mjura: https://github.com/crowbar/barclamp-crowbar/pull/1361

crowbar_framework/app/views/nodes/_form.html.haml
```diff
@@ -108,6 +108,43 @@
...
+              %tr
+                %td
+                  = all_disks.fetch(unique_name, "-")
+                %td
+                  = unique_name
+                %td
+                  %span.form-control{ :disabled => "disabled" }
+                    = beautify_disk_owner(disk["owner"])
+
+            - @node.disk_roles(refresh : true).each do |unique_name, disk_role|
```
mjura:
```
F, [2016-01-12T08:06:50.477630 #20009] FATAL -- : 
SyntaxError (/opt/dell/crowbar_framework/app/views/nodes/_form.html.haml:135: syntax error, unexpected ':', expecting ')'
 @node.disk_roles(refresh : true).each do |unique_name, disk_role|
```

jsuchome: Hm, crowbar guys offered this way, I think. Could you pass `@node.disk_roles({ :refresh => true })` instead?

vuntz: Remove the space before `:`.

tboerger: yeah, `- @node.disk_roles(refresh: true).each do |unique_name, disk_role|
` should be correct.

crowbar_framework/app/views/nodes/_form.html.haml
tboerger: So when do we start to use the modern variant?

jsuchome:
This looked like a nice idea at some point...

PR #163, opened 2015-12-09
Minimal work to enable Leap nodes

vuntz:
This change only makes it possible to enable Leap when allocating a
node (assuming that the install media is available), but several other
changes are needed to have working Leap support.


rsalevsky:
+1

nkrinner:
+1

dirkmueller:
+1

PR #165, opened 2015-12-11
Support allocating Debian 8 nodes

vuntz:
This seems to mostly work fine now; it requires some more love (I added some FIXME) and it conflicts with the PR for Leap support, so marking as WIP.

But I could possibly split some of the commits for earlier merge if there's interest.


dirkmueller:
adding "do not merge" before M8 is released

vuntz:
Let's close for now, no interest in pushing this further.

PR #168, opened 2015-12-15
Fix color of list headings

rsalevsky:
Color before fix:
cinder_before

Color after fix:
cinder_fixed

@kwwii Are you fine with this change? I changed the color to the same as before but I don't know if this change was by intention. The Problem was that the trash bin item was not visible.


dirkmueller:
Please add the bug reference (949927) to the commit message.

mjura:
It is still not enough GREEN colors, but bug is fixed

+1

aplanas:
+1 when the commit message is changed

rsalevsky:
Please not merge yet, I'll wait for Ken's feedback.

tboerger:
:+1:

rsalevsky:
Ken reviewed the changes, so let's merge.

PR #171, opened 2015-12-16
Only use bootproto "none" for bond slaves

rhafer:
Should be "static" (with no IP Address) for OVS slaves.


aplanas:
+1

vuntz:
+1, but I'm starting to feel the logic is a bit complicated and it might be worth adding some comments or links to what the file would look like in the different cases. Maybe it's just me, though?

rhafer:
> but I'm starting to feel the logic is a bit complicated and it might be worth adding some comments or links to what the file would look like in the different cases. Maybe it's just me, though?

@vuntz It's not only you. I tried something in: https://github.com/crowbar/crowbar-core/pull/173
Though I am still unsure if that is really a big improvement ;) .

PR #172, opened 2015-12-17
Revert "Old cli tools wrapper rewrite"

toabctl:
Reverts crowbar/crowbar-core#129

### DIFF
--- a/bin/barclamp_lib.rb
+++ b/bin/barclamp_lib.rb
@@ -1,6 +1,6 @@
 #
 # Copyright 2011-2013, Dell
-# Copyright 2013-2015, SUSE Linux GmbH
+# Copyright 2013-2014, SUSE LINUX Products GmbH
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


aplanas:
Reasons: https://github.com/crowbar/crowbar-core/pull/129#issuecomment-165244279 and https://github.com/crowbar/crowbar-core/pull/129#issuecomment-165302524

MaximilianMeister:
:+1: for now unless there is a quicker fix (but maybe we have to adapt the gating to use `crowbar_batch` and deploy `pacemaker` as well ?)

tboerger:
So somebody else can take this pr. How should I work if even the test environment doesn't work like expected? Core components like pacemaker are not tested at all? Awesome...

toabctl:
> So somebody else can take this pr. How should I work if even the test environment doesn't work like expected? Core components like pacemaker are not tested at all? Awesome...

We have only basic testing due to limited hardware resources. But you can run the tests multiple times with different parameters...

PR #175, opened 2015-12-17
Raise CROWBAR_THREADS to 10 (bsc#952171)

rsalevsky:
We need to raise the workers because deploying a proposal and accessing
the web ui at the same time is not possible. Also we need to raise the
workers instead of threads due to crowbar is not thread safe.

PS: This is the amount of workers which I found out while try and error.

### DIFF
--- a/chef/cookbooks/crowbar/files/default/crowbar.service
+++ b/chef/cookbooks/crowbar/files/default/crowbar.service
@@ -14,7 +14,7 @@ TimeoutStopSec=10
 WorkingDirectory=/opt/dell/crowbar_framework
 
 Environment="CROWBAR_ENV=production"
-Environment="CROWBAR_THREADS=1"
+Environment="CROWBAR_THREADS=10"
 Environment="CROWBAR_WORKERS=5"
 Environment="CROWBAR_LISTEN=127.0.0.1"
 Environment="CROWBAR_PORT=3000"
--- a/chef/cookbooks/crowbar/templates/default/sysconfig.crowbar.erb
+++ b/chef/cookbooks/crowbar/templates/default/sysconfig.crowbar.erb
@@ -11,10 +11,10 @@ CROWBAR_ENV="production"
 ## Path:            Crowbar
 ## Description:     Maximum number of threads
 ## Type:            integer
-## Default:         1
+## Default:         10
 ## ServiceRestart:  crowbar
 # Sets the maximum number of threads used by the web server.
-CROWBAR_THREADS="1"
+CROWBAR_THREADS="10"
 
 ## Path:            Crowbar
 ## Description:     Maximum number of worker processes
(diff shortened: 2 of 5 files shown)


tboerger:
Have you tested this setting for the mentioned xen deployments?

rhafer:
I fear this will significantly raise our memory requirements. It seems that a single puma worker for crowbar is using at least 100MB if (non-shared) RAM. So adding 12 additional worker to the default setup mean an additionally 1.2GB of RAM required just for puma. I am not strictly against that, but this feels like a lot.

I've no idea about puma, but is it possible to use threads instead? Or some more lightweight Concurrency model? (IIRC rainbows offered a couple of different Concurrency models for different types of load)

aplanas:
@rhafer puma can use threads, you can configure the min and max number of threads per worker. I think that the problem is that maybe crowbar is not thread-safe, but I am not sure if this was tested

vuntz:
Can you elaborate on the issues we would have with threads, but that we wouldn't have with workers?

tboerger:
Within standup we decided to keep the workers at 5 and instead give 4 threads a try with a full mkcloud run.

dirkmueller:
also, doesn't that depend on the number of nodes that are running in the cloud? it seems to me that we'll hit the same issue with nodes+1 in the cloud

rsalevsky:
@rhafer @tboerger raising the threads seams to work, did some manual testing and gating works too

tboerger:
:+1:

rhafer:
I (similar to @dirkmueller 's question) still like to understand the limitations of this. With 5 worker and 10 threads it means now we can have ~50 requests running in parallel, right? And more than that will cause crowbar to block again? (I don't think this is an issue currently, but it might become one in the future)

And are we sure that crowbar is really thread-safe (enough) so that this doesn't cause any problems?

tboerger:
> I (similar to @dirkmueller 's question) still like to understand the limitations of this. With 5 worker and 10 threads it means now we can have ~50 requests running in parallel, right? And more than that will cause crowbar to block again? (I don't think this is an issue currently, but it might become one in the future)
> 
> And are we sure that crowbar is really thread-safe (enough) so that this doesn't cause any problems?

Of course it's limiting the number of nodes as well, but it have been pretty similar for rainbows in the past... For bigger deployments we have to fix different places, like getting rid of applying proposals in realtime instead of async execution.

rhafer:
> but it have been pretty similar for rainbows in the past

Actually AFAIK rainbows had a completely different concurrency model which didn't block one worker (thread  or process) for every long running request. (Which AFAIK was the whole point why rainbows was used).

I am fine with this change in general I am just trying to understand the implications of it.

dirkmueller:
@tboerger but what do you suggest when we're about 6 months too late for switching to an async model?

aplanas:
5 workers with 10 threads give 50 concurrent requests, that is not so bad for an admin tool. Maybe a little better option here is to provide a thread pool (min:max).

+1 for me

rsalevsky:
This PR is not about discussing the switch to puma and if this change was right or wrong. 

Raising the threads works and fixes the concurrency problem. When there is no objection I will merge this soon.

dirkmueller:
@rsalevsky there were several open questions here, e.g. the threads vs workers thing

PR #177, opened 2016-01-07
Introduce "crowbar-upgrade" state

rhafer:
This basically forward porting the required pieces from Tex.


crowbar_framework/vendor/assets/javascripts/jquery/ledUpdate.js
```diff
@@ -71,6 +71,7 @@
             val.status.ready,
             val.status.failed,
             val.status.unknown,
+            val.status.crowbar-upgrade,
```
tboerger: That's simply invalid :)

crowbar_framework/app/controllers/nodes_controller.rb
```diff
@@ -262,6 +262,7 @@ def status
                 "pending" => 0,
                 "unready" => 0,
                 "building" => 0,
+                "crowbar-upgrade" => 0,
```
tboerger: What about naming it simply `upgrade`?

tboerger:
Please fix the offenses, afterwards I will give a proper review.

crowbar_framework/app/controllers/nodes_controller.rb
jsuchome: This is port of https://github.com/crowbar/barclamp-crowbar/pull/1358.
We might need `upgrade` for other tasks, that's why we've chosen `crowbar-upgrade`

tboerger: Then you have to fix invalid javascript like `val.status.crowbar-upgrade` as this results in `val.status.crowbar - upgrade` (upgrade is an undefined variable)

tboerger: BTW, as this is anyway just for displaying on the frontend even other upgrade states can be grouped within that.

jsuchome: Hm, that looks correct.

tboerger: It's not correct. It's INVALID, e.g. https://github.com/crowbar/crowbar-core/pull/177#discussion_r49061027
You can not use a dash within an object attribute.

jsuchome: My _correct_ was meant to your comment :-)

rhafer:
> Please fix the offenses, afterwards I will give a proper review.

You might wanna review https://github.com/crowbar/barclamp-crowbar/pull/1358 then, because that's where it is coming from. Sorry, should have pointed to that earlier.

tboerger:
> You might wanna review crowbar/barclamp-crowbar#1358 then, because that's where it is coming from. Sorry, should have pointed to that earlier.

Sure, but still a lot offenses as those checks are not enabled on barclamp-crowbar.

rhafer:
> Sure, but still a lot offenses as those checks are not enabled on barclamp-crowbar.

Yeah, I'll handle those when updating with the latest change from the other PR.

rhafer:
It might be that we'll need to add at least a no-op crowbar_upgrade.rb recipe as well. Since nodes coming from Tex will have that in their run_list. I am not sure yet if we ever need to run chef-client on the nodes though before transitioning to the next upgrade state. So I am leaving that out for now.

crowbar_framework/app/models/node_object.rb
```diff
@@ -498,17 +498,20 @@ def status
...
-      "building"  #yellow
+      "failed"    # flashing red
+    when "hardware-installing", "hardware-install", "hardware-installed",
+         "hardware-updated", "hardware-updating"
+      "building"  # yellow
+    when "crowbar_upgrade"
+      "crowbar_upgrade" # blue
     else # including: installing, installed, reinstall, reboot, recovering, readying, applying
-      "unready"   #spinner
+      "unready"   # spinner
```
rhafer: Hm, interesting. Before that change you complained about missing space after ```#```

jsuchome:
Actually, maybe we could use the recipe for crowbar_upgrade.rb to perform the next actions that are needed for the nodes (e.g. shutting down OpenStack services before the system upgrade)

rhafer:
> Actually, maybe we could use the recipe for crowbar_upgrade.rb to perform the next actions that are needed for the nodes (e.g. shutting down OpenStack services before the system upgrade)

Yeah. So let's just add the role and recipe as part of the PR that really needs it. This one currently doesn't I think.

jsuchome:
+1

crowbar_framework/app/models/node_object.rb
aplanas: maybe is referring to the spaces between the string and the comment

vuntz:
+1. For reference, the crowbar-upgrade role is in https://github.com/crowbar/crowbar-core/pull/179

PR #178, opened 2016-01-08
Some fixes in apply_role for handling of nodes removed from proposal

vuntz:


crowbar_framework/app/models/service_object.rb
```diff
@@ -1187,6 +1187,9 @@ def apply_role(role, inst, in_queue)
       return [405, message]
     end
 
+    # Invalidate cache as apply_role_pre_chef_call can save nodes
+    pre_cached_nodes = {}
```
MaximilianMeister: with this change can we get rid of the subsequent use in line 1224?

```
- nobj = pre_cached_nodes[node] || NodeObject.find_node_by_name(node)
+ nobj = NodeObject.find_node_by_name(node)
```

tboerger:
I have not tested it but codewise :+1:

aplanas:
some gating problems (tempest: `TestServerBasicOps.test_server_basicops`) But the proposals are applied, so I guess that is unrelated

crowbar_framework/app/models/service_object.rb
vuntz: Actually, yes and no: we should change the line you mention, but not in that way: the cache can still be re-filled and that line should participate in that effort.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1225,7 +1226,8 @@ def apply_role(role, inst, in_queue)
       pids = {}
       unless non_admin_nodes.empty?
         non_admin_nodes.each do |node|
-          nobj = pre_cached_nodes[node] || NodeObject.find_node_by_name(node)
+          pre_cached_nodes[node] ||= NodeObject.find_node_by_name(node)
```
MaximilianMeister: maybe I am a bit blind here, but how can `pre_cached_nodes[node]` ever be not empty at this point?

vuntz: Right now, it's always empty. But I wrote the code so that in case this changes, it will still work as intended.

rsalevsky:
+1

aplanas:
+1

PR #179, opened 2016-01-08
Execute actions on nodes that are being prepared for upgrade

jsuchome:
Nodes should already be in crowbar_upgrade state from the previous steps
executed still with previous SUSE Cloud version. Now we want to execute
further actions on them, but not anything else from normal runlists.

See https://trello.com/c/LtOPxaRz/616-5-s34p8-enhance-crowbar-upgrade-mode-to-shut-down-openstack


chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,114 @@
...
+return unless node[:platform] == "suse"
+
+if node["crowbar_wall"]["crowbar_openstack_upgrade"]
+
+  # Actions to be run last, when admin node is already new SUSE Cloud version (6)
+  # Nodes will be restarted and their system upgraded after this.
+
+  # put HA nodes into maintenance mode (TODO is it needed? It wasn't in 4-5 upgrade...)
+
+  # Stop pacemaker resources in all clusters
```
jsuchome: In upgrade 4-5, we were doing this from admin server by ssh-ing to the founder nodes first, so we could be sure resouces were stopped before the next action. (https://github.com/SUSE-Cloud/suse-cloud-upgrade/blob/master/lib/suse-cloud-upgrade-4-to-5-pre#L313)

Now, doing that here from recipe, we cannot be sure which node goes first.

So is it wrong to stop pacemaker resources here? Is it enough to use maintenance mode instead of this? Or do we need to add some orchestration to ensure founder nodes are first? @vuntz ?

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,114 @@
...
+# saved in the node structure.
+
+return unless node[:platform] == "suse"
+
+if node["crowbar_wall"]["crowbar_openstack_upgrade"]
+
+  # Actions to be run last, when admin node is already new SUSE Cloud version (6)
+  # Nodes will be restarted and their system upgraded after this.
+
+  # put HA nodes into maintenance mode (TODO is it needed? It wasn't in 4-5 upgrade...)
```
rhafer: Do you intend to stop corosync as well? If yes, you can't put the nodes into maintenance mode, because stopping corosync would cause fencing. (See https://bugzilla.suse.com/show_bug.cgi?id=917593 for details)

jsuchome: I don't know :-(. See also my own comment https://github.com/crowbar/crowbar-core/pull/179#discussion_r49316944

jsuchome: As mentioned in the referenced bug report, we could put whole cluster into maintenance mode. I'm still not sure if it is needed, though.

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
vuntz: Interesting case. First, you don't have access to the sync mark API since you don't depend on the crowbar-pacemaker cookbook and nothing else depends on it. So one solution would be to depend on that cookbook -- but that's relatively bad as we're in the crowbar cookbook here and that shouldn't depend on something like this. Maybe a suggests would work.

Another solution is to actually run what you're running here on all nodes of the cluster, not just the founder. It shouldn't matter much to run on all nodes: it would be important if we were looking at the exit code for crm (because in that case, two crm calls in parallel could possibly conflict with each other). But we do it in a bash resource, so it won't have an impact. The only reason to run on the founder only is to either avoid running the code multiple times (on different nodes) or to guarantee that it's run only once per cluster because the code can create issue. But for "crm resource stop", as mentioned, that should be fine.

But in all cases: yes, you do need to change something because you can't risk stopping the services on some nodes before the pacemaker resources are stopped.

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
vuntz: Hrm, I'm actually unsure for that one. I'd leave it out for now and see if we have issues. (Putting the cluster in maintenance mode has other interesting side-effects like: which code triggers the "go out of maintenance mode"?)

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
rhafer: > So one solution would be to depend on that cookbook -- but that's relatively bad as we're in the crowbar cookbook here and that shouldn't depend on something like this.

True. Though we could add a role (+recipe) to the pacemaker barclamp (or some other that fit better) that has "crowbar_upgrade" in elements_states. That recipe could to do the shutdown pieces for the cluster resources.
Am I thinking to complicated again?

vuntz: How would the nodes get that role?

rhafer: Argh, we're doing a crowbar_upgrade state transition already on the Cloud 5 side. Otherwise that would have been my answer.

Now it is: ```NodeObject.rebuild_run_list```

chef/data_bags/crowbar/template-crowbar.json
```diff
@@ -25,17 +25,24 @@
     "crowbar": {
       "crowbar-revision": 10,
       "crowbar-applied": false,
+      "schema-revision": 11,
```
jsuchome: OK, this is confusing. Should I bump `"crowbar-revision"`, or use new `"schema-revision"` here? There currently is 010 migrate file...

crowbar_framework/app/models/node_object.rb
```diff
@@ -724,7 +724,11 @@ def rebuild_run_list
     crowbar["run_list_map"] = {} if crowbar["run_list_map"].nil?
 
     # Cull by state
-    map = crowbar["run_list_map"].select { |k,v| v["states"].include?("all") or v["states"].include?(self.crowbar["state"]) }
+    map = crowbar["run_list_map"].select do |k, v|
+      v["states"].include?(state) ||
```
vuntz: As mentioned in https://github.com/crowbar/barclamp-crowbar/pull/1362, I think using `state` is wrong.

chef/data_bags/crowbar/template-crowbar.json
vuntz: Wow, it's a bug I guess. Should be schema-revision, not crowbar-revision.

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,114 @@
...
+      for i in /etc/init.d/openstack-* \
+               /etc/init.d/openvswitch-switch \
+               /etc/init.d/ovs-usurp-config-* \
+               /etc/init.d/drbd \
+               /etc/init.d/openais;
+      do
+        if test -e $i
+        then
+          initscript=`basename $i`
+          $i
```
vuntz: I guess this should be `$i stop`?

jsuchome: Ah, yes. Now you now I've not tested it yet :-)

tboerger:
:+1:

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,113 @@
...
+        for resource in $(crm configure show | grep ^$type | cut -d " " -f2); do
+          crm resource stop $resource
+        done
+      done
+    EOF
+    only_if { ::File.exist?("/usr/sbin/crm") }
+  end
+
+  # Stop openstack services
+  bash "stop HA and openstack services" do
```
aplanas: Silly question to understand better this: why not to use the `service` resource here?

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,113 @@
...
+  # Actions to be run last, when admin node is already new SUSE Cloud version (6)
+  # Nodes will be restarted and their system upgraded after this.
+
+  # put HA node into maintenance mode (TODO is it needed? It wasn't in 4-5 upgrade...)
+
+  # Stop pacemaker resources in all clusters
+  bash "stop pacemaker resources" do
+    code <<-EOF
+      for type in clone ms primitive; do
+        for resource in $(crm configure show | grep ^$type | cut -d " " -f2); do
```
aplanas: do we have a LWRP in crowbar-ha to do this? or maybe we do not want to require it in this recipe?

aplanas:
maybe the last codeclimate complains makes sense to fix

aplanas:
basically +1 from me

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
jsuchome: We do not require anything like that, and this is in crowbar barclamp which usually does not require anything else

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
jsuchome: Hm, I'm not sure... first we do not know which services are running there, so we are looking for them with that pattern... that way it would be ugly to use first bash and than go back to chef resource.

But maybe... we actually have the node data, so maybe we could look inside and actually know which services should be running? But i'm not sure if we have such mapping somewhere in the node or we would need manually created mapping (like `:keystone => [ "openstack-keystone"] ` etc.)

nkrinner: I looked into this too when I worked on something similar earlier in the upgrade work. And as far as I found out, it is not easy to extract information about about running services from the node data. There was also some inconsistent naming, iirc. There is an unfinished PR somewhere, where Max started to work on a LWRP to do that. But work on that has not been finished so far and to me it looked like too much work to reuse that code.

nkrinner:
+1 lgtm

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,113 @@
...
+    only_if { ::File.exist?("/usr/sbin/crm") }
+  end
+
+  # Stop openstack services
+  bash "stop HA and openstack services" do
+    code <<-EOF
+      for i in /etc/init.d/openstack-* \
+               /etc/init.d/openvswitch-switch \
+               /etc/init.d/ovs-usurp-config-* \
+               /etc/init.d/drbd \
```
rhafer: Hm, if we really need to dump the postgresql database at some point we might actually need to keep drbd running on the nodes until that happened.

Questions is at what point do we want to dump the database?

jsuchome: So shouldn't we do that before this step? Or do you want to shutdown everything but the database?

aplanas: IMHO the dump needs to be done in this step exactly: after stopping openstack and before stopping the rest of the services.

Also in a HA env, not sure if shutting down also openais will cause problems with the pgsql agent

rhafer: Hm, right. I guess that means we need to add yet another step to the upgrade process?

jsuchome: Maybe not. This may be silly, but what about:

1. during shutting down pacemaker resources, do not shut down postgress ones
2. find out if we are at the node that is actually running the DB (db_node = true)
3. if that is true, do the db dump
4. during shutting down openstack, do not shutdown openais and drbd
4a. only when when dbnode == true, after the dump is finished, shut down also openais and drbd

This could work without any additional state or orchestration, or am I missing something?

aplanas: @jsuchome this can work if there are not any other openstack service running in the cluster outside HA. Also before the `pg_dump` we need to wait that the rest of the nodes already stopped the openstack services (HA and no HA ones, including the ones outside the cluster)

 But is cool that we can identify the node that contains the database!

As I see, the basic schema is (please, I am aware that my mental model for HA and Chef is incorrect):

1. Stop openstack pacemaker resources
2. Stop non HA openstack resources
3. Wait (sync) until no openstack service is running in any node of the cloud
4. `pg_dump` and copy the database somewhere else
5. Stop the rest of the pacemaker services, inc, the database, drdb and openais

sayalilunkad: Just a little bit of re shuffling:
1. shutdown pacemaker services related to openstack 
2. shutdown openstack services
3. if dbnode == true -> take backup
4. shutdown other pacemaker services
5. shutdown openais and drbd

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,113 @@
...
+      do
+        if test -e $i
+        then
+          $i stop
+        fi
+      done
+    EOF
+  end
+
+elsif node["crowbar_wall"]["crowbar_upgrade"]
```
jsuchome: We actually do not need stuff from this point bellow, because that's currently Cloud5 only related. I've added it here in case we'd like to use it for future product upgrade, but it might be better to remove it if it is confusing...

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
jsuchome: @rhafer @aplanas what do you think abouthttps://gist.github.com/jsuchome/59ba5b0f81ccba9d3f59#file-gistfile1-txt-L14 ?

jsuchome: @rhafer @aplanas @sayalilunkad What about this?

http://pastebin.nue.suse.com/19360/src

rhafer: * /tmp if obviously not a good place for storing a backup :), but I guess that was just for a quick test. For now I'd probably just put it on the drbd volume where the real database lives as well.
* Is there a way to check if enough space is available where we put the db dump? (I guess there is no 100% reliable way). This  also raises the question what to do when there's not enough space available.
* the etherpad mentions we should offer the user a possibility to (manually) download the db dump (if we'd store it on drbd we'd need to do that before stopping the drbd resources.

sayalilunkad: There is a command to find the size of the database in postgres so we can somehow get an idea of the size of the backup and then check if there is enough space. But in the case where there is not enough space I am not sure what we can do..

sayalilunkad: Ok this looks like a better solution, we get the exact size of the backup by running this sudo -u postgres pg_dumpall $PGD_OPTIONS | wc -c

aplanas: @jsuchome some comments:

1. According to the code, the `pg_dump` is only done if we are in HA, but if the database is not in a cloud, the dump is never done
2. We are doing the dump meanwhile there are openstack services running that can change the database meanwhile

I think that we are close now

aplanas: Ah, for the space problem I will use gzip, this is how usually do my backups

jsuchome: @aplanas re 2., openstack services should be shut down by pacemaker. Anyway, I'll update this PR soon

rhafer: Hm, interesting. Though I guess it's also pretty time consuming on a large DBs.

jsuchome: BTW, we're commenting on the wrong else branch, this one is about disabling services, not stopping them. See the top of file for the stopping actions

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,152 @@
...
+    # if we have drbd, find out the device
+    if ::Kernel.system("service drbd status")
+      device = node["drbd"]["rsc"]["postgresql"]["device"]
+      dump_location = `grep #{device} /etc/mtab | cut -d ' ' -f 2`
+      dump_location += "/db.dump"
+    end
+
+    bash "dump database content" do
+      code <<-EOF
+        su - postgres -c 'pg_dumpall > #{dump_location}
```
sayalilunkad: Missing quote in the end. Should be su - postgres -c 'pg_dumpall > #{dump_location}'

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
aplanas: @jsuchome uhm, maybe I am missing something, but there are three kind of openstack services that are still running at this point:

1. openstack HA services in the same cluster but in different nodes (is a race)
2. openstack non-HA services running inside the cluster
3. openstack HA and non-HA services running in different nodes / clusters

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,152 @@
...
+
+if node["crowbar_wall"]["crowbar_openstack_upgrade"]
+
+  # Actions to be run last, when admin node is already new SUSE Cloud version (6)
+  # Nodes will be restarted and their system upgraded after this.
+
+  # put HA node into maintenance mode (TODO is it needed? It wasn't in 4-5 upgrade...)
+
+  # 1. Find the node which runs database
+  db_node = ::Kernel.system("service postgresql status")
```
aplanas: Oh cool! This works for non HA and HA?

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
jsuchome: 1+2 for the same cluster, all stop actions have been done before the DB dump. For HA services, pacemaker should stop them also at the other nodes in cluster

If they are in different clusters, that's true, they still might be running. You are right, we need some loop here. If we'd need to wait until all recipes are finished here, we might need to introduce yet another state (as also @rhafer mentioned if user want so access the DB dump)...

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
jsuchome: I believe so

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
jsuchome: @rhafer 
> the etherpad mentions we should offer the user a possibility to (manually) download the db 
> dump (if we'd store it on drbd we'd need to do that before stopping the drbd resources.

What about after the new pacemaker barclamp is reapplied (which starts drbd), and before user reapplies database barclamp? Is it too complicated?

rhafer: It might just be too late. E.g. a failure during the OS Upgrade (or a user error) could have destroyed the data.

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  # 3. Dump the database at the DB node
+  # FIXME check the available space before: sudo -u postgres pg_dumpall | wc -c
+  # FIXME find the correct place for the dump (drbd?)
+
+  dump_location = "/var/lib/pgsql/db-before-cloud6.dump"
+
+  # if we have drbd, find out the device
+  # FIXME what if we have shared NFS?
+  if ::Kernel.system("service drbd status")
```
rhafer: Why is this needed at all? If I understand correctly, we are already sure that postgresql is running here. And AFAIK we always store the data in the same location (/var/lib/pgsql). Can't we just assume that if postgresql is running here the correct device is mounted there?

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -32,6 +32,55 @@ def role_constraints
...
+
+      # mark the position in the upgrade process
+      node["crowbar_wall"]["crowbar_openstack_shutdown"] = true
+      node.save
+    end
+
+    # Adapt current proposal, so only non-db nodes have crowbar-upgrade role
+    # (crowbar-upgrade role is currently assigned to all nodes)
+    proposal = ProposalObject.find_proposal("crowbar", "default")
+    proposal["deployment"]["crowbar"]["elements"]["crowbar-upgrade"] -= db_nodes
```
rhafer: So if I understand correctly, this step will basically do nothing on the cluster when everything is deployed in a single cluster?

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
jsuchome: If we are really using always this location, than indeed this is not necessary.

crowbar_framework/app/models/crowbar_service.rb
jsuchome: Hm, right. So you're proposing to skip the commit when the "crowbar-upgrade" role list is is empty at this point, right?

jsuchome: Done in latest commit

rhafer: Well actually I was wondering what's the reason to not shutdown the non-db service on that cluster anyway? So that crowbar-db-dump would really just be about the database. And remove some of the duplication in the two recipes.

jsuchome: I'm not sure if I understand your proposal. I think that splitting the whole process into 3 pieces:
- shut down all non-db
- dump the db (only after previous step is surely done at all nodes)
- shut down the db (only after user has a chance to retrieve the dump)

makes it work in all special HA and non-HA cases, even if in some situations it could be done simpler.

The recipe cannot be just about the database anyway, we also need to shut down corosync at the end.

jsuchome: Or you mean that I should run the first round on all nodes, not just DB ones? That's what I also considered, but in that case I would need to add additional checks if node has db role or not into the recipe and I thought that assigning DB nodes different role directly makes it cleaner.

rhafer: > Or you mean that I should run the first round on all nodes, not just DB ones?

Yes that's what I meant. But I am ok if you think your current approach is easier.

I'd still prefer if we could reduce the code duplication. Would it be possible to move e.g. the part that shuts down all non-database resource into a separate recipe and include that ( ```include_recipe``` ) in the other recipes?

jsuchome: Yes, I thought about moving the duplicate parts into include, it should be next step.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -32,6 +32,79 @@ def role_constraints
...
+
+    # This proposal could return some error if there's not enough space for DB dump
+    # Controller must show the error and be able to call the function again once the problem
+    # is resolved
+    proposal_commit("default", false, false)
+
+    proposal = ProposalObject.find_proposal("crowbar", "default")
+    if proposal["deployment"]["crowbar"]["crowbar-status"] == "failed"
+      raise proposal["deployment"]["crowbar"]["crowbar-failed"]
+    end
```
jsuchome: This looks wrong to me, BTW. Is there any better way to show user just the exception I've raised from the chef recipe (see `raise` in `recipes/crowbar-db-dump.rb`) ?

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -32,6 +32,79 @@ def role_constraints
...
+
+    return true if proposal["deployment"]["crowbar"]["elements"]["crowbar-upgrade"].empty?
+
+    # Commit proposal so the shutdown actions from crowbar-upgrade get executed for non-db nodes
+    proposal_commit("default", false, false)
+  end
+
+  def dump_database
+    db_nodes = []
+    NodeObject.find("state:crowbar_upgrade").each do |node|
```
rhafer: ```db_nodes =  NodeObject.find("state:crowbar_upgrade AND roles:database-server")``` should also do it.

crowbar_framework/app/models/crowbar_service.rb
jsuchome: @rhafer , @vuntz ?

crowbar_framework/app/models/crowbar_service.rb
jsuchome: No, because nodes have currently only `crowbar-upgrade` as role. But looking into `run_list_map` saves the day.

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
```diff
@@ -0,0 +1,113 @@
...
+
+  ruby_block "check available space" do
+    block do
+      require "pathname"
+      path = Pathname.new(dump_location).split.first.to_s
+      available = `df | grep "#{path}" | sed "s/ \\+/ /g" | cut -d " " -f 4`
+      db_size = `su - postgres -c 'pg_dumpall | wc -c'`
+      if db_size.to_i > available.to_i
+        Chef::Log.fatal("DB size is: #{db_size}, available space on #{path} is #{available}")
+        raise "Not enough space on #{path} for the Database dump!"
```
sayalilunkad: Maybe it is a good idea to check here if there is enough space else where on the node that makes it possible to take the backup there and if not then raise the error. Also would be nice to add 'Please free some space and try again'  or something similar when the error is raised.

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,60 @@
...
+# saved in the node structure.
+
+return unless node[:platform] == "suse"
+
+if node["crowbar_wall"]["crowbar_openstack_shutdown"]
+
+  # Actions to be run last, when admin node is already new SUSE Cloud version (6)
+  # Nodes will be restarted and their system upgraded after this.
+
+  # 2. Stop all pacemaker resources
```
sayalilunkad: Probably change the numbering for the steps here now.

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,60 @@
...
+      for type in clone ms primitive; do
+        for resource in $(crm configure show | grep ^$type | cut -d " " -f2); do
+          crm resource stop $resource
+        done
+      done
+    EOF
+    only_if { ::File.exist?("/usr/sbin/crm") }
+  end
+
+  # 3. Stop openstack services and corosync
```
sayalilunkad: same as above

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
jsuchome: 1. I'm not sure. The state of the node might be quite different from one user setup to another, so I'm not sure we can find good solution by guessing.

2. Yes, but that's task for the UI part, not chef recipe

crowbar_framework/app/models/crowbar_service.rb
rhafer: Hm, ok. But it should still have the proposal role. Then this would work (unless I am missunderstanding rebuild_run_list change):

```db_nodes = NodeObject.find("state:crowbar_upgrade AND roles:database-config-default")```

Comment (jsuchome) [inline crowbar_framework/app/models/crowbar_service.rb]: I'll try that once I have my setup back

File: crowbar_framework/app/models/crowbar_service.rb (hunk 14 above)
Comment (rhafer) [inline crowbar_framework/app/models/crowbar_service.rb]: Doesn't proposal_commit returns [405, "<error message>" ] when the chef-client run on any of the nodes fails. The error message should be the same (log excerpt) as what is stored in the "crowbar-failed" attribute.

But I guess you're look for a way to return a more specific error message from the client?

Comment (jsuchome) [inline crowbar_framework/app/models/crowbar_service.rb]: It doesn't, although I expected something like that :-( So I assume I'm not doing something right...

File: crowbar_framework/app/models/crowbar_service.rb (hunk 13 above)
Comment (jsuchome) [inline crowbar_framework/app/models/crowbar_service.rb]: Done

File: crowbar_framework/app/models/crowbar_service.rb (hunk 14 above)
Comment (jsuchome) [inline crowbar_framework/app/models/crowbar_service.rb]: It's because crowbar_service.rb is overriding `apply_role` method, and although it calls `super` in it, it ignores the result of the one called from superclass: 

https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/crowbar_service.rb#L219

**Comment by jsuchome** [comment]
I removed WIP tag, I think it currently works as it should.

The remaining issue is the error reporting from the chef-client run - that seems like something more generic, so maybe better part of different PR.

**Comment by jsuchome** [comment]
> The remaining issue is the error reporting from the chef-client run - that seems like something
> more generic, so maybe better part of different PR.

This is based on @vuntz idea: http://pastebin.suse.de/19720

**Comment by jsuchome** [comment]
Latest update: deleted `element_run_list_order` on schema downgrade (spotted by @nkrinner )

**Comment by rhafer** [comment]
> The remaining issue is the error reporting from the chef-client run - that seems like something more generic, so maybe better part of different PR.

You mean something like what you proposed here: http://pastebin.suse.de/19720? 
Yeah, I am fine adding that with another PR.

File: crowbar_framework/app/models/crowbar_service.rb [hunk 19]
```diff
@@ -32,6 +32,73 @@ def role_constraints
...
+    # Adapt current proposal, so only non-db nodes have crowbar-upgrade role
+    # (crowbar-upgrade role is currently assigned to all nodes)
+    proposal = ProposalObject.find_proposal("crowbar", "default")
+    proposal["deployment"]["crowbar"]["elements"]["crowbar-upgrade"] -= db_nodes
+    proposal.save
+
+    return true if proposal["deployment"]["crowbar"]["elements"]["crowbar-upgrade"].empty?
+
+    # Commit proposal so the shutdown actions from crowbar-upgrade get executed for non-db nodes
+    proposal_commit("default", false, false)
```
rhafer: I guess we need error handling here as well, right?

jsuchome: Well, yes, if you mean the generic one (there are even more calls of `proposal_commit` I've added). But we do not expect any specific error raised from the recipe (`crowbar-upgrade` in this case)

nkrinner:
+1

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -32,6 +32,73 @@ def role_constraints
     end
   end
 
+  def shutdown_openstack_at_nodes
```
rhafer: I know I am being a pain, but I guess the name is now somewhat misleading. Because it only handles non-database nodes. Not very nice but more precise: ```shutdown_services_at_non_db_nodes```. Feel free to use something else.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -32,6 +32,73 @@ def role_constraints
...
+    db_nodes = []
+    NodeObject.find("state:crowbar_upgrade").each do |node|
```
rhafer: Do we need to make sure here, to only run node nodes that have  ["crowbar_wall"]["crowbar_upgrade"] and non of the other upgrade related flags? E.g. what can go wrong if we reapply the recipe to a node that already had "crowbar_openstack_shutdown" set?

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -32,6 +32,73 @@ def role_constraints
...
+
+    # Commit proposal so the shutdown actions from crowbar-upgrade get executed for non-db nodes
+    proposal_commit("default", false, false)
+  end
+
+  def dump_database
+    # Find the nodes with Database role
+    # (we can't look at node.roles, because that list has been reduced to just crowbar-upgrade item)
+    db_nodes = NodeObject.find("state:crowbar_upgrade AND run_list_map:database-server")
+
```
rhafer: Here, we might want to also only run this on nodes that have ```["crowbar_wall"]["crowbar_openstack_shutdown"]``` ?

crowbar_framework/app/models/crowbar_service.rb
jsuchome: I think I had removing "crowbar_upgrade" there before and forgot to commit it...

rhafer:
Any idea about that travis failure?

nkrinner:
The travis failure should be fixable with updated settings here: https://github.com/crowbar/crowbar-core/tree/master/crowbar_framework/spec/fixtures

crowbar_framework/app/models/crowbar_service.rb
jsuchome: There's a check for this value in the crowbar-db-dump recipe

rhafer:
Overall I am fine with this. I am still not a real fan of that orchestration via flags on ```crowbar_wall``` but I guess we discussed that already ;) So let's keep it for now.

crowbar_framework/app/models/crowbar_service.rb
jsuchome: Hm, ok, my comment was not relevant. So, what could go wrong? I do not think anything should go wrong, because in this stage, recipe only shuts down the services (non-DB). If they are  already stopped, that's fine (we have such case anyway in case of HA cluster)

chef/data_bags/crowbar/template-crowbar.json
```diff
@@ -18,24 +18,37 @@
...
     "crowbar": {
-      "crowbar-revision": 10,
+      "crowbar-revision": 0,
       "crowbar-applied": false,
+      "schema-revision": 11,
       "element_states": {
-        "crowbar": [ "all" ]
+        "crowbar": [ "all" ],
+        "crowbar-upgrade": [ "readying", "ready", "applying", "crowbar_upgrade" ],
+        "crowbar-db-dump": [ "readying", "ready", "applying", "crowbar_upgrade" ]
```
rhafer: Oh, I almost forgot this. Why do you want to run these for anything else but "crowbar_upgrade" here. I know we need it for tex to be able to restart chef-client, but here I am pretty sure we only need "crowbar_upgrade".

jsuchome: Good point. I've not tested it, but it makes sense.

crowbar_framework/app/models/crowbar_service.rb
rhafer: Ok, but shouldn't we react in someway if ```dump_database``` is called and non of the nodes have the right state? Doesn't it indicate that something went wrong? It might be a good idea to tell the user about that.

BTW,  ```dump_database``` is also a pretty ambiguous name here. Can we change this to something that is indicating that we're dumping the openstack database on the client nodes? (sorry, once more)

jsuchome: The question is, what exactly to tell user in such case... I'll try something

jsuchome: Because AFAIK, we do not expect going back in the wizard, right?

crowbar_framework/app/models/crowbar_service.rb
rhafer: Ok.

rhafer:
> Not sure yet what ficture exactly needs to get updated.

I guess it's crowbar_framework/spec/fixtures/data_bags/template-crowbar.json

rhafer:
+1 once the ci is green

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
```diff
@@ -0,0 +1,87 @@
...
+elsif node["crowbar_wall"]["crowbar_db_shutdown"]
+
+  # Now, stop the database.
+  # In case of HA setup, also stop remaining pacemaker resources and HA services.
+  bash "stop the database" do
+    code <<-EOF
+      if test -e /usr/sbin/crm; then
+        crm resource stop postgresql
+        crm resource stop fs-postgresql
+        crm resource stop vip-admin-database-default-data
```
vuntz: Coming with a late review, but these lines are not fully correct:

- there can be a `g-postgresql` group to stop
- `vip-admin-database-default-data` is the name in your setup, but `default-data` comes from the name of the pacemaker proposal; so you should somehow find the right name
- you should stop the ms and drbd resources too (in case drbd is used)

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
```diff
@@ -0,0 +1,87 @@
...
+      fi
+      for i in /etc/init.d/openstack-* \
```
vuntz: Didn't we already stop openstack bits in the recipe that is included from this one?

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,44 @@
...
+
+return unless node[:platform_family] == "suse"
+
+if node["crowbar_wall"]["crowbar_openstack_shutdown"]
+
+  include_recipe "crowbar::stop-services-before-upgrade"
+
+  # Stop DRBD and corosync.
+  # (Note that this node is not running database)
+  bash "stop HA and openstack services" do
```
vuntz: s/and openstack//

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -32,6 +32,86 @@ def role_constraints
...
+    proposal["deployment"]["crowbar"]["elements"]["crowbar-upgrade"] = []
+    proposal["deployment"]["crowbar"]["elements"]["crowbar-db-dump"] = db_nodes.map(&:name)
+    proposal.save
+
+    # This proposal could return some error if there's not enough space for DB dump
+    # Controller must show the error and be able to call the function again once the problem
+    # is resolved
+    commit_and_check_proposal
+
+    # If the function succeedes, we need to tell user where to find the database dump
```
vuntz: s/succeedes/succeeds/ ;-)

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -32,6 +32,86 @@ def role_constraints
...
+    # Commit proposal so the shutdown actions from crowbar-upgrade get executed for non-db nodes
+    commit_and_check_proposal
+  end
+
+  def dump_openstack_database
+    # Find the nodes with Database role
+    # (we can't look at node.roles, because that list has been reduced to just crowbar-upgrade item)
+    db_nodes = NodeObject.find("state:crowbar_upgrade AND run_list_map:database-server")
+
+    raise "There does not seem to exist any node running the database." if db_nodes.empty?
```
vuntz: Should we really raise an exception? Isn't it an issue in case we upgrade a Cloud 5 setup with no openstack deployment?

rhafer: Hm, seems I didn't made my point clear enough when requesting some error check here (https://github.com/crowbar/crowbar-core/pull/179#discussion_r50143198).

My intention was to have a check to make sure here, that all nodes returned here have the right thing in ["crowbar_wall"]["crowbar_openstack_shutdown"]. Mainly to prevent dump_openstack_database from doing anything before shutdown_services_at_non_db_nodes was called.

> Should we really raise an exception? Isn't it an issue in case we upgrade a Cloud 5 setup with no openstack deployment?

Is there much value in allowing to upgrade a Cloud 5 setup with no openstack deployment? Are you thinking about a ceph-only setup?

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
jsuchome: Not in one special case: when everything is in the same cluster as DB node, or directly on one node.

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
jsuchome: Hm, so I will again need a check for running resources before stopping.

(ms should be already stopped before)

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
jsuchome: True

crowbar_framework/app/models/crowbar_service.rb
vuntz: > Is there much value in allowing to upgrade a Cloud 5 setup with no openstack deployment? Are you thinking about a ceph-only setup?

I'm not thinking about anything specific setup, just one without database. I agree it's not that important, and would be fine skipping that. But if there's an easy fix...

jsuchome: > My intention was to have a check to make sure here, that all nodes returned here have the 
> right thing in ["crowbar_wall"]["crowbar_openstack_shutdown"]. Mainly to prevent 
> dump_openstack_database from doing anything before
> shutdown_services_at_non_db_nodes was called.

But again, this is checked in the recipe: https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb#L30 (and the else branch)

jsuchome: I have no idea about that exception either: it doesn't really help user in any way.

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
vuntz: > (ms should be already stopped before)

Hrm, I'm not sure it can, as the postgresql primitive depends on it. I might be wrong, though.

jsuchome: Anyway, I'm including it in a stopping loop in the followup PR

PR #180, opened 2016-01-12
Add root_password_hash to provisioner template

dirkmueller:
This makes it easier to discover the setting when you're looking
for it in the raw mode.


tboerger:
If that really helps :+1:

PR #185, opened 2016-01-14
Ucs fixes (bsc#961761)

dirkmueller:


vuntz:
+1

aplanas:
+1

crowbar_framework/app/controllers/ucs_controller.rb
```diff
@@ -303,6 +303,12 @@ def truncate(message)
   def sendXML(xmlString = "")
     uri = URI.parse(@ucs_url)
     http = Net::HTTP.new(uri.host, uri.port)
+    http.use_ssl = False
```
tboerger: We are not on python, that should be `false` and not `False`

crowbar_framework/app/controllers/ucs_controller.rb
```diff
@@ -303,6 +303,12 @@ def truncate(message)
...
+    if uri.scheme == "https"
+      http.use_ssl = True
```
tboerger: We are not on python, that should be `true` and not `True`

crowbar_framework/app/controllers/ucs_controller.rb
tboerger: Beside that setting it to false is anyway a waste.

PR #187, opened 2016-01-15
Added method to disable all non-core Proposals

rhafer:
This is needed to keep the non-core Cookbooks from being executed when a
node is coming back to "ready" after performing and OS upgrade. It's supposed
to be called before running the actual upgrade (i.e. before calling
prepare_nodes_for_os_upgrade()). Unfortunately we can't use
ServiceObject.destroy_active() for this, because that would cause (queued)
chef-client runs on the non-upgrades nodes. We don't want that here.

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -262,6 +262,27 @@ class CrowbarService < ServiceObject
     base
   end
 
+  def disable_non_core_proposals
+    # Find all non-core proposals and remove all roles that belong
+    # to those proposals from the nodes
+    active_non_core_roles = RoleObject.find_roles_by_name("*-config-*").reject(&:core_role?)
+    active_non_core_roles.each do |role|
+      @logger.debug("deactivating proposal #{role.name}")
+
+      roles_to_remove = role.elements.keys
+      roles_to_remove << role.name
+
+      roles_to_remove.each do |delete_role|
+        nodes = NodeObject.find("roles:#{delete_role}")
+        nodes.each do |node|
+          node.delete_from_run_list(delete_role)
+          node.save
+        end
+      end
+      role.destroy
+    end
+  end
+
   def prepare_nodes_for_os_upgrade
     upgrade_nodes = NodeObject.all.reject { |node| node.admin? || node[:platform] == "windows" }
 


crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -176,6 +176,27 @@ def create_proposal
     base
   end
 
+  def disable_non_core_proposals
```
rhafer: I am not completely sure if CrowbarService is really the right place for this. I am interested in feedback.

vuntz:
> This is needed to keep the non-core Cookbooks from being executed when a
> node is coming back to "ready" after performing and OS upgrade.

Oh, hrm.

Don't we need to also mark the proposals as not applied anymore? In which case maybe we need to combine both actions somehow (and do some apply_role to do the magic, but without running chef-client). Just an idea, might be more complicated than your approach.

rhafer:
> Don't we need to also mark the proposals as not applied anymore? 

Not sure what you mean by this, but when I delete in the ```*-config-*``` roles (that's what the ```role.destroy``` the proposals show up as not being applied in the UI. Do you mean something else?

vuntz:
@rhafer I meant exactly that. I just didn't read that closely enough.

So... I would possibly put this in a module in `lib/`... But @tboerger @rsalevsky @MaximilianMeister know better than me.

crowbar_framework/app/models/crowbar_service.rb
MaximilianMeister: As we have the architecture now with CrowbarService overriding/extending the ServiceObject I think it makes sense here.
Not sure if we manipulate much on NodeObjects in the lib either.

rhafer:
This should be ready for more review now. I opted for keeping the code in crowbar_service, because @MaximilianMeister is ok with it. And because that we're currently have almost all of the upgrade related calls.

MaximilianMeister:
codewise +1

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -262,6 +262,27 @@ def create_proposal
...
+      @logger.debug("deactivating proposal #{role.name}")
+
+      roles_to_remove = role.elements.keys
+      roles_to_remove << role.name
+
+      roles_to_remove.each do |delete_role|
+        nodes = NodeObject.find("roles:#{delete_role}")
+        nodes.each do |node|
+          node.delete_from_run_list(delete_role)
+          node.save
```
jsuchome: Just thinking: aren't we saving node(s) quite often this way? Wouldn't it be better to iterate over all nodes, for each one delete all roles that need to be removed and than save the node only once?

(But maybe it's not relevant or node.save is so fast that we do not need to bother...)

jsuchome:
This looks good to me. +1 (but not tested)

crowbar_framework/app/models/crowbar_service.rb
rhafer: Hm, you're right. I'll add that as a followup item to trello. But I'll merge this variant for now.

PR #188, opened 2016-01-15
Enable checking for provided repositories, whether or not they are enabled

jsuchome:


crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -156,6 +156,8 @@ def provided_and_enabled?(feature, platform = nil, arch = nil)
             next unless provided_features.include? feature
             found = true
 
+            next unless check_enabled
```
vuntz: I think you can `break`.

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -129,19 +129,19 @@ def check_all_repos
         repochecks
       end
 
-      def provided_and_enabled?(feature, platform = nil, arch = nil)
+      def provided_and_enabled?(feature, platform = nil, arch = nil, check_enabled = true)
```
vuntz: I'm somehow a bit uneasy with the fact that we allow disabling a feature of this method that is in the name of the method. I don't know if that's okay or not from a style perspective. (An alternative would be to move the content of this method to a helper and call the helper from here).

tboerger:
:+1:

jsuchome:
@vuntz better this way?

vuntz:
> @vuntz better this way?

Yes! I would just make the helper method private or protected (so just move it down in a later commit).

vuntz:
Thanks for the changes, +1

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -130,18 +130,59 @@ def check_all_repos
...
+
+      private
+
+      def provided_with_enabled?(feature, platform = nil, arch = nil, check_enabled = true)
         answer = false
 
         if platform.nil?
           all_platforms.each do |p|
-            if provided_and_enabled?(feature, p)
+            if provided_and_enabled?(feature, p, arch, check_enabled)
```
nkrinner: Maybe I am overlooking something here, but this calls the provided_and_enabled? method which does nothing but calling this method  (provided_with_enabled?) again, and also has the wrong number of arguments?

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -130,18 +130,59 @@ def check_all_repos
...
               answer = true
               break
             end
           end
         elsif arch.nil?
           arches(platform).each do |a|
-            if provided_and_enabled?(feature, platform, a)
+            if provided_and_enabled?(feature, platform, a, check_enabled)
```
nkrinner: Same here

nkrinner:
+1

vuntz:
+1

PR #189, opened 2016-01-18
Update dashboard css

rsalevsky:
TODO:
- [ ]  refine CSS to not affect other panels

### DIFF
--- a/crowbar_framework/app/assets/javascripts/barclamps/crowbar/application.js
+++ b/crowbar_framework/app/assets/javascripts/barclamps/crowbar/application.js
@@ -16,6 +16,26 @@
  */
 
 jQuery(document).ready(function($) {
+  function nodeStatus() {
+    $.ajax({
+      type: "GET",
+      dataType: "json",
+      url: Routes.index_barclamp_path("machines"),
+      success: function(data) {
+        $.each(data.nodes, function( key, node ) {
+          $("[data-alias="+node.alias+"]").removeClass (function (index, css) {
+            return (css.match (/(^|\s)status-border-\S+/g) || []).join(" ");
+          });
+          $("[data-alias="+node.alias+"]").addClass("status-border-"+node.status);
+        });
+      }
+    });
+
+    setTimeout(nodeStatus, 10000);
+  }
+
+  nodeStatus();
+
   $('[data-group-add]').live('submit', function(event) {
     var $el = $(event.target);
     var $input = $el.find('input[name=group]');
--- a/crowbar_framework/app/assets/stylesheets/content/panel.scss
+++ b/crowbar_framework/app/assets/stylesheets/content/panel.scss
@@ -1,0 +1,56 @@
+.panel-heading {
+  border-bottom: 0;
+}
+
+.panel .panel-heading h3 {
+  margin: 0;
+  padding: 0;
+}
+
+.panel>.list-group .list-group-item, .panel>.panel-collapse>.list-group .list-group-item {
+  border: 1px;
+  border-color: $status-border-gray;
+  border-left: 7px;
+  border-radius: 5px;
+  border-style: solid;
+  box-shadow: 1px 1px 0 $gray-lighter;
+  margin: 5px;
+}
+
+.panel-heading+.list-group .list-group-item:first-child {
+  border-top-width: 1px;
+}
+
+.panel>.list-group:last-child .list-group-item:last-child, .panel>.panel-collapse>.list-group:last-child .list-group-item:last-child{
+  border-bottom: 1px solid #ccc; ;
+}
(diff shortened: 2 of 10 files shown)


crowbar_framework/app/assets/javascripts/barclamps/crowbar/application.js
```diff
@@ -16,6 +16,26 @@
  */
 
 jQuery(document).ready(function($) {
+  function nodeStatus() {
```
tboerger: Why do you drop a generic plugin for this kind of fixed poor written function?

crowbar_framework/app/helpers/nodes_helper.rb
```diff
@@ -20,48 +20,28 @@ def nodes_by_role(role)
...
-
-    content_tag(
-      :span,
-      "",
-      :title => tooltip,
-      "data-piechart" => values.join(","),
-      "data-tooltip" => "true"
+  def link_to_node(node, handle)
+    link_to(
+      node_icon(node) + " " + truncate(node[:alias], length: 25),
```
tboerger: No, simply no... The tag helper avoids this ugly kind of string concat and you are bringing in this ugly stuff again...

crowbar_framework/app/helpers/nodes_helper.rb
```diff
@@ -20,48 +20,28 @@ def nodes_by_role(role)
...
-      result.push group[:status]["unknown"]
-
-      if group[:status]["building"]
-        result.push group[:status]["unready"] + group[:status]["pending"] + group[:status]["building"]
-      else
-        result.push group[:status]["unready"] + group[:status]["pending"]
-      end
-      result.push group[:status]["crowbar_upgrade"]
+  def node_icon(node)
+    return icon_tag "rocket" if node[:alias] == "crowbar"
```
tboerger: That's a pretty bad indicator, you can give any alias to the admin node. You are forced to check `admin?` to see if that is an admin node.

crowbar_framework/app/views/nodes/_node_groups.html.haml
```diff
@@ -7,16 +7,19 @@
     .col-xs-3{ "data-id" => group_name, "data-droppable" => @draggable.to_s, "data-group" => group_name }
       .panel.panel-default
         .panel-heading
-          %h2
-            = piechart_for group
```
tboerger: IMHO we should keep a generic indicator if one of the nodes is in a different state... maybe not this kind of piechart but still some indicator that makes it directly obvious that something is not green.

crowbar_framework/app/views/nodes/_node_groups.html.haml
```diff
@@ -7,16 +7,19 @@
...
+          %h3
             = truncate(group_name || t("unknown"), :length => 25)
 
         %ul.list-group
           - group[:nodes].values.sort_by { |handle| @nodes[handle][:alias] }.each do |handle|
             - @nodes[handle].tap do |node|
-              %li.list-group-item{ "data-id" => handle.parameterize, "data-draggable" => @draggable.to_s, "data-update" => group_change_path(:id => handle.parameterize, :group => "{0}".html_safe) }
-                %span.led{ :class => node[:status], :title => t(node[:state], :scope => "state", :default => node[:state].titlecase), "data-node" => handle }
-                = link_to(truncate(node[:alias], :length => 25), node_path(handle), :title => node[:description])
+              %li.list-group-item{ "data-id" => handle.parameterize,
```
tboerger: As you are anyway touching this old code, please use a real hash syntax.

```ruby
%li.list-group-item{ data: { id: handle.parameterize, ... } }
```

crowbar_framework/app/assets/stylesheets/content/panel.scss
```diff
@@ -0,0 +1,62 @@
+$status-border-gray: #ddd;
```
tboerger: No variable definitions within that stylesheets.

tboerger:
Sorry but you have no clue about CSS cascading, right?
Beside that you should just use the list items within a pnael instead of overwriting the global listings and panels...

tboerger:
Ah, I forgot... For the current state :-1:

crowbar_framework/app/assets/javascripts/barclamps/crowbar/application.js
rsalevsky: The old generic plugin did a complete page reload, this is not what we want for this simple status change.

tboerger: Than update the damn unobtrusive plugin instead of writing spaghetti code...

PR #190, opened 2016-01-18
network: Don't specify which flavor of openvswitch kmp to install

vuntz:
Let zypper install the right one automatically.

### DIFF
--- a/chef/cookbooks/network/attributes/default.rb
+++ b/chef/cookbooks/network/attributes/default.rb
@@ -19,7 +19,7 @@ when "suse"
                                    "vlan"]
   default[:network][:ovs_pkgs] = ["openvswitch",
                                   "openvswitch-switch",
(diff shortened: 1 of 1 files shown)


aplanas:
+1

tboerger:
:+1:

rhafer:
@vuntz Does it really work with Xen? I was wondering if the right kmp actually is installed before the node is rebooting. Or does installing the xen kernel also pull in the ovs kmp?

vuntz:
@rhafer I triggered a new gating job with xen, so let's see what happens. You're right that the xen kmp will only be installed on the reboot, but this might still be okay (hopefully ovs fails to start, but chef-client will start it).

vuntz:
Obsoleted by https://github.com/crowbar/crowbar-core/pull/517

PR #191, opened 2016-01-18
Fixups for Crowbar Backup/Restore

MaximilianMeister:
https://trello.com/c/9tQEIY2K/573-5-p1-5-finishing-crowbar-backup-restore-upgrade-pr

### DIFF
--- a/crowbar_framework/app/models/backup.rb
+++ b/crowbar_framework/app/models/backup.rb
@@ -36,7 +36,10 @@ class Backup < ActiveRecord::Base
   validates :size,
     presence: true
 
-  validate :validate_file_extension, :validate_version, :validate_hostname
+  validate :validate_chef_file_extension,
+    :validate_version,
+    :validate_hostname,
+    :validate_upload_file_extension
 
   def path
     self.class.image_dir.join(filename)
@@ -135,7 +138,7 @@ class Backup < ActiveRecord::Base
     path.delete if path.exist?
   end
 
-  def validate_file_extension
+  def validate_chef_file_extension
     Dir.glob(data.join("knife", "**", "*")).each do |file|
       next if Pathname.new(file).directory?
       next if File.extname(file) == ".json"
@@ -158,5 +161,11 @@ class Backup < ActiveRecord::Base
     unless system_hostname == backup_hostname
       errors.add(:base, I18n.t("backups.validation.hostnames_not_identical"))
     end
+  end
+
+  def validate_upload_file_extension
+    return if !file || (file && file.original_filename.match(/tar\.gz$/))
+
+    errors.add(:base, I18n.t("backups.validation.invalid_file_extension"))
   end
 end
--- a/crowbar_framework/app/models/backup.rb
+++ b/crowbar_framework/app/models/backup.rb
@@ -114,7 +114,7 @@ class Backup < ActiveRecord::Base
   end
 
   def save_archive
-    self.name = file.original_filename.split(".").first
+    self.name = file.original_filename.remove(".tar.gz")
 
     if path.exist?
       errors.add(:filename, I18n.t(".invalid_filename", scope: "backups.index"))
--- a/crowbar_framework/app/models/backup.rb
+++ b/crowbar_framework/app/models/backup.rb
@@ -39,7 +39,7 @@ class Backup < ActiveRecord::Base
   validate :validate_file_extension, :validate_version, :validate_hostname
 
   def path
-    self.class.image_dir.join("#{name}.tar.gz")
+    self.class.image_dir.join(filename)
   end
 
   def filename
@@ -117,7 +117,7 @@ class Backup < ActiveRecord::Base
     self.name = file.original_filename.remove(".tar.gz")
 
     if path.exist?
-      errors.add(:filename, I18n.t(".invalid_filename", scope: "backups.index"))
+      errors.add(:filename, I18n.t(".invalid_filename_exists", scope: "backups.index"))
       return false
     end
 
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -43,9 +43,11 @@ module Crowbar
               file = Pathname.new(file)
               next unless file.extname == ".json"
               record = JSON.load(file.read)
-              if file.basename.to_s.include?("-default.json") && type == :databags
-                bc_name = file.basename.to_s.split("-").first
-                Proposal.create(barclamp: bc_name, name: "default", properties: record)
+              filename = file.basename.to_s
+              if proposal?(filename) && type == :databags
+                bc_name, prop = filename.split("-")
+                prop.gsub!(/.json$/, "")
+                Proposal.create(barclamp: bc_name, name: prop, properties: record.raw_data)
                 SchemaMigration.run_for_bc(bc_name)
               else
                 record.save
@@ -97,6 +99,10 @@ module Crowbar
         Crowbar::Migrate.migrate!
 
         true
+      end
+
+      def proposal?(filename)
+        !filename.match(/(_network\.json$)|(^template-(.*).json$)/)
       end
     end
   end
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -42,7 +42,7 @@ module Crowbar
             file = Pathname.new(file)
             next unless file.extname == ".json"
             record = JSON.load(file.read)
-            if file.basename.to_s.include?("-default.json")
+            if file.basename.to_s.include?("-default.json") && type == :databags
               bc_name = file.basename.to_s.split("-").first
               Proposal.create(barclamp: bc_name, name: "default", properties: json)
               SchemaMigration.run_for_bc(bc_name)
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -26,7 +26,7 @@ module Crowbar
       end
 
       def restore
-        [:crowbar, :knife, :database].each do |component|
+        [:restore_crowbar, :run_installer, :restore_chef, :restore_database].each do |component|
           ret = send(component)
           return ret unless ret == true
         end
@@ -36,7 +36,7 @@ module Crowbar
 
       protected
 
-      def knife
+      def restore_chef
         [:nodes, :roles, :clients, :databags].each do |type|
           Dir.glob(@data.join("knife", type.to_s, "**", "*")).each do |file|
             file = Pathname.new(file)
@@ -55,7 +55,7 @@ module Crowbar
         true
       end
 
-      def crowbar
+      def restore_crowbar
         Crowbar::Backup::Base.restore_files.each do |source, destination|
           # keep the permissions of the files that are already in place
           system(
@@ -66,6 +66,10 @@ module Crowbar
           )
         end
 
+        true
+      end
+
+      def run_installer
         Crowbar::Installer.install!
         sleep(1) until Crowbar::Installer.successful? || Crowbar::Installer.failed?
         return false if Crowbar::Installer.failed?
@@ -80,7 +84,7 @@ module Crowbar
         true
       end
 
-      def database
+      def restore_database
         SerializationHelper::Base.new(YamlDb::Helper).load(
           @data.join("crowbar", "production.yml")
         )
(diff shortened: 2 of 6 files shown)


crowbar_framework/app/models/backup.rb
```diff
@@ -114,10 +117,10 @@ def create_archive
   end
 
   def save_archive
-    self.name = file.original_filename.split(".").first
+    self.name = file.original_filename.remove(".tar.gz")
```
vuntz: Thanks.

This makes me wonder: should we have a validation when uploading a tarball that the extension is `.tar.gz`?

crowbar_framework/app/models/backup.rb
```diff
@@ -23,6 +23,9 @@ class Backup < ActiveRecord::Base
   before_validation :save_or_create_archive, on: :create
   after_validation :delete_archive, unless: -> { errors.empty? }
   after_destroy :delete_archive
+  after_create do
```
vuntz: I think `after_create` is not good enough as we call `data` possibly at other times (not just when creating) -- for instance, when restoring.

vuntz:
Thanks a lot for the cleanups!

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -36,13 +36,13 @@ def restore
...
         [:nodes, :roles, :clients, :databags].each do |type|
           Dir.glob(@data.join("knife", type.to_s, "**", "*")).each do |file|
             file = Pathname.new(file)
             next unless file.extname == ".json"
             record = JSON.load(file.read)
-            if file.basename.to_s.include?("-default.json")
+            if file.basename.to_s.include?("-default.json") && type == :databags
               bc_name = file.basename.to_s.split("-").first
               Proposal.create(barclamp: bc_name, name: "default", properties: json)
               SchemaMigration.run_for_bc(bc_name)
```
vuntz: Replying to https://github.com/crowbar/crowbar-core/pull/144#discussion_r50018136 here (easier to get the discussion in the new PR):

if you look two lines below here, we do a `record.save`. That's a chef object that is being saved, and that will fail if chef is down. We try to deal with such exceptions elsewhere in the code (`Net::HTTPServerException` or `Errno::ECONNREFUSED`, should not be too hard to find examples).

We should just catch the exception and raise some nice "Restore failed" exception somehow.

crowbar_framework/app/models/backup.rb
```diff
@@ -23,6 +23,9 @@ class Backup < ActiveRecord::Base
...
+    data.rmtree
```
rsalevsky: This is not needed, Rails will delete the tmpdir when the instance is collected by the garbage collector.

MaximilianMeister: when I checked, my `/tmp` folder has not been cleaned up. when exactly does this garbage collection happen?

crowbar_framework/lib/crowbar/backup/restore.rb
MaximilianMeister: thanks for the hint

crowbar_framework/app/models/backup.rb
MaximilianMeister: we have that in the frontend at least, this is taken care of by the fileinput.js

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -36,13 +36,13 @@ def restore
...
+      def restore_chef
         [:nodes, :roles, :clients, :databags].each do |type|
           Dir.glob(@data.join("knife", type.to_s, "**", "*")).each do |file|
             file = Pathname.new(file)
             next unless file.extname == ".json"
             record = JSON.load(file.read)
-            if file.basename.to_s.include?("-default.json")
+            if file.basename.to_s.include?("-default.json") && type == :databags
               bc_name = file.basename.to_s.split("-").first
               Proposal.create(barclamp: bc_name, name: "default", properties: json)
```
rhafer: This needs to be ```properties: record```, I guess?

MaximilianMeister: right, how could this ever have worked? :-/

crowbar_framework/app/models/backup.rb
MaximilianMeister: I am a bit lost here. I don't have any idea how to achieve this.
* add the `data.rmtree` to more callbacks (wouldn't help)
* use `ObjectSpace.define_finalizer` (discouraging) see e.g. [here](https://www.ruby-forum.com/topic/17155)

how can we fix this?

MaximilianMeister:
@vuntz the owner/permission issue I would like to handle in another PR, would that be fine with you? i removed the WIP for now on this one

vuntz:
> @vuntz the owner/permission issue I would like to handle in another PR, would that be fine with you?

Sure, I'm all for non-monster PRs :-)

MaximilianMeister:
rebased. maybe that caused the gating in tempest to fail...

crowbar_framework/app/models/backup.rb
rsalevsky: We should also validate the extension in the model. Just adapt this https://github.com/MaximilianMeister/crowbar-core/blob/fixup/crowbar-backup/crowbar_framework/app/models/backup.rb#L34 regex.

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -26,7 +26,7 @@ def initialize(backup)
       end
 
       def restore
-        [:crowbar, :knife, :database].each do |component|
+        [:restore_crowbar, :run_installer, :restore_chef, :restore_database].each do |component|
```
rsalevsky: We are already in the restore lib so drop the restore prefix.

MaximilianMeister: I have taken @vuntz suggestion [here](https://github.com/crowbar/crowbar-core/pull/144/files#r49711356)
to me personally it does not really matter

rsalevsky: @tboerger suggested to remove it

vuntz: Should we rename this method to have no name, then? ;-) Note that I don't care too much -- I prefer the new naming, and don't understand what's good in the old one but will live with anything.

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -36,26 +36,32 @@ def restore
...
-            else
-              record.save
+      def restore_chef
+        begin
+          [:nodes, :roles, :clients, :databags].each do |type|
+            Dir.glob(@data.join("knife", type.to_s, "**", "*")).each do |file|
+              file = Pathname.new(file)
+              next unless file.extname == ".json"
+              record = JSON.load(file.read)
+              if file.basename.to_s.include?("-default.json") && type == :databags
```
rhafer: There's a problem with this. Not all proposal end in ```-default.json```. All proposals that can have multiple instances have the instance name here, instead of ```default```. This is especially true for the pacemaker proposals (but there are a few others as well)

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -36,26 +36,32 @@ def restore
...
+                bc_name = file.basename.to_s.split("-").first
```
rhafer: Hm, for some reason databags restored from Tex still contain that old ```bc-``` prefix. Isn't that supposed to migrated somewhere? (At least it doesn't seem to work)

MaximilianMeister: this was the PR where the renaming took place https://github.com/crowbar/crowbar-core/pull/46

Not sure about the migration, @vuntz do we have such a thing or do we have to implement it yet for the upgrade case?

crowbar_framework/lib/crowbar/backup/restore.rb
MaximilianMeister: right, so the proposals will have a name that the `[user has defined].json`? then we maybe have to figure out the right regex. I will have a look

crowbar_framework/lib/crowbar/backup/restore.rb
rsalevsky: The upgrade should remove this prefix: https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/upgrade.rb#L56-L62

crowbar_framework/lib/crowbar/backup/restore.rb
vuntz: I somehow feel that you can just restore all data bags as proposals when upgrading from Cloud 5. Or is there any data bag that you want to filter out?

rsalevsky: Will this work when you restore from Cloud 6 as well?

vuntz: No, but this code path (moving a proposal from data bag to proposal to database) is specific to the cloud 5 upgrade...

rhafer: Actually there are the network databags (that store the ip allocations) which are not really proposals.

vuntz: Good catch. We can filter them out with a match on `/_network$/`.

crowbar_framework/lib/crowbar/backup/restore.rb
rhafer: Right, it's just not executed. It seems ```supported?``` is returning false (https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/upgrade.rb#L32)

rsalevsky: Oh good catch, @MaximilianMeister can you implement that you abort the restore in case that the version is not the same as the running crowbar version and an upgrade is not possible?

rhafer: Actually the bug seems that ```version``` is a float, while ```supported``` expects it to be a string

crowbar_framework/app/models/backup.rb
MaximilianMeister: isn't it `file.original_filename` rather than `name`? I guess we need a custom validation for the file extension

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -36,26 +36,32 @@ def restore
...
+                Proposal.create(barclamp: bc_name, name: "default", properties: record)
```
rhafer: This is still not right. I think we need ```record.raw_data``` here. Otherwise the json will include that "json_class" and "chef_type" stuff, which ```Proposal``` doesn't like.

crowbar_framework/lib/crowbar/backup/restore.rb
MaximilianMeister: > We can filter them out with a match on /_network$/.

so we filter all `_network` and all `bc-template*` ? then the rest is the proposals that we want? would that be fine?

rhafer: Yeah, I think that shoudl work.

crowbar_framework/lib/crowbar/backup/restore.rb
rhafer: I can confirm that ```record.raw_data``` works here.

crowbar_framework/app/models/backup.rb
MaximilianMeister: hmm this one is still nagging... @vuntz shall I drop it until we have a solution (we should at least track it in a card)? it shouldn't block the other fixes IMO

rhafer:
I am fine with merging this in the current state and fixing the remaining issues in subsequent pull request as this brings as a lot closer to a working restore of tex 5 backups. 

+1

crowbar_framework/app/models/backup.rb
```diff
@@ -159,4 +165,10 @@ def validate_hostname
       errors.add(:base, I18n.t("backups.validation.hostnames_not_identical"))
     end
   end
+
+  def validate_upload_file_extension
+    return if file && file.original_filename.match(/tar\.gz$/)
```
MaximilianMeister: that's not correct yet

MaximilianMeister:
I have dropped the `after_create` where we remove the tmpdir until we have a better solution, and I fixed the `tar.gz` backend validation which was broken. would be good if we could merge it to proceed with the remaining issues

vuntz:
+1

> I have dropped the after_create where we remove the tmpdir until we have a better solution

Should we uncheck the item about that in the card?

MaximilianMeister:
> Should we uncheck the item about that in the card?

done, thanks for the hint. can someone merge this? i am not allowed...

rhafer:
still +1

PR #192, opened 2016-01-19
workaround random mksquashfs race bsc#962397

bmwiedemann:
https://bugzilla.suse.com/show_bug.cgi?id=962397
The instsys we use with autoyast
downloads some RPMs from HTTP and converts these to squashfs
using mksquashfs. Since that uses multiple cores,
it seems that there sometimes is a race
that produces corrupt filesystems and thus breaks the install.

we only started using more than 1 core recently in
https://github.com/SUSE-Cloud/automation/commit/0ff56c95bbc932f2b89a

### DIFF
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -259,6 +259,7 @@ if not nodes.nil? and not nodes.empty?
         when os =~ /^(open)?suse/
           append << "install=#{install_url} autoyast=#{node_url}/autoyast.xml"
           append << "ifcfg=dhcp4 netwait=60"
+          append << "squash=0" # workaround bsc#962397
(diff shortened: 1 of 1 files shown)


toabctl:
+1

aplanas:
I would put a comment indicating that `squash=0` parameter comes from this bug, so we can follow and revert it once is fixed

bmwiedemann:
@aplanas I would not because
a) we have git blame to find the comprehensive commit message and
b) the bug cannot be fixed in SLES-12-SP1 or older, because those are already released.

aplanas:
@bmwiedemann I understand the second point, but IMHO a git entry can't replace a comment in the code explaining the motivation of things that are not evident for the reader ...

bmwiedemann:
added bug comment

toabctl:
still +1

aplanas:
thanks!!!

+1

vuntz:
+1

PR #194, opened 2016-01-20
Cleanup stopping openstack and pacemaker

jsuchome:
Followup for https://github.com/crowbar/crowbar-core/pull/179

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
@@ -63,17 +63,23 @@ if node["crowbar_wall"]["crowbar_openstack_shutdown"]
 
 elsif node["crowbar_wall"]["crowbar_db_shutdown"]
 
-  # Now, stop the database.
-  # In case of HA setup, also stop remaining pacemaker resources and HA services.
+  # Stop remaining pacemaker resources
+  bash "stop pacemaker resources" do
+    code <<-EOF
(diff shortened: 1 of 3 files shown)


chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
```diff
@@ -63,17 +63,23 @@
 
 elsif node["crowbar_wall"]["crowbar_db_shutdown"]
 
-  # Now, stop the database.
-  # In case of HA setup, also stop remaining pacemaker resources and HA services.
+  # Stop remaining pacemaker resources
+  bash "stop pacemaker resources" do
+    code <<-EOF
+      for type in clone ms primitive; do
```
jsuchome: Note that I'm not listing `group` here; this is the same as in 4-5 upgrade script where we hadn't group as well.

jsuchome:
@vuntz this should address your comments from previous PR

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
vuntz: Ah, interesting, I thought we were. I guess it works because we stop one item in the group.

vuntz:
Thanks for the cleanups, +1.

jsuchome:
@vuntz are you sure? I didn't address your `g-postgresql` comment...

rhafer:
+1

PR #196, opened 2016-01-20
Skip binary files from backup (bsc#961464)

dirkmueller:
/var/chef/backup is collected by supportutils and we don't want
to collect megabytes of binary garbage in there.


vuntz:
+1

toabctl:
+1

toabctl:
gating fails: ```Cannot find a resource for f on suse version 12.1```

toabctl:
@dirkmueller needs some rework.

PR #197, opened 2016-01-20
Pass the specific error message from chef recipe to crowbar rails app

jsuchome:
This is currently only used by upgrade process, but potentially could offer solution for other cases when we want to report an error from chef recipe.

(Note that this PR contains 2 different solutions for resetting the error flag of node.)

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
@@ -49,8 +49,12 @@ when "openstack_shutdown"
       available = `df #{path} | tail -n 1 | sed "s/ \\+/ /g" | cut -d " " -f 4`
       db_size = `su - postgres -c 'pg_dumpall | wc -c'`
       if db_size.to_i > available.to_i
-        Chef::Log.fatal("DB size is: #{db_size}, available space on #{path} is #{available}")
-        raise "Not enough space on #{path} for the Database dump!"
+        message = "Not enough space for the Database dump on node #{node.name}!\n" \
+          "Database size is: #{db_size}B, available space on #{path} is only #{available}B."
+        Chef::Log.fatal(message)
+        node["crowbar_wall"]["chef_error"] = message
+        node.save
+        raise message
       end
     end
   end
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -358,6 +358,10 @@ class CrowbarService < ServiceObject
   def apply_role (role, inst, in_queue)
     @logger.debug("Crowbar apply_role: enter")
     answer = super
+    if answer.first != 200
+      @logger.error("Crowbar apply_role: super apply_role finished with error")
+      return answer
+    end
     @logger.debug("Crowbar apply_role: super apply_role finished")
 
     role = role.default_attributes
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -45,9 +45,25 @@ class CrowbarService < ServiceObject
   # Unfortunatelly we need to explicitely look at crowbar-status of the proposal
   # because apply_role from this model ignores errors from superclass's apply_role.
   def commit_and_check_proposal
-    proposal_commit("default", false, false)
+    answer = proposal_commit("default", false, false)
+    # check if error message is saved in one of the nodes
+    if answer.first != 200
+      found_errors = []
+      NodeObject.find("state:crowbar_upgrade").each do |node|
+        error = node["crowbar_wall"]["chef_error"] || ""
+        next if error.empty?
+        found_errors.push error
+        @logger.error("Chef run ended with an error on the node #{name}: #{error}")
+        node["crowbar_wall"]["chef_error"] = ""
+        node.save
+      end
+      unless found_errors.empty?
+        raise found_errors.join("\n")
+      end
+    end
 
     proposal = Proposal.where(barclamp: "crowbar", name: "default").first
+    # there could be different error than one raised from a recipe
     if proposal["deployment"]["crowbar"]["crowbar-status"] == "failed"
       raise proposal["deployment"]["crowbar"]["crowbar-failed"]
     end


crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -309,7 +322,10 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
   def apply_role (role, inst, in_queue)
     @logger.debug("Crowbar apply_role: enter")
     answer = super
-    @logger.debug("Crowbar apply_role: super apply_role finished")
+    if answer.first != 200
+      @logger.error("Crowbar apply_role: super apply_role finished with error")
+      return answer
+    end
```
vuntz: Just wondering (and you probably know better since you looked at such logs recently): do we need to keep the debug log if we got a `200`?

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -40,8 +40,21 @@ def role_constraints
...
+    answer = proposal_commit("default", false, false)
+
+    # check if error message is saved in one of the nodes
+    if answer.first != 200
+      NodeObject.find("state:crowbar_upgrade").each do |node|
+        error = node["crowbar_wall"]["chef-error"] || ""
+        next if error.empty?
+        @logger.error("Chef run ended with an error on the node #{name}: #{error}")
+        node["crowbar_wall"]["chef-error"] = ""
+        node.save
```
vuntz: I'm trying to think when is the best way to reset the error. This of course works right now, but if we want to have a generic solution, we need to find the right location for this. Maybe we should reset at the beginning of apply_role instead?

crowbar_framework/app/models/crowbar_service.rb
jsuchome: Oh, I was planning to leave it there

crowbar_framework/app/models/crowbar_service.rb
vuntz: Or maybe we should reset it at the end of a successful chef-client run?

jsuchome: Yes, i was expecting discussion about this. Though this should work, resetting before the call seems more right, yes.

jsuchome: Hm, also, if the error happens to be on more nodes, this current way would reset it only on the first one. So it's obviously wrong.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1125,6 +1125,12 @@ def apply_role(role, inst, in_queue)
 
       save_it = false
 
+      # reset possible previous error flags at the node
```
jsuchome: So, this is the place where we iterate over nodes during apply_role, so I thought it might be a good place to reset the error flag.

However, what if the apply_role after the one which failed is working with different set of nodes? That we would have some nodes where the error message is still saved, right? After all, it might look that reseting the flag _after_ the apply role (like in previous commits, still present in this PR) might be better idea...

vuntz:
Hrm, I just stumbled upon that: it seems we have the `crowbar-failed` attribute that is used for something similar. But I'm not 100% it's used for nodes, for roles or for proposals... Can you maybe quickly check that? (or I'll do it when I'll come back to this PR if you're busy -- no problem)

jsuchome:
@vuntz yes, I'm already using it as a backup: https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/crowbar_service.rb#L45

It's related to proposal, not node (which isn't necessary a problem0.

The main problem I see with it is that it includes that whole page full of errors (like the line in chef recipe code which caused the failure etc.) Which is of course useful, but not as a simple hint to user that something is wrong and for which we could offer specific advice for solution.

jsuchome:
@rhafer do you have any opinion about this one?

rhafer:
Right "crowbar-failed" is AFAIK meant for Propsoal specific error. We use it also as a crutch to pass the full error log from chef-client when that fails.

I think it's a good idea to also have the possibility to set very specific error messages on nodes.

Regarding clearing that flag, couldn't we just try to do that a the very end of the successful chef-client run? Similar to how we clear the maintenance mode flag for HA clusters? (Chef::Handler)

jsuchome:
I do not see any perfect solution on the horizon. Maybe we could go with the original one from first commit here (https://github.com/jsuchome/crowbar-core/commit/c11d66664d9897ea6f1a75c111743f890334368d - reset the flag right after spotting it) and think about something better for the future.

Note that conflicts with different chef-clirnt runs are not happening here, as chef-clients are disabled on the nodes with crowbar_upgrade state.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -40,8 +40,24 @@ def role_constraints
...
+        next if error.empty?
+        found_errors.push error
+        @logger.error("Chef run ended with an error on the node #{name}: #{error}")
+        node["crowbar_wall"]["chef-error"] = ""
+        node.save
+      end
+    end
+    unless found_errors.empty?
+      raise found_errors.join("\n") # this will not work for flash messages...
+    end
```
vuntz: Does it make sense to move all the use of  `found_errors` inside the `if` block?

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
```diff
@@ -49,8 +49,12 @@
       available = `df #{path} | tail -n 1 | sed "s/ \\+/ /g" | cut -d " " -f 4`
       db_size = `su - postgres -c 'pg_dumpall | wc -c'`
       if db_size.to_i > available.to_i
-        Chef::Log.fatal("DB size is: #{db_size}, available space on #{path} is #{available}")
-        raise "Not enough space on #{path} for the Database dump!"
+        message = "Not enough space for the Database dump on node #{node.name}!
+Database size is: #{db_size}B, available space on #{path} is only #{available}B."
+        Chef::Log.fatal(message)
+        node["crowbar_wall"]["chef-error"] = message
```
vuntz: The attribute should probably be `chef_error` and not `chef-error` (as we try to avoid dashes in attribute names).

crowbar_framework/app/models/service_object.rb
vuntz: So I'd remove this for now since I guess it needs some more thinking. Since you reset the attribute in CrowbarService already, we should be safe for now.

vuntz:
(btw, this needs a rebase)

I agree we should keep it simple and only use the error attribute for your need right now.

Would you mind creating a followup card in trello to design how to propagate error messages better?

crowbar_framework/app/models/crowbar_service.rb
jsuchome: Yes, makes sense

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
```diff
@@ -49,8 +49,12 @@
       available = `df #{path} | tail -n 1 | sed "s/ \\+/ /g" | cut -d " " -f 4`
       db_size = `su - postgres -c 'pg_dumpall | wc -c'`
       if db_size.to_i > available.to_i
-        Chef::Log.fatal("DB size is: #{db_size}, available space on #{path} is #{available}")
-        raise "Not enough space on #{path} for the Database dump!"
+        message = "Not enough space for the Database dump on node #{node.name}!
+Database size is: #{db_size}B, available space on #{path} is only #{available}B."
```
rhafer: Hm, that's not exactly nice ;)

rhafer:
If possible fix that message string format somehow, otherwise I think this is ok.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -45,9 +45,25 @@ def role_constraints
...
+      NodeObject.find("state:crowbar_upgrade").each do |node|
+        error = node["crowbar_wall"]["chef_error"] || ""
+        next if error.empty?
+        found_errors.push error
+        @logger.error("Chef run ended with an error on the node #{name}: #{error}")
+        node["crowbar_wall"]["chef_error"] = ""
+        node.save
+      end
+      unless found_errors.empty?
+        raise found_errors.join("<br/>").html_safe
```
vuntz: I don't think you should use html here as this would look nice when the error appears in the json-based API.

There must be a way to have the `\n` bits translated to `<br/>` later on, but not sure how...

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
jsuchome: What do you mean exactly?

aspiers: Split the long string via concatenation.

vuntz:
+1

rhafer:
+1

PR #198, opened 2016-01-20
Fix db:cleanup task to work with empty databases

rsalevsky:


tboerger:
:+1: Of course it works ;)

MaximilianMeister:
> Of course it works ;)

It works? then +1

PR #199, opened 2016-01-21
preserve ownership/permissions on backup and restore

MaximilianMeister:
cc @vuntz

### DIFF
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -47,7 +47,14 @@ module Crowbar
               if proposal?(filename) && type == :databags
                 bc_name, prop = filename.split("-")
                 prop.gsub!(/.json$/, "")
-                Proposal.create(barclamp: bc_name, name: prop, properties: record.raw_data)
+                proposal = Proposal.where(
+                  barclamp: bc_name
+                ).first_or_initialize(
+                  barclamp: bc_name,
+                  name: prop
+                )
+                proposal.properties = record.raw_data
+                proposal.save
                 SchemaMigration.run_for_bc(bc_name)
               else
                 record.save
--- a/crowbar_framework/lib/crowbar/upgrade.rb
+++ b/crowbar_framework/lib/crowbar/upgrade.rb
@@ -14,6 +14,8 @@
 # limitations under the License.
 #
 
+require "tempfile"
+
 module Crowbar
   class Upgrade
     attr_accessor :data
@@ -22,6 +24,10 @@ module Crowbar
       @backup = backup
       @data = @backup.data
       @version = @backup.version
+      @status = {
+        status: :ok,
+        errors: []
+      }
     end
 
     def upgrade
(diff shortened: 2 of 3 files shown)


vuntz:
`system("sudo", "chown", "-R", "crowbar:crowbar", dest)` in export.rb will need to be removed.

crowbar_framework/app/models/backup.rb
```diff
@@ -100,13 +100,21 @@ def create_archive
 
     Crowbar::Backup::Export.new(dir).export
     Dir.chdir(dir) do
-      ar = Archive::Compress.new(
+      cmd = [
+        "tar",
+        "--same-permissions",
+        "--same-owner",
```
vuntz: There's no option in libarchive for that?

vuntz:
Also, when we extract the backup, the files are created with the wrong ownership (maybe correct permissions, though).

crowbar_framework/app/models/backup.rb
MaximilianMeister: the problem I thought started with trying to pack files that don't belong to crowbar. I assumed libarchive does not have the permissions to pack them, that's why I used a system call (which in return makes me wonder why I haven't used sudo here...).
I guess I was wrong. It seems we can keep the code here as it was, I will have to test it

MaximilianMeister:
> system("sudo", "chown", "-R", "crowbar:crowbar", dest) in export.rb will need to be removed.

right, thanks I totally forgot that one

MaximilianMeister:
> Also, when we extract the backup, the files are created with the wrong ownership (maybe correct permissions, though).

I think the only place where the tar flags matter is the extract case. from the manpage
```
       -p, --preserve-permissions, --same-permissions
              extract information about file permissions (default for superuser)
       --same-owner
              try extracting files with the same ownership as exists in the archive (default for superuser)
```

so I kept the `create_archive` as it is (using libarchive), and just changed the `extract` method to use tar.
`rubygem-archive` does have a flag for permission but not for ownership it seems. https://github.com/erikh/archive/blob/master/lib/archive/extract.rb#L36

vuntz:
Cool. However, no you can't rely on garbage collection to remove the extracted directory, since you need sudo to remove it.

MaximilianMeister:
> Cool. However, no you can't rely on garbage collection to remove the extracted directory, since you need sudo to remove it.

that makes sense, but this is only the case for the `tmpdir` that was created by `extract`, so where would i hook that? the memoization confuses a bit, how can I know that I can clean the `tmpdir`?

rsalevsky:
I don't like this approach, I prefer when we collect the permission while https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/backup/export.rb#L125 in an yaml file and just use `sudo` to reset the permissions in https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/backup/restore.rb#L67. It's less code changes and the  code is also cleaner.

:-1:

vuntz:
@rsalevsky that's what I discussed with @MaximilianMeister some days ago: what you suggest is reinventing a format to backup (and restore) files with all metadata. Do we really want to do that ourselves and make sure we get everything right, while tar does it already and is known to work well?

MaximilianMeister:
> I don't like this approach, I prefer when we collect the permission while https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/backup/export.rb#L125 in an yaml file and just use sudo to reset the permissions in https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/backup/restore.rb#L67. It's less code changes and the code is also cleaner.

can you submit your idea? to me it feels like more work/code than this way. and you should know the model better now as you have refactored it

rsalevsky:
@vuntz @MaximilianMeister Ruby gives us access to the meta data so getting and setting this data should be no problem. I did it already in an other project and there was no problem. But I see the point that we should not take the risk so shortly while the release, I will give my +1 when we make sure that we take a look at this after the release.

vuntz:
+1

rhafer:
This needs a rebase, but you might want to wait with that until #227 is merged.

PR #200, opened 2016-01-21
Setup wizard with basic function calls to test the upgrade process

rsalevsky:


dirkmueller:
OK to merge with me (assuming gating passes)

crowbar_framework/app/controllers/setups_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+#
+
+class SetupsController < ApplicationController
+  skip_before_filter :enforce_installer
+  before_filter :hide_navigation
+
+  def show
+  end
+
+  def stop
```
jsuchome: I would add something like (may be part of followup PR):

```ruby
@service_object = CrowbarService.new logger

while true do
  @service_object.shutdown_openstack_at_nodes
  begin
    @service_object.dump_database
    break
  rescue => e
    flash[:alert] = e.message
  end
end

@service_object.finalize_openstack_shutdown
```

rhafer:
+1 and merging

PR #201, opened 2016-01-21
migrate horizon renaming and refactor knife upgrade

MaximilianMeister:
* rename files/content from nova_dashboard to horizon
* getting rid of DRY
* make more use of the Pathname Object

I hope I caught all the specific places/files where we need to rename/sed nova_dashboard to horizon...
I have just performed a `grep -R nova_dashboard *` and a `find . | grep nova_dashboard` inside the cloud5 backup.

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade.rb
+++ b/crowbar_framework/lib/crowbar/upgrade.rb
@@ -40,31 +40,49 @@ module Crowbar
 
     def knife_files
       @data.join("knife", "databags", "barclamps").rmtree
+      knife_path = @data.join("knife")
 
-      crowbar_databags_path = @data.join("knife", "databags", "crowbar")
+      crowbar_databags_path = knife_path.join("databags", "crowbar")
       crowbar_databags_path.children.each do |file|
-        if file == "bc-template-nova_dashboard.json"
-          crowbar_databags_path.join(file).rename(
-            crowbar_databags_path.join(file.sub!("nova_dashboard", "horizon"))
-          )
+        file_path = crowbar_databags_path.join(file)
 
-          file_content = File.read(crowbar_databags_path.join(file))
-          file_content.gsub!("nova_dashboard", "horizon")
-          File.open(crowbar_databags_path.join(file), "w") { |content| content.puts file_content }
+        if file.basename.to_s =~ /^bc-nova_dashboard-(.*)\.json$/
+          new_file = filename_replace(file_path, "nova_dashboard", "horizon")
+          filecontent_replace(new_file, "nova_dashboard", "horizon")
+          file_path = new_file
         end
 
-        next unless file.to_s.match("bc-.*.json")
-        new_file = file.sub("bc-", "")
-        crowbar_databags_path.join(file).rename(new_file)
+        next unless file_path.basename.to_s =~ /^bc-(.*).json$/
+        file_path = filename_replace(file_path, "bc-", "")
+        filecontent_replace(file_path, "bc-", "")
+      end
 
-        file_content = crowbar_databags_path.join(new_file).read
-        file_content.gsub!("bc-", "")
-        File.open(crowbar_databags_path.join(new_file), "w") { |content| content.puts file_content }
+      roles_path = knife_path.join("roles")
+      roles_path.children.each do |file|
+        case file.basename.to_s
+        when /^nova_dashboard-(.*).json$/
+          new_file = filename_replace(file, "nova_dashboard", "horizon")
+          filecontent_replace(new_file, "nova_dashboard", "horizon")
+        when /^crowbar-(.*).json$/
+          filecontent_replace(new_file, "nova_dashboard", "horizon")
+        end
       end
     end
 
     def crowbar_files
       FileUtils.touch(@data.join("crowbar", "production.yml"))
+    end
+
+    def filename_replace(file, search, replace)
+      new_file = file.sub(search, replace)
+      file.rename(new_file)
+      new_file
+    end
+
+    def filecontent_replace(file, search, replace)
+      file_content = file.read
+      file_content.gsub!(search, replace)
+      file.open("w") { |content| content.puts file_content }
     end
   end
 end


crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -40,31 +40,54 @@ def supported?
...
+      file = roles_path.join("nova_dashboard-server.json")
+      new_file = rename_file(file, "nova_dashboard", "horizon")
+      replace_file_content(new_file, "nova_dashboard", "horizon")
     end
 
     def crowbar_files
       FileUtils.touch(@data.join("crowbar", "production.yml"))
     end
+
+    def rename_file(file, search, replace)
```
rsalevsky: The name is not exactly what the method does, it's a rename with a regex. So this should be reflected in the name and then we can move this to our utils lib.

MaximilianMeister: sounds reasonable, but do we have any other case where we need that?

we could also name it very specifically `nova_dashboard_to_horizon` and get rid of all the arguments in the caller, they already look very DRY in this file.

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -40,31 +40,54 @@ def supported?
...
-        if file == "bc-template-nova_dashboard.json"
-          crowbar_databags_path.join(file).rename(
-            crowbar_databags_path.join(file.sub!("nova_dashboard", "horizon"))
-          )
+        file_path = crowbar_databags_path.join(file)
 
-          file_content = File.read(crowbar_databags_path.join(file))
-          file_content.gsub!("nova_dashboard", "horizon")
-          File.open(crowbar_databags_path.join(file), "w") { |content| content.puts file_content }
+        if file == "bc-template-nova_dashboard.json"
```
rhafer: We need to do this for the proposal. Not the template. I think the templates don't need to be restored at all.

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -40,31 +40,54 @@ def supported?
...
+          filecontent_replace(new_file, "nova_dashboard", "horizon")
+          file_path = new_file
         end
 
-        next unless file.to_s.match("bc-.*.json")
-        new_file = file.sub("bc-", "")
-        crowbar_databags_path.join(file).rename(new_file)
+        next unless file_path.to_s =~ "bc-.*.json"
+        file_path = filename_replace(file_path, "bc-", "")
+        filecontent_replace(file_path, "bc-", "")
```
rhafer: Hm, isn't this a bit risky? Some attribute values might actually really contain "bc-" which doesn't need to be replaced.

MaximilianMeister: yes that is a bit ambiguous, but I checked all files and there was no other `bc-` value in the files. but I can just add some more characters then it's a bit safer

crowbar_framework/lib/crowbar/upgrade.rb
MaximilianMeister: ah right, thanks

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -40,31 +40,54 @@ def supported?
...
+      end
 
-        file_content = crowbar_databags_path.join(new_file).read
-        file_content.gsub!("bc-", "")
-        File.open(crowbar_databags_path.join(new_file), "w") { |content| content.puts file_content }
+      barclamps_databag_path = knife_path.join("barclamps")
```
rhafer: That path (..../knife/barclamps) doesn't exist in the data dump AFAICS. I guess you meant (.../knife/databags/barclamps), bugt we're removing that directory on line 42. So something is wrong here. I don't think we need those files for the upgrade.

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -40,31 +40,54 @@ def supported?
...
+          filecontent_replace(new_file, "nova_dashboard", "horizon")
+        when "tempest.json" || "openstack.json"
+          filecontent_replace(new_file, "nova_dashboard", "horizon")
+        end
       end
+
+      roles_path = knife_path.join("roles")
+      file = roles_path.join("nova_dashboard-server.json")
+      new_file = filename_replace(file, "nova_dashboard", "horizon")
+      filecontent_replace(new_file, "nova_dashboard", "horizon")
```
rhafer: What about the proposal role? It's called nova_dashboard-config-default. I guess that needs to be renamed as well.

crowbar_framework/lib/crowbar/upgrade.rb
rhafer: Well. Someone might just set a name for a cinder backend that contains this. Or the password for the default keystone user. So yeah, I think match for at least "bc-<barclamp>" would be better here. (Though we need to verify to not miss something by doing so).

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -40,31 +40,60 @@ def supported?
...
+          filecontent_replace(new_file, "nova_dashboard", "horizon")
+          file_path = new_file
         end
 
-        next unless file.to_s.match("bc-.*.json")
-        new_file = file.sub("bc-", "")
-        crowbar_databags_path.join(file).rename(new_file)
+        next unless file_path.to_s =~ /^bc-(.*).json$/
+        file_path = filename_replace(file_path, "bc-", "")
+        filecontent_replace(file_path, "bc-", "")
```
MaximilianMeister: any idea how we can make this less ambiguous? maybe if it starts with a `"`? it's a bit tricky

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -40,31 +40,60 @@ def supported?
...
-        if file == "bc-template-nova_dashboard.json"
-          crowbar_databags_path.join(file).rename(
-            crowbar_databags_path.join(file.sub!("nova_dashboard", "horizon"))
-          )
+        file_path = crowbar_databags_path.join(file)
 
-          file_content = File.read(crowbar_databags_path.join(file))
-          file_content.gsub!("nova_dashboard", "horizon")
-          File.open(crowbar_databags_path.join(file), "w") { |content| content.puts file_content }
+        if file =~ /^bc-nova_dashboard-(.*)\.json$/
```
rhafer: ```file.to_s```

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -40,31 +40,60 @@ def supported?
...
+        end
       end
     end
 
     def crowbar_files
       FileUtils.touch(@data.join("crowbar", "production.yml"))
     end
+
+    def filename_replace(file, search, replace)
+      new_file = file.sub(search, replace)
```
rhafer: Not sure. But here as well? ```file.to_s```

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -40,31 +40,60 @@ def supported?
...
-        crowbar_databags_path.join(file).rename(new_file)
+        next unless file_path.to_s =~ /^bc-(.*).json$/
+        file_path = filename_replace(file_path, "bc-", "")
+        filecontent_replace(file_path, "bc-", "")
+      end
 
-        file_content = crowbar_databags_path.join(new_file).read
-        file_content.gsub!("bc-", "")
-        File.open(crowbar_databags_path.join(new_file), "w") { |content| content.puts file_content }
+      barclamps_databag_path = knife_path.join("databags", "barclamps")
```
rhafer: As said, this directory is removed in line 42.
Do we really need it for restore?

crowbar_framework/lib/crowbar/upgrade.rb
MaximilianMeister: no, actually here not. `sub` is an instance method of `Pathname`

crowbar_framework/lib/crowbar/upgrade.rb
MaximilianMeister: I don't know that. @rsalevsky why does this remove happen in 42?

rhafer: I am pretty sure we don't need those files.

MaximilianMeister: @rhafer okay then I will remove the whole block?

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -40,31 +40,49 @@ def supported?
...
-        if file == "bc-template-nova_dashboard.json"
-          crowbar_databags_path.join(file).rename(
-            crowbar_databags_path.join(file.sub!("nova_dashboard", "horizon"))
-          )
+        file_path = crowbar_databags_path.join(file)
 
-          file_content = File.read(crowbar_databags_path.join(file))
-          file_content.gsub!("nova_dashboard", "horizon")
-          File.open(crowbar_databags_path.join(file), "w") { |content| content.puts file_content }
+        if file.to_s =~ /^bc-nova_dashboard-(.*)\.json$/
```
rhafer: Hm, that's still not quite right. ```file.basename.to_s``` should do it.

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -40,31 +40,49 @@ def supported?
...
+          new_file = filename_replace(file_path, "nova_dashboard", "horizon")
+          filecontent_replace(new_file, "nova_dashboard", "horizon")
+          file_path = new_file
         end
 
-        next unless file.to_s.match("bc-.*.json")
-        new_file = file.sub("bc-", "")
-        crowbar_databags_path.join(file).rename(new_file)
+        next unless file_path.to_s =~ /^bc-(.*).json$/
```
rhafer: Same here.

crowbar_framework/lib/crowbar/upgrade.rb
MaximilianMeister: thanks. I broke it by using `^` in the regex

vuntz:
+1

rhafer:
> any idea how we can make this less ambiguous? maybe if it starts with a "? it's a bit tricky

How about matching for "bc-<basename-of-current-file>" would that missing anything?
I am fine to address this later, but please track it in Trello so we don't forget about it.

+1

MaximilianMeister:
> How about matching for "bc-" would that missing anything?

maybe I am misunderstanding, but `filecontent_replace(file_path, "bc-", "")` does exactly that

rhafer:
> maybe I am misunderstanding, but filecontent_replace(file_path, "bc-", "") does exactly that

Argh, github ate the most important part of my comment. I meant:

```
bc-<barclamp-name>
```

I guess barclamp-name can be derived from the current filename.

MaximilianMeister:
> I am fine to address this later, but please track it in Trello so we don't forget about it.

done

PR #202, opened 2016-01-21
crowbar: Fix node runlist sorting to be predictable

vuntz:
If two roles have the same priority, there was nothing that was
guaranteeing a fixed order. So sort by priority and then name.

This bug could result in issues for HA clusters where one member of the
cluster runs one role while others are running another role with the
same priority -- triggering timeouts on the sync marks.

cc @aspiers; this should fix https://ci.suse.de/job/openstack-mkcloud/15117/console


aplanas:
+1

jsuchome:
Seems to make sense. +1

aspiers:
Wow, how did we not see that before?  Good catch!

PR #203, opened 2016-01-21
Do not allow change to shutdown state from crowbar_upgrade

jsuchome:
(we need to reboot the nodes for upgrade, but without changing the state)

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -152,6 +152,13 @@ class CrowbarService < ServiceObject
       if node.crowbar["state"] != state
         @logger.debug("Crowbar transition: state has changed so we need to do stuff for #{name} to #{state}")
 
+        # Do not allow change to shutdown state from crowbar_upgrade
+        # (we need to reboot the nodes for upgrade, but without changing the state)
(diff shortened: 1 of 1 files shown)


jsuchome:
Do we need this in tex as well, in case user reboots the node manually?

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -152,6 +152,13 @@ def transition(inst, name, state)
       if node.crowbar["state"] != state
         @logger.debug("Crowbar transition: state has changed so we need to do stuff for #{name} to #{state}")
 
+        # Do not allow change to shutdown state from crowbar_upgrade
+        # (we need to reboot the nodes for upgrade, but without changing the state)
+        if node.crowbar["state"] == "crowbar_upgrade" && state == "shutdown"
+          @logger.debug("Crowbar transition: current state is crowbar_upgrade; change not allowed")
```
rhafer: Could you tweak the log message to
 'current node state is "crowbar_upgrade" ignoring change to "shutdown"'

rhafer:
Minor nitpick. Apart from that +1

jsuchome:
@vuntz , any comments?

nkrinner:
+1

PR #204, opened 2016-01-21
Add backend calls for 'shutdown of services' step

jsuchome:
So, this is obviously wrong from UI perspective, handling of error from `dump_database` is not correct ... (  but maybe that could be solved in the final UI?)

At least the right calls to API are there..

### DIFF
--- a/crowbar_framework/app/controllers/setups_controller.rb
+++ b/crowbar_framework/app/controllers/setups_controller.rb
@@ -22,7 +22,31 @@ class SetupsController < ApplicationController
   end
 
   def stop
-    # TODO: place function call
+    # Stop services at the nodes
+    @service_object = CrowbarService.new logger
+
+    @service_object.shutdown_services_at_non_db_nodes
+
+    # Dump of the database could fail with the message that there's not enough space
+    # We need to show the error to user so he can manually prepare the space on node
+    # and than redo this step.
+    begin
+      @service_object.dump_openstack_database
+    rescue => e
+      flash[:alert] = e.message
+      # FIXME: redirect to previous step?
+    end
+
+    # FIXME: now user needs to be told where is the database dump located in case
+    # he/she wants to manually fetch it
+    redirect_to setup_path
+  end
+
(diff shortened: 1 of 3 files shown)


crowbar_framework/app/controllers/setups_controller.rb
```diff
@@ -22,7 +22,31 @@ def show
...
+      flash[:alert] = e.message
+      # FIXME: redirect to previous step?
+    end
+
+    # FIXME: now user needs to be told where is the database dump located in case
+    # he/she wants to manually fetch it
+    redirect_to setup_path
+  end
+
+  def finalize_stop
```
rsalevsky: I would name it just `finalize` and the route in the routes.rb is missing. https://github.com/jsuchome/crowbar-core/blob/a444d3d6689496e188df103550f046b2b55e0a7b/crowbar_framework/config/routes.rb#L148

crowbar_framework/app/controllers/setups_controller.rb
```diff
@@ -22,7 +22,31 @@ def show
...
+    # Stop services at the nodes
+    @service_object = CrowbarService.new logger
+
+    @service_object.shutdown_openstack_at_nodes
+
+    # Dump of the database could fail with the message that there's not enough space
+    # We need to show the error to user so he can manually prepare the space on node
+    # and than redo this step.
+    begin
+      @service_object.dump_database
```
rhafer: Isn't the method called "dump_openstack_database" meanwhile?

jsuchome: It is...

rhafer:
> there's a new finalize_stop action which is not called from anywhere (I don't even know if it should be separate controller action)

ti should be called from button Step 6 I think.

rhafer:
Looks good to me. CI is still failing because of the initial patch in not in D:C:6:S yet.

rhafer:
+1

nkrinner:
+1

PR #206, opened 2016-01-21
Use link_to instead of button_to

rsalevsky:
This change has the advantage that the user can see the URL before
clicking on it.

### DIFF
--- a/crowbar_framework/app/helpers/backups_helper.rb
+++ b/crowbar_framework/app/helpers/backups_helper.rb
@@ -16,16 +16,15 @@
 
 module BackupsHelper
   def download_backup_button(backup)
-    button_to(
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/helpers/backups_helper.rb
```diff
@@ -16,7 +16,7 @@
 
 module BackupsHelper
   def download_backup_button(backup)
-    button_to(
+    link_to(
       t("download"),
       download_backup_path(backup.id),
       method: :get,
```
tboerger: Drop that :)

tboerger:
I prefer the links anyway, so :+1:

MaximilianMeister:
+1

PR #208, opened 2016-01-21
Use pxe config for correct mac address (admin interface)

rhafer:
node.macaddress can't be used for that as that might be changing. Depending on
which interface is used for the default route it might even point to a
ovs-bridge or bond device. This fix is required for the operating system
upgrade process.


vuntz:
+1

dguitarbite:
+1

PR #210, opened 2016-01-22
Added API calls for repositories checks to the UI

jsuchome:


jsuchome:
`repositories` should actually be called silently, and the page be shown only if the check fails. Hm.

rhafer:
+1

jsuchome:
I'm not sure if we still need this because the final wizard UI seems to be merged

rsalevsky:
I stared working to integrate all this backend stuff into the final wizard.

jsuchome:
@rsalevsky ah, thanks!

rsalevsky:
See #221

PR #211, opened 2016-01-22
Backup: Search by name and id

rsalevsky:
This makes it possible to download an image by the name. Specilly for
the crowbar-client this is very helpful.

### DIFF
--- a/crowbar_framework/app/models/backup.rb
+++ b/crowbar_framework/app/models/backup.rb
@@ -168,4 +168,20 @@ class Backup < ActiveRecord::Base
 
     errors.add(:base, I18n.t("backups.validation.invalid_file_extension"))
   end
+
+  class << self
+    def find_using_id_or_name(name)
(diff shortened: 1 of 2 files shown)


tboerger:
:+1: as this is a snippet i gave you from another project :D

crowbar_framework/app/models/backup.rb
```diff
@@ -168,4 +168,20 @@ def validate_upload_file_extension
 
     errors.add(:base, I18n.t("backups.validation.invalid_file_extension"))
   end
+
+  class << self
+    def find_using_name(name)
```
MaximilianMeister: so this method can be used for finding records via name **or** id?

MaximilianMeister: can we name it `find_using_name_or_id(name_or_id)`?

MaximilianMeister:
+1

PR #213, opened 2016-01-22
make sure to update the (core) proposals that are already created

MaximilianMeister:
cc @rhafer

### DIFF
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -47,7 +47,14 @@ module Crowbar
               if proposal?(filename) && type == :databags
                 bc_name, prop = filename.split("-")
                 prop.gsub!(/.json$/, "")
-                Proposal.create(barclamp: bc_name, name: prop, properties: record.raw_data)
+                proposal = Proposal.where(
+                  barclamp: bc_name
+                ).first_or_initialize(
+                  barclamp: bc_name,
+                  name: prop
+                )
+                proposal.properties = record.raw_data
+                proposal.save
                 SchemaMigration.run_for_bc(bc_name)
               else
                 record.save
--- a/crowbar_framework/lib/crowbar/upgrade.rb
+++ b/crowbar_framework/lib/crowbar/upgrade.rb
@@ -65,14 +65,13 @@ module Crowbar
           file_path = new_file
         when /^bc-network-(.*)\.json$/
           json = JSON.load(file.read)
-          # override with migrated attributes
-          attributes = SchemaMigration.migrated_attributes("network", json)
-          if attributes.nil?
+          attributes_deployment = SchemaMigration.migrate_proposal_from_json("network", json)
+          if attributes_deployment.nil?
             @status[:status] = :internal_server_error
             @status[:errors].push "Cannot upgrade the network proposal with #{file}"
             next
           end
-          json["attributes"]["network"] = attributes
+          json["attributes"]["network"] = attributes_deployment.first
           network_proposal = Tempfile.new("network")
           network_proposal.write JSON.pretty_generate(attributes: json["attributes"])
           network_proposal.close
--- a/crowbar_framework/lib/crowbar/upgrade.rb
+++ b/crowbar_framework/lib/crowbar/upgrade.rb
@@ -24,6 +24,10 @@ module Crowbar
       @backup = backup
       @data = @backup.data
       @version = @backup.version
+      @status = {
+        status: :ok,
+        errors: []
+      }
     end
 
     def upgrade
@@ -62,7 +66,13 @@ module Crowbar
         when /^bc-network-(.*)\.json$/
           json = JSON.load(file.read)
           # override with migrated attributes
-          json["attributes"]["network"] = SchemaMigration.migrated_attributes("network", json)
+          attributes = SchemaMigration.migrated_attributes("network", json)
+          if attributes.nil?
+            @status[:status] = :internal_server_error
+            @status[:errors].push "Cannot upgrade the network proposal with #{file}"
+            next
+          end
+          json["attributes"]["network"] = attributes
           network_proposal = Tempfile.new("network")
           network_proposal.write JSON.pretty_generate(attributes: json["attributes"])
           network_proposal.close
--- a/crowbar_framework/lib/crowbar/upgrade.rb
+++ b/crowbar_framework/lib/crowbar/upgrade.rb
@@ -14,6 +14,8 @@
 # limitations under the License.
 #
 
+require "tempfile"
+
 module Crowbar
   class Upgrade
     attr_accessor :data
@@ -57,6 +59,21 @@ module Crowbar
           new_file = filename_replace(file_path, "nova_dashboard", "horizon")
           filecontent_replace(new_file, "nova_dashboard", "horizon")
           file_path = new_file
+        when /^bc-network-(.*)\.json$/
+          json = JSON.load(file.read)
+          # override with migrated attributes
+          json["attributes"]["network"] = SchemaMigration.migrated_attributes("network", json)
+          network_proposal = Tempfile.new("network")
+          network_proposal.write JSON.pretty_generate(attributes: json["attributes"])
+          network_proposal.close
+          cmd = [
+            "sudo",
+            "cp",
+            network_proposal.path,
+            "/etc/crowbar/network.json"
+          ]
+          system(*cmd)
+          network_proposal.unlink
         end
 
         next unless file_path.basename.to_s =~ /^bc-(.*).json$/
--- a/crowbar_framework/lib/schema_migration.rb
+++ b/crowbar_framework/lib/schema_migration.rb
@@ -58,7 +58,7 @@ module SchemaMigration
     end
   end
 
-  def self.migrated_attributes(bc_name, json)
+  def self.migrate_proposal_from_json(bc_name, json)
     template = Proposal.new(barclamp: bc_name)
 
     return if template.nil?
@@ -71,8 +71,8 @@ module SchemaMigration
     attributes = json["attributes"][bc_name]
     deployment = json["deployment"][bc_name]
 
-    # return attributes
-    migrate_object(bc_name, template, all_scripts, attributes, deployment).first
+    # return migrated attributes and deployment
+    migrate_object(bc_name, template, all_scripts, attributes, deployment)
   end
 
   private
--- a/crowbar_framework/lib/schema_migration.rb
+++ b/crowbar_framework/lib/schema_migration.rb
@@ -58,6 +58,23 @@ module SchemaMigration
     end
   end
 
+  def self.migrated_attributes(bc_name, json)
+    template = Proposal.new(barclamp: bc_name)
+
+    return if template.nil?
+    return if template["deployment"].nil?
+    return if template["deployment"][bc_name].nil?
+
+    all_scripts = find_scripts_for_bc(bc_name)
+    return if all_scripts.empty?
+
+    attributes = json["attributes"][bc_name]
+    deployment = json["deployment"][bc_name]
+
+    # return attributes
+    migrate_object(bc_name, template, all_scripts, attributes, deployment).first
+  end
+
   private
 
   def self.data_bags_path


vuntz:
+1

rhafer:
Somehow I think it would be much better to restore the core proposals before actually running the installer, if that is somehow possible. Otherwise it will be every easy to break the upgrade in weird ways. Especially if the network barclamp settings  before running the upgrade don't match exactly what's in the backup (post install changes of that are not really working).

E.g. in (simplified) steps:
1. Restore core baclamps proposals to /etc/crowbar/ and run the migrations on the json files. I guess we just need to put the "attributes" section into /etc/crowbar/
2. Run installer
3. Restore non core barclamps and update the "deployment" sections of the core barclamps 

@vuntz I'd really appreciate your thought here as well.

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -47,7 +47,9 @@ def restore_chef
               if proposal?(filename) && type == :databags
                 bc_name, prop = filename.split("-")
                 prop.gsub!(/.json$/, "")
-                Proposal.create(barclamp: bc_name, name: prop, properties: record.raw_data)
+                proposal = Proposal.first_or_initialize(barclamp: bc_name, name: prop)
```
rhafer: Somehow this doesn't seem to work. While this now updates existing proposals successfully it somehow does not create new one if there wasn't one before. Didn't figure out why this happens yet.

vuntz:
@rhafer hrm, I think you're correct, especially for the network barclamp.

It however means changes to the install process because install-suse-cloud will try to enforce network.json, I think...

rhafer:
@vuntz Doesn't https://github.com/crowbar/crowbar/blob/master/scripts/install-chef-suse.sh#L821 mean that at least crowbar, dns, network, provisioner and ntp are also supported?

vuntz:
@rhafer yes, but from all of these, only network has settings which cannot be changed after the install. All the other ones should work fine with the current workflow (although I agree it's suboptimal).

crowbar_framework/lib/crowbar/backup/restore.rb
rhafer: It needs to be:

```Proposal.where(barclamp: bc_name).first_or_initialize(barclamp: bc_name, name: prop)```

Otherwise we will just overwrite the first Proposal (which ever that is) again and again.

Comment (MaximilianMeister) [inline crowbar_framework/lib/crowbar/backup/restore.rb]: thanks I changed it now to
```ruby
proposal = Proposal.where(barclamp: bc_name, name: prop).first_or_initialize
```

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -52,6 +54,22 @@ def knife_files
...
+        if file.basename.to_s =~ /^bc-network-(.*)\.json$/
+          json = JSON.load(file.read)
+          Tempfile.open("network") do |content|
+            content.write JSON.pretty_generate(json["attributes"])
+            cmd = [
+              "sudo",
+              "echo",
+              content.read,
+              ">",
+              "/etc/crowbar/network.json"
```
vuntz: Why not copy the file? Sounds a bit safer than redirecting the output of an `echo`.

MaximilianMeister: you're right. will do

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -47,7 +47,9 @@ def restore_chef
               if proposal?(filename) && type == :databags
                 bc_name, prop = filename.split("-")
                 prop.gsub!(/.json$/, "")
-                Proposal.create(barclamp: bc_name, name: prop, properties: record.raw_data)
+                proposal = Proposal.where(barclamp: bc_name, name: prop).first_or_initialize
```
rhafer: I think this is still wrong. See e.g.:

```https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/service_object.rb#L775``` for how it is called elsewhere.

Comment (MaximilianMeister) [inline crowbar_framework/lib/crowbar/backup/restore.rb]: right, sorry next time i will test it before. I thought Rails would be so smart to know what I want... will fix

File: crowbar_framework/lib/schema_migration.rb [hunk 4]
```diff
@@ -58,6 +58,28 @@ def self.run_for_bc bc_name
...
+    return if all_scripts.empty?
+
+    begin
+      validator = CrowbarValidator.new(data_bags_path.join("template-#{bc_name}.schema"))
+    rescue StandardError => e
+      raise "Failed to load databag schema for #{bc_name}: #{e.message}"
+    end
+
+    proposal = Proposal.where(barclamp: bc_name, name: prop).first
+    proposal.raw_data = json
```
MaximilianMeister: @rhafer can we just feed the databag here into raw_data?

rhafer: Doesn't just work if you pass ```json``` instead of ```proposal``` into ```migrate_proposal```

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -84,5 +104,11 @@ def filecontent_replace(file, search, replace)
       file_content.gsub!(search, replace)
       file.open("w") { |content| content.puts file_content }
     end
+
+    def validate_proposal(barclamp, proposal)
+      so = ServiceObject.new(Rails.logger)
+      so.bc_name = barclamp
```
jsuchome: Is this required? Shouldn't it be created by ServiceObject?

MaximilianMeister: default is "unknown"

jsuchome: Ah, OK

rhafer:
Sorry, but can we leave the validation part out for now and handle that in a separate PR? Sorry for even bringing it up here. It requires some more thoughts and we should really get the rest of this merged.

crowbar_framework/lib/schema_migration.rb
```diff
@@ -58,6 +58,23 @@ def self.run_for_bc bc_name
...
+    template = Proposal.new(barclamp: bc_name)
+
+    return if template.nil?
+    return if template["deployment"].nil?
+    return if template["deployment"][bc_name].nil?
+
+    all_scripts = find_scripts_for_bc(bc_name)
+    return if all_scripts.empty?
+
+    attributes = json["attributes"]
```
rhafer: This needs to be ```json["attributes"][bc_name]```

crowbar_framework/lib/schema_migration.rb
```diff
@@ -58,6 +58,23 @@ def self.run_for_bc bc_name
...
+    deployment = json["deployment"]
```
rhafer: Same here: json["deployment"][bc_name]

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -52,6 +54,24 @@ def knife_files
           file_path = new_file
         end
 
+        if file.basename.to_s =~ /^bc-network-(.*)\.json$/
+          json = JSON.load(file.read)
+          # override with migrated attributes
+          json["attributes"] = SchemaMigration.migrated_attributes("network", json)
```
rhafer: With the suggested changes in ```migrated_attributes``` you need to assign the result to json["attributes"]["network"].

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -52,6 +54,22 @@ def knife_files
...
+          Tempfile.open("network") do |content|
+            content.write JSON.pretty_generate(json["attributes"])
```
vuntz: I was confused for one second because `content` is actually not the content, but a file object. Would it make sense to rename that variable?

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -52,6 +54,24 @@ def knife_files
...
```
rhafer: We need the written json file to start with 
```
{
  "attributes":
    "network":
....
}
```

Currently it starts at ``` { "network": ... } ```

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -52,6 +54,24 @@ def knife_files
...
+            cmd = [
+              "sudo",
+              "cp",
+              content.path,
+              "/etc/crowbar/network.json"
+            ]
+            system(*cmd)
+          end
+          # finally validate network proposal
+          validate_proposal("network", json)
```
rhafer: As said, lets leave this out for now and worry about that later.

crowbar_framework/lib/crowbar/upgrade.rb
rhafer: You probably just need to reassigne the result of migrated_attributes to a new Hash.

```
attributes_migrated = {}
attributes_migrated["attributes"] = SchemaMigration.migrated_attributes("network", json)
```
 and then ```JSON.pretty_generate(attributes_migrated)```

I haven't tried that though.

MaximilianMeister:
> Sorry, but can we leave the validation part out for now and handle that in a separate PR? Sorry for even bringing it up here. It requires some more thoughts and we should really get the rest of this merged.

sure, i took out the commit.

I also addressed all the comments

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -52,6 +54,22 @@ def knife_files
...
+          json["attributes"]["network"] = SchemaMigration.migrated_attributes("network", json)
+          Tempfile.open("network") do |f|
+            f.write JSON.pretty_generate(attributes: json["attributes"])
+            cmd = [
+              "sudo",
+              "cp",
+              f.path,
+              "/etc/crowbar/network.json"
+            ]
+            system(*cmd)
```
rhafer: For some reason the copy in /etc/crowbar/ is of size 0 in my test setup. I gues you're missing the sync or close or something like that before actually copying the file here.

rhafer: Confirmed. If I put ```f.close``` before the cmd the file in not empty anymore. But I am sure there's a more correct way to address this.

vuntz:
+1

crowbar_framework/lib/schema_migration.rb
```diff
@@ -58,6 +58,23 @@ def self.run_for_bc bc_name
...
+    return if template["deployment"][bc_name].nil?
+
+    all_scripts = find_scripts_for_bc(bc_name)
+    return if all_scripts.empty?
+
+    attributes = json["attributes"][bc_name]
+    deployment = json["deployment"][bc_name]
+
+    # return attributes
+    migrate_object(bc_name, template, all_scripts, attributes, deployment).first
```
vuntz: Actually: `migrate_object` can return `nil`. Should we deal with this?

MaximilianMeister: probably we should, would this be fatal then? wondering what do do in that case

rhafer:
+1

MaximilianMeister:
@rhafer before merging let me propose another commit on top

crowbar_framework/lib/schema_migration.rb
MaximilianMeister: this is still the wrong place for using `.first` will fix

MaximilianMeister:
> Should this be localizable? (Also, I think it's better to put "Cannot" than "Could not" in error messages)

when the i18n namespaces for the controllers are done we can add this into the step that it belongs to I would say.

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -57,6 +63,27 @@ def knife_files
           new_file = filename_replace(file_path, "nova_dashboard", "horizon")
           filecontent_replace(new_file, "nova_dashboard", "horizon")
           file_path = new_file
+        when /^bc-network-(.*)\.json$/
+          json = JSON.load(file.read)
+          # override with migrated attributes
+          attributes = SchemaMigration.migrate_proposal_from_json("network", json)
```
rhafer: I guess you should rename ```attributes``` here. ```migrate_proposal_from_json``` does no longer return only the attributes.

MaximilianMeister: done and rebased

rhafer:
+1

rhafer:
@vuntz are you still good with this?

vuntz:
yes, still +1.

PR #215, opened 2016-01-22
Adds init-scripts section to autoyast-upgrade profile

dguitarbite:


chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
```diff
@@ -91,6 +91,25 @@
         </source>
       </script>
     </chroot-scripts>
+    <init-scripts config:type="list">
+      <!-- bugfix bnc#886238: https://bugzilla.novell.com/show_bug.cgi?id=886238 -->
+      <script>
+       <filename>autoyast_set_hostentries.sh</filename>
```
rhafer: Please remove this one. We don't need that for the upgrade. It would create duplicate entries in /etc/hosts  I think.

dguitarbite: You are right, removing them.

rhafer:
+1

mmnelemane:
+1

PR #216, opened 2016-01-22
Wrap parameters

rsalevsky:
Obsoletes https://github.com/crowbar/crowbar-client/pull/64

### DIFF
--- a/crowbar_framework/app/controllers/backups_controller.rb
+++ b/crowbar_framework/app/controllers/backups_controller.rb
@@ -15,6 +15,7 @@
 #
 
 class BackupsController < ApplicationController
+  wrap_parameters :backup, include: [:file]
   skip_before_filter :enforce_installer
   before_action :set_backup, only: [:destroy, :restore, :download]
 
--- a/crowbar_framework/config/initializers/wrap.rb
+++ b/crowbar_framework/config/initializers/wrap.rb
@@ -16,7 +16,7 @@
 #
 
 ActiveSupport.on_load :action_controller do
-  wrap_parameters format: [] if respond_to? :wrap_parameters
+  wrap_parameters format: [:json] if respond_to? :wrap_parameters
 end
 
 ActiveSupport.on_load :active_record do


vuntz:
Typo in commit message: wrappend

MaximilianMeister:
stupid question, why do we create a namespace `backup` and having to do this wrapping afterwards? doesn't say in the commit message.

rsalevsky:
@MaximilianMeister With this we make sure that we just trust allowed params and at the same time we make sure that we have all required data. `:file` is in this case very special because it's just temporary and it's not part of the database.

vuntz:
I don't understand why this is needed. Is this purely for convenience? Or is this to allow clients written in a lazy way?

rsalevsky:
This change is needed that we can send json requests with params without a namespace. For html requests it is recommended to namespace them but for json requests it is in general not wanted because it makes the api requests more complex. The rails wrapper just ensures that both ways works for json.

We need this PR to unblock backup & restore gating and fix backup commands in crowbarctl.

MaximilianMeister:
> This change is needed that we can send json requests with params without a namespace. For html requests it is recommended to namespace them but for json requests it is in general not wanted because it makes the api requests more complex. The rails wrapper just ensures that both ways works for json.

makes sense to me, can you please add this to the commit message?

and would be good if @tboerger could review this

jdsn:
> and would be good if @tboerger could review this

@MaximilianMeister it was @tboerger who proposed this patch

jdsn:
:+1:

rsalevsky:
Updated commit message.

vuntz:
> This change is needed that we can send json requests with params without a namespace. For html requests it is recommended to namespace them but for json requests it is in general not wanted because it makes the api requests more complex. The rails wrapper just ensures that both ways works for json.

I'm fine with the change, but why don't we do this for all controllers and only here?

rsalevsky:
We do it for all controllers, this is done by the first commit. The second commit is just that we also include `:file` into the wrapping because this is not done by default due to https://github.com/rsalevsky/crowbar-core/blob/wrap_parameters/crowbar_framework/app/models/backup.rb#L21.

vuntz:
Ah okay, thanks for the explanation.

So you get a LGTM from me, unfortunately, I'm not comfortable enough for a +1. (but feel free to change my LGTM to a +1 if you get a rails expert approving this)

tboerger:
Even if it's already merged, :+1: :)

PR #218, opened 2016-01-25
At some point we have to unset db_synced flags

jsuchome:
@rhafer this is another part of backend that needs to be called at some point


jsuchome:
@rsalevsky please look also at this one. We don't need this PR, but we need this function call in the final wizard as well

rsalevsky:
@jsuchome I opened #221. Can you comment there where I should add this call?

PR #219, opened 2016-01-25
remove unwanted attributes.rails as it cannot be restored

MaximilianMeister:
cc @rhafer

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade.rb
+++ b/crowbar_framework/lib/crowbar/upgrade.rb
@@ -46,6 +46,14 @@ module Crowbar
       crowbar_databags_path.children.each do |file|
         file_path = crowbar_databags_path.join(file)
 
+        if file.basename.to_s =~ /^bc-crowbar-(.*)\.json$/
+          json = JSON.load(file.read)
+          if json["attributes"]["rails"]
+            json["attributes"].delete("rails")
+            file.open("w") { |content| content.puts JSON.pretty_generate(json) }
+          end
+        end
+
         if file.basename.to_s =~ /^bc-nova_dashboard-(.*)\.json$/
           new_file = filename_replace(file_path, "nova_dashboard", "horizon")
           filecontent_replace(new_file, "nova_dashboard", "horizon")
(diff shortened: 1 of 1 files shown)


crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -46,7 +46,14 @@ def knife_files
       crowbar_databags_path.children.each do |file|
         file_path = crowbar_databags_path.join(file)
 
-        if file.basename.to_s =~ /^bc-nova_dashboard-(.*)\.json$/
+        case file.basename.to_s
+        when /^bc-crowbar-(.*)\.json$/
+          json = JSON.load(file.read)
+          if json["raw_data"]["attributes"]
+            json["raw_data"]["attributes"].delete("rails")
```
rhafer: As ```json``` is of type Chef::DataBagItem  this should be:
```json["attributes"].delete("rails")``` or  ```json.raw_data["attributes"].delete("rails")```
I'd prefer the first.

rhafer:
+1

vuntz:
Isn't that covered by chef/data_bags/crowbar/migrate/crowbar/010_drop_rails.rb?

rhafer:
> Isn't that covered by chef/data_bags/crowbar/migrate/crowbar/010_drop_rails.rb?

@vuntz Yeah, but it can't work. Because the "rails" attribute is outside the "attributes.crowbar" tree. And the migration only get's to see everything inside that hierarchy.

vuntz:
@rhafer then I'm missing the "why is this unwanted?" explanation in the commit message :-)

rhafer:
> then I'm missing the "why is this unwanted?" explanation in the commit message :-)

@MaximilianMeister can you add something like this to the commit message:

The "rails" attributes have been removed from the crowbar proposal. We're deleting them here because it's not possible to delete them using a Schema Migration as they were located outside the "attributes.crowbar" hierarchy in the proposal.

jsuchome:
+1 (and I also vote for more detailed commit message)

MaximilianMeister:
@jsuchome @rhafer @vuntz commit message updated

rhafer:
+1 again

vuntz:
@MaximilianMeister hrm, it seems your lines in the commit messages are a bit too long ;-)

But +1

MaximilianMeister:
@rhafer @vuntz gating is green, can someone merge this please?

PR #221, opened 2016-01-25
Upgrader backend

rsalevsky:
https://github.com/crowbar/crowbar-openstack/pull/224 is required

### DIFF
--- a/crowbar_framework/app/controllers/installer/upgrades_controller.rb
+++ b/crowbar_framework/app/controllers/installer/upgrades_controller.rb
@@ -38,7 +38,7 @@ module Installer
           @backup = Backup.new(params.permit(:file))
           if @backup.save
             format.html do
-              redirect_to install_upgrade_url
+              redirect_to restore_upgrade_url
             end
           else
             format.html do
@@ -54,7 +54,7 @@ module Installer
       end
     end
 
-    def install
+    def restore
       @current_step = 4
       if request.post?
         respond_to do |format|
--- a/crowbar_framework/app/controllers/installer/upgrades_controller.rb
+++ b/crowbar_framework/app/controllers/installer/upgrades_controller.rb
@@ -73,8 +73,14 @@ module Installer
       @current_step = 5
       if request.post?
         respond_to do |format|
-          format.html do
-            redirect_to services_upgrade_url
+          if view_context.check_repos?
+            format.html do
+              redirect_to services_upgrade_url
+            end
+          else
+            format.html do
+              redirect_to repos_upgrade_url
+            end
           end
         end
       else
--- a/crowbar_framework/app/controllers/installer/upgrades_controller.rb
+++ b/crowbar_framework/app/controllers/installer/upgrades_controller.rb
@@ -21,6 +21,7 @@ module Installer
     skip_before_filter :enforce_installer
     before_filter :hide_navigation
     before_filter :set_progess_values
+    before_filter :set_service_object, only: [:services, :backup]
 
     def show
       respond_to do |format|
@@ -95,8 +96,17 @@ module Installer
 
       if request.post?
         respond_to do |format|
-          format.html do
-            redirect_to backup_upgrade_url
+          begin
+            @service_object.shutdown_services_at_non_db_nodes
+            @service_object.dump_openstack_database
+            format.html do
+              redirect_to backup_upgrade_url
+            end
+          rescue => e
+            format.html do
+              flash[:alert] = e.message
+              redirect_to services_upgrade_url
+            end
           end
         end
       else
@@ -110,8 +120,17 @@ module Installer
       @current_step = 7
       if request.post?
         respond_to do |format|
-          format.html do
-            redirect_to nodes_upgrade_url
+          begin
+            @service_object.finalize_openstack_shutdown
+            Openstack::Upgrade.unset_db_synced
+            format.html do
+              redirect_to nodes_upgrade_url
+            end
+          rescue => e
+            format.html do
+              flash[:alert] = e.message
+              redirect_to backup_upgrade_url
+            end
           end
         end
       else
@@ -148,6 +167,10 @@ module Installer
     end
 
     protected
+
+    def set_service_object
+      @service_object = CrowbarService.new(logger)
+    end
 
     def set_progess_values
       @min_step = 1
--- a/crowbar_framework/app/controllers/installer/upgrades_controller.rb
+++ b/crowbar_framework/app/controllers/installer/upgrades_controller.rb
@@ -20,6 +20,7 @@ module Installer
   class UpgradesController < ApplicationController
     skip_before_filter :enforce_installer
     before_filter :hide_navigation
+    before_filter :set_progess_values
 
     def show
       respond_to do |format|
@@ -30,6 +31,7 @@ module Installer
     end
 
     def start
+      @current_step = 3
       if request.post?
         respond_to do |format|
           @backup = Backup.new(params.permit(:file))
@@ -50,7 +52,9 @@ module Installer
         end
       end
     end
+
     def install
+      @current_step = 4
       @backup = Backup.all.first
       if request.post?
         respond_to do |format|
@@ -66,6 +70,7 @@ module Installer
     end
 
     def repos
+      @current_step = 5
       if request.post?
         respond_to do |format|
           format.html do
@@ -80,6 +85,8 @@ module Installer
     end
 
     def services
+      @current_step = 6
+
       if request.post?
         respond_to do |format|
           format.html do
@@ -94,6 +101,7 @@ module Installer
     end
 
     def backup
+      @current_step = 7
       if request.post?
         respond_to do |format|
           format.html do
@@ -108,6 +116,7 @@ module Installer
     end
 
     def nodes
+      @current_step = 8
       if request.post?
         respond_to do |format|
           format.html do
@@ -122,6 +131,7 @@ module Installer
     end
 
     def finish
+      @current_step = 9
       respond_to do |format|
         format.html
       end
@@ -132,6 +142,11 @@ module Installer
     end
 
     protected
+
+    def set_progess_values
+      @min_step = 1
+      @max_step = 9
+    end
 
     def hide_navigation
       @hide_navigation = true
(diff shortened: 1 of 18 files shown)


crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -72,10 +131,19 @@ def services
     end
 
     def backup
+      @current_step = 8
       if request.post?
         respond_to do |format|
-          format.html do
-            redirect_to nodes_upgrade_url
+          begin
+            @service_object.finalize_openstack_shutdown
```
jsuchome: Please add `Openstack::Upgrade.unset_db_synced` here... at least until we find out better place.

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -30,6 +32,45 @@ def show
...
+          end
+        end
+      else
+        respond_to do |format|
+          format.html
+        end
+      end
+    end
+
+    def install
```
jsuchome: I do not understand this step, what is it supposed to do?

jsuchome:
Please add at least some meaningful (while still temporary) text to backup.description, like 
*Now OpenStack services have been shut down.
Find the backup of OpenStack database at node XY, as `/var/lib/pgsql/db-before-cloud6.dump`.*

crowbar_framework/app/controllers/installer/upgrades_controller.rb
rsalevsky: This steps shows the installer while crowbar backup restore.

rsalevsky:
> Please add at least some meaningful (while still temporary) text to backup.description, like 
Now OpenStack services have been shut down.
Find the backup of OpenStack database at node XY, as /var/lib/pgsql/db-before-cloud6.dump.

Can you say from where I can get the information where the database backup is located?

sayalilunkad:
You can get the location of backup from here https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/crowbar/attributes/default.rb#L18

jsuchome:
> You can get the location of backup from here ...

Yes, that's the location on the node, but we still need to find out correct node. I think you could just use that premininary text anyway - I do not think this PR will be closed only when all the stuff is really finished, so it's better than nothing.

jsuchome:
If http://imagebin.suse.de/2182/img is the correct view of this, why do we have first two steps there? (Prepare Upgrade and Data download). That seems like something that belongs to Cloud5 UI

rhafer:
> But than the location of the dump file is on shared device so it should not matter which of the nodes we pick.

That's not quite true. Because the device is only mounted on one of the nodes at a given time.
We could save the location in a node attribute and search for that. That's not 100% accurate either because in theory the active node might have changed in the meantime (it's unlikely to happen, but ...)

rsalevsky:
> If http://imagebin.suse.de/2182/img is the correct view of this, why do we have first two steps there? (Prepare Upgrade and Data download). That seems like something that belongs to Cloud5 UI

That's true, you have all steps in both Cloud versions. It should show that the steps belongs together. In Cloud 5 you can just go till step 2 and on Cloud 6 you start with step 3 then.

jsuchome:
> We could save the location in a node attribute and search for that.

OK, I'll try to adapt backend for that. But let's not delay this PR by waiting for it

jsuchome:
> That's true, you have all steps in both Cloud versions

Hm, I find it a bit confusing, but if others think it is OK, I'm not against.

crowbar_framework/app/helpers/installer/upgrades_helper.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      true
+    end
+
+    def check_repos
+      check_ha_repo != false && check_ceph_repo != false
+    end
+
+    def database_node
+      @node ||= NodeObject.find("state:crowbar_upgrade AND run_list_map:database-server").first
```
jsuchome: OK, so with https://github.com/crowbar/crowbar-core/pull/232 , this could be used to find the real node. I'm not sure if it is correct (had no chance to test it) but it might be
``@node = NodeObject.find("crowbar_upgrade_db_dumped_here:true")`

jsuchome:
I do not have opportunity to test it  now, but it looks correct to me. Let's solve new/followup issues by new PR.

crowbar_framework/app/helpers/installer/upgrades_helper.rb
```diff
@@ -0,0 +1,79 @@
...
+
+      unless Crowbar::Repository.provided_and_enabled?("ceph")
+        Openstack::Upgrade.enable_repos_for_feature("ceph", logger)
+      end
+
+      true
+    end
+
+    def check_repos
+      check_ha_repo != false && check_ceph_repo != false
```
tboerger: Please do it ruby conventional to suffix boolean return value methods mit `?`, beside that just check for `check_ha_repos? && check_ceph_repo?`

crowbar_framework/app/helpers/installer/upgrades_helper.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    def check_repos_button
+      if check_repos
+        continue_button
+      else
+        button_tag icon_tag(:refresh, t(".recheck")), class: "btn btn-primary"
+      end
+    end
+
+    def panel_body_color(boolen)
```
MaximilianMeister: s/boolen/boolean but why a boolean here? why is this helper needed if it just returns one of two strings?

MaximilianMeister: maybe you just call `check_ha_repo` from the helper then you don't need to pass an argument

crowbar_framework/app/helpers/installer/upgrades_helper.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    def check_ha_repo?
+      return nil unless Proposal.where(barclamp: "pacemaker")
+      return false unless Crowbar::Repository.provided?("ha")
+
+      unless Crowbar::Repository.provided_and_enabled?("ha")
+        Openstack::Upgrade.enable_repos_for_feature("ha", logger)
+      end
+
+      true
```
rhafer: This is wrong. At least when ```Openstack::Upgrade.enable_repos_for_feature``` fails. E.g. because the repo not really there.

Same is true for ```check_ceph_repo?```

crowbar_framework/app/helpers/installer/upgrades_helper.rb
```diff
@@ -0,0 +1,79 @@
...
+    def alert_type(boolean)
+      if boolean
+        "alert-success"
+      else
+        "alert-danger"
+      end
+    end
+
+    def check_ha_repo?
+      return nil unless Proposal.where(barclamp: "pacemaker")
```
rhafer: ```Proposal.where(barclamp: "pacemaker")``` always returns something that evaluates to  ```true``` you might try with:

```Proposal.where(barclamp: "pacemaker").first``` that will evaluate to ```nil``` in case there is no proposal.

File: crowbar_framework/app/helpers/installer/upgrades_helper.rb (hunk 6 above)
Comment (jsuchome) [inline crowbar_framework/app/helpers/installer/upgrades_helper.rb]: If repo is not really there, shouldn't it  exit at Crowbar::Repository.provided?("ha")

File: crowbar_framework/config/locales/crowbar/en.yml [hunk 8]
```diff
@@ -551,6 +551,7 @@ en:
         prepare: 'Prepare upgrade'
         download: 'Data download'
         start: 'Start upgrade'
+        install: 'Install Crowbar'
```
jsuchome: No, this does not make sense. Why should we talk about *Installing Crowbar* after data restore is done?

crowbar-upgrade-ui-after-restore

crowbar_framework/app/helpers/installer/upgrades_helper.rb
rhafer: I am not really sure about the semantics of that method. ```provided?``` Seems to just check whether that feature is provided by one of the repos mentioned in the repos*.yml file in crowbar_framework/config (or /etc/crowbar)
It doesn't check wether that repo actually is accessible AFAICS. But maybe @MaximilianMeister  or @vuntz  can clarify on that.

jsuchome: The same function is called from `ceph_service.rb` and `pacemaker_service.rb` so I would expect it returns wrong results there as well...

jsuchome: @rhafer See https://github.com/crowbar/crowbar-core/pull/256 for possible fix...

rhafer: > The same function is called from ceph_service.rb and pacemaker_service.rb so I would expect it returns wrong results there as well...

It's calling provided_and_enabled there which seems to de something slightly different. (But maybe that's what you want here as well?

jsuchome: Yes, but `provided_and_enabled?` doesn't check repository presense as well

rhafer:
Isn't clicking "continue" after selecting a backup tarball on the "Start Upgrade" Page supposed to show:
1. Progess of the upload (I thinke we could live without that)
2. Redirectry ot the "Restore Crowbar" page and show the Installer Steps, this would be really helpful as the installer takes quite some time to complete.

BTW, I am fine with addressing that in a follow up. Same is true for the missing OS upgrade steps backend calls.

rsalevsky:
> 1. Progess of the upload (I thinke we could live without that)

This is not really needed it should just takes seconds, what we can do easily is including a spinner.

> 2. Redirectry ot the "Restore Crowbar" page and show the Installer Steps, this would be really helpful as the installer takes quite some time to complete.

We have the steps in the view but we can't display them all because we don't have enough space. But I will address this in a follow up PR where I integrate the restore steps.

The PR is big enough so I prefer doing all additional changes in a new PR.

rhafer:
> The PR is big enough so I prefer doing all additional changes in a new PR.

I agree. +1

MaximilianMeister:
+1

PR #222, opened 2016-01-25
Fix updating the PXE config for discovered nodes

rhafer:
They don't have an admin IP allocated in that state, so we can't use that for
finding out the correct mac address.


vuntz:
+1

rsalevsky:
+1

toabctl:
+1

PR #224, opened 2016-01-25
Fix restoring directories

rhafer:
When restoring a directory we only want the contents of the directory be
restored and not the directory itself, otherwise we get e.g. ".ssh" restored
into ".ssh/.ssh".

### DIFF
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -66,10 +66,20 @@ module Crowbar
       def restore_crowbar
         Crowbar::Backup::Base.restore_files.each do |source, destination|
           # keep the permissions of the files that are already in place
+          src_path = @data.join("crowbar", source)
+          dest_path = Pathname.new(destination)
+          # If source and destination are both directories we just need to
+          # copy the contents of source, not the directory itself.
+          src_string = if dest_path.directory? && src_path.directory?
+            "#{srcpath}/."
+          else
+            srcpath.to_s
+          end
+
           system(
             "sudo", "-i",
             "cp", "-r", "--no-preserve=mode,ownership",
-            @data.join("crowbar", source).to_s,
+            src_string,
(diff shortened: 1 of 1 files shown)


jsuchome:
+1 (I'm assuming that if `destination` is a directory, we make sure it is created)

rhafer:
Good question, currently AFAIK we're not creating directories on restore. Though I think all of them are existing already when restoring. @MaximilianMeister ?

MaximilianMeister:
> I'm assuming that if destination is a directory, we make sure it is created

is `cp -r` not doing that?

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -66,10 +66,12 @@ def restore_chef
       def restore_crowbar
         Crowbar::Backup::Base.restore_files.each do |source, destination|
           # keep the permissions of the files that are already in place
+          srcpath = @data.join("crowbar", source)
```
MaximilianMeister: MINOR: for consistency I would name it `src_path`

MaximilianMeister: and delete the comment above. this will hopefully change with https://github.com/crowbar/crowbar-core/pull/199

jsuchome:
@MaximilianMeister Well if we are copying `cp -r #{srcpath}/. destination` that we copy the content of srcpath directory into destination directory and that must exist before, right?

rhafer:
I guess I'll just rework it to copy without the "." if destination does not exist

vuntz:
+1

jsuchome:
+1

vuntz:
Gating fails in cct because of a timeout when installing packages. Definitely not related to this, and the rest of the gating job would not have found anything in this PR, so merging.

rhafer:
Argh. There's a type in the merged code :(

Will submit a fixup.

PR #225, opened 2016-01-26
ensure interface_list is a list (bsc#963540)

bmwiedemann:
https://bugzilla.suse.com/show_bug.cgi?id=963540

### DIFF
--- a/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+++ b/chef/cookbooks/barclamp/libraries/barclamp_library.rb
@@ -201,7 +201,7 @@ module BarclampLibrary
         return [nil, nil] if intf_to_if_map[conduit].nil?
 
         c_info = intf_to_if_map[conduit]
-        interface_list = c_info["if_list"]
+        interface_list = c_info["if_list"] || []
         team_mode = c_info["team_mode"] rescue nil
 
         return [interface_list[0], interface_list, nil] if interface_list.size == 1
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -1046,7 +1046,7 @@ class NodeObject < ChefObject
     return [nil, nil] if intf_to_if_map[conduit].nil?
 
     c_info = intf_to_if_map[conduit]
-    interface_list = c_info["if_list"]
+    interface_list = c_info["if_list"] || []
     team_mode = c_info["team_mode"] rescue nil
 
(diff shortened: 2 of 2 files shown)


chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -201,7 +201,7 @@ def self.lookup_interface_info(node, conduit, intf_to_if_map = nil)
         return [nil, nil] if intf_to_if_map[conduit].nil?
 
         c_info = intf_to_if_map[conduit]
-        interface_list = c_info["if_list"]
+        interface_list = c_info["if_list"] || []
```
vuntz: Wouldn't we crash a few lines later if it's `nil` (when we do `interface_list.size == 1`)?

vuntz:
cc @rhafer as the bug is related to what we try to fix in https://github.com/crowbar/crowbar-core/pull/222 (which suggests that the fix there is incomplete, I guess)

chef/cookbooks/barclamp/libraries/barclamp_library.rb
rhafer: I am pretty sure this is not going to fix the problem. If interface_list was nil here before chef would have thrown and "NoMethodError: undefined method `size' for nil:NilClass" exception on line 204. If you initialize to [] it will bail out on  line 219. Unless I am overlooking something.

rhafer:
Looking at the logs in the bug, I see a lot of SQL Busy Errors. So this might really be a different thing.

bmwiedemann:
#222 is not merged yet, so maybe it is the same root cause?

rhafer:
The failure that was causing #222 was slightly different and the fix in #222 would not solve the problem in bug#963540

chef/cookbooks/barclamp/libraries/barclamp_library.rb
bmwiedemann: yes, indeed. So this is not where the nil value came from.

dirkmueller:
Perhaps we should simply move the invocation of find_node_boot_mac_addresses() to the places when we are sure the node has an admin network assigned?

rhafer:
If the node didn't have an admin IP assigned I am pretty sure it would have failed in a different way (thats what #222 is supposed to address). This case is somewhat different and it might really be related to the SQL Exceptions (e.g the node might have lost a role because of that).

@bmwiedemann Are you able to reproduce that failure somewhere where we could take a closer look?

PR #226, opened 2016-01-26
Revert runlist reduction for crowbar_upgrade

jsuchome:
introduced by https://github.com/jsuchome/crowbar-core/commit/03f83fc3744449a51916bd07dcd7c5bfb706657a

We need to make sure that all the Proposal Roles (-config-default) (especially those
of the core barclamps) stay assigned to the nodes during the upgrade process.
Otherwise chef-client on the admin server (e.g. provisioner::update_nodes) will fail.

(This is same as https://github.com/crowbar/barclamp-crowbar/pull/1368 for tex)

### DIFF
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -727,12 +727,10 @@ class NodeObject < ChefObject
     crowbar["run_list_map"] = {} if crowbar["run_list_map"].nil?
 
     # Cull by state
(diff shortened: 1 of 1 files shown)


rhafer:
+1

vuntz:
+1

@rhafer is your +1 still valid (not sure if only commit message changed or code also changed)?

rhafer:
Yeah, I still good with it. +1

PR #229, opened 2016-01-26
Abort if cstruct can not be loaded

dirkmueller:
Apparently under some circumstances installation of cstruct fails
(maybe has something to do with invalid repositories being listed)
and then the rest goes berserk. Abort early instead.


vuntz:
I don't mind the change, but when this happens, we should investigate what's going on; chef should fail if the package cannot be installed. Maybe the zypper bug with wrong exit code is back (or maybe the update for it is not out?)

dirkmueller:
@vuntz right, in one case that I was debugging it was the zypper-bug-with-wrong exit code somehow. I was not able to pinpoint it because the thing failed at a completely different place..

PR #230, opened 2016-01-26
Update the target_platform attribute before upgrading the OS

rhafer:
The nodes (including the admin node) still have the old target_platform set
(e.g. suse-11.3) change that to the new default OS.

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -307,8 +307,11 @@ class CrowbarService < ServiceObject
 
   def prepare_nodes_for_os_upgrade
     upgrade_nodes = NodeObject.all.reject { |node| node.admin? || node[:platform] == "windows" }
+    admin_node = NodeObject.admin_node
 
     upgrade_nodes.each do |node|
+      node["target_platform"] = admin_node["provisioner"]["default_os"]
+      node.save
       node.set_state("os-upgrading")
     end
 
--- a/crowbar_framework/lib/crowbar/upgrade.rb
+++ b/crowbar_framework/lib/crowbar/upgrade.rb
@@ -100,6 +100,18 @@ module Crowbar
           filecontent_replace(file, "nova_dashboard", "horizon")
         end
       end
+
+      # find admin node and update target_platform
+      nodes_path = knife_path.join("nodes")
+      nodes_path.children.each do |file|
+        json = JSON.load(file.read)
+        next unless json["crowbar"] && json["crowbar"]["admin_node"]
+        json.delete "target_platform"
+        json["provisioner"].delete "default_os"
+        file.open("w") do |node|
+          node.write(JSON.pretty_generate(json))
+        end
+      end
     end
 
     def crowbar_files


crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -291,8 +291,13 @@ def disable_non_core_proposals
 
   def prepare_nodes_for_os_upgrade
     upgrade_nodes = NodeObject.all.reject { |node| node.admin? || node[:platform] == "windows" }
+    admin_node = NodeObject.admin_node
+    admin_node["target_platform"] = admin_node["provisioner"]["default_os"]
+    admin_node.save
```
vuntz: Shouldn't that bit be done a bit earlier? (like, on restore)

rhafer: @vuntz You mean just for the admin node? Or for all nodes?

I guess we could do it also on restore. Might a bit more complicated to implement it there though.

vuntz: Just for the admin node. An option is to just unset it during the restore; there's code elsewhere that will automatically reset it to the right value, I think.

rhafer: Ah, ok. Yeah, I guess that makes sense.

rhafer:
move the code for the admin node to upgrade.rb. Not fully tested yet.

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -100,6 +100,17 @@ def knife_files
           filecontent_replace(file, "nova_dashboard", "horizon")
         end
       end
+
+      # find admin node and update target_platform
+      nodes_path = knife_path.join("nodes")
+      nodes_path.children.each do |file|
+        json = JSON.load(file.read)
+        next unless json["crowbar"] && json["crowbar"]["admin_node"]
+        json["target_platform"] = "suse-12.1"
```
vuntz: I'm fine with this, but just wondering out loud: do we possibly have any way to avoid the hard-coding?

rhafer: Yeah, I wondered about that as well. Couldn't come up with a good idea yet. Do we have any (non-clumsy) way to map from the current information in /etc/os-release to a platform string that chef likes? I guess ohai provides that, but it's probably a bit too much to run ohai here?

vuntz: Yeah, running ohai is a bit overkill. So what about my earlier suggestion: simply `json.delete("target_platform")` and let the provisioner cookbook fill that again?

vuntz: Btw, we should probably unset `node[:provisioner][:default_os]` for the admin node, to have that be updated too (although it seems it will be updated anyway right now, but that could be a bug).

rhafer: > So what about my earlier suggestion: simply json.delete("target_platform") and let the provisioner cookbook fill that again?

Yeah, doing that now. I also clear the default_os now as suggested. Seems to work fine.

vuntz:
+1

aplanas:
+1

PR #232, opened 2016-01-27
Indicate the node where the dump is actually located

jsuchome:
cc @rhafer

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
@@ -36,7 +36,13 @@ when "openstack_shutdown"
 
   # If postgres is not running here, it means we're in the DB cluster and database runs
   # on the other node: let the other node take care of the rest.
-  return unless ::Kernel.system("service postgresql status")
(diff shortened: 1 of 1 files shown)


rhafer:
+1 , If this is really what upgrade UI needs.

dirkmueller:
shouldn't that be an override attribute? Just a guess..

sayalilunkad:
probably need to define this somewhere node[:crowbar][:upgrade][:db_dumped_here]. Not sure though.

jsuchome:
@sayalilunkad no, it's not needed (we already define :upgrade submap)
@dirkmueller you mean that we would not need to reset it on the other node? Might be good, but currently I'm glad that this works

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
```diff
@@ -68,6 +74,10 @@
     EOF
   end
 
+  # we have to indicate the node where the dump is actually located
+  node[:crowbar][:upgrade][:db_dumped_here] = true
```
vuntz: `node.set` please

vuntz:
+1

PR #233, opened 2016-01-27
preserve ownership/permissions on backup and restore

MaximilianMeister:
discussion happened in https://github.com/crowbar/crowbar-core/pull/199

cc @rhafer @vuntz


rhafer:
+1 (the other +1 is in #199 so merging)

PR #236, opened 2016-01-27
add missing chef files

MaximilianMeister:


MaximilianMeister:
rebased

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -70,27 +79,39 @@ def restore_chef
...
+          "#{src_path}/."
+        else
+          src_path.to_s
+        end
+
+        system(
+          "sudo", "-i",
+          "cp", "-a",
+          src_string,
+          @data.join("crowbar", source).to_s,
```
rhafer: You need to remove this line. The source is already in ```src_string```

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -40,6 +47,8 @@ def restore_chef
         begin
           [:nodes, :roles, :clients, :databags].each do |type|
             Dir.glob(@data.join("knife", type.to_s, "**", "*")).each do |file|
+              # skip client "crowbar"
+              next if type == :clients && file.basename.to_s =~ /^crowbar.json$/
```
rhafer: ```file``` is not a Pathname here yet.

Just moving the next line in front of this one should help

crowbar_framework/lib/crowbar/backup/restore.rb
MaximilianMeister: sorry, manual rebase mistake

rsalevsky:
Maybe it's better when we transform the file list into a hash and add this just as `key: value`.

MaximilianMeister:
> Maybe it's better when we transform the file list into a hash and add this just as key: value.

I agree a Hash is always nicer, but I think it is fine as an Array as long as you don't have more than two values in each entry.

MaximilianMeister:
rebased

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -70,27 +80,38 @@ def restore_chef
         true
       end
 
+      def restore_files(source, destination)
+        # keep the permissions of the files that are already in place
+        src_path = @data.join("crowbar", source)
+        dest_is_dir = system("sudo", "-i", "test", "-d", destination)
```
aplanas: Ahm not sure, but `sudo -i` will put you in /root (in a different working directory that you are now)

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -70,27 +80,38 @@ def restore_chef
...
+        # If source and destination are both directories we just need to
+        # copy the contents of source, not the directory itself.
+        src_string = if dest_is_dir && src_path.directory?
+          "#{src_path}/."
+        else
+          src_path.to_s
+        end
+
+        system(
+          "sudo", "-i",
```
aplanas: same here?

crowbar_framework/lib/crowbar/backup/restore.rb
rhafer: IIRC ```destination``` is an absolute path here. But I guess you're still right the ```-i``` shouldn't be needed here. (And in the other place)

MaximilianMeister:
@aplanas @rhafer  dropped "-i"

rhafer:
+1

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -70,27 +80,38 @@ def restore_chef
...
+          "#{src_path}/."
+        else
+          src_path.to_s
+        end
+
+        system(
+          "sudo",
+          "cp", "-a",
+          src_string,
+          destination
```
aplanas: bit weird line layout?

aplanas:
+1 for me

rsalevsky:
+1

PR #238, opened 2016-01-28
Fix login to UCS Manager

mmnelemane:
Resolve errors during login to UCS manager from crowbar UI.
s/True/true/ and s/False/false/.


vuntz:
+1

dirkmueller:
+1

aplanas:
+1

PR #239, opened 2016-01-28
refine ambiguous regex for replacing file content

MaximilianMeister:
the regex "bc-" is not exact enough as there could be another
string in the json file, that contains "bc-", e.g. a cinder-backend
password.
make sure to determine the barclamps name and use it to replace
only patterns in the file that have "bc-<bc_name>"

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade.rb
+++ b/crowbar_framework/lib/crowbar/upgrade.rb
@@ -88,8 +88,13 @@ module Crowbar
         end
 
         next unless file_path.basename.to_s =~ /^bc-(.*).json$/
+        # better determination of bc-<barclamp_name> to replace in the file
+        # it could be that some other unrelated string in the json contains "bc-"
+        bc_name_search = file_path.basename.to_s.rpartition("-").first
+        bc_name_replace = bc_name_search.split("-").last
+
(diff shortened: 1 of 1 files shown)


MaximilianMeister:
@aplanas sorry for the late answer, I have investigated a bit. AFAICS we cannot have the case that a proposal name has a "-" in it.

I have tried it with pacemaker, which supports naming a proposal with something else than `default` and I got
`Invalid service name proposa-1. Please limit name to [A-z0-9_] without spaces.`

so we should be good to go here.

aplanas:
+1

MaximilianMeister:
I will make a rebase and trigger gating once more.

rsalevsky:
+1

aplanas:
@rsalevsky to soon to merge IMHO, there where complains in the gating before a rebase (a 500 internal error in crowbar) There was a change that is the same issue related with model validations that we see in different PR, but now I can't be sure.

rsalevsky:
@aplanas This part is not covered by the gating, I run in manual on my machine and it succeeded.

vuntz:
Would it have been an option to actually only replace things that were in quotes? That would also remove some false positive.

PR #241, opened 2016-01-28
create tarball with root, drop archive gem

MaximilianMeister:


MaximilianMeister:
I assume we don't need archive anymore? can anyone confirm?

rhafer:
> I assume we don't need archive anymore? can anyone confirm?

@rsalevsky  It was introduced only for the backup restore feature, right ? If yes, it can go now.

rsalevsky:
Yes, was just for this usecase.

MaximilianMeister:
> Yes, was just for this usecase.

we will probably have to remove it from the product as well then

rhafer:
+1

rsalevsky:
+1

PR #243, opened 2016-01-28
Adds failure clause to prepare_nodes_for_upgrade

dguitarbite:


crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -338,10 +339,21 @@ def prepare_nodes_for_os_upgrade
...
+          upgrade_nodes_failed.push(node.name)
+        end
+      else
+        @logger.error("Upgrade failed for #{node.name}. Node not ready for reboot")
+        upgrade_nodes_failed.push(node.name)
       end
     end
-    true
+    if upgrade_nodes_failed.any?
+      return false, upgrade_nodes_failed
```
rsalevsky: Please don't return an array.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -338,10 +339,21 @@ def prepare_nodes_for_os_upgrade
...
+    else
+      return true
```
rsalevsky: Write something like:

```ruby
return false if upgrade_nodes_failed.any?

true
```

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -307,6 +307,7 @@ def disable_non_core_proposals
   def prepare_nodes_for_os_upgrade
     upgrade_nodes = NodeObject.all.reject { |node| node.admin? || node[:platform] == "windows" }
     admin_node = NodeObject.admin_node
+    upgrade_nodes_failed = []
```
rsalevsky: Move this to line between line 339 and 340.

crowbar_framework/app/models/crowbar_service.rb
dguitarbite: Please suggest me a better way to do the same? How do I return a list of things without returning an array?

crowbar_framework/app/models/crowbar_service.rb
dguitarbite: Lets keep it there. We may need it for future.

crowbar_framework/app/models/crowbar_service.rb
dguitarbite: Its the same. I am not going to change it. There is nothing wrong with the current syntax. Performance wise and code wise.

rhafer: > Write something like:
> 
> ```return false if upgrade_nodes_failed.any?```

@rsalevsky  Actually the goal is to return something more that just false in case the reboot of one or more of the nodes failed (e.g. the names for the nodes that failed). So that the Controller could you that value to display a more usefull error message than "something failed".

aplanas: Actually I found this quite dangerous because the `false` results is always evaluated as a `true` in a Boolean context. IMHO is better to return, for this method, always an array and force the unpack in every call.

dguitarbite: @aplanas Not exactly sure if I understand this in the right way but is this something specific to ruby? Its a bit difficult to grasp how even Boolean evaluation checks for `empty returns` over `something in there is true` ... I always saw programming languages interpret Boolean as either `true` or `false`! I will update this if what you say applies for Ruby. #weird!

aplanas: @dguitarbite Is like in Python (at least in my tests with `irb`):

* `true` is evaluated as `true`
* `[false, [list, of, nodes]]` is evaluated also as `true`

So maybe we can do as in Python: return always `[true, nil]` or `[false, [list, of, nodes]]`, so in the caller side you always use `result, list_of_nodes = object.prepare_nodes_for_os_upgrade(...)`

But I am sure that a Ruby expert can provide a confirmation is this an expected idiom in Ruby too.

dguitarbite: @aplanas Ill update this. Anyways I see that the code which is calling this method is not ready to handle this change at the moment. So this change should be good starting point for the other methods calling this one. Ill update it.

crowbar_framework/app/models/crowbar_service.rb
rsalevsky: Just return `upgrade_nodes_failed` without `false` otherwise you will have a nested array which is not useful in this case.

crowbar_framework/app/models/crowbar_service.rb
rsalevsky: Let's do it in this way:

``ruby
return upgrade_nodes_failed if upgrade_nodes_failed.any?

true
```

With this we have the nodes which failed and we do it in the ruby way.

rhafer: @rsalevsky Hm, though ```true``` and [ "some", "list" ] are both evaluating to ```true``` in boolean context. Would it then be better to return ```nil``` in case of success?

crowbar_framework/app/models/crowbar_service.rb
rsalevsky: Then we can move it in the future for now we follow the common ruby style. We decided to not merge easy fixable style issues.

crowbar_framework/app/models/crowbar_service.rb
rsalevsky: @rhafer Right, we can do this or return an empty array then we just need to check if the array has elements.

dguitarbite: @rsalevsky If the array is empty it will return ``true``

crowbar_framework/app/models/crowbar_service.rb
dguitarbite: I cannot find common ruby style anywhere in the official ruby docs. Link please.

crowbar_framework/app/models/crowbar_service.rb
rhafer: @dguitarbite Right, and that's what we just mentioned as problematic. Because it's going to be difficult for the caller to test if the call was successful or not. Please either return ```nil``` or just the empty ```upgrade_nodes_failed``` list in case of success.

crowbar_framework/app/models/crowbar_service.rb
rsalevsky: Sorry, missed the `upgrade_nodes.each do |node|` block so moving the line would be anyway wrong.

crowbar_framework/app/models/crowbar_service.rb
dguitarbite: @rhafer Ill return an empty list. So I can remove the if clause at the return statement block of the method.

rsalevsky:
+1

rhafer:
+1

PR #245, opened 2016-01-28
Rework and fix "disable_non_core_proposals"

rhafer:
Relying on chef search to find relevant nodes which have the element roles
assinged doesn't work as the nodes are in "crowbar_upgrade" state, where they
the lack the element roles.
This implementation also reduces the amount of node.save calls to one per node.

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -288,21 +288,20 @@ class CrowbarService < ServiceObject
     # Find all non-core proposals and remove all roles that belong
     # to those proposals from the nodes
     active_non_core_roles = RoleObject.find_roles_by_name("*-config-*").reject(&:core_role?)
-    active_non_core_roles.each do |role|
-      @logger.debug("deactivating proposal #{role.name}")
-
-      roles_to_remove = role.elements.keys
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -288,21 +288,20 @@ def disable_non_core_proposals
...
+      active_non_core_roles.each do |role|
+        roles_to_remove = role.elements.keys
+        roles_to_remove << role.name
+        roles_to_remove.each do |delete_role|
           node.delete_from_run_list(delete_role)
-          node.save
         end
       end
-      role.destroy
+      node.save
```
aplanas: As I understand in the original code, the role is also removed, but not in the new version. Is this change OK?

jsuchome:
LGTM

crowbar_framework/app/models/crowbar_service.rb
MaximilianMeister: happens below in 304

MaximilianMeister:
maybe re-add some logging, but codewise +1

vuntz:
+1

aplanas:
+1

PR #246, opened 2016-01-28
Leave some upgrade related role for nodes in crowbar_upgrade state. 

jsuchome:
some upgrade role assigned, so it does not lose its state.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -44,6 +44,8 @@ when "openstack_shutdown"
       done
     EOF
   end
+when "done_openstack_shutdown", "wait_for_openstack_shutdown"
+  Chef::Log.debug("Nothing to do on this node, waiting for others to finish their work...")
 else
   Chef::Log.warn("Invalid upgrade step given: #{upgrade_step}")
 end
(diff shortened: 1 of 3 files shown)


jsuchome:
Now I'm basically losing some clarity from using 2 different roles (crowbar-upgrade and crowbar-db-dump) ... :-(

rhafer:
The commit message is a bit confusing (at least the headline in much too long). Otherwise I think this is ok.

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -44,6 +44,8 @@
       done
     EOF
   end
+when "done_openstack_shutdown", "wait_for_openstack_shutdown"
+  Chef::Log.debug("Nothing to do at this node, waiting for others to finish their work...")
```
vuntz: s/at this node/on this node/

vuntz:
So I must admit reading the code change is clearer than the commit message, but I have no real suggestion on how to improve the message. So +1 (would be nice to fix the small nitpick)

rhafer:
I am wondering if it would be simpler now to remove the "crowbar-db-dump" role completely and orchestrate the whole things base on different "crowbar_upgrade_step" values for different nodes.

But in the end I think it doesn't matter too much. So +1

jsuchome:
@rhafer yes, I was hinting on that in my first comment, that different roles have little sense here.
Only that the db and non-db code is divided into different recipes, so there's still small value left

vuntz:
> @rhafer yes, I was hinting on that in my first comment, that different roles have little sense here.
> Only that the db and non-db code is divided into different recipes, so there's still small value left

Well, this can be done with some include_recipe magic without different roles.

Can we possibly investigate this as a followup? (not in this pr)

jsuchome:
Yes, I'll try to cleanup it after this one is merged

PR #249, opened 2016-01-29
Raise busy timeout.

dirkmueller:
When qa_crowbarsetup.sh renames the nodes in a loop, we're hitting
the transaction timeout as that takes more than 5s. Raise the limit
for now.

### DIFF
--- a/crowbar_framework/config/database.yml
+++ b/crowbar_framework/config/database.yml
@@ -17,8 +17,8 @@
 
 default: &default
   adapter: sqlite3
-  pool: 5
-  timeout: 5000
+  pool: 3
+  timeout: 30000
 
 development:
   <<: *default


crowbar_framework/config/database.yml
```diff
@@ -17,8 +17,8 @@
 
 default: &default
   adapter: sqlite3
-  pool: 5
-  timeout: 5000
+  pool: 3
```
aplanas: why decrease the pool?

vuntz:
+1, but I doubt this is enough to fix it (in my attempts, raising the timeout didn't solve the issue).

dirkmueller:
I think this will not solve anything, see mail to cloud-devel. something is fucked with the way we use activerecord

dirkmueller:
closing

PR #250, opened 2016-01-29
simplify ping_succeeds and firewall_disabled

bmwiedemann:
system call already returns a bool
so we can just return it as is

@rsalevsky might want to review this change of his code

### DIFF
--- a/crowbar_framework/lib/crowbar/checks/network.rb
+++ b/crowbar_framework/lib/crowbar/checks/network.rb
@@ -67,19 +67,11 @@ module Crowbar
       end
 
       def firewall_disabled?
-        if system("sudo LANG=C iptables -n -L | grep -qvE '^$|^Chain [^ ]|^target     prot'")
-          return false
-        end
-
-        true
+        !system("sudo LANG=C iptables -n -L | grep -qvE '^$|^Chain [^ ]|^target     prot'")
       end
 
       def ping_succeeds?
-        unless system("ping -c 1 #{fqdn} > /dev/null 2>&1")
-          return false
-        end
-
-        true
+        system("ping -c 1 #{fqdn} > /dev/null 2>&1")
       end
 
       def hostname


crowbar_framework/lib/crowbar/checks/network.rb
```diff
@@ -67,19 +67,11 @@ def ip_configured?
       end
 
       def firewall_disabled?
-        if system("sudo LANG=C iptables -n -L | grep -qvE '^$|^Chain [^ ]|^target     prot'")
-          return false
-        end
-
-        true
+        return ! system("sudo LANG=C iptables -n -L | grep -qvE '^$|^Chain [^ ]|^target     prot'")
```
aplanas: Maybe these spaces between "target" and "prot" and be expressed also as a regex (I know that is not related with this PR)

aplanas: Also I agree with houndci: the `return` is redundant

crowbar_framework/lib/crowbar/checks/network.rb
```diff
@@ -67,19 +67,11 @@ def ip_configured?
...
       end
 
       def ping_succeeds?
-        unless system("ping -c 1 #{fqdn} > /dev/null 2>&1")
-          return false
-        end
-
-        true
+        return system("ping -c 1 #{fqdn} > /dev/null 2>&1")
```
aplanas: same

rsalevsky:
The changes are not ruby style conform, from my point it also makes it not easier to understand.

bmwiedemann:
addressed hound style issues

dirkmueller:
@rsalevsky can you explain the ruby style issue?

aplanas:
+1 for me

vuntz:
+1, but would like to hear more about the style issue.

rsalevsky:
It's fine now. +1

PR #252, opened 2016-01-29
Set a timeout for sessions

dirkmueller:


vuntz:
So I think we want your change, but we need more than that. See http://guides.rubyonrails.org/security.html#session-expiry

vuntz:
See https://github.com/crowbar/crowbar-core/pull/257 for session sweeping.

+1 for your patch still.

aplanas:
+1 if this together with #257 address the full session problem

dirkmueller:
@vuntz do we still need this one when the other one is merged?

vuntz:
> @vuntz do we still need this one when the other one is merged?

Yes, your patch sets the timeout in the cookie. Which we want :-)

dirkmueller:
ok then..

tboerger:
What? 4 hours only?

PR #253, opened 2016-01-29
crowbar: Switch sqlite database to use write-ahead logging

vuntz:
Write-ahead logging seems to be more efficient, and apparently makes the
following SQLite3::BusyException much less frequent:
```
E, [2016-01-29T09:50:45.167185 #27521] ERROR -- : SQLite3::BusyException: database is locked: begin transaction
F, [2016-01-29T09:50:45.168484 #27521] FATAL -- :
ActiveRecord::StatementInvalid (SQLite3::BusyException: database is locked: begin transaction):
  config/initializers/sqlite_transaction_monkey_patch.rb:27:in `block in begin_db_transaction'
  config/initializers/sqlite_transaction_monkey_patch.rb:27:in `begin_db_transaction'
```
For more details, see https://www.sqlite.org/wal.html

### DIFF
--- a/crowbar_framework/config/database.yml
+++ b/crowbar_framework/config/database.yml
@@ -19,6 +19,7 @@ default: &default
   adapter: sqlite3
   pool: 5
   timeout: 5000
+  options: "PRAGMA journal_mode=WAL;"
 
 development:
   <<: *default


dirkmueller:
+1

aplanas:
+1

aplanas:
Maybe we can try also with "PRAGMA locking_mode=EXCLUSIVE;" for VFS without shared memory extensions

vuntz:
> Maybe we can try also with "PRAGMA locking_mode=EXCLUSIVE;" for VFS without shared memory extensions

It should only be needed for legacy VFSes, which I assumed was not a problem. Do you have more information on this?

dirkmueller:
lets wait with merging this for a minute..

aplanas:
No, and that is the reason to suggest a more conservative configuration. But I am more than fine to try this.

tboerger:
Than I'm pretty sure we can drop the mentioned initializer as it seems to be not working anymore.

vuntz:
> Than I'm pretty sure we can drop the mentioned initializer as it seems to be not working anymore.

It's still needed: without `:immediate`, things are worse because we don't even block trying to get a lock until it's too late and then there's no way to wait for the lock.

PR #254, opened 2016-01-29
crowbar: Fix strings in installer

vuntz:
The strings were not found in the library used; I assume it's because
the library is not the controller, so doesn't get the magic prefix.

### DIFF
--- a/crowbar_framework/lib/crowbar/installer.rb
+++ b/crowbar_framework/lib/crowbar/installer.rb
@@ -24,7 +24,7 @@ module Crowbar
         else
           return {
             status: 501,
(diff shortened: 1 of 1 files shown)


vmoravec:
This fix is already part of https://github.com/crowbar/crowbar-core/pull/221 .
Btw I could only have seen this translation issue only in when chef was failing, see here https://ci.suse.de/view/Cloud/view/QA/job/cloud-mkphyscloud-qa-scenario-2c/107/console and here http://pastebin.nue.suse.com/20294/src

vuntz:
> This fix is already part of #221 .

Patch there is not covering all strings... @rsalevsky ?

dirkmueller:
+1

PR #255, opened 2016-01-29
Do not mix chef calls and proposal saves

dirkmueller:
Using chef calls while the propsal is still being modified
can cause delays that would cause the transaction to be delayed
so long that other workers get a busy exception. Shorten the lifetime
of the critical section.


vuntz:
I don't understand.

Does doing something like `Proposal.where(barclamp: "crowbar", name: inst).first` lock the database?

vuntz:
So code looks correct; not sure if it will help us, but I'm fine trying.

crowbar_framework/app/models/provisioner_service.rb
```diff
@@ -72,22 +72,24 @@ def transition(inst, name, state)
...
-        # Reload the roles
-        db.reload
         role = RoleObject.find_role_by_name "provisioner-config-#{inst}"
       end
 
+      result = false
       @logger.debug("Provisioner transition: Make sure that base is on everything: #{name} for #{state}")
-      result = add_role_to_instance_and_node("provisioner", inst, name, db, role, "provisioner-base")
+      Proposal.where(barclamp: "provisioner", name: inst).first do |db|
+        result = add_role_to_instance_and_node("provisioner", inst, name, db, role, "provisioner-base")
```
bmwiedemann: https://ci.suse.de/view/Cloud/view/Builds/job/openstack-mkcloud/15836/ failed with
```
Error: service dhcpd missing
```

bmwiedemann: https://ci.suse.de/view/Cloud/view/Builds/job/openstack-mkcloud/15840/ has the same problem

tboerger:
> Using chef calls while the propsal is still being modified
> can cause delays that would cause the transaction to be delayed
> so long that other workers get a busy exception. Shorten the lifetime
> of the critical section.

I really don't understand how `find_by` should be related...

vuntz:
@dirkmueller this has conflicts now. But I'm not sure we still need this...

vuntz:
Do not merge for M11.

PR #256, opened 2016-01-29
Check if repository is physically available.

jsuchome:
I'm not sure why this check was missing, possibly there was a reason for that? @MaximilianMeister

### DIFF
--- a/crowbar_framework/lib/crowbar/repository.rb
+++ b/crowbar_framework/lib/crowbar/repository.rb
@@ -197,9 +197,10 @@ module Crowbar
             next unless provided_features.include? feature
             found = true
 
-            break unless check_enabled
-
             r = new(platform, arch, repo)
+            answer &&= r.available?
+
+            break unless check_enabled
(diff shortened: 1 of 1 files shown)


MaximilianMeister:
can't say without investigating. better ask @vuntz who wrote it. maybe he knows a quick answer

rhafer:
Makes sense to me. +1

jsuchome:
@vuntz do you know this code?

vuntz:
+1; the code change is really just a fix from the previous PR (https://github.com/crowbar/crowbar-core/pull/188/files)

rhafer:
+1

jsuchome:
@vuntz I'm not against your +1, but it is not a fix of previous PR, the check for availability of repo was never really there.

vuntz:
Well, it was kind of implicit wit the `active?` check :-)

jsuchome:
Hm, to me `active?` seems to only look at yml file value, but maybe I'm missing something

PR #257, opened 2016-01-29
crowbar: Add rake task and cron job to sweep sessions from database

vuntz:
This is not done automatically by rails, and is important to do. See
http://guides.rubyonrails.org/security.html#session-expiry


aplanas:
+1

rsalevsky:
+1

vuntz:
Hrm, should we change the crowbar cookbook to also install such a cron job?

tboerger:
https://github.com/rails/activerecord-session_store/pull/37/files :)

dirkmueller:
+1

PR #258, opened 2016-01-29
Use only one role for upgrade related tasks

jsuchome:
and decide what to do based on the value of crowbar_upgrade_step.

So, this is a followup of https://github.com/crowbar/crowbar-core/pull/246 where we realized there's no more a need to have more roles since we have to always give some upgrade role to all nodes.


jsuchome:
I've just tested it with latest Cloud6 code, and it seems to work well. @rhafer , @vuntz ?

sayalilunkad:
+1

rhafer:
+1

PR #259, opened 2016-01-29
provisioner: Do not set product key for Hyper-V install (bsc#964250)

vuntz:
The installer fails because it doesn't recognize the product key, but we
don't need one for Hyper-V.

https://bugzilla.suse.com/show_bug.cgi?id=964250


toabctl:
+1 (assuming you tested it)

vuntz:
> (assuming you tested it)

Yes :-)

dirkmueller:
+1

PR #261, opened 2016-02-01
crowbar: Improve logging to also contain a thread "id"

vuntz:
It's not really an ID (I think it's the pointer to the thread object),
but it's good enough to allow us finding which log line is for a
specific thread.

With threads, logs look like this:

  I, [2016-02-01T14:27:51.597500 #9963:0x00000002dece50]  INFO -- Completed 200 OK in 28ms (Views: 1.0ms | ActiveRecord: 0.2ms)

Without threads, they look like this:

  I, [2016-02-01T14:39:47.425383 #11702]  INFO -- Completed 200 OK in 544ms (Views: 0.8ms | ActiveRecord: 0.0ms)

### DIFF
--- a/crowbar_framework/lib/crowbar/logger/formatter.rb
+++ b/crowbar_framework/lib/crowbar/logger/formatter.rb
@@ -18,20 +18,57 @@ module Crowbar
   module Logger
     class Formatter < ::Logger::Formatter
       include ActiveSupport::TaggedLogging::Formatter
+      FORMAT = "%s, [%s#%d] %5s -- %s\n"
+      THREAD_FORMAT = "%s, [%s#%d:%s] %5s -- %s\n"
 
-      def call(severity, timestamp, progname, msg)
-        final_message = if msg.is_a? String
+      def call(severity, time, progname, msg)
+        threads = ENV["CROWBAR_THREADS"]
+        # if env var is not set, default is multi-threaded
+        if threads.to_s.empty? || threads.to_i > 1
+          # there's no API to get an id from a thread object, so let's cheat
+          thread_id = Thread.current.inspect.gsub(/^#<Thread:([^ ]*) .*/, "\\1")
+          format(
+            THREAD_FORMAT,
+            severity[0..0],
+            format_datetime(time),
+            $$,
+            thread_id,
+            severity,
+            msg2str(msg)
+          )
+        else
+          format(
+            FORMAT,
+            severity[0..0],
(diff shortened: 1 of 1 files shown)


crowbar_framework/lib/crowbar/logger/formatter.rb
```diff
@@ -18,20 +18,55 @@ module Crowbar
   module Logger
     class Formatter < ::Logger::Formatter
       include ActiveSupport::TaggedLogging::Formatter
+      FORMAT = "%s, [%s#%d] %5s -- %s\n"
+      THREAD_FORMAT = "%s, [%s#%d:%s] %5s -- %s\n"
 
-      def call(severity, timestamp, progname, msg)
-        final_message = if msg.is_a? String
+      def call(severity, time, progname, msg)
+        if (ENV["CROWBAR_THREADS"].to_i || 10) > 1
```
aplanas: If `CROWBAR_THREADS` is not set (empty string or `nil`), or is set to 0, is expected to have 10 threads? Also if this is the puma default, I also expect to remove this magic number from the check.

vuntz: > If CROWBAR_THREADS is not set (empty string or nil), or is set to 0, is expected to have 10 threads?

```
irb(main):001:0> "".to_i
=> 0
irb(main):002:0> nil.to_i
=> 0
```

> Also if this is the puma default, I also expect to remove this magic number from the check.

So I'm using the same magic as in config/puma.rb. I'm not really sure if we can do something else to get the default? Any idea?

That being said, my logic is wrong. Should probably be `(ENV["CROWBAR_THREADS"] || 10).to_i`

aplanas: maybe a more explicit approach? 10 is kind of irrelevant here:

```ruby
threads = ENV["CROWBAR_THREADS"]
if threads.to_s.empty? || threads.to_i > 1?
  ...
else
  ...
end
```

And you are right!

```
irb(main):001:0> 0 || 10
=> 0
```

but this looks like works:

```
irb(main):005:0> 0 | 10
=> 10
```

vuntz: Thanks, I like your approach. Implemented!

aplanas:
@vuntz thanks!!

+1

tboerger:
Seem to make sense... :+1:

PR #262, opened 2016-02-01
Move restore process to background

rsalevsky:
**Not tested!**


tboerger:
In general I don't think that it will work to save the thread reference within the rails cache, and I don't believe anyway that it's a good idea to save others than native types in such a cache method.

So IMHO :-1:

rsalevsky:
Superseded by #269

PR #263, opened 2016-02-01
More complete handling of cluster with remote nodes in webui

vuntz:


jsuchome:
+1 (specifically for renaming Remotes to Remote nodes... )

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -138,7 +142,7 @@ en:
       confirm_delete_proposal: 'Are you sure you wish to delete this proposal?'
       available_clusters: 'Available Clusters'
       available_clusters_hint: 'Here you can see all available clusters.'
-      available_remotes: 'Available Remotes'
+      available_remotes: 'Available Clusters with Remote Nodes'
```
aplanas: nitpick that is clearly for a different PR: we use title and lower case sentences a bit random

vuntz: Actually, if you look at the UI, it is relatively consistent. It's just that the strings don't appear in the same kind of labels. But if you find any such issue, yes, we should fix that! (it's a thing I find very annoying too)

crowbar_framework/app/models/role_object.rb
```diff
@@ -28,24 +28,38 @@ def cluster_roles(roles = RoleObject.all)
     end
   end
 
+  def cluster_remote_roles(roles = RoleObject.all)
+    @cluster_remote_roles ||= begin
```
aplanas: Please, bare with my rusty Ruby, but is this `begin` superfluous here, where there is only one expression?

aplanas:
+1

crowbar_framework/app/models/role_object.rb
vuntz: TBH, I don't know. I think I'm used to the `@variable ||= begin` way, but indeed it might not be required.

PR #265, opened 2016-02-02
rename postgres data directory after data dump

sayalilunkad:
Needed to start new version of postgresql.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -95,6 +95,25 @@ when "dump_openstack_database"
 
 when "db_shutdown"
 
+  # Stopping postgresql service in HA and non-HA
+  if File.exist?("/usr/sbin/crm")
+    bash "Stop postgresql" do
+      code <<-EOF
+        crm stop postgresql
+      EOF
+    end
+  else
+    service "postgresql" do
+      action :stop
+    end
+  end
+
+  bash "Rename data directory of postgres" do
+    code <<-EOF
+      mv /var/lib/pgsql/data /var/lib/pgsql/data.old
+    EOF
+  end
+
   # Stop remaining pacemaker resources
   bash "stop pacemaker resources" do
     code <<-EOF
@@ -108,18 +127,12 @@ when "db_shutdown"
     only_if { ::File.exist?("/usr/sbin/crm") }
   end
 
-  # Stop the database and corosync
-  bash "stop the database" do
-    code <<-EOF
-      for i in /etc/init.d/drbd \
-               /etc/init.d/openais \
-               /etc/init.d/postgresql;
-      do
-        if test -e $i; then
-          $i stop
-        fi
-      done
-    EOF
+  # Stop other services
+  service "drbd" do
+    action :stop
+  end
+  service "openais" do
+    action :stop
   end
 when "done_openstack_shutdown", "wait_for_openstack_shutdown"
   Chef::Log.debug("Nothing to do on this node, waiting for others to finish their work...")


chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
```diff
@@ -104,8 +107,11 @@
           $i stop
         fi
       done
+      rm -rf /var/lib/pgsql/data
```
jsuchome: could you add a comment explaining that this call will be needed for non-HA case?

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
```diff
@@ -83,6 +83,9 @@
   # Stop remaining pacemaker resources
   bash "stop pacemaker resources" do
     code <<-EOF
+      crm stop postgresql
+      # Delete postgres data directory
```
jsuchome: Could you elaborate on why exactly is it needed?

aplanas:
also I think that psql store the db data in `/var/lib/pgsql/X.Y`, so there is not collision when a new version of postgresql call initdb

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
sayalilunkad: The postgres data dir needs to be deleted after we stop postgresql server but before we stop fs-postgresql service.

jsuchome: Yes, but please add this into that comment :-)

sayalilunkad:
@aplanas nope it is always in /var/lib/pgsql/data.  the version directory is only in /usr/lib/postgresqlX.Y/

aplanas:
@sayalilunkad ouch true, is in `/etc/sysconfig/postgresql`, sad

sayalilunkad:
@aplanas that's the reason also why it needs to be deleted :/

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
```diff
@@ -106,6 +110,10 @@
       done
     EOF
   end
+  # Deletes postgres data dir in non-HA scenario
+  directory "/var/lib/pgsql/data" do
+    action :delete
```
rhafer: According to the docs (https://docs.chef.io/resource_directory.html) this doesn't work recursively by default. You'd need to add ```recursive true``` here.

tboerger: Correct!

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
```diff
@@ -83,6 +83,10 @@
   # Stop remaining pacemaker resources
   bash "stop pacemaker resources" do
     code <<-EOF
+      crm stop postgresql
+      # HA: Deletes postgres data directory after stopping postgresql service
+      # but before stopping fs-postgresql.
+      rm -rf /var/lib/pgsql/data
```
mjura: Can we archive `/var/lib/pgsql/data` before removing this?

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
mjura: Can we archive `/var/lib/pgsql/data` before removing this?

aplanas: @mjura If I understand correctly this is done very after the dump is done, but maybe is good idea to rename the directory instead of removing it, and remove it one the restore was done.

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
tboerger: Why do you remove it within that block if you have later on the directory delete?

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
sayalilunkad: @mjura @rhafer I am fine with moving the data to some directory like /var/lib/pgsql/data.old instead of deleting it and then deleting it after successful restore. Any thoughts?

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
sayalilunkad: @tboerger that is only for HA.

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
jsuchome: Can't we meet the problem with space?

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
tboerger: But the other part gets anyway executed?

jsuchome: It does not hurt, does it? There might be some check, for example `not_if { ::File.exist?("/usr/sbin/crm") }` but it's not really necessary

tboerger: It hurts so far that it's weird and makes maintaining the code more difficult.

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
rhafer: @mjura In the UI we tell the user to download the database dump, we could also give him the opportunity to grab a copy of that data dir (it would require some rearrangements though). What would be the reason to keep the data dir on the device as well? One reason against it is disk space issues.

@vuntz I am a somewhat undecided on this, could you please be the tie-breaker ;)

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
rhafer: > But the other part gets anyway executed?

That's true, but in the HA case it's too late then, because the fs-postgresql resource is stopped and /var/lib/pgsql/data/ is umounted already.

That being said I guess the code could be re-arranged to first stop postgresql (using crm in HA and init-script in non-HA) then delete/move the directory and after that stop the rest of the services. I am not quite sure if that will make the code easier though.

chef/cookbooks/crowbar/recipes/crowbar-db-dump.rb
vuntz: > @vuntz I am a somewhat undecided on this, could you please be the tie-breaker ;)

I'm not convinced we would really hit any issue with disk space (do we have any figure that would indicate that?), and it feels safer to keep an old copy in case there's some broken user manipulation. At least, that's what I would do if I were to do things manually ;-)

rhafer: > I'm not convinced we would really hit any issue with disk space (do we have any figure that would indicate that?)

@vuntz I don't have any numbers of course. But we might (very roughly) need 3x the diskspace for the database after the upgrade than we did before. (SQL dump + old database + new database) I wouldn't bet that all systems that are upgraded have that much of a reserve. I guess it's another thing to document then.

vuntz: If you do have serious concerns, I'm fine removing the directory. It's really a question of whether we can afford another safety net; it's not strictly required.

rhafer: > If you do have serious concerns

As said, I am undecided on this. But as we now already have already 2.5 people leaning towards keeping it, let's keep it and just rename the directory after creating the dump and stopping postgresql. In the end we just need to make sure the the new postgresql has a clean data directory when reapplying the database barclamp.
"Better be safe than sorry" I guess.

We can still tell the user to remove it/move it elsewhere if the isn't enough diskspace when we try to reload the dump.

rhafer:
Hrm, sorry for causing you more work, but I just merged: https://github.com/crowbar/crowbar-core/pull/258 so this one needs to be rebased.

rhafer:
+1

aplanas:
+1

rhafer:
Not sure why the codeclimate stuff didn't run here. But every thing else is green. Merging.

PR #267, opened 2016-02-03
Insert missing upgrade steps

rhafer:


crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -156,6 +156,9 @@ def backup
     def nodes
       @current_step = 8
       if request.post?
+        @service_object = CrowbarService.new(Rails.logger)
```
jsuchome: Actually I think you only need to adapt `before_filter :set_service_object` at the beginning

rhafer:
@rsalevsky This are the missing step to trigger the OS upgrade, it's by no means complete, lacks error handling and shouldn't be merged (we just needed something simple for testing). You might want to use it as a base of implementing the rest of step (updating the counters and switch to the last page after the nodes are ready).

crowbar_framework/app/controllers/installer/upgrades_controller.rb
rhafer: Ah, thanks for the hint. I'll update.

jsuchome:
btw, I'd be in favor of merging it anyway, in case anyone will be testing the upgrade workflow...

rhafer:
@MaximilianMeister  any opinion?

MaximilianMeister:
@rhafer do we have a card for that in trello? @rsalevsky is not here until next monday.

rhafer:
@MaximilianMeister https://trello.com/c/Z7q97UwY

rsalevsky:
@rhafer I prefer to put both commands in a `begin ... rescue` block. Like i did it here: [upgrades_controller.rb#L112-L123](https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/controllers/installer/upgrades_controller.rb#L112-L123)

rhafer:
@rsalevsky Done. Do you think we should merge this and you add the counter stuff in a later PR?

rsalevsky:
@rhafer Definitely yes. +1

jsuchome:
+1

aplanas:
+1

PR #269, opened 2016-02-03
Adapt Restore step

MaximilianMeister:
put restore process in the background and show progress in the view


MaximilianMeister:
just saw that the javascript/ajax doesn't work as expected yet. setting to WIP

MaximilianMeister:
here is a little screencast of how it looks currently https://w3.suse.de/~mmeister/upgrade_restore.webm

rhafer:
I am not 100% sure it it's related to the changes in this pull request (I am pretty certain it is), but now any errors that happen during the restore are simply not visible any more (not even in the logs).

E.g. I had the proposal validation fail for nova after running the SchemaMigration  (because of: https://github.com/crowbar/crowbar-openstack/pull/255) and the restore just didn't complete, without indication of an error. Not even in the logs.

MaximilianMeister:
If it is, then it is probably due to the `Thread`. but no logs really sounds weird...

btw I have restored until cinder (no HA) without any errors

MaximilianMeister:
could it be that we don't have the `production.out` until after the installation?

rhafer:
I don't think we have it at all anymore. stdout of puma is now ending up in the journal (or /var/log/messages). But those errormessage should go to production.log, shoudn't they? The last thing I see in the journal is btw:

```Feb 04 08:34:37 crowbar puma[30472]: D, [2016-02-04T08:34:37.246206 #30504] DEBUG -- : Restoring proposal nova-default.json```
But nothing about the failed migration.

**Comment by MaximilianMeister** [comment]
maybe there is a problem with the inherited logger from `Crowbar::Backup::Base`. Could be that STDOUT is somehow detached in a Thread, can you try with `Rails.logger` in restore.rb?

**Comment by rhafer** [comment]
Using Rails.logger cause the stuff to endup in production.log, but still no error messages.

**Comment by MaximilianMeister** [comment]
> Using Rails.logger cause the stuff to endup in production.log, but still no error messages.

upgrade.rb needs to be adapted, we dont have any error handling yet, just for a few selected files from the backup

File: crowbar_framework/lib/crowbar/backup/restore.rb [hunk 1]
```diff
@@ -67,19 +152,33 @@ def restore_chef
...
+                    msg: msg
+                  }
+                end
               else
                 record.save
               end
             end
           end
+
+          @status[:restore_chef] = { status: :ok, msg: "" }
```
rhafer: Doesn't this overwrite the status that is set when an error happens during Schema migration?

rhafer:
Something is still borked with the error handling. If a Schema migration failed I just get this in the UI:

```Restore failed. Please check the log at /var/log/crowbar/production.log and /var/log/crowbar/install.log ```

but it's supposed to give the full error message, isn't it? (At least the error shows up in the log now)

**Comment by rhafer** [comment]
If that is somehow more difficult to fix I am fine to address the in a later pull request.

File: crowbar_framework/lib/crowbar/backup/restore.rb (hunk 1 above)
Comment (MaximilianMeister) [inline crowbar_framework/lib/crowbar/backup/restore.rb]: right, thanks

**Comment by MaximilianMeister** [comment]
> but it's supposed to give the full error message, isn't it?

no it is not supposed to. it is not that trivial to pass the exact error message to the controller due to that it runs in the background... I would vote for fixing it later as well (due to all the remaining important tasks)

File: crowbar_framework/lib/crowbar/backup/restore.rb [hunk 2]
```diff
@@ -67,22 +156,36 @@ def restore_chef
...
+
+          @status[:restore_chef] ||= { status: :ok, msg: "" }
         rescue Errno::ECONNREFUSED
           raise Crowbar::Error::ChefOffline.new
         rescue Net::HTTPServerException
           raise "Restore failed"
         end
 
-        Rails.logger.info("Re-runnig chef-client locally to apply changes from imported proposals")
+        Rails.logger.info("Re-running chef-client locally to apply changes from imported proposals")
```
rhafer: Oops sorry for that.

rhafer:
+1

MaximilianMeister:
gating failed in build due to patching restore.rb

will make a rebase

jdsn:
Even after rebase the restore.rb diff does not apply on my just installed crowbarnode (develcloud6) (maybe I am missing some changes from Staging).
But after applying manually the upload was successful and the restore was triggered successfully - anyhow the restore did not succeed.

@MaximilianMeister I'll keep  the machine running, lets look into this on Monday.

MaximilianMeister:
seems the patch applied, as gating passed

@jdsn I will have a look on monday what failed, but @rhafer had also a failure at nova migration, but likely nothing that was introduced by this code.

rsalevsky:
+1

jdsn:
@MaximilianMeister sure the gating passed, but the gating job did not test the new functionality, it just does not break anything else :)

PR #271, opened 2016-02-04
After DB shutdown, do not allow any more actions with crowbar-upgrade role

jsuchome:
Set the upgrade step to the poing where no further action is done
even when the upgrade recipes are accidentally executed

So, this is the simple solution (er, workaround) for https://trello.com/c/vfykMmnv/787-upgrade-remove-crowbar-upgrade-roles-before-the-upgrade-of-the-nodes


rhafer:
I think it useful to merge this until we have a real fix (simplifies testing/fixing the subsequent steps)
+1

MaximilianMeister:
+1

PR #272, opened 2016-02-04
crowbar: Disable check for interface with IPv6 address for our hostname

vuntz:
This doesn't matter too much at the moment, as we don't really depend on
IPv6 anywhere, and it's possible that people would have broken DNS
records or broken network config for IPv6 (doc team just hit that, for
instance).

cc @MaximilianMeister

### DIFF
--- a/crowbar_framework/lib/crowbar/checks/network.rb
+++ b/crowbar_framework/lib/crowbar/checks/network.rb
@@ -59,9 +59,10 @@ module Crowbar
         unless ipv4_configured || ipv4_addrs.empty?
           return false
         end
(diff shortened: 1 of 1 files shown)


vuntz:
Btw, it's really bad that this just makes crowbar fail to start; an error message in the web interface would be really much better...

dirkmueller:
+1

MaximilianMeister:
but if we don't depend on it at all, why not get rid of the check? 

+1 either way

dirkmueller:
@MaximilianMeister it might be useful in the future with ipv6 enablement upcoming. I don't particularly care either way..

PR #274, opened 2016-02-04
Revert "Set CROWBAR_THREADS to 10 to allow concurrent access (bsc#952…

dirkmueller:
…171)"

Apparently the threads make triggering of the sql transactions
deadlocks much more likely. Since PUMA doesn't do real threads
but user space threading, I don't think this is a good idea anyway..

This reverts commit fc31aaa9782e179a3c66fd0d7476a2b66290f0c8.

### DIFF
--- a/chef/cookbooks/crowbar/files/default/crowbar.service
+++ b/chef/cookbooks/crowbar/files/default/crowbar.service
@@ -14,7 +14,7 @@ TimeoutStopSec=10
 WorkingDirectory=/opt/dell/crowbar_framework
 
 Environment="CROWBAR_ENV=production"
-Environment="CROWBAR_THREADS=10"
+Environment="CROWBAR_THREADS=1"
 Environment="CROWBAR_WORKERS=5"
 Environment="CROWBAR_LISTEN=127.0.0.1"
 Environment="CROWBAR_PORT=3000"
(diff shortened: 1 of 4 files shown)


aplanas:
I still believe that there is something else. I am testing reverting https://github.com/crowbar/crowbar-core/commit/38672baeee11084a9e281fd94756df5d2632121b in a local deployment, but for some reason I am no lucky enough to reproduce the original issue

rhafer:
Just to repeat once more here. This is going to cause blocking for xen and backup/restore so I am against merging it currently.

tboerger:
As mentioned on IRC I think we have to increase the number of workers, even if that means we should consume more memory.

rsalevsky:
This is definitely not the reason why it breaks otherwise it makes no sense that it worked for 1 1/2 months. I think what we see now is that we changed / intoduced something which is not worker / thread save. We should search for this instead of just reverting the PR blindly.

vuntz:
> it makes no sense that it worked for 1 1/2 months

do we know that for sure?

vuntz:
https://github.com/crowbar/crowbar-core/pull/285 makes things much better.

vuntz:
Do not merge for M11.

dirkmueller:
not needed anymore with #285

PR #278, opened 2016-02-04
Upgrade additions

tboerger:
As requested by @kwwii I have made some additions and restructuring to the upgrade UI. As you can see there is a single missing part for async node upgrade triggering, this will be done my @MaximilianMeister.


crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -153,20 +156,23 @@ def backup
...
+
       if request.post?
         respond_to do |format|
           begin
-            @service_object.disable_non_core_proposals
-            @service_object.prepare_nodes_for_os_upgrade
+            # TODO(must): Do the following actions async in background
+            # @service_object.disable_non_core_proposals
+            # @service_object.prepare_nodes_for_os_upgrade
+
```
rhafer: Why, async? disable_non_core_proposals and prepare_nodes_for_os_upgrade won't take too long I think. ```prepare_nodes_for_os_upgrade``` in not blocking while performing the upgrade. It will block until the nodes succeed (or failed) calling the reboot command.

In any case, please leave those steps enabled here even if it's not their final location (and add ```infos``` to ```set_service_object```)

rhafer:
BTW, based on this PR I've replaced a lot of the UI text with something that should fit better. Please have a look at: https://github.com/rhafer/crowbar-core/commit/a30d6fc4e6b7dd85dc9aa795e487ac605dc6fa20

tboerger:
I have applied the requested stuff from @kwwii. 

> This is not quite correct. If there is no Pacemaker proposal it means that we don't need the repository at all. I guess that's why it returned nil in the past. The same it true for the ceph check.

Than @rsalevsky should use different method names as a method with a `?` suffix have to ALWAYS return a boolean value by ruby convention.

I'm out of this PR as I will only be available on monday, otherwise I'm on vacation.

crowbar_framework/app/controllers/installer/upgrades_controller.rb
tboerger: > Why, async? disable_non_core_proposals and prepare_nodes_for_os_upgrade won't take too long I think. prepare_nodes_for_os_upgrade in not blocking while performing the upgrade. It will block until the nodes succeed (or failed) calling the reboot command.
> 
> In any case, please leave those steps enabled here even if it's not their final location (and add infos to set_service_object)

This should be async because the following step will do AJAX requests to retrieve the currently still left to update nodes.

rhafer: > This should be async because the following step will do AJAX requests to retrieve the currently still left to update nodes.

Right, but the real update just starts after "reboot" was called and by that prepare_nodes_for_os_upgrade() is already finished. Then we load the next page which does the AJAX magic, because that's where it takes really long (AutoYaST ...).

rhafer:
> Than @rsalevsky should use different method names as a method with a ? suffix have to ALWAYS return a boolean value by ruby convention.

That might be true. Though it's probably not the best reason to break the functionality here. :) Don't get me wrong, I am all for following the conventions and fixing that, but then please in a proper way.

> I'm out of this PR as I will only be available on monday, otherwise I'm on vacation.

That's for letting us know. Have a nice vacation.

@MaximilianMeister or @rsalevsky  are you able to pick this up?

MaximilianMeister:
> @MaximilianMeister or @rsalevsky are you able to pick this up?

@rhafer Can we get https://github.com/crowbar/crowbar-core/pull/269 merged first? It touches some areas, that are used here as well, and testing would be better if all steps before upgrade (and info) would be more or less done

rhafer:
> Can we get #269 merged first? It touches some areas, that are used here as well, and testing would be better if all steps before upgrade (and info) would be more or less done

Yeah, sure. Makes sense.

rsalevsky:
Superseded by #298

PR #279, opened 2016-02-04
Switch from puma to passenger

vuntz:
Doing some preliminary tests with passenger. So far, I'm not able to reproduce the SQLite3::BusyException exception...

This patch is not fully tested (I configured things manually for my testing).

We should still discuss whether the switch makes sense, of course.


tboerger:
And who will maintain this monster? Maintaining passenger is a pain in the ass. Beside that we need to drop it with cloud 7 again when we introduce websockets as they are unsupported by the free passenger version. So I'm really not convinced...

vuntz:
> And who will maintain this monster?

It's already maintained internally, so it comes for free.

vuntz:
> Beside that we need to drop it with cloud 7 again when we introduce websockets as they are unsupported by the free passenger version.

"WebSockets work great on both the open source variant of Phusion Passenger, as well as on Phusion Passenger Enterprise. "

https://github.com/phusion/passenger-ruby-websocket-demo

vuntz:
And again, this needs to be discussed, as mentioned in the PR. I only opened the PR to allow people to test things.

rsalevsky:
I have the same experience with passenger as @tboerger. I have two projects running on passenger and both makes just problems. One is terrible slow while parallel requests and the other one throws random error messages after some time.

tboerger:
Ok, so they changed the websocket thing. But I still have bad experience related to performance and stability like @rsalevsky have

crowbar_framework/lib/crowbar/logger/formatter.rb
```diff
@@ -24,7 +24,7 @@ class Formatter < ::Logger::Formatter
       def call(severity, time, progname, msg)
         threads = ENV["CROWBAR_THREADS"]
         # if env var is not set, default is multi-threaded
-        if threads.to_s.empty? || threads.to_i > 1
+        if !threads.to_s.empty? || threads.to_i > 1
```
tboerger: `if threads.to_s.present? || threads.to_i > 1`

tboerger:
A general question, why do you stick with a standalone passenger instead of the apache module?

vuntz:
> A general question, why do you stick with a standalone passenger instead of the apache module?

It's mentioned in the commit message: because we have cookbooks that can reload or restart apache, using the module is actually pretty bad as the cookbooks can kill the rails app. And this actually happens during the installer... So we can't do that.

vuntz:
This was an experiment; unless we decide to really do it, let's close this.

PR #281, opened 2016-02-05
Retry transactions when SQLite3 is busy

aplanas:
After Balazs patch, all transactions are IMMEDIATE, adquiring
both read and write locks.  From time to time we can see
transactions that try to get both locks for more that the timeout.

This patch retry 5 times the transaction before giving up.

Fixes bsc#963738

### DIFF
--- a/crowbar_framework/config/initializers/sqlite_transaction_monkey_patch.rb
+++ b/crowbar_framework/config/initializers/sqlite_transaction_monkey_patch.rb
@@ -24,7 +24,16 @@ module ActiveRecord
   module ConnectionAdapters
     class SQLite3Adapter < AbstractAdapter
       def begin_db_transaction
-        log("begin transaction",nil) { @connection.transaction(:immediate) }
+        log("begin transaction", nil) do
+          5.times do
+            begin
+              @connection.transaction(:immediate)
+              break
+            rescue SQLite3::BusyException => e
+              logger.info "begin transaction fails #{e}"
+            end
+          end
+        end
       end
     end
   end


aplanas:
I try to use `retry`, but I always get `LocalJumpError: no block given (yield)` error when the exception is captured.

This is still under testing.

vuntz:
@aplanas I think I tried a patch like this some time ago and didn't manage to get it to work :/

aplanas:
@vuntz actually the last incarnation of this patch worked well in my tests (detected 4 blocks and each one was retried with success) Now I am using a proper `retry` instead of `times`

vuntz:
@aplanas https://github.com/vuntz/crowbar-core/commit/cb3e0cb153c25b1d679b552634c4fdfe23666a11 was my attempt that was still hitting the bug.

aplanas:
yep, the same workaround

aplanas:
Thanks to @vuntz, I put this patch under a more stressful test, where the tests are transitioned to the discovering state, the proposals reset and the nodes put again to the discovered state in an endless loop.

This command shows that sometimes the transactions fails, but are recovered
  # grep transaction /var/log/crowbar/production.log

This commands shows that there are not SQLite3::BusyException
  # grep SQLite3 /var/log/crowbar/production.log

Every time the 8 nodes are discovered
(c16 is available for testing)

crowbar_framework/config/initializers/sqlite_transaction_monkey_patch.rb
```diff
@@ -24,7 +24,16 @@ module ActiveRecord
...
     class SQLite3Adapter < AbstractAdapter
       def begin_db_transaction
-        log("begin transaction",nil) { @connection.transaction(:immediate) }
+        log("begin transaction", nil) do
+          5.times do
+            begin
+              @connection.transaction(:immediate)
+              break
+            rescue SQLite3::BusyException => e
+              logger.info "begin transaction fails #{e}"
```
vuntz: Can you add a `:` before the exception message?

vuntz:
> This commands shows that there are not SQLite3::BusyException
> # grep SQLite3 /var/log/crowbar/production.log

That's because your patch is hiding it  completely: after the 5th attempt, the exception is silently ignored :/

aplanas:
> That's because your patch is hiding it completely: after the 5th attempt, the exception is silently ignored :/

Very true. I also tested another one that `raise e` at the end and print the counter. In those test only retry one ... I will update.

vuntz:
But it does indeed work better than when I tried with my old patch. Weird :/

Hrm, it could be because here we don't have the 8 nodes connecting to the server every 10 or 15 seconds (like sleshammer does). Would be the next thing to test, I guess.

vuntz:
Anyway, it improves things at least a bit, so +1.

vuntz:
> But it does indeed work better than when I tried with my old patch. Weird :/

Ah, I think the move to WAL might explain it. I didn't have both changes when I tried.

aplanas:
@vuntz I do not have vpn access and I can't update the patch.

Because https://github.com/vuntz/crowbar-core/commit/cb3e0cb153c25b1d679b552634c4fdfe23666a11 is using the same approach, maybe we can re-test the old one and if behaves like this, close it in favor of yours (so no need to wait until Monday)

aplanas:
> Ah, I think the move to WAL
This is what I think, yes.

aplanas:
Also to increase the speed of UPDATE / INSERT we can use some configurations that help WAL like

```
PRAGMA synchronous=OFF
PRAGMA journal_mode=MEMORY
```

Or even disable the journal

vuntz:
I added this on top of my previous stress test:
```sh
while true; do
  for i in `seq 1 10`; do
    crowbar machines list &
  done
  wait
done
```

I was able to trigger the sqlite exception ~~once~~ ~~twice~~ three times. So it doesn't fully solve it. A mistake in my previous patch was that the sleep was constant (0.1); adding some random to this, it might help a bit (to avoid threads retrying together).

vuntz:
Btw, that's what I use to watch if things go wrong: `tail -f /var/log/crowbar/production.log | grep -E "ARTOI: updating proposal|begin transaction|commit transaction|BusyException"`

aplanas:
Can you try also with disabling synchronous write and putting the journal in memory?

vuntz:
Just for the record, that's what I'm testing on c16 right now: https://github.com/crowbar/crowbar-core/compare/master...vuntz:retry-on-sqlite-busy?expand=1

It's slightly better, but maybe by pure luck: I reached the fourth retry twice.

> Can you try also with disabling synchronous write and putting the journal in memory?

Unfortunately, these settings are dangerous for data integrity (in case of system crash); I don't think we would want to ship with them :/

vuntz:
I tried the stress test again with the current code (no workaround) to see how bad it was; I hit the exception four times in less than two minutes.

I ran with the patch from my branch for much longer in total, and in the end only hit the issue once so far. I did see some retries (and I think we don't see the logs from the session bits unless there's an exception, so there might be more retries hidden).

So all in all, I think it's a good improvement, and a patch like this is definitely wanted as long as we stay with sqlite.

aplanas:
Yep, IMHO the real solution is replace SQLite, but meanwhile this can be enough. @vuntz lets close this in favor of https://github.com/crowbar/crowbar-core/compare/master...vuntz:retry-on-sqlite-busy?expand=1

vuntz:
Superseded by https://github.com/crowbar/crowbar-core/pull/285

PR #283, opened 2016-02-05
Delete pacemaker resources after stopping.

jsuchome:
They could be re-created once the barclamps are applied again.

1. Maybe it will not be needed, let's see...
2. Maybe some cibadmin command could do it better: cibadmin -E ?

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -31,6 +31,11 @@ when "openstack_shutdown"
 
   include_recipe "crowbar::stop-services-before-upgrade"
 
+  execute "delete pacemaker resources in non-DB cluster" do
+    command "cibadmin -E -f"
+    only_if { ::File.exist?("/usr/sbin/cibadmin") }
+  end
+
   # Stop DRBD and corosync.
   # (Note that this node is not running database)
   service "drbd" do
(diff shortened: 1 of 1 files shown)


jsuchome:
This really does not seem to work. When I try to delete everything, I get the complain that drbd is still running (although it was stopped before). `cibadmin -E -f` might really be better option

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -31,6 +31,11 @@
 
   include_recipe "crowbar::stop-services-before-upgrade"
 
+  execute "delete pacemaker resources in non-DB cluster" do
+    command "cibadmin -E -f"
+    only_if { ::File.exist?("/usr/sbin/crm") }
```
vuntz: Maybe you should check for cibadmin, not crm?

Note that this will remove not just resources and constraints, but also nodes. But I guess that's okay...

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -102,6 +107,7 @@
           crm resource stop $resource
         done
       done
+      cibadmin -E -f
```
vuntz: Only for readability: I'd use an `execute` chef resource like what you did above. This makes it easier to compare the two code paths.

jsuchome:
Finally, after some testing, this seem to work well

vuntz:
+1

aplanas:
+1

PR #285, opened 2016-02-06
crowbar: Retry transaction when sqlite is busy (bsc#963738)

vuntz:
This is of course not a perfect solution, and can still fail after the
retries.

I still believe that there's a deeper issue because every time this
happens, it seems that two threads (not necessarily in the same worker,
though) are entering a deadlock while acquiring the sqlite lock while
this clearly should not happen (one should acquire the lock, while the
other would be blocked on acquiring).

But it makes the original bug extremely hard to reproduce now with the
following stress test:

 - having 8 nodes being rediscovered in a loop (this requires removing
   the nodes from the core proposals, and transitioning them from
   discovering to discovered)
 - having 10 requests to crowbar listing the nodes (and therefore
   creating sessions in the database) in a loop

https://bugzilla.suse.com/show_bug.cgi?id=963738

Co-Authored-By: Alberto Planas \<aplanas@gmail.com\>

### DIFF
--- a/crowbar_framework/config/initializers/sqlite_transaction_monkey_patch.rb
+++ b/crowbar_framework/config/initializers/sqlite_transaction_monkey_patch.rb
@@ -24,7 +24,21 @@ module ActiveRecord
   module ConnectionAdapters
     class SQLite3Adapter < AbstractAdapter
       def begin_db_transaction
-        log("begin transaction",nil) { @connection.transaction(:immediate) }
+        log("begin transaction", nil) do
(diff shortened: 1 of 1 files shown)


vuntz:
For the record: please see the long discussion in https://github.com/crowbar/crowbar-core/issues/281 (more or less the same patch, but @aplanas was not able to update the patch during the week-end).

aplanas:
+1

nkrinner:
+1

PR #286, opened 2016-02-08
crowbar: Fix typo in log

vuntz:
Also, not sure why we "re-run" instead of simply "run" ;-)


vuntz:
Ah, it's fixed in https://github.com/crowbar/crowbar-core/pull/269. Let's not create a useless conflict with that PR.

PR #287, opened 2016-02-08
dont use dots for public-IP DNS names

bmwiedemann:
dots make it much harder to use wildcard SSL certs,
because those can only have a \* at the beginning
and dont support _._.domain


vuntz:
-1: this is what is done everywhere in crowbar for all networks. So if we want such a change, I expect some more changes in the network cookbook/barclamp to happen.

And in general, it's too late to do it for this release.

PR #289, opened 2016-02-08
Check if database proposal exists in CrowbarService::dump_openstack_database

mjura:
Check if database proposal exists in case,
when somebody is trying to upgrade undeployed nodes
or only with Ceph cluster

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -105,9 +105,8 @@ class CrowbarService < ServiceObject
 
     # Check if database proposal exists in case when somebody is
     # trying to upgrade undeployed nodes or only with Ceph cluster
-    db_proposal = Proposal.where(barclamp: "database").first
-    msg = "There does not seem to exist any node running the database."
-    raise msg if db_proposal && db_nodes.empty?
+    db_proposal = Proposal.find_by(barclamp: "database")
+    raise I18n.t("model.service.database_missing") if db_proposal && db_nodes.empty?
 
     # This proposal could return some error if there's not enough space for DB dump
     # Controller must show the error and be able to call the function again once the problem
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -103,7 +103,11 @@ class CrowbarService < ServiceObject
       node.save
     end
 
-    raise "There does not seem to exist any node running the database." if db_nodes.empty?
+    # Check if database proposal exists in case when somebody is
+    # trying to upgrade undeployed nodes or only with Ceph cluster
+    db_proposal = Proposal.where(barclamp: "database").first
+    msg = "There does not seem to exist any node running the database."
+    raise msg if db_proposal && db_nodes.empty?
 
     # This proposal could return some error if there's not enough space for DB dump
     # Controller must show the error and be able to call the function again once the problem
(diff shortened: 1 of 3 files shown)


jsuchome:
Wouldn't it be better to do the check for proposal at the beginning and skip the search for db nodes?

mjura:
> Wouldn't it be better to do the check for proposal at the beginning and skip the search for db nodes?

Unfortunately, no because we skip 
```
      node["crowbar_wall"]["crowbar_upgrade_step"] = step
      node.save
```
which is needed to move forward upgrade process.

jsuchome:
Hm, OK, I think it could stay this way.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -103,7 +103,10 @@ def dump_openstack_database
       node.save
     end
 
-    raise "There does not seem to exist any node running the database." if db_nodes.empty?
+    # Check if database proposal exists in case when somebody is trying to upgrade undeployed nodes
+    # or only with Ceph cluster
+    db_proposal = Proposal.where(barclamp: "database", name: "default").first
```
vuntz: Can you move that line outside of the loop? Also, please don't look for a proposal with name `default` as the name could be different; just look for any database proposal.

mjura:
Was tested, works fine. Please comment

crowbar_framework/app/views/installer/upgrades/backup.html.haml
```diff
@@ -17,8 +17,11 @@
               = t(".description")
 
             %p
-              = database_node.name
-              = database_backup_path
+              - if upgrade_database_node
+                = upgrade_database_node.name
+                = upgrade_database_backup
+              - else
+                No database to dump
```
jsuchome: Could you make full sentence out of this?

vuntz:
@mjura can you rebase?

mjura:
@vuntz : It was rebased

mjura:
Ready to go

jsuchome:
+1c (the texts should be translated, but I'm not sure if we need to go through them now...)

vuntz:
If the text is not made localizable in this PR, then please create a followup task somewhere to make sure it will be.

mjura:
I've added translation, but I've to test

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -103,7 +103,11 @@ def dump_openstack_database
       node.save
     end
 
-    raise "There does not seem to exist any node running the database." if db_nodes.empty?
+    # Check if database proposal exists in case when somebody is
+    # trying to upgrade undeployed nodes or only with Ceph cluster
+    db_proposal = Proposal.where(barclamp: "database").first
```
MaximilianMeister: MINOR: you could also use 
`db_proposal = Proposal.find_by(barclamp: "database")` 
or
`db_proposal = Proposal.find_by(barclamp: "database", name: "default")`

rsalevsky:
@mjura Can you integrate #308 and close my PR?

crowbar_framework/app/models/crowbar_service.rb
vuntz: (don't do the second suggestion, first one is fine, I guess)

mjura: Ok, I've added first one

mjura:
@rsalevsky : I've cherry-picked your changed and rebased it

vuntz:
+1

rsalevsky:
+1

PR #291, opened 2016-02-08
Return true if repos are present (or not needed)

jsuchome:
cc @rsalevsky @rhafer

### DIFF
--- a/crowbar_framework/app/helpers/installer/upgrades_helper.rb
+++ b/crowbar_framework/app/helpers/installer/upgrades_helper.rb
@@ -87,6 +87,8 @@ module Installer
       if upgrade_ceph_repo_needed?
         return false unless upgrade_ceph_repo?
       end
+
+      true
     end
 
     def restored?


jsuchome:
I'm not sure if 2nd commit is correct way to do it. However, it also should not be right to require user to **Recheck** when everything is already correctly in place.

rsalevsky:
@jsuchome I would say we should postpone the second commit. There was no decision yet if we should skip steps or just say that everything is fine.

jsuchome:
Well, either way, currently the page does not say "that everything is fine", but tells me that HA repos are missing.

rsalevsky:
Let's just merge the first commit and I will implement something for the second.

jsuchome:
OK, deleted https://github.com/jsuchome/crowbar-core/commit/8986c79efee52b0e6fb27e26fca0e5091421df2c

rhafer:
+1

rsalevsky:
+1

PR #292, opened 2016-02-09
Various improvements to the exported files controller

vuntz:


rsalevsky:
+1

crowbar_framework/app/controllers/support_controller.rb
```diff
@@ -103,22 +102,29 @@ def export_supportconfig
...
-      RoleObject.all.each { |r| r.export }
-      Proposal.all.each { |p| p.export }
-
       filename = "crowbar-chef-#{Time.now.strftime("%Y%m%d-%H%M%S")}.tgz"
 
-      pid = Process.fork do
-        exports = Dir.glob(Rails.root.join("db", "*.json").to_s)
-        cmd     = ["tar", "-czf", Rails.root.join("tmp", filename).to_s, *exports]
+      tmpdir = Dir.mktmpdir
+      tmpdirpath = Pathname.new(tmpdir)
```
MaximilianMeister: MINOR: i would call it `tmpdir_path` for consistency

or as an alternative just use `tmpdir = Dir.mktmpdir(Pathname.new(tmpdir))` and reference it with `.to_s` where needed

crowbar_framework/config/routes.rb
```diff
@@ -62,13 +62,11 @@
 
   #support paths
   get "utils(.:format)", controller: "support", action: "index", as: "utils"
-  get "utils/files/:id(.:format)", controller: "support", action: "destroy", constraints: { id: /[^\/]+/ }, as: "utils_files"
+  delete "utils/files/:id(.:format)", controller: "support", action: "destroy", constraints: { id: /[^\/]+/ }, as: "utils_files"
```
MaximilianMeister: does this require some change in the cli?

MaximilianMeister:
just unsure about the cli part (don't know if we even offer it)
other than that it looks good

+1

crowbar_framework/config/routes.rb
tboerger: No, it's not part of the CLI yet.

PR #293, opened 2016-02-09
Reorganize Utilities menu a bit

vuntz:


rsalevsky:
+1

MaximilianMeister:
+1 (but i think you will need to update the branding branch right after the merge)

vuntz:
> (but i think you will need to update the branding branch right after the merge)

It doesn't conflict, so we're good it seems.

crowbar_framework/config/navigation.rb
```diff
@@ -33,9 +33,11 @@
     level1.item :barclamps, t("nav.barclamps.title"), barclamp_modules_path do |level2|
       level2.item :all, t("nav.barclamps.all"), barclamp_modules_path
       level2.item :crowbar, t("nav.barclamps.crowbar"), index_barclamp_path(controller: "crowbar")
-      level2.item :queue, t("nav.barclamps.queue"), deployment_queue_path
     end
     level1.item :utils, t("nav.utils.title"), utils_path do |level2|
+      level2.item :queue, t("nav.utils.queue"), deployment_queue_path
+      level2.item :queue, t("nav.utils.repositories"), repositories_path
+      level2.item :queue, t("nav.utils.backup"), backups_path
```
rhafer: @vuntz  Copy and paste error? That ```:queue``` an all lines?

tboerger: This gets overwritten with barclamp_install but it should be nice to have that fixed.

rhafer: It was fixed in: https://github.com/crowbar/crowbar-core/pull/303

tboerger: Thanks

PR #294, opened 2016-02-09
Do not stop openvswitch-switch before upgrade

jsuchome:
Stopping openvswitch  before upgrade causes the nodes to become unreachable and upgrade gets stuck. This patch helps to avoid it.


sayalilunkad:
+1 works for me.

vuntz:
Do I guess right that without this, the node is no more reachable? (might be good to mention it in the commit message)

jsuchome:
Yes, that might be the problem, although I have not met this before (I usualy didn't have neutron config). So I'd hope @sayalilunkad or @rhafer would offer change to my commit message...

rhafer:
@jsuchome how about this:

Keep openvswitch running during the upgrade preparation
Otherwise the nodes might loose connectivity to the admin node

jsuchome:
Thanks, messsage updated.

vuntz:
+1

aplanas:
+1

PR #295, opened 2016-02-09
Block the UI when restoring _not_ from upgrade

MaximilianMeister:
as the restore now happens in the background (see https://github.com/crowbar/crowbar-core/pull/269), a restore from /utils/backup comes back immediately, which is not really wanted.

restore_ongoing
restore_failed

btw. if we would have restored in the background here as well, the UI would have to be adapted to show the progress, maybe even create a new view (that we also link to from the installer startpage) to show progress. This is the easy solution now due to the remaining open tasks.

### DIFF
--- a/crowbar_framework/app/assets/javascripts/application.js
+++ b/crowbar_framework/app/assets/javascripts/application.js
@@ -1,5 +1,22 @@
 //= require_self
 
+function setBlockUI(events, selector) {
+  selector.live(events, function(event) {
+    $.blockUI({
+      css: {
+        border: 'none',
+        padding: '15px',
+        backgroundColor: '#000',
+        '-webkit-border-radius': '10px',
+        '-moz-border-radius': '10px',
+        opacity: .5,
+        color: '#fff'
+      },
+      message: $(event.target).data('blockui')
+    });
+  });
+}
+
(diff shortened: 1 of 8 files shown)


crowbar_framework/app/assets/javascripts/application.js
```diff
@@ -78,7 +78,7 @@ jQuery(document).ready(function($) {
     ]
   });
 
-  $('[data-blockui]').live('submit', function(event) {
+  $('[data-blockui]').live('click submit', function(event) {
```
rsalevsky: This will cause problems see https://github.com/crowbar/barclamp-crowbar/pull/1380

MaximilianMeister:
> This will cause problems see crowbar/barclamp-crowbar#1380

I have changed the javascript now a bit to set the blockui for different events and selectors

vwallfahrer:
Tested it for Cloud5 and it worked.

vuntz:
Not an expert, but looks good to me.

rsalevsky:
+1

aplanas:
+1

tboerger:
https://github.com/crowbar/barclamp-crowbar/pull/1386#discussion_r52871673

PR #296, opened 2016-02-09
Fix deployment queue (bsc#965610 + cleanups)

vuntz:


crowbar_framework/lib/crowbar/deployment_queue.rb
```diff
@@ -17,25 +17,23 @@ module Crowbar
...
+      logger.debug("queue proposal: enter for #{bc}:#{inst}")
       delay = []
       pre_cached_nodes = {}
       begin
         lock = acquire_lock("queue")
-        if @proposal_queue.barclamp == bc && @proposal_queue.name == inst
-          preexisting_queued_item = @proposal_queue.properties
-        end unless @proposal_queue.nil?
+
+        queued_proposal = ProposalQueue.find_by(barclamp: bc, name: inst)
```
MaximilianMeister: would it be possible to have 2 queues for one barclamp?

vuntz: For one proposal in a barclamp, no.

For one barclamp, yes: two different proposals.

vuntz:
Rebased only to re-trigger travis which was blocked...

bmwiedemann:
+1 - did you give it some manual testing? e.g. in qa_crowbarsetup proposal step we could leave out the commits and thus generate a queue easily

aplanas:
+1

vuntz:
> did you give it some manual testing?

Yes

 > e.g. in qa_crowbarsetup proposal step we could leave out the commits and thus generate a queue easily

I'm unsure we want to do that, though; you have to keep in mind that right now, anything in the queue is applied as if the first proposal was still applying. So we would possibly hit some timeout. But maybe something to play with, still...

crowbar_framework/lib/crowbar/deployment_queue.rb
```diff
@@ -197,25 +193,21 @@ def process_queue
...
-    def commit_proposals(list)
-      list.map do |item|
-        logger.debug("process queue: item to do: #{item.inspect}")
-        bc = item[:barclamp]
-        inst = item[:inst]
+    def commit_proposal(bc, inst)
+      logger.debug("process queue: committing item: #{bc}:#{inst}")
 
-        service = eval("#{bc.camelize}Service.new logger")
+      service = eval("#{bc.camelize}Service.new logger")
```
tboerger: `eval` is simply evil and simply poor code... And such a simple case can be done so easy without using `eval`.

```ruby
"#{bc}_service".classify.constantize.new logger
```

crowbar_framework/lib/crowbar/deployment_queue.rb
```diff
@@ -197,25 +193,21 @@ def process_queue
...
+      # This will call apply_role and chef-client.
+      # Params: (inst, in_queue, validate_after_save)
+      status, message = service.proposal_commit(inst, true, false)
 
-        logger.debug("process queue: item #{item.inspect}: results #{message.inspect}")
+      logger.debug("process queue: committed item #{bc}:#{inst}: results = #{message.inspect}")
 
-        # FIXME: this is perhaps no longer needed
-        $htdigest_reload = true
+      # FIXME: this is perhaps no longer needed
```
tboerger: This is indeed not needed anymore as the thdigest happens inside of apache and not anymore in Crowbar.

crowbar_framework/lib/crowbar/deployment_queue.rb
vuntz: Sure; but as you can see it was not introduced in this PR. Do you want to submit a PR for this? Or shall I do it?

crowbar_framework/lib/crowbar/deployment_queue.rb
vuntz: Same, not introduced in this PR. Do you want to submit a PR for this? Or shall I do it?

crowbar_framework/lib/crowbar/deployment_queue.rb
tboerger: Feel free to do it, got to catch up with the pull requests and emails.

PR #297, opened 2016-02-09
crowbar: Make "Default Filesystem" option only visible on SUSE platforms

vuntz:
This is only respected by the autoyast profile.

Also, disable the select tag if the node is allocated because, well,
it's too late to change in that case.


toabctl:
if that works, +1

jsuchome:
+1

vuntz:
> if that works

it does :-)

aplanas:
+1

PR #299, opened 2016-02-09
bsc#949273 stop green led when back in ready state

MaximilianMeister:
https://bugzilla.suse.com/show_bug.cgi?id=949273


crowbar_framework/vendor/assets/javascripts/jquery/ledUpdate.js
```diff
@@ -268,6 +268,12 @@
 
   LedUpdate.prototype.destroy = function() {
     $('.led.unready, .led.in_process, .led.spin').destroy();
+    // workaround https://bugzilla.suse.com/show_bug.cgi?id=949273
+    // to stop the green led from moving after node is in ready state
+    $('.led.ready').sprite({
```
vuntz: Should we also do that for the others (.led.failed, .led.pending, .led.waiting, led.red)?

vuntz:
@MaximilianMeister thanks a lot for this! Just wondering out loud: do you think we should wait for @tboerger for merging this (in the meantime, we can add the patch to the package)? Or you're happy merging the way it is (assuming we get +1)?

crowbar_framework/vendor/assets/javascripts/jquery/ledUpdate.js
MaximilianMeister: not sure what that should bring, but this happens already in L262 if it is what you mean.

MaximilianMeister:
>  Just wondering out loud: do you think we should wait for @tboerger for merging this (in the meantime, we can add the patch to the package)? Or you're happy merging the way it is (assuming we get +1)?

I wouldn't wait, would be good to have one more confirmation that it works (best with a multinode setup) as i have just tested it with the crowbar node. @vuntz can you do a quick test?

crowbar_framework/vendor/assets/javascripts/jquery/ledUpdate.js
vuntz: > not sure what that should bring

In case the same issue happens for other states than ready.



> but this happens already in L262 if it is what you mean.

Nope, that's not the same led states :-)

MaximilianMeister: ouh I should read the classes :) yes thanks for the hint I will add it

MaximilianMeister:
btw the QA for this is just applying any proposal, and then switching to the root page and observe the led's. if anyone (with a multinode setup) can confirm that would be great

jgrassler:
+1, Works on my multi node setup (pretty vanilla mkcloud parameters, built with plain step).

vuntz:
Hrm; it seems it kills the animation for me when the node is applying: I just get a simple yellow led.

vuntz:
Btw, no need to apply a proposal to test, just open the node dashboard in the browser and do something like in a terminal:
```
crowbar crowbar transition $node applying
crowbar crowbar transition $node ready
```

@MaximilianMeister that should work with the admin server too, so no need for a multinode setup.

MaximilianMeister:
> Hrm; it seems it kills the animation for me when the node is applying: I just get a simple yellow led.

I had this as well, but only when i added `.led.unready, .led.in_process, .led.spin` to the selector.
Like it is now it has been working fine on multiple tests. That's weird.

> that should work with the admin server too, so no need for a multinode setup.

I know, that's how I tested it, I was only not sure how it behaves when there are more than one object found by the selector...
anyways this updateLed library should maybe be replaced some day. It is not really good to read and I think we could do sth better in less than ~300 LOC

vuntz:
Hrm, I wonder if it could be browser specific... I was testing with FF.

MaximilianMeister:
> Hrm, I wonder if it could be browser specific... I was testing with FF.

that might be, I was testing only with chrome...

MaximilianMeister:
> Hrm, I wonder if it could be browser specific... I was testing with FF.

@vuntz I have tested this now with FF and it works for me... what shall we do?

MaximilianMeister:
just for the record, we decided to revive https://github.com/crowbar/crowbar-core/pull/189
@rsalevsky will try to come up with something to replace this broken led functionality.

it would also look much better, and we wouldn't need to worry about this bug anymore

cyntss:
@MaximilianMeister I finally solved this problem!! yes.. almost 9 pm and that is how long it took me lol.

You can find the issue solved at: http://crowbar.g9.cloud.suse.de/

Since I dont have a local copy of the files and i worked with ssh and vim only, I'll copy-paste the code form vendor/assets/javascripts/jquery/ledUpdate.js

cyntss:
Find it here @MaximilianMeister  http://pastebin.nue.suse.com/21266

aplanas:
@cyntss thanks!!

MaximilianMeister:
@cyntss thanks for not giving up on this :-) I just tested it here at home, the transitioning to `ready` works well! but it seems the first 30sec after transitioning to `applying` the led stays yellow (and on status 200), and only after 30sec it starts to animate like it should (and on status 304). both in FF and Chrome. did you see that too?

cyntss:
I didnt check that because i was always refreshing to see the Applying status. I will fix that, it should be a smaller fix for this case :)

cyntss:
Ok... now this is the RIGHT solution.. no more javascript for animation.

in applicacion.scss (or even better if we create a file called animations.scss and then add this file in applicacion.scss), we need to paste the following code:

http://pastebin.nue.suse.com/21334

cyntss:
and this is for the JS file ledUpdate.js

http://pastebin.nue.suse.com/21338

MaximilianMeister:
tested, and it works just fine :-)

the animation is also much smoother now.
spritely was removed in a second commit. I guess we don't need it anymore. I haven't found anymore references elsewhere.

crowbar_framework/vendor/assets/javascripts/jquery/ledUpdate.js
```diff
@@ -242,6 +235,9 @@
   };
 
   LedUpdate.prototype.update = function(element, clazz, title, callback) {
+    var self = this;
+    var animatedClasses = ['unready', 'in_process', 'spin'];
```
MaximilianMeister: @cyntss I guess we can drop this 2 variables? looks like a remainder from debugging

crowbar_framework/vendor/assets/javascripts/jquery/spritely.js
```diff
@@ -1,539 +0,0 @@
-/*
```
vuntz: Can you keep this file just for now? Just to be safe (in case it's used in some hidden corner we forgot about).

vuntz:
Also, I checked on crowbar.g9 (I assume the fix is deployed there); I see the animated led moving 1 or 2 pixels left-right. Is that expected?

MaximilianMeister:
> Also, I checked on crowbar.g9 (I assume the fix is deployed there); I see the animated led moving 1 or 2 pixels left-right. Is that expected?

right, though I believe that was not introduced by this patch. I reverted the patch on g9 and i see the same... but would be good to see it on a fresh cloud. can you please redeploy g9 when you have time ?

vuntz:
> right, though I believe that was not introduced by this patch.

I don't see that on a system that wasn't touched for this bug.

MaximilianMeister:
@cyntss any idea why the led moves 1-2 pixels to the right when animating?

tboerger:
After dropping `sprite` and the 2 debugging variables :+1:
Cool!

MaximilianMeister:
> After dropping sprite...

I already had a drop commit, but see here why I took it out again https://github.com/crowbar/crowbar-core/pull/299#discussion-diff-52790726

and the crooked LED is still an issue....

crowbar_framework/vendor/assets/javascripts/jquery/spritely.js
tboerger: > Can you keep this file just for now? Just to be safe (in case it's used in some hidden corner we forgot about).

I am absolutely sure that this is not used ANYWHERE else!

vuntz: @tboerger my comment was really to avoid any risk for the release; since this won't make it in the release (but for an update), I'm fine removing the file now.

tboerger:
To make Hound and Codeclimate happy this should help:

```patch
diff --git a/.hound.scss.yml b/.hound.scss.yml
index f9831c8..ea52c04 100644
--- a/.hound.scss.yml
+++ b/.hound.scss.yml
@@ -2,5 +2,8 @@ linters:
   NestingDepth:
     max_depth: 6
 
+  LeadingZero:
+    style: exclude_zero
+
 exclude:
   - "crowbar_framework/vendor/**/*"
```

tboerger:
To avoid the vendor prefix offenses you should use https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/mixins/_vendor-prefixes.scss

tboerger:
And to make the keyframes better manageable you should make something like http://joshbroton.com/quick-fix-sass-mixins-for-css-keyframe-animations/

MaximilianMeister:
@cyntss @vuntz the problem with the pixel shifting is the image, it needs to be reworked a bit, right now it is not really proportional and it seems with css you cannot tweak that anymore

crooked_led

@tboerger will talk to @kwwii as he could adapt the image with photoshop, I wanted to do it with GIMP but I am not an expert

crowbar_framework/vendor/assets/stylesheets/mixins/_keyframes.scss
```diff
@@ -0,0 +1,14 @@
+@mixin keyframes($animation-name) {
+  @-webkit-keyframes #{$animation-name} {
+    @content;
```
cyntss: This is a very cool feature form sass.. less still doesnt have anything like it. :+1:

crowbar_framework/app/assets/stylesheets/content/animation.scss
```diff
@@ -0,0 +1,10 @@
+.led.unready,
+.led.spin,
+.led.in_process {
+  animation: play 0.8s steps(8) infinite;
```
tboerger: This is still not using the bootstrap mixin for animations, e.g. `@include animation(play .8s steps(8) infinite)`

crowbar_framework/vendor/assets/stylesheets/mixins/_keyframes.scss
tboerger: Yeah, that's a pretty cool SASS/SCSS feature 8)

tboerger:
:+1:

MaximilianMeister:
@cyntss @vuntz now with some more tweaking with some help from @tboerger it finally works without pixelshifting!

vuntz:
+1

MaximilianMeister:
btw, gating fails because gating can't patch binary diffs (introduced by https://github.com/MaximilianMeister/crowbar-core/commit/ed9b244c146c761b4c71835aa4874aa22052a4d5)

PR #300, opened 2016-02-09
restore: don't treat repos as proposals

MaximilianMeister:
during testing an exported repo (ptf.json) was treated as a proposal
so just save repos

not sure if there is a better way to determine repos (without having to look into the file)

### DIFF
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -163,7 +163,7 @@ module Crowbar
 
               record = JSON.load(file.read)
               filename = file.basename.to_s
-              if proposal?(filename) && type == :databags
+              if proposal?(file) && type == :databags
                 Rails.logger.debug "Restoring proposal #{filename}"
                 bc_name, prop = filename.split("-")
                 prop.gsub!(/.json$/, "")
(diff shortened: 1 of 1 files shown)


crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -163,7 +163,7 @@ def restore_chef
 
               record = JSON.load(file.read)
               filename = file.basename.to_s
-              if proposal?(filename) && type == :databags
+              if proposal?(filename) && !repo?(file) && type == :databags
```
vuntz: I feel a better check is to look if the file is in the crowbar databag; if it's in any other databag, then it's not a proposal. So I'd improve `proposal?` accordingly.

MaximilianMeister: sounds good, but determination would also be by matching the file.to_s, or do you have another way? 

i could write it like this
```ruby
      def proposal?(file)
        !file.basename.to_s.match(/(_network\.json$)|(^template-(.*).json$)|(^queue\.json$)/) && \
          !crowbar_databag?(file)
      end

      def crowbar_databag?(file)
        file.to_s =~ /knife\/databags\/crowbar/
      end
```

vuntz: Sounds good, except that you don't want to negate the call to `crowbar_databag?` I think :-)

MaximilianMeister: true. maybe i will even not use a separate method, as it is just 1 line

vuntz:
Assuming it works: +1

MaximilianMeister:
> Assuming it works

this has been tested on a QA machine and it lead to https://github.com/crowbar/crowbar-core/pull/300 so it passed that point

MaximilianMeister:
@vuntz btw I am always testing my patches, if not, i will make a note

aplanas:
+1

PR #301, opened 2016-02-09
Adapt ownership of rabbitmq logs before upgrade

rhafer:
On SLE12 rabbitmq startup and shutdown logs will be created by the user
rabbitmq. Adjust ownership after shutdown so that it can be cleanly started
on SLE12.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/stop-services-before-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/stop-services-before-upgrade.rb
@@ -49,3 +49,15 @@ bash "stop OpenStack services" do
     done
   EOF
 end
+
+# On SLE11 the startup and shutdown logs are created as root, adjust
+# ownership so that rabbitmq on SLE12 can write to them after upgrade. They
+# are created by the rabbitmq user there. We're not deleting them here, cause
+# the could still contain helpful information for debugging.
+["shutdown_log", "shutdown_err", "startup_log", "startup_err"].each do |logfile|
+  file "/var/log/rabbitmq/#{logfile}" do
+    user "rabbitmq"
+    group "rabbitmq"
+    only_if { File.exist? "/var/log/rabbitmq/#{logfile}" }
+  end
+end


vuntz:
I'm fine with that, but I guess I would have used the chef way:

```ruby
file foo do
  user rabbitmq
  group rabbitmq
  only_if { File.exist? foo }
end
```

rhafer:
> I'm fine with that, but I guess I would have used the chef way

Argh. Right. This is a bit embarrassing I just didn't think about that ```only_if``` condition. Otherwise I would have done it that way from the beginning. I'll adjust it.

sayalilunkad:
I agree with vuntz. Also we could reduce a lot of code if we just change the ownership for the entire directory if that does not hurt anything else which I think it shouldn't.

rhafer:
> if that does not hurt anything else which I think it shouldn't.

It shouldn't, but I am not really sure about that. I'd prefer to just adapt the files we need.

rhafer:
> I would have used the chef way

done. Looks much nicer. But I haven't tested it yet.

chef/cookbooks/crowbar/recipes/stop-services-before-upgrade.rb
```diff
@@ -49,3 +49,15 @@
     done
   EOF
 end
+
+# On SLE11 the startup and shutdown logs are created as root, adjust
+# ownership so that rabbitmq on SLE12 can write to them after upgrade. They
+# are created by the rabbitmq user there. We're not deleting them here, cause
+# the could still contain helpful information for debugging.
+["shutdown_log", "shutdown_err", "startup_log", "startup_err"].each do |logfile|
+  file "/var/log/rabbitmq/#{logfile}" do
```
jsuchome: Could you use a variable for `"/var/log/rabbitmq/#{logfile}"` ?

vuntz:
+1

sayalilunkad:
+1

PR #302, opened 2016-02-09
the file in the backup is called database.yml

MaximilianMeister:


rsalevsky:
+1

vuntz:
+1

aplanas:
+1

MaximilianMeister:
sorry for adding more code after the review but we were still lacking some exception handling at that place. could you please have another look @aplanas @rsalevsky @vuntz ?

MaximilianMeister:
btw I hit the following error when trying to restore a backup which was taken from a cloud after a failed restore... 

```
D, [2016-02-09T16:34:40.752088 #8977:0x00000005f286e0] DEBUG -- Restoring Crowbar database
E, [2016-02-09T16:34:40.753554 #8977:0x00000005f286e0] ERROR -- SQLite3::SQLException: near "TRUNCATE": syntax error: TRUNCATE "proposals"
```

That is how I came up with the Exception handling

rsalevsky:
Still +1

aplanas:
+1

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -266,9 +267,18 @@ def run_installer
...
         Rails.logger.debug "Restoring Crowbar database"
-        SerializationHelper::Base.new(YamlDb::Helper).load(
-          @data.join("crowbar", "production.yml")
-        )
+        begin
+          SerializationHelper::Base.new(YamlDb::Helper).load(
+            @data.join("crowbar", "database.yml")
+          )
+        rescue SQLite3::SQLException => e
+          Rails.logger.debug "Failed to load database.yml - #{e}"
```
vuntz: Using " - " is a bit unusual... Can you write it the usual way: `Failed to load database.yml: #{e}`?

MaximilianMeister:
> Using " - " is a bit unusual... Can you write it the usual way: Failed to load database.yml: #{e}?

done, as this was just a change in the log, do we need to wait for gating? would be good to have it in master as QA has to test this stuff now

last gating passed btw

PR #305, opened 2016-02-10
provisioner: Fix chef crash on admin server when forgetting nodes

vuntz:
I've seen that several times, but each time I didn't have time to debug
and this change was "good enough". It's still somehow correct because if
you read the code path that sets the interface list, it can be nil when
a conduit cannot be resolved, for instance.

Of course, now that I'm trying to reproduce, I'm not able to...

cc @rhafer

### DIFF
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -16,7 +16,7 @@
 def find_node_boot_mac_addresses(node, admin_data_net)
   # If we don't have an admin IP allocated yet using node.macaddress is
   # our best guess for the boot macaddress
-  return [node.macaddress] if admin_data_net.nil?
+  return [node.macaddress] if admin_data_net.nil? || admin_data_net.interface_list.nil?
   result = []
   admin_interfaces = admin_data_net.interface_list
   admin_interfaces.each do |interface|


vuntz:
Note that I'm mixed on merging this, tbh. I'm unhappy not being able to debug further... On the other hand, this totally kills the admin server when it happens, so it's pretty bad.

vuntz:
I think that's the same issue that @bmwiedemann was hitting when he submitted https://github.com/crowbar/crowbar-core/pull/225 -- and I'm pretty sure it's not because of a missing package.

bmwiedemann:
+1 the change fixes more things than it breaks (likely 0). I'm just not sure, if this is the best place to fix.

rhafer:
+1 I think it's good to merge this. But this deserves some further analysis (post release I guess). Would be really good to find out the root cause for this.

> I think that's the same issue that @bmwiedemann was hitting when he submitted #225 -- and I'm pretty sure it's not because of a missing package.

Then it's a different problem, because #225 was clearly caused by the ohai attributes not being populated correctly (because of the missing package).

aplanas:
Can help to add and `if` to check this condition and write a warning in the logs? So we can detect how frequently is happening?

+1

vuntz:
I created https://trello.com/c/t5BD661T as a followup to investigate this.

PR #306, opened 2016-02-10
Fixes for shutting down pacemaker resources

jsuchome:
Not yet tested, still investigating it this can help...

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -98,8 +98,7 @@ when "dump_openstack_database"
 
 when "db_shutdown"
 
-  # Stop remaining pacemaker resources
-  bash "stop pacemaker resources" do
+  bash "stop remaining pacemaker resources" do
     code <<-EOF
       for type in clone ms primitive; do
         for resource in $(crm configure show | grep ^$type | cut -d " " -f2);
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -104,7 +104,7 @@ when "db_shutdown"
       for type in clone ms primitive; do
         for resource in $(crm configure show | grep ^$type | cut -d " " -f2);
         do
-          crm resource stop $resource
+          crm --force --wait resource stop $resource
         done
       done
     EOF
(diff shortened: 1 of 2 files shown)


chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -104,7 +104,7 @@
       for type in clone ms primitive; do
         for resource in $(crm configure show | grep ^$type | cut -d " " -f2);
         do
-          crm resource stop $resource
+          crm --force --wait resource stop $resource
```
aplanas: Is not `--wait` alone enough? This `--force` is for user interaction, is happening with this `--wait` option?

jsuchome: It's true that `--force` was likely not needed so far. But it's probably better to be sure if some user interaction would be required...

aplanas:
+1

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
vuntz: We actually always need `--force`, it was a bug to not have it. The `--wait` is the imrpovement here

vuntz:
+1

jsuchome:
New significant commit -> resetting your +1 counts

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -31,6 +31,18 @@
...
+  bash "stop rabbitmq resources" do
+    code <<-EOF
+        for resource in g-rabbitmq rabbitmq fs-rabbitmq ms-drbd-rabbitmq drbd-rabbitmq;
+        do
+          crm --force --wait resource stop $resource
+        done
+      done
+    EOF
+    only_if { ::File.exist?("/usr/sbin/crm") }
+  end
```
vuntz: Oh, I thought you would do it in the other recipe. But okay :-)

sayalilunkad: I am not completely sure but I think it is possible to use execute blocks here instead similar to how it is done here https://github.com/crowbar/crowbar-core/pull/301

jsuchome:
So, first testing shows that the resources were stopped and deleted without any problem. I'll try yet different setup...

vuntz:
+1 for this, although I believe it's fine to put the new bash resource in stop-services-before-upgrade.rb instead (and avoid code duplication).

vuntz:
Still +1.

aplanas:
+1

vuntz:
@jsuchome are you okay merging? (You might want to force push to fix the hound comment and just merge without waiting again for gating).

jsuchome:
@vuntz rebased and fixed hound issue; but I can't merge it myself before another gating passes

PR #307, opened 2016-02-10
run migrations before initializing the database (and afterwards)

MaximilianMeister:
also the SQLite3Exception came from the migrator, not the initializer
add some more error handling in case the database initializer fails

### DIFF
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -267,16 +267,7 @@ module Crowbar
 
       def restore_database
         Rails.logger.debug "Restoring Crowbar database"
-        begin
-          Crowbar::Migrate.migrate!
-        rescue SQLite3::SQLException => e
-          Rails.logger.debug "Failed to migrate database: #{e}"
-          set_failed
-          @status[:restore_database] = {
-            status: :not_acceptable,
-            msg: I18n.t("backups.index.migrate_database_failed")
-          }
-        end
+        migrate_database(:before)
 
         begin
           SerializationHelper::Base.new(YamlDb::Helper).load(
@@ -291,12 +282,25 @@ module Crowbar
           }
         end
 
+        migrate_database(:after)
+
         @status[:restore_database] ||= { status: :ok, msg: "" }
       end
 
       def proposal?(file)
         !file.basename.to_s.match(/(_network\.json$)|(^template-(.*).json$)|(^queue\.json$)/) && \
           file.to_s =~ /knife\/databags\/crowbar/
+      end
+
+      def migrate_database(time)
+        Crowbar::Migrate.migrate!
+      rescue SQLite3::SQLException => e
+        Rails.logger.debug "Failed to migrate database #{time} loading: #{e}"
+        set_failed
+        @status[:restore_database] = {
+          status: :not_acceptable,
+          msg: I18n.t("backups.index.migrate_database_failed")
+        }
       end
     end
   end
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -268,10 +268,21 @@ module Crowbar
       def restore_database
         Rails.logger.debug "Restoring Crowbar database"
         begin
+          Crowbar::Migrate.migrate!
+        rescue SQLite3::SQLException => e
+          Rails.logger.debug "Failed to migrate database: #{e}"
+          set_failed
+          @status[:restore_database] = {
+            status: :not_acceptable,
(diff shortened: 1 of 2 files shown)


crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -268,10 +268,21 @@ def run_installer
...
+          Rails.logger.debug "Failed to migrate database: #{e}"
+          set_failed
+          @status[:restore_database] = {
+            status: :not_acceptable,
+            msg: I18n.t("backups.index.migrate_database_failed")
+          }
+        end
+
+        begin
           SerializationHelper::Base.new(YamlDb::Helper).load(
```
rsalevsky: I think we should keep this in the `begin` block.

MaximilianMeister: you mean both statements in one begin, rescue ?

rsalevsky: yep

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -268,10 +268,18 @@ def run_installer
...
             @data.join("crowbar", "database.yml")
           )
         rescue SQLite3::SQLException => e
+          Rails.logger.debug "Failed to migrate database: #{e}"
+          set_failed
+          @status[:restore_database] = {
+            status: :not_acceptable,
+            msg: I18n.t("backups.index.migrate_database_failed")
+          }
+        rescue StandardError => e
```
MaximilianMeister: maybe I have to be more specific here now after putting everything in one begin block... If Migrate throws another error than SQLite, we will have a misleading log entry

MaximilianMeister: @rsalevsky do you mind if I change back to two begin blocks? Or do you have an idea which Errors i could catch for the YamlDB Serialization?

aplanas: Actually I think @rsalevsky's suggestion was not optimal for this case, unless there is an specific exception to the `load` case.

MaximilianMeister: @aplanas I changed it back now

crowbar_framework/lib/crowbar/backup/restore.rb
MaximilianMeister: this will lead to some problems at Exception handling, as discussed here https://github.com/crowbar/crowbar-core/pull/307#discussion_r52455950

aplanas:
IMHO +1

rsalevsky:
+1

mjura:
+1

vuntz:
Sorry, I have trouble understanding why we migrate before restoring the database. Don't we want to restore what's inside the database (so after restoring)?

MaximilianMeister:
> Sorry, I have trouble understanding why we migrate before restoring the database. Don't we want to restore what's inside the database (so after restoring)?

As I understood the problem here is that there is no database to restore before we migrate. in other words migrate creates the db @rsalevsky can you confirm that?

vuntz:
But then we would still want to migrate after restoring...

MaximilianMeister:
> But then we would still want to migrate after restoring...

that shouldn't hurt at least, but I lack some rails knowledge here, maybe @tboerger has an idea?

I could add it for now though, then we are at least on the safe side...

MaximilianMeister:
> But then we would still want to migrate after restoring...

@vuntz done

vuntz:
I'd still like to hear from @rsalevsky how he feels; maybe I'm wrong about the need of a post-restore migration.

rsalevsky:
I also have a lack of knowledge here. As I understood the problem the yaml_db backup has no database schema so it just imports the data to the existing tables. But at this time we just did a `rails db:create` but we didn't migrated so the database has not tables. At first we need to migrate to create the tables and then we can import the data so do a migrate after importing isn't required.

At the moment this should not hurt at all due to in the upgrade case the database is anyway empty.

vuntz:
Ok. I'm fine either way. @MaximilianMeister up to you :-)

MaximilianMeister:
@vuntz @rsalevsky IMO lets keep it like it is and merge it

vuntz:
Thinking more about it: I think what this tells us is that we need to have somewhere in the backup data information about the current migration for the database. This way we can migrate up to that point before restoring the db, restore the db, and then run the remaining migrations.

@MaximilianMeister do you want to tackle that now, or should I create a card?

MaximilianMeister:
@vuntz I will create a card

MaximilianMeister:
@vuntz https://trello.com/c/8Y7uiCOB/919-restore-backup (I was so free to just quote you)

PR #309, opened 2016-02-10
exclude swift databag

MaximilianMeister:
any idea where this databag comes from?

### DIFF
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -285,9 +285,11 @@ module Crowbar
       end
 
       def proposal?(file)
-        !file.basename.to_s.match(
-          /(_network\.json$)|(^template-(.*).json$)|(^queue\.json$)|(^swift\.json$)/
-        ) && file.to_s =~ /knife\/databags\/crowbar/
+        return false if file.basename.to_s =~ /^template-(.*).json$/ || \
+            !file.to_s =~ /knife\/databags\/crowbar/
+        raw_data = JSON.parse(file.read)["raw_data"]
+        return false if raw_data.nil?
+        raw_data.key?("attributes") && raw_data.key?("deployment")
       end
     end
   end
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -285,8 +285,9 @@ module Crowbar
       end
 
       def proposal?(file)
-        !file.basename.to_s.match(/(_network\.json$)|(^template-(.*).json$)|(^queue\.json$)/) && \
-          file.to_s =~ /knife\/databags\/crowbar/
+        !file.basename.to_s.match(
+          /(_network\.json$)|(^template-(.*).json$)|(^queue\.json$)|(^swift\.json$)/
+        ) && file.to_s =~ /knife\/databags\/crowbar/
       end
     end
   end


mjura:
+1

rsalevsky:
+1

vuntz:
+1

MaximilianMeister:
rebased

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -285,8 +285,9 @@ def restore_database
       end
 
       def proposal?(file)
-        !file.basename.to_s.match(/(_network\.json$)|(^template-(.*).json$)|(^queue\.json$)/) && \
-          file.to_s =~ /knife\/databags\/crowbar/
+        !file.basename.to_s.match(
+          /(_network\.json$)|(^template-(.*).json$)|(^queue\.json$)|(^swift\.json$)/
+        ) && file.to_s =~ /knife\/databags\/crowbar/
```
vuntz: Coming back to this after some more thoughts... I don't have a Cloud 5 data dump here, so I cannot check, but I think all proposals are actually starting with `bc-`. I feel that would be the right check, actually.

Is there a data dump somewhere to look at for this?

MaximilianMeister: In the dump they still have the `bc-` prefix, that's correct. but in `upgrade.rb` which runs before restore, they are renamed. https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/upgrade.rb#L96

rhafer: How about actually loading the file and check if it has an "attributes" and a "deployment" section? That way we can be pretty sure it's a proposal.

vuntz:
Okay, +1 for now and happy to get an improvement in a followup PR (or here if you have time, of course :-)).

MaximilianMeister:
> How about actually loading the file and check if it has an "attributes" and a "deployment" section? That way we can be pretty sure it's a proposal

@rhafer @vuntz could you please check the last commit? if it is better like that I can squash it. I personally like the idea by @rhafer

vuntz:
@MaximilianMeister looks good on the paper, +1 if you tested it :-)

MaximilianMeister:
> @MaximilianMeister looks good on paper, +1 if you tested it :-)

sorry, was still not quite correct. now it should be fine

MaximilianMeister:
actually not.. as we don't filter out templates anymore

MaximilianMeister:
@rhafer @vuntz not sure what is better now. shall we take it like that, or shall i drop the latest commit?

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -285,8 +285,10 @@ def restore_database
       end
 
       def proposal?(file)
-        !file.basename.to_s.match(/(_network\.json$)|(^template-(.*).json$)|(^queue\.json$)/) && \
-          file.to_s =~ /knife\/databags\/crowbar/
+        return false if file.basename.to_s =~ /^template-(.*).json$/
```
vuntz: Can you add back the `file.to_s =~ /knife\/databags\/crowbar/` check? This is definitely a requirement. And then, with the check of the content of the data bag item, I think we're good.

+1 after that.

MaximilianMeister:
> Can you add back the `file.to_s =~ /knife\/databags\/crowbar/` check? 

done

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -285,7 +285,10 @@ def restore_database
       end
 
       def proposal?(file)
-        !file.basename.to_s.match(/(_network\.json$)|(^template-(.*).json$)|(^queue\.json$)/) && \
+        return false if file.basename.to_s =~ /^template-(.*).json$/
+        raw_data = JSON.parse(file.read)["raw_data"]
+        return false if raw_data.nil?
+        raw_data.key?("attributes") && raw_data.key?("deployment") && \
           file.to_s =~ /knife\/databags\/crowbar/
```
vuntz: Heh, I would have added that check with the template one to avoid loading the file, but +1.

MaximilianMeister:
> Heh, I would have added that check with the template one to avoid loading the file, but +1.

updated

rhafer:
+1

vuntz:
Still +1

PR #312, opened 2016-02-11
crowbar: Fix typo and add comment explaining the rabbitmq specific code

vuntz:


jsuchome:
+1

aplanas:
+1

vuntz:
Gating fails because the other PR that got merged is not in the package, so we can't build a test package. Anyway to check with gating => merging.

PR #313, opened 2016-02-11
Remove the vhost config file of nova dashboard before upgrade

rhafer:
The config file doesn't work with apache 2.4 anymore and cause
the service to fail to start when reapplying the horzion barclamp.


rhafer:
I not really proud of this, but I currently don't see another way to fix it when looking at the apache restart/reload dance we do in the horizon cookbook.

vuntz:
Why is it not possible to do at the beginning of the horizon cookbook?

vuntz:
I don't like it, but okay, +1.

vuntz:
Nobody else wants to review it... I'll bite the bullet and just merge this.

PR #317, opened 2016-02-11
Not all listed rabbitmq resources have to be on the node

jsuchome:


jsuchome:
Yes, good point, but on the other hand:

1. we are trying to force stop the resource anyway
2. we could check for each resource being present, but wouldn't it be too slow? 
(I've thought about doing just one check for all, like `f crm configure show | grep -q rabbitmq; then`, but e.g. drbd-rabbitmq resource still doesn't have to be there...

vuntz:
Can you elaborate on the issue? Is it the return value at the end (which makes the resource fail), or is it the noise in the logs (stderr redirection)?

TBH, I think it's fine to check for the resources being present. The upgrade process is slow anyway...

jsuchome:
stderr likely is not needed, but the resource fails if I try to delete something that is not present

vuntz:
Well, you could fix it by adding a true as the last command after the loop ;-) But I think doing the check for existence is good.

chef/cookbooks/crowbar/recipes/stop-services-before-upgrade.rb
```diff
@@ -40,7 +40,9 @@
   code <<-EOF
     for resource in g-rabbitmq rabbitmq fs-rabbitmq ms-drbd-rabbitmq drbd-rabbitmq;
     do
-      crm --force --wait resource stop $resource
+      if crm configure show | grep -q $resource; then
```
vuntz: Please do `if crm configure show $resource` instead. Your grep is matching too many things ("grep rabbitmq" for instance)

vuntz:
+1

aplanas:
+1

PR #320, opened 2016-02-11
On restore check for existing proposals with by name

rhafer:
Otherwise it's not possible to restore mulitple proposals of barclamps that
allow multiple instances (e.g. pacemaker)

### DIFF
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -168,7 +168,8 @@ module Crowbar
                 bc_name, prop = filename.split("-")
                 prop.gsub!(/.json$/, "")
                 proposal = Proposal.where(
-                  barclamp: bc_name
+                  barclamp: bc_name,
+                  name: prop
                 ).first_or_initialize(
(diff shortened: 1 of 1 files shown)


rhafer:
This is untested currently.

There is one thing I am currently unsure about, will this actually really re-execute the schema migrations when uploading the second proposal? (It should by I am not sure if that really works.)

cc @jsuchome @MaximilianMeister

vuntz:
+1

aplanas:
+1

MaximilianMeister:
> There is one thing I am currently unsure about, will this actually really re-execute the schema migrations when uploading the second proposal? (It should by I am not sure if that really works.)

this will actually migrate every proposal of pacemaker in the database. so when you save the second pacemaker proposal, at the next step it will actually (try to) run the migrations for the first one again, but as it is already migrated, the migration scripts will not be run.

I left out the proposal name, because we were worried only about the core proposals back then, which only come with the default proposal. https://github.com/crowbar/crowbar-core/pull/213

+1

MaximilianMeister:
just tested it with two pacemaker proposals, and both are restored

PR #321, opened 2016-02-11
Use the same units when comparing for enough diskspace

rhafer:
"df" is returning kilobytes (making that more explicit by adding the -k
option).  The db_dump we use returns bytes. Convert that to kilobytes and add
some additional reserve (currently 10MB).

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -70,11 +70,14 @@ when "dump_openstack_database"
     block do
       require "pathname"
       path = Pathname.new(dump_location).split.first.to_s
-      available = `df #{path} | tail -n 1 | sed "s/ \\+/ /g" | cut -d " " -f 4`
+      available = `df -m #{path} | tail -n 1 | sed "s/ \\+/ /g" | cut -d " " -f 4`
       db_size = `su - postgres -c 'pg_dumpall | wc -c'`
+      # Use Megabytes and add some reserve (30 MiB)
+      db_size = db_size.to_i / 1024 / 1024 + 30
       if db_size.to_i > available.to_i
         message = "Not enough space for the Database dump on node #{node.name}!\n" \
-          "Database size is: #{db_size}B, available space on #{path} is only #{available}B."
+          "Database size is: #{db_size}MiB, available space on #{path} " \
+          "is only #{available.to_i}MiB."
         Chef::Log.fatal(message)
         node["crowbar_wall"]["chef_error"] = message
         node.save


chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -70,11 +70,13 @@
...
       path = Pathname.new(dump_location).split.first.to_s
-      available = `df #{path} | tail -n 1 | sed "s/ \\+/ /g" | cut -d " " -f 4`
+      available = `df -k #{path} | tail -n 1 | sed "s/ \\+/ /g" | cut -d " " -f 4`
       db_size = `su - postgres -c 'pg_dumpall | wc -c'`
+      # Use kilobytes and add some reserve (10 MB)
+      db_size = db_size.to_i / 1024 + 10240
       if db_size.to_i > available.to_i
         message = "Not enough space for the Database dump on node #{node.name}!\n" \
-          "Database size is: #{db_size}B, available space on #{path} is only #{available}B."
+          "Database size is: #{db_size}kB, available space on #{path} is only #{available.to_i}kB."
```
aplanas: I am reasonably sure that is using binary and not decimal units, so is `KB`, `KiB` or `K`

aplanas:
except for the comment: +1

jsuchome:
Is it fair to say to user that DB size is actual size + (hidden) 10 MB? I mean, maybe we should mention the reserve in the message.

But in general +1.

vuntz:
Wouldn't it be better to talk in MB?

rhafer:
Changed to MB and increased the server a little more 30MB. If the disk is really that full after the dump there will likely be troubles anyway.

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -70,11 +70,13 @@
...
       path = Pathname.new(dump_location).split.first.to_s
-      available = `df #{path} | tail -n 1 | sed "s/ \\+/ /g" | cut -d " " -f 4`
+      available = `df -m #{path} | tail -n 1 | sed "s/ \\+/ /g" | cut -d " " -f 4`
       db_size = `su - postgres -c 'pg_dumpall | wc -c'`
+      # Use Megabytes and add some reserve (30 MB)
+      db_size = db_size.to_i / 1024 / 1024 + 30
       if db_size.to_i > available.to_i
         message = "Not enough space for the Database dump on node #{node.name}!\n" \
-          "Database size is: #{db_size}B, available space on #{path} is only #{available}B."
+          "Database size is: #{db_size}MB, available space on #{path} is only #{available.to_i}MB."
```
jsuchome: As @aplanas already pointed out, it should be **MiB**

jsuchome:
But +1 anyway

rhafer:
Ok, changed to MiB to make the IEC happy :)

vuntz:
+1

aplanas:
+1

PR #324, opened 2016-02-11
crowbar: Fix blockui appearing nearly all the time (bsc#966356)

vuntz:
The selector evaluation should be done when we call live, not before.

https://bugzilla.suse.com/show_bug.cgi?id=966356

### DIFF
--- a/crowbar_framework/app/assets/javascripts/application.js
+++ b/crowbar_framework/app/assets/javascripts/application.js
@@ -1,7 +1,7 @@
 //= require_self
 
 function setBlockUI(events, selector) {
-  selector.live(events, function(event) {
+  $(selector).live(events, function(event) {
     $.blockUI({
       css: {
         border: 'none',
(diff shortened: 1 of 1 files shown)


vuntz:
I'm not 100% sure it's correct (would need to trigger a blockui to know that :-))

crowbar_framework/app/assets/javascripts/application.js
```diff
@@ -95,9 +95,9 @@ jQuery(document).ready(function($) {
     ]
   });
 
-  setBlockUI('submit', $('[data-blockui]'));
-  setBlockUI('click', $('body.backups, [data-blockui]'));
-  setBlockUI('click', $('installer-upgrades, [data-blockui]'));
+  setBlockUI('submit', '[data-blockui]');
+  setBlockUI('click', 'body.backups, [data-blockui]');
```
vuntz: Btw, I'm not sure about the selector: https://api.jquery.com/multiple-selector/ vs https://api.jquery.com/descendant-selector/

I was expecting the latter, not the former...

jsuchome:
It does not seem to help during editing of proposal...

vuntz:
@jsuchome did you recompile the assets and restart crowbar?

jsuchome:
Yes, both

MaximilianMeister:
I think the problem was not passing the selector, it was the logic in the chaining.

this should fix it https://github.com/crowbar/crowbar-core/pull/328

vuntz:
Thanks, closing this one.

PR #325, opened 2016-02-12
crowbar: Fix system() call when creating backup

vuntz:
It was failing when the name of the archive had a space, and... it was
allowing command injection.

### DIFF
--- a/crowbar_framework/app/controllers/logging_controller.rb
+++ b/crowbar_framework/app/controllers/logging_controller.rb
@@ -20,7 +20,7 @@ class LoggingController < BarclampController
     ctime=Time.now.strftime("%Y%m%d-%H%M%S")
     @file = "crowbar-logs-#{ctime}.tar.bz2"
     pid = fork do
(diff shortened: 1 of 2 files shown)


aplanas:
+1

bmwiedemann:
+1

MaximilianMeister:
as I see now the structure of the tarball has changed (I am just guessing that it is due to this PR due to the timestamp)

```
drwxr-xr-x crowbar/crowbar   0 2016-02-12 13:51 crowbar/
drwxr-xr-x crowbar/crowbar   0 2016-02-12 13:51 crowbar/root/
...
...
```

and now it has a different structure, was that wanted?

```
drwx------ crowbar/crowbar   0 2016-02-15 09:06 ./
drwxr-xr-x crowbar/crowbar   0 2016-02-15 09:06 ./crowbar/
...
...
```

vuntz:
It's a side-effect of moving from "tar czf foo.tar *" to "tar czf foo.tar ." -- but it doesn't harm.

PR #326, opened 2016-02-12
crowbar: Avoid exporting backups in the backup

vuntz:
This just makes the size go exponential.

### DIFF
--- a/crowbar_framework/lib/crowbar/backup/export.rb
+++ b/crowbar_framework/lib/crowbar/backup/export.rb
@@ -123,7 +123,13 @@ module Crowbar
           else
             # copy files with higher permissions
             dest = data_dir.join(destination).to_s
-            system("sudo", "cp", "-a", source, dest)
+            if source == "/var/lib/crowbar"
+              # avoid doing an export of the existing backups, to avoid growing
+              # size of backups
+              system("sudo", "rsync", "-a", "#{source}/", "--exclude", "backup", dest)
+            else
+              system("sudo", "cp", "-a", source, dest)
+            end
           end
         end
       end
--- a/crowbar_framework/config/initializers/yaml_db.rb
+++ b/crowbar_framework/config/initializers/yaml_db.rb
@@ -19,7 +19,7 @@ module SerializationHelper
   class Dump
     def self.tables
       ActiveRecord::Base.connection.tables.reject do |table|
-        ["schema_info", "schema_migrations", "sessions"].include?(table)
+        ["schema_info", "schema_migrations", "sessions", "backups"].include?(table)
       end
     end
   end


aplanas:
+1

MaximilianMeister:
the only good thing to have the backups in the backup would be if I restore in a non upgrade context. because after restore, my backup is gone and that looks a bit weird to the user...

tboerger:
I would prefer to always use rsync.

vuntz:
> the only good thing to have the backups in the backup would be if I restore in a non upgrade context. because after restore, my backup is gone and that looks a bit weird to the user...

Can't we automatically re-import the backup we're restoring from?

rsalevsky:
Should be possible to re-import the current backup. But I would expect that I have all backups after restoring... Not sure if the user expects the same. Maybe we should add a checkbox for including the backups?

MaximilianMeister:
> Can't we automatically re-import the backup we're restoring from?

we would have to add the database entry manually, and not do the `cleanup` at the end of a successful restore. or somehow use the `upload` mechanism?

crowbar_framework/lib/crowbar/backup/export.rb
```diff
@@ -123,7 +123,13 @@ def crowbar
           else
             # copy files with higher permissions
             dest = data_dir.join(destination).to_s
-            system("sudo", "cp", "-a", source, dest)
+            if source == "/var/lib/crowbar"
+              # avoid doing an export of the existing backups, to avoid growing
+              # size of backups
+              system("sudo", "rsync", "-a", "#{source}/", "--exclude", "backup", dest)
+            else
+              system("sudo", "cp", "-a", source, dest)
```
vuntz: That's just the old code :-) I can change it to use rsync too.

nicolasbock: You would save yourself an `else` :)

vuntz: Not really, because we don't want to exclude backup subdirectories in other directories :-)

nicolasbock: Well, that's true I suppose.

jsuchome:
+1

vuntz:
@skazi0 @jsuchome did one of you try the patch? It wasn't tested since February, so I'm simply wondering if it still works as intended.

jsuchome:
No, I did not test it :-(

vuntz:
> No, I did not test it :-(

Okay, I'll see if I can test it before merging.

vuntz:
Still works, merging.

bmwiedemann:
was cherry-picked to 3.0 in

```
commit 2d527b4563f9ec0e4eb5d247a317c3dcde329e38
Author:     Vincent Untz <vuntz@suse.com>
AuthorDate: Fri Feb 12 01:35:29 2016 +0100
Commit:     Maximilian Meister <mmeister@suse.de>
CommitDate: Mon Dec 12 16:43:10 2016 +0100
```

PR #327, opened 2016-02-12
crowbar: Add missing files to backup

vuntz:
At least on tex, we backup these files...

### DIFF
--- a/crowbar_framework/lib/crowbar/backup/base.rb
+++ b/crowbar_framework/lib/crowbar/backup/base.rb
@@ -96,7 +96,7 @@ module Crowbar
         end
 
         def export_files
-          restore_files.push(restore_files_after_install.first).map(&:reverse)
+          restore_files.concat(restore_files_after_install).map(&:reverse)
         end
       end
     end


rhafer:
I think this is not the right fix.  The needed files are listed in ```restore_files_after_install``` you just have to fix creating the list in ```export_files``` https://github.com/vuntz/crowbar-core/blob/fix-backup/crowbar_framework/lib/crowbar/backup/base.rb#L114

I seem that just takes first item of ```restore_files_after_install```

vuntz:
Ah, yes, thanks.

vuntz:
A quick test with no node and no proposal shows that it seems to work.

rhafer:
+1

rsalevsky:
+1

MaximilianMeister:
does the 2nd commit help after all?

vuntz:
> does the 2nd commit help after all?

I've removed it.

aplanas:
+1

MaximilianMeister:
+1

PR #328, opened 2016-02-12
bsc#966356 selectors were OR chained, not AND

MaximilianMeister:
https://bugzilla.suse.com/show_bug.cgi?id=966356

### DIFF
--- a/crowbar_framework/app/assets/javascripts/application.js
+++ b/crowbar_framework/app/assets/javascripts/application.js
@@ -96,8 +96,8 @@ jQuery(document).ready(function($) {
   });
 
   setBlockUI('submit', $('[data-blockui]'));
-  setBlockUI('click', $('body.backups, [data-blockui]'));
-  setBlockUI('click', $('installer-upgrades, [data-blockui]'));
+  setBlockUI('click', $('body.backups').find('[data-blockui]'));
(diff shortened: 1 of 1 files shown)


jsuchome:
Seems to help. +1

vuntz:
Would you mind updating the commit message. It feels like it should be something like:

```
crowbar: Fix blockui appearing nearly all the time (bsc#966356)

selectors were OR chained, not AND

https://bugzilla.suse.com/show_bug.cgi?id=966356
```

vuntz:
So this more or less reverts https://github.com/crowbar/crowbar-core/commit/2bd6bf0dd73d14bcb93069eb4bbe7fbf0cf964fb

@rsalevsky what was the reason behind that change?

vuntz:
+1

MaximilianMeister:
https://github.com/crowbar/crowbar-core/commit/2bd6bf0dd73d14bcb93069eb4bbe7fbf0cf964fb has worked on the pages where we wanted it to work (before with the descendant selector it somehow didn't)

but it has not been noticed that it also works on the barclamp pages, where we have $([data-blockui]) as well, but we didn't want it there...

rsalevsky:
+1 tested in the wizard

PR #330, opened 2016-02-12
Call cibadmin only if openais is still running

jsuchome:
We might need to reapply the recipe (after some error)
and cibadmin call fails if we stopped openais already.


vuntz:
+1

jsuchome:
(it's probably in obscure cases only, so you don't need to include it in M11)

rhafer:
+1

PR #331, opened 2016-02-12
Refactor dropdown menu for power actions (#bnc955810)

vwallfahrer:
- Dropdown menu collapse after action
- Shows flash message
- Shows blockUI while rebooting

### DIFF
--- a/crowbar_framework/app/assets/javascripts/application.js
+++ b/crowbar_framework/app/assets/javascripts/application.js
@@ -1,22 +1,5 @@
 //= require_self
 
-function setBlockUI(events, selector) {
-  selector.live(events, function(event) {
-    $.blockUI({
-      css: {
-        border: 'none',
-        padding: '15px',
-        backgroundColor: '#000',
-        '-webkit-border-radius': '10px',
-        '-moz-border-radius': '10px',
-        opacity: .5,
-        color: '#fff'
-      },
-      message: $(event.target).data('blockui')
-    });
-  });
-}
-
 jQuery(document).ready(function($) {
   $('textarea.editor').each(function() {
     var cm = CodeMirror.fromTextArea(this, {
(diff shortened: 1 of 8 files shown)


vuntz:
Why do we need the blockui for this?

vwallfahrer:
@vuntz  It takes a few seconds until the page is reloaded and node starts to reboot.

crowbar_framework/app/controllers/nodes_controller.rb
```diff
@@ -307,19 +307,42 @@ def hit
...
+        else
+          case action
+          when "reinstall", "reset", "update", "delete"
+            machine.set_state(action)
+          when "reboot", "shutdown", "poweron", "powercycle", "poweroff", "identify", "allocate"
+            machine.send(action)
+          else
+            render json: { error: "Invalid hit request '#{action}'" }, status: 500 && return
+          end
+            head :ok
```
tboerger: As hound say, too much indention :)

crowbar_framework/app/controllers/nodes_controller.rb
```diff
@@ -307,19 +307,42 @@ def hit
...
+        end
+      end
+
+      format.html do
+        if machine.nil?
+          flash[:alert] = "Could not find node '#{name}'"
+          redirect_to nodes_path, status: 404
```
tboerger: I think that should be `dashboard_url`. Hint: Always use the `_url` helpers for redirects.

tboerger:
In general I would just drop the `remote: true` flags from the links to make it real requests instead of AJAX.

vuntz:
Travis failed, can you check it out?

vwallfahrer:
@vuntz  yes, I will do it

vuntz:
@vwallfahrer can you maybe squash the commits?

crowbar_framework/app/assets/javascripts/application.js
```diff
@@ -95,9 +78,35 @@ jQuery(document).ready(function($) {
     ]
   });
 
-  setBlockUI('submit', $('[data-blockui]'));
-  setBlockUI('click', $('body.backups').find('[data-blockui]'));
-  setBlockUI('click', $('body.installer-upgrades').find('[data-blockui]'));
```
vuntz: Aren't we breaking the blockui bit for backup/installer here?

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -442,6 +442,7 @@ en:
       powercycle: 'Power Cycle'
       poweroff: 'Power Off'
       identify: 'Identify'
+      blockui_message: 'Rebooting nodes...'
```
jsuchome: Is it nodes, or just one node? I believe Power action is only available from 1 node view....

jsuchome:
It seems to work well - block message is shown while rebooting

crowbar_framework/config/locales/crowbar/en.yml
vwallfahrer: @jsuchome  you are right. I will change that

crowbar_framework/app/assets/javascripts/application.js
MaximilianMeister: yes, this needs to be adapted https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/helpers/backups_helper.rb#L42

jsuchome:
I can't judge javascript changes, but as for functionality, LGTM

crowbar_framework/app/assets/javascripts/application.js
vuntz: I see the backup part was done, but not the installer part.

tboerger:
After fixing all blockui parts within the whole application you get my :+1:

PR #333, opened 2016-02-12
also backup /etc/crowbar to restore before the installation

MaximilianMeister:
install-chef-suse will pick up the files from there


MaximilianMeister:
@vuntz  @rhafer are there some files that need special treatment here, or is it enough to just place them there before the installer?

vuntz:
> is it enough to just place them there before the installer?

Should be enough; will test your patch right now.

vuntz:
Works, +1.

vuntz:
Works, +1.

aplanas:
+1

PR #334, opened 2016-02-12
Cleanup old openvswitch config during the upgrade

jsuchome:
An attempt to fix https://bugzilla.suse.com/show_bug.cgi?id=966452

### DIFF
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -341,7 +341,10 @@ if not nodes.nil? and not nodes.empty?
                       architecture: arch,
                       is_ses: storage_available && !cloud_available,
                       crowbar_join: "#{os_url}/crowbar_join.sh",
-                      default_fs: mnode[:crowbar_wall][:default_fs] || "ext4")
+                      default_fs: mnode[:crowbar_wall][:default_fs] || "ext4",
+                      needs_openvswitch:
+                        (mnode[:network] && mnode[:network][:needs_openvswitch]) || false
+            )
           end
 
         when os =~ /^(hyperv|windows)/
--- a/chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
@@ -90,6 +90,24 @@
           ]]>
         </source>
       </script>
+<% if @needs_openvswitch -%>
+      <script>
+        <chrooted config:type="boolean">true</chrooted>
+        <debug config:type="boolean">true</debug>
+        <filename>openvswitch_upgrade</filename>
+        <source>
+          <![CDATA[
+          # Remove the stuff that is not needed any more on SLE12
+          chkconfig -d openvswitch-switch
+          rm -f /etc/init.d/ovs-ifup-br-*
+
+          systemctl enable openvswitch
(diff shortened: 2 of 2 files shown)


chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -341,7 +341,8 @@ def find_node_boot_mac_addresses(node, admin_data_net)
                       architecture: arch,
                       is_ses: storage_available && !cloud_available,
                       crowbar_join: "#{os_url}/crowbar_join.sh",
-                      default_fs: mnode[:crowbar_wall][:default_fs] || "ext4")
+                      default_fs: mnode[:crowbar_wall][:default_fs] || "ext4",
+                      needs_openvswitch: mnode[:network] && mnode[:network][:needs_openvswitch] || false)
```
vuntz: Can you add some `()` here, it's always a bit ambiguous to have `&&` and `||` in the same expression.

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
```diff
@@ -90,6 +90,20 @@
...
+        <filename>openvswitch_upgrade</filename>
+        <source>
+          <![CDATA[
+          chkconfig -d openvswitch-switch
+          systemctl enable openvswitch
+          rm /etc/init.d/ovs-ifup-br-*
+          ]]>
+        </source>
+      </script>
+<% end -%>
```
vuntz: Hrm, should that script be before the other script? Or it doesn't matter?

vuntz: Oh, it's not in the init-scripts, but chroot-scripts. I guess it doesn't matter, then.

rhafer: It should not matter. It's still in the installer and not using and ovs bridges at that point in time.

jsuchome: It actually runs after it, inside chroot (`chrooted` set to true, unlike the other script)

vuntz:
+1, but would like to hear feedback from testing ;-)

jsuchome:
Yes, that's why it is still marked as WIP. I'm currently installing Cloud6, so it will take some time before I get to this part

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
```diff
@@ -90,6 +90,24 @@
...
+        <filename>openvswitch_upgrade</filename>
+        <source>
+          <![CDATA[
+          # Remove the stuff that is not needed any more on SLE12
+          chkconfig -d openvswitch-switch
+          rm /etc/init.d/ovs-ifup-br-*
+
+          systemctl enable openvswitch
+
+          rm /etc/openvswitch/conf.db* /etc/openvswitch/.conf.db*
```
rhafer: I think it better to add a ```-f``` to both of the ```rm``` calls.

rhafer:
I was now able to reproduce the problem (it really only happens if there are repos that cannot be access via the admin node)

And I can confirm that the latest patch fixes it for me.

+1

jsuchome:
I can also confirm that the last iteration of this was working.

PR #335, opened 2016-02-12
crowbar: Remove from Backup link from installer

vuntz:
This just doesn't work, let's fix it later.


MaximilianMeister:
+1 (assuming the one comment # works for the next lines in the subtree)

vuntz:
> (assuming the one comment # works for the next lines in the subtree)

It does :-)

vuntz:
This is critical to get in; so merging even though there's only one +1.

PR #336, opened 2016-02-12
Reject incoming DNS requests while restoring/upgrading (bsc#966419)

rhafer:
This is needed to avoid a running cloud accessing the (incomplete) DNS info
that the admin node would handout when the installer runs as that might cause
services to fail when upgrading the admin node.

Ideally we would reconfigure bind to only listen on 127.0.0.1 during that
stage, but that requires some more changes as dnsmasq is already running on
that (to be done in a later commit).

https://bugzilla.suse.com/show_bug.cgi?id=966419

### DIFF
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -348,6 +348,36 @@ end
 # We would like to bind service only to ip address from admin network
 admin_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
 
+# When we're restoring the admin node from backup or upgrade data,
+# reject incoming DNS traffic to avoid sending wrong results to running
+# clients.
+# FIXME: A cleaner approach would be to restrict the listener address of
+# the nameserver to just the loopback interface while restoring. That however
+# needs coordination with dnsmasq which might be listening there already.
+if node["crowbar"]["admin_node"] && ::File.exist?("/var/lib/crowbar/install/restore_steps")
+  admin_if = node[:crowbar_wall][:network][:nets][:admin].first
+  bash "block incoming DNS request during install/upgrade" do
+    code <<-EOH
+      iptables -I INPUT -i #{admin_if} -p udp -m udp --dport 53 \
+        -j REJECT -m comment --comment "UPGRADE_DNSBLOCK"
+      iptables -I INPUT -i #{admin_if} -p tcp -m tcp --dport 53 \
+        -j REJECT -m comment --comment "UPGRADE_DNSBLOCK"
+    EOH
+    not_if "iptables -L INPUT | grep -q UPGRADE_DNSBLOCK"
+  end
+elsif node["crowbar"]["admin_node"]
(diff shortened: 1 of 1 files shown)


chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -347,6 +347,31 @@ def make_zone(zone)
 
 # We would like to bind service only to ip address from admin network
 admin_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+# if we're the admin node and no other node exists yet just listen on loop back
```
vuntz: Update comment for restore-only

rhafer: Oops, that's completely bogus anyway and from different attempt to fix the problem. I'll rephrase.

rhafer:
Fixed comment.

I also removed the check for the number of nodes, it's not needed anymore and actually had the potential of being racey.

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -348,6 +348,36 @@ def make_zone(zone)
...
+  bash "block incoming DNS request during install/upgrade" do
+    code <<-EOH
+      iptables -I INPUT -i #{admin_if} -p udp -m udp --dport 53 \
+        -j REJECT -m comment --comment "UPGRADE_DNSBLOCK"
+      iptables -I INPUT -i #{admin_if} -p tcp -m tcp --dport 53 \
+        -j REJECT -m comment --comment "UPGRADE_DNSBLOCK"
+    EOH
+    not_if "iptables -L INPUT | grep -q UPGRADE_DNSBLOCK"
+  end
+else
```
vuntz: You can probably do `elsif node["crowbar"]["admin_node"]`

rhafer: Right. Done.

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -348,6 +348,36 @@ def make_zone(zone)
 # We would like to bind service only to ip address from admin network
 admin_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
 
+# When we're restoring the admin node from backup or upgrade data,
+# reject incoming DNS traffic to avoid sending wrong results to running
+# clients.
+# FIXME: A cleaner approach would be to restrict the listner address of
```
vuntz: s/listner/listener/ but I can live with it

vuntz:
+1

vuntz:
@rhafer do you think it makes sense to have a followup card to clean this up?

rhafer:
> do you think it makes sense to have a followup card to clean this up?

Yes. I'd also leave the bug open. (With lower priority)

vuntz:
+1

aplanas:
+1

rhafer:
FYI, I just did another successful test run with this patch on a complete new deployment.

PR #337, opened 2016-02-12
crowbar-upgrade: avoid propagation of error status

aplanas:
`/etc/init.d/openais status` if is `stop`, will return `7` (error), and will be propagated.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -31,8 +31,12 @@ when "openstack_shutdown"
 
   include_recipe "crowbar::stop-services-before-upgrade"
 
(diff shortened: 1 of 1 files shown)


vuntz:
+1

sayalilunkad:
+1

jsuchome:
Ouch. Well, than maybe I should use `if /etc/init.d/openais status` instead. 

Because if we put `|| true` there, than we do not need that check for service status at all

aplanas:
New approach based on @vuntz and @jsuchome suggestions!

jsuchome:
+1 based on the code (but not tested by myself)

vuntz:
Can someone test it?

sayalilunkad:
I am testing it. Will take around half an hour.

sayalilunkad:
+1 works for me

PR #338, opened 2016-02-12
crowbar: Use restoration instead of restore when it's not a verb

vuntz:


aplanas:
+1 for 'restoration process' (checked with some automatic translators)

jsuchome:
I also like it more.

PR #339, opened 2016-02-12
Delete backup tmpdir

rsalevsky:


crowbar_framework/config/initializers/tmpdir.rb
```diff
@@ -0,0 +1,30 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Dir
+  class << self
+    alias_method :o_mktmpdir, :mktmpdir
+
+    def Dir.mktmpdir(prefix_suffix=nil, tmp_dir=nil, *rest)
+      tmp_dir = Rails.root.join("tmp").to_s if tmp_dir.nil?
```
vuntz: Why not use the system tmp dir?

rsalevsky: Donn't know it's on the card see https://trello.com/c/v8GWRrgw/600-3-s36p9-cleanup-tmpdirs-after-usage.

crowbar_framework/app/controllers/backups_controller.rb
```diff
@@ -54,6 +54,8 @@ def create
         end
       end
     end
+  ensure
+    @backup.cleanup
```
vuntz: We don't need to check that @backup is not nil?

crowbar_framework/app/models/backup.rb
```diff
@@ -100,6 +100,11 @@ def upgrade
     end
   end
 
+  def cleanup
+    FileUtils.rm_rf(@data) if @data
```
MaximilianMeister: `@data.rmtree` doesn't work here?

tboerger:
Even if it get's cleaned after process gets restarted :+1:

crowbar_framework/config/initializers/tmpdir.rb
```diff
@@ -0,0 +1,37 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Dir
+  class << self
+    # set alias to orginal method to use it after overwrite
+    alias_method :o_mktmpdir, :mktmpdir
```
tboerger: http://apidock.com/rails/ActiveSupport/CoreExtensions/Module/alias_method_chain

crowbar_framework/config/initializers/tmpdir.rb
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Dir
+  class << self
+    def Dir.mktmpdir_with_rails_context(prefix_suffix = nil, tmp_dir = nil, *rest)
+      # set tmp dir to application tmp dir
+      tmp_dir = Rails.root.join("tmp").to_s if tmp_dir.nil?
+      # get the caller of the method
```
MaximilianMeister: i would just say
# determine the caller

crowbar_framework/config/initializers/tmpdir.rb
```diff
@@ -0,0 +1,35 @@
...
+      call = caller_locations(1, 1).first.label
+      # set suffix to the caller
```
MaximilianMeister: i would either drop this comment or say
`# set caller prefix to the directory name`

crowbar_framework/config/initializers/tmpdir.rb
```diff
@@ -0,0 +1,35 @@
...
+      # this makes debugging easier in case there is a tmp dir not correctly removed
```
MaximilianMeister: `in case a tmp dir is not correctly removed`

MaximilianMeister:
+1

crowbar_framework/config/initializers/tmpdir.rb
```diff
@@ -0,0 +1,35 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Dir
+  class << self
+    def Dir.mktmpdir_with_rails_context(prefix_suffix = nil, tmp_dir = nil, *rest)
```
tboerger: `def mktmpdir_with_rails_context(prefix_suffix = nil, tmp_dir = nil, *rest)`

vuntz:
>> Why not use the system tmp dir?
> Don't know it's on the card see https://trello.com/c/v8GWRrgw/600-3-s36p9-cleanup-tmpdirs-after-usage.

So it's not really a good answer :-) I'd like to get a good explanation of why in the commit message.

MaximilianMeister:
> So it's not really a good answer :-) I'd like to get a good explanation of why in the commit message.

I would say the system's tmp dir could be a problem if the admin chooses to clean it from time to time by a cronjob or with systemd, and as the directories we create there are really application specific, it makes sense to put them under the rails tmp dir.

we already use Rails/tmp for the file locks, so basically we apply the same reason that was used there (whatever it was)

vuntz:
Thanks for the answer; would be a good thing to add to the commit message.

> the system's tmp dir could be a problem if the admin chooses to clean it from time to time by a cronjob or with systemd

The tmp dirs we're creating are really temporary, and a cron job that cleans /tmp only removes old files, not all files. So not sure it applies. On the contrary, I would say the issue is that if we lack proper cleaning (like right now, before this patch), such cron jobs would be what save us and using the rails tmp dir wouldn't help :-)

MaximilianMeister:
>  On the contrary, I would say the issue is that if we lack proper cleaning (like right now, before this patch), such cron jobs would be what save us and using the rails tmp dir wouldn't help :-)

that is true, nevertheless cleaning after "ourselves" would be much better in that case IMO. My main problem that I see was (even though unlikely) that at the time of restore or backup, someone could decide to clean the `/tmp` (which is a common procedure) and that would crash crowbar. in the rails tmp the admin will likely not touch anything

vuntz:
> nevertheless cleaning after "ourselves" would be much better in that case IMO

I agree, of course. I'm just saying the usual tmp dir may have another layer of cleaning that helps us.

Anyway, what really matters to me is that it's explained in the commit message.

MaximilianMeister:
`Beside this it gives us the possibility to find the method when a tmpdir is not deleted.`

i would delete this from the commit message, as it is a small detail, or rephrase it a bit

vuntz:
+1

MaximilianMeister:
still +1 (and thanks for the changes)

rsalevsky:
rebased

vuntz:
Can you backport to 3.0?

PR #340, opened 2016-02-12
don't allow restoring backup from different crowbar version

MaximilianMeister:
but only when it is not an upgrade


tboerger:
:+1:

rsalevsky:
+1

PR #341, opened 2016-02-14
Protect admin server against discovered nodes with missing ohai data

vuntz:
It seems there's a bug somewhere that can cause a node to not have ohai data. And this can result in breaking the chef run on the admin server, which is really bad.

So protect against this.


rhafer:
+1 (I assume we don't really know yet when exactly that problem with missing ohai data happens?)

aplanas:
I am a bit anxious about this workaround. I guess there will be a bug report and a trello card to find and fix the root cause.

Meanwhile +1

vuntz:
> I assume we don't really know yet when exactly that problem with missing ohai data happens?

Nope; I didn't manage to reproduce this -- it happened on a discovered node with no reason :/

> I am a bit anxious about this workaround.

It's actually not a workaround, I would say: you can do some kind of DoS on the admin server without this.

tboerger:
:+1:

PR #342, opened 2016-02-14
ipmi: Allocate address on bmc network after allocation

vuntz:
Do not allocate an address for discovered nodes (which might not be
used), but only when the node gets allocated (reaching
hardware-installing state).

### DIFF
--- a/crowbar_framework/app/models/ipmi_service.rb
+++ b/crowbar_framework/app/models/ipmi_service.rb
@@ -71,6 +71,16 @@ class IpmiService < ServiceObject
       bmc_role = node.admin? ? "bmc-nat-router" : "bmc-nat-client"
       result = add_role_to_instance_and_node("ipmi", inst, name, db, role, bmc_role)
 
+      @logger.debug("IPMI transition: leaving from discovered state for #{name}")
+      a = [200, { name: name }] if result
+      a = [400, "Failed to add role to node"] unless result
+      return a
+    end
+
+    # node got allocated, ensure that node can use BMC network by allocating an
+    # IP / enabling the interface
+    if state == "hardware-installing"
+      @logger.debug("IPMI transition: hardware-installing state for #{name}")
       ns = NetworkService.new @logger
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/ipmi_service.rb
```diff
@@ -71,6 +71,16 @@ def transition(inst, name, state)
...
 
+      @logger.debug("IPMI transition: leaving from discovered state for #{name}")
+      a = [200, { name: name }] if result
+      a = [400, "Failed to add role to node"] unless result
+      return a
+    end
+
+    # node got allocated, ensure that node can use BMC network by allocating an
+    # IP / enabling the interface
+    if state == "hardware-installing"
```
rhafer: Don't you need to add "hardware-installing"  to the "transition_list" in the proposal to make the is work?

vuntz: You're correct, I think.

vuntz: fixed

crowbar_framework/app/models/ipmi_service.rb
```diff
@@ -71,6 +71,16 @@ def transition(inst, name, state)
       bmc_role = node.admin? ? "bmc-nat-router" : "bmc-nat-client"
       result = add_role_to_instance_and_node("ipmi", inst, name, db, role, bmc_role)
 
+      @logger.debug("IPMI transition: leaving from discovered state for #{name}")
+      a = [200, { name: name }] if result
+      a = [400, "Failed to add role to node"] unless result
```
MaximilianMeister: maybe you could write
```ruby
a = if result
  [200, { name: name }]
else
  [400, "Failed to add role to node"]
end
```

tboerger:
```
return if result
  [200, { name: name }]
else
  [400, "Failed to add role to node"]
end
```

vuntz:
Obsoleted by https://github.com/crowbar/crowbar-core/pull/516

PR #343, opened 2016-02-15
respect the migration level in the backup

MaximilianMeister:
backup the migration level in the meta.yml, and restore it before loading the database

### DIFF
--- a/crowbar_framework/app/models/backup.rb
+++ b/crowbar_framework/app/models/backup.rb
@@ -33,6 +33,9 @@ class Backup < ActiveRecord::Base
       message: "allows only letters and numbers"
     }
 
+  validates :migration_level,
+    presence: true
+
   validate :validate_chef_file_extension,
     :validate_version,
     :validate_hostname,
@@ -158,6 +161,7 @@ class Backup < ActiveRecord::Base
     @data = Pathname.new(dir)
     self.version = ENV["CROWBAR_VERSION"]
     self.size = path.size
+    self.migration_level = ActiveRecord::Migrator.current_version
   end
 
   def save_archive
@@ -182,6 +186,7 @@ class Backup < ActiveRecord::Base
     self.version = meta["version"].to_s
     self.size = path.size
     self.created_at = Time.zone.parse(meta["created_at"])
+    self.migration_level = meta["migration_level"]
   end
 
   def delete_archive
--- a/crowbar_framework/app/models/backup.rb
+++ b/crowbar_framework/app/models/backup.rb
@@ -186,7 +186,8 @@ class Backup < ActiveRecord::Base
     self.version = meta["version"].to_s
     self.size = path.size
     self.created_at = Time.zone.parse(meta["created_at"])
-    self.migration_level = meta["migration_level"]
+    # 20151222144602_create_backups.rb
+    self.migration_level = meta["migration_level"] || 20151222144602
   end
 
   def delete_archive
--- a/crowbar_framework/db/migrate/20160215120522_add_migration_level_to_backups.rb
+++ b/crowbar_framework/db/migrate/20160215120522_add_migration_level_to_backups.rb
@@ -1,0 +1,5 @@
+class AddMigrationLevelToBackups < ActiveRecord::Migration
+  def change
+    add_column :backups, :migration_level, :integer, limit: 8, default: 20151222144602
+  end
+end
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -23,6 +23,7 @@ module Crowbar
         @backup = backup
         @data = @backup.data
         @version = @backup.version
+        @migration_level = @backup.migration_level
         @status = {}
         @thread = nil
       end
@@ -287,7 +288,7 @@ module Crowbar
 
       def restore_database
         Rails.logger.debug "Restoring Crowbar database"
-        migrate_database(:before)
+        migrate_database(:before, @migration_level)
 
         begin
           SerializationHelper::Base.new(YamlDb::Helper).load(
@@ -315,8 +316,12 @@ module Crowbar
         raw_data.key?("attributes") && raw_data.key?("deployment")
       end
 
-      def migrate_database(time)
-        Crowbar::Migrate.migrate!
+      def migrate_database(time, migration_level = nil)
+        if migration_level
+          Crowbar::Migrate.migrate_to(migration_level)
+        else
+          Crowbar::Migrate.migrate!
+        end
       rescue SQLite3::SQLException => e
         Rails.logger.debug "Failed to migrate database #{time} loading: #{e}"
         set_failed
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -288,7 +288,17 @@ module Crowbar
 
       def restore_database
         Rails.logger.debug "Restoring Crowbar database"
-        migrate_database(:before, @migration_level)
+        if ActiveRecord::Migrator.get_all_versions.include? @migration_level
+          migrate_database(:before, @migration_level)
+        else
+          Rails.logger.error "Cannot migrate to #{@migration_level}. Migration unknown"
+          set_failed
+          @status[:restore_database] = {
+            status: :not_acceptable,
+            msg: I18n.t("backups.index.restore_database_failed")
+          }
+          return
+        end
 
         begin
           SerializationHelper::Base.new(YamlDb::Helper).load(
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -311,6 +311,7 @@ module Crowbar
             status: :not_acceptable,
             msg: I18n.t("backups.index.restore_database_failed")
           }
+          return
         end
 
         migrate_database(:after)
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -275,7 +275,7 @@ module Crowbar
         sleep(1) until Crowbar::Installer.successful? || Crowbar::Installer.failed?
 
         if Crowbar::Installer.failed?
-          Rails.logger.debug "Crowbar Installation Failed"
+          Rails.logger.error "Crowbar Installation Failed"
           set_failed
           @status[:run_installer] = {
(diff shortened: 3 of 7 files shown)


crowbar_framework/app/models/backup.rb
```diff
@@ -167,6 +171,7 @@ def save_archive
     self.version = meta["version"]
     self.size = path.size
     self.created_at = Time.zone.parse(meta["created_at"])
+    self.migration_level = meta["migration_level"]
```
vuntz: Please add a fallback: this patch won't be in the release, so we need to assume there'll be backups without this metadata.

rsalevsky: Ouch, this means that we take the risk that a backup restore / upgrade will fail.

vuntz: Why? We know what the fallback should be.

crowbar_framework/lib/crowbar/migrate.rb
```diff
@@ -31,6 +31,14 @@ def migrate!
           )
         end
       end
+
+      def migrate_to(level)
+        puts "Migrating database schema to #{level}..."
+        ActiveRecord::Migrator.migrate(
+          ActiveRecord::Migrator.migrations_paths,
+          level.to_i
```
MaximilianMeister: just to mention, that also seems to work for down migrations. so when a Cloud6 backup is older than the `current_version` it will migrate down

MaximilianMeister:
@vuntz I have created a new migration now, instead of touching the one that is already frozen for the release

tboerger:
:+1:

rsalevsky:
+1

vuntz:
Is it still untested? (see PR title)

MaximilianMeister:
> Is it still untested? (see PR title)

yes it is, I will test it today later on

MaximilianMeister:
>> Is it still untested? (see PR title)

> yes it is, I will test it today later on

tested and works. on the first migration it migrates to what it says in meta.yml, and on the second migration all the way up to the latest

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -296,8 +297,12 @@ def proposal?(file)
         raw_data.key?("attributes") && raw_data.key?("deployment")
       end
 
-      def migrate_database(time)
-        Crowbar::Migrate.migrate!
+      def migrate_database(time, migration_level = nil)
+        if migration_level
+          Crowbar::Migrate.migrate_to(migration_level)
```
vuntz: What happens if that migration level is not known yet? (ie: use new backup on old code)

MaximilianMeister: good point, but is this really possible in the real life? that implies that a user somehow has downgraded his cloud. or where would the user get that backup from?

vuntz: > good point, but is this really possible in the real life? that implies that a user somehow has downgraded his cloud. or where would the user get that backup from?

If we had a db migration over the lifetime of the product, it could happen if the user does a backup with updates, and restores without updates.

MaximilianMeister: > If we add a db migration over the lifetime of the product, it could happen if the user does a backup with updates, and restores without updates.

that's the point that I don't understand, why would he restore without the updates? wouldn't that mean that the updates were rolled back?

vuntz: > that's the point that I don't understand, why would he restore without the updates?

Because a user does mistakes.

MaximilianMeister: back to the question 

> What happens if that migration level is not known yet? (ie: use new backup on old code)

I think I we were not on the same page, I thought you mean the migration_level attribute, or were you talking about the migration_level in the db?

if it is about the attribute: 
wouldn't it just be ignored then? it depends on how old the code is I guess. but if it is at the code state of release, then the migration just runs two times without the migration_level distinction.

if it is about the database entry: 
we could just add a check if that migration exists in `migrator.get_all_versions` and when it doesn't exist we return and show a message in the UI.
would that be fine with you?

vuntz: Let me rephrase my point: if the migration level attribute in the backup is higher than what the rails app knows, then I think we should either do not accept to upload this backup or fail early when trying to restore the backup.

I hope it helps? (I'm a bit confused by how you split the options, but that's probably an issue in my brain ;-))

MaximilianMeister: > I hope it helps?

I hope https://github.com/MaximilianMeister/crowbar-core/commit/5ad4362e3ea165174fcac82b1888bf97638f5781 helps :-)

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -268,14 +269,23 @@ def run_installer
...
-        migrate_database(:before)
+        if ActiveRecord::Migrator.get_all_versions.include? @migration_level
+          migrate_database(:before, @migration_level)
+        else
+          Rails.logger.error "Cannot migrate to #{@migration_level}. Migration unknown"
+          set_failed
+          @status[:restore_database] = {
+            status: :not_acceptable,
+            msg: I18n.t("backups.index.restore_database_failed")
+          }
```
vuntz: Should we `return` here?

crowbar_framework/lib/crowbar/backup/restore.rb
vuntz: Thanks, it does!

crowbar_framework/lib/crowbar/backup/restore.rb
MaximilianMeister: yes we should :-) thanks

vuntz: Not related to your change here, but now this raises the question: should we return also after https://github.com/crowbar/crowbar-core/pull/343/files#diff-f70b9c61f1886e37f29a33a5988e642eR291 ? :-)

MaximilianMeister: heh, I think this was forgotten when we introduced the second migrate_database call... thanks again

vuntz:
Thanks for all the fixes, +1.

rsalevsky:
+1

I guess some parts needs to go to cloud 5?

MaximilianMeister:
normal gating is green. but I triggered rebuild with `plain crowbarbackup crowbarrestore testsetup`
https://ci.suse.de/job/openstack-mkcloud/17319/console

MaximilianMeister:
https://github.com/crowbar/crowbar-core/pull/355 should fix the issue in the gating, so I guess we would have to wait for it to be merged

crowbar_framework/app/models/backup.rb
```diff
@@ -172,6 +176,7 @@ def save_archive
     self.version = meta["version"]
     self.size = path.size
     self.created_at = Time.zone.parse(meta["created_at"])
+    self.migration_level = meta["migration_level"] || ActiveRecord::Migrator.last_migration.version
```
vuntz: `ActiveRecord::Migrator.last_migration.version` is wrong here; I think it should a hard-coded version which is the latest one we have as of today.

MaximilianMeister: >  I think it should a hard-coded version which is the latest one we have as of today.

so that we come to a state of cloud6 without updates? what would happen with that hard coded value when we add another migration?

vuntz: > so that we come to a state of cloud6 without updates?

(not sure what you mean)

> what would happen with that hard coded value when we add another migration?

Well, that's where the second call to migrate is used :-) This is what happens:

- option a: backup doesn't have this migration level metadata. It means that it was created before this PR got merged, which means that the migration level is the hardcoded one I suggest. (It cannot be lower than that due to the fact that the backup feature was non-existing before)

- option b: backup has this metadata. It means our first call to migrate moves us to this level, we do the restore, then we call migrate to the latest level.

I'm not sure if what I'm saying is clear. Does that make sense?

crowbar_framework/app/models/backup.rb
```diff
@@ -172,6 +176,8 @@ def save_archive
     self.version = meta["version"]
     self.size = path.size
     self.created_at = Time.zone.parse(meta["created_at"])
+    # 20151222144602_create_backups.rb
```
MaximilianMeister: @vuntz to have some comment, can you help me here a bit to find a reasonable explanation for this in your words?

MaximilianMeister:
gating will not make sense until https://github.com/crowbar/crowbar-core/pull/359 is addressed

MaximilianMeister:
backup restore gating succeeded

crowbar_framework/db/migrate/20160215120522_add_migration_level_to_backups.rb
```diff
@@ -0,0 +1,5 @@
+class AddMigrationLevelToBackups < ActiveRecord::Migration
+  def change
+    add_column :backups, :migration_level, :integer, limit: 8
```
vuntz: Do we need to set somehow the default value for existing backups?

MaximilianMeister: we could use `default: 20151222144602` like for the backup upload in https://github.com/crowbar/crowbar-core/pull/343/files#diff-408ac208d889893305de65518aa046a2R180 ?

MaximilianMeister: this can also be changed in another migration if needed with `change_column_default` in the future

vuntz: I'd use that as a default, yes.

vuntz:
+1

rsalevsky:
Still +1

MaximilianMeister:
rebased

vuntz:
Still +1

MaximilianMeister:
rebased

MaximilianMeister:
@vuntz can you merge this?

vuntz:
> @vuntz can you merge this?

We're missing the +1 after the rebase.

Here's mine: +1.

rsalevsky:
+1

vuntz:
@MaximilianMeister feel free to backport to 3.0.

MaximilianMeister:
> We're missing the +1 after the rebase.

@vuntz sorry for the nagging, but why do we need another +1 after a rebase?

PR #344, opened 2016-02-15
Remove the crowbar_upgrade role from all nodes when it is no longer needed

jsuchome:
@rhafer fist test shows that this approach is actually working

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -135,6 +135,11 @@ class CrowbarService < ServiceObject
       node["crowbar_wall"]["crowbar_upgrade_step"] = "done_openstack_shutdown"
       node.save
     end
+
+    # ... and remove all nodes from the "crowbar-upgrade" role in proposal
+    proposal = Proposal.where(barclamp: "crowbar", name: "default").first
+    proposal["deployment"]["crowbar"]["elements"]["crowbar-upgrade"] = []
+    proposal.save
   end
 
   #
@@ -292,9 +297,19 @@ class CrowbarService < ServiceObject
     # Find all non-core proposals and remove all roles that belong
     # to those proposals from the nodes
     active_non_core_roles = RoleObject.find_roles_by_name("*-config-*").reject(&:core_role?)
+
+    # Add crowbar-upgrade role and crowbar-config-default roles to handle
+    # crowbar-upgrade removal from nodes
+    upgrade_role = RoleObject.find_role_by_name("crowbar-upgrade")
+    crowbar_config_role = RoleObject.find_role_by_name("crowbar-config-default")
+
+    roles_to_disable = active_non_core_roles
+    roles_to_disable << upgrade_role unless upgrade_role.nil?
+    roles_to_disable << crowbar_config_role
+
     upgrade_nodes = NodeObject.all.reject(&:admin?)
     upgrade_nodes.each do |node|
-      active_non_core_roles.each do |role|
+      roles_to_disable.each do |role|
         roles_to_remove = role.elements.keys
         roles_to_remove << role.name
         roles_to_remove.each do |delete_role|
@@ -304,8 +319,10 @@ class CrowbarService < ServiceObject
       node.save
     end
 
-    # and finally delete the roles itself
+    # finally delete the roles itself
     active_non_core_roles.each(&:destroy)
+    crowbar_config_role.elements["crowbar-upgrade"] = []
+    crowbar_config_role.save
   end
 
   def prepare_nodes_for_os_upgrade


jsuchome:
... or maybe not. Do we have to do this call once the nodes are ready again?

jsuchome:
No, this is wrong, currently it moves the nodes into the ready state...

jsuchome:
What if we call this snippet at some point when the nodes are alrady in 'ready' state, but we still have a control over the upgrade process UI? I'm thinking about that point when all nodes are upgraded and UI shows the button "Go to your barclamps"

rhafer:
I've been thinking about actually putting this into the ```disable_non_core_proposals``` or ```prepare_nodes_for_os_upgrade``` steps. Or into a separate step between them. And trying to do it without actual needing to commit the proposal.

Basically we'd need to:
* edit the proposal similar to what your patch does (but without commiting it)
* edit the proposal role in similar fashion
* remove the roles crowbar-config-default and crowbar-upgrade from the nodes' run_list (similar to what ```disable_non_core_proposals``` already does for the non-core barclamp.

jsuchome:
Do you mean something like this? But `delete_from_run_list` only marks the role for deletion, which actually happens in `rebuild_run_list` - shouldn't I call this explicitely?

jsuchome:
Hm, it seems that this was enough, actually... last change seems to work, crowbar-upgrade is gone from runlist and crowbar proposal does have empty list of elements for crowbar-upgrade...

rhafer:
But I guess the node still have the crowbar-config-default role assigned, right? I think we need to remove that as well from the nodes (apart the admin node).

rhafer:
And I assume the crowbar-config-default role still had the nodes listed in the deployment section. That one needs to be updated as well as we don't commit the proposal.

jsuchome:
I believe it's working correctly now. I'll have the setup with this patch running for some time, in case anyone wants to take a  look.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -305,7 +307,14 @@ def disable_non_core_proposals
     end
 
     # and finally delete the roles itself
-    active_non_core_roles.each(&:destroy)
+    roles_to_disable.each do |role|
+      if role.name == "crowbar-config-default"
+        role.elements["crowbar-upgrade"] = []
+        role.save
+      else
+        role.destroy
```
rhafer: This will also delete the "crowbar-upgrade" role. We shouldn't do that. I think ```active_non_core_roles.each(&:destroy)``` and updating crowbar-config-default is enough.

jsuchome: You mean `active_non_core_roles.each(&:destroy)` called on the original roles only, right, without crowbar-config-default and crowbar-core?
I thought deleting "crowbar-upgrade" role is save, we do not need it any more.

rhafer: ```crowbar-upgrade``` is an element role of the crowbar barclamp and part of the package. If we want to delete that we'd need a schema migration for that. But currently I don't think we want that.

vuntz:
@jsuchome gating failed while installing crowbar; can you double-check it's not caused by your PR?

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -129,12 +129,10 @@ def finalize_openstack_shutdown
     # Commit the same proposal, we have changed the upgrade step for DB nodes
     commit_and_check_proposal
 
-    # Finally, set the upgrade step to the point where no further action is done
-    # even when the upgrade recipes are accidentally executed
-    NodeObject.find("state:crowbar_upgrade AND roles:database-config-default").each do |node|
-      node["crowbar_wall"]["crowbar_upgrade_step"] = "done_openstack_shutdown"
-      node.save
-    end
```
vuntz: I'd like to keep that, just to be that we don't run anything again in case the user adds the node to crowbar-upgrade again.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -305,7 +307,14 @@ def disable_non_core_proposals
...
     # and finally delete the roles itself
-    active_non_core_roles.each(&:destroy)
+    roles_to_disable.each do |role|
+      if role.name == "crowbar-config-default"
+        role.elements["crowbar-upgrade"] = []
+        role.save
+      else if role.name != "crowbar-upgrade"
+        role.destroy
+      end
+    end
```
vuntz: This special handling here is weird. What about:

```ruby
### earlier in the patch
active_non_core_roles = RoleObject.find_roles_by_name("*-config-*").reject(&:core_role?)
# Add crowbar-upgrade role to delete and crowbar-config-default to modify
upgrade_role = RoleObject.find_role_by_name("crowbar-upgrade")
crowbar_config_role = RoleObject.find_role_by_name("crowbar-config-default")

roles_to_disable = active_non_core_roles
roles_to_disable << upgrade_role unless upgrade_role.nil?
roles_to_disable << crowbar_config_role
...
### where I'm commenting
active_non_core_roles.each(&:destroy)
crowbar_config_role.elements["crowbar-upgrade"] = []
crowbar_config_role.save
```

jsuchome: Yes... the point is, I want that previous code where `node.delete_from_run_list` is executed to be run for all roles here, so also for crowbar-config-default and crowbar-upgrade

jsuchome: Ah, ok, we could keep active_non_core_roles, right.

vuntz: Yes, it feels more readable to me at least.

vuntz:
+1

jsuchome:
Tested the last version of code: it works as expected.

rhafer:
+1

PR #345, opened 2016-02-16
better handling of invalid tarballs on upload

MaximilianMeister:


crowbar_framework/app/models/backup.rb
```diff
@@ -159,14 +159,19 @@ def save_archive
       return false
     end
 
+    begin
+      meta = YAML.load_file(data.join("meta.yml"))
+      self.version = meta["version"]
+      self.size = path.size
+      self.created_at = Time.zone.parse(meta["created_at"])
+    rescue
```
vuntz: Shouldn't you rescue for StandardError only?

MaximilianMeister: i have changed it now to not run into the error in the first place

rsalevsky:
+1

crowbar_framework/app/models/backup.rb
```diff
@@ -159,14 +159,19 @@ def save_archive
       return false
     end
 
+    if data.join("meta.yml").exist?
+      meta = YAML.load_file(data.join("meta.yml"))
```
vuntz: What if it's not a valid yaml file?

tboerger: That requires a begin and rescue block with a catch for the correct error class.

tboerger: Beside that I would avoid this nesting level. Maybe something like this is better:

```ruby
unless data.join("meta.yml").exist?
  errors.add(:file_content, I18n.t("backups.index.meta_missing"))
  return false
end

begin
  meta = YAML.load_file(
    data.join("meta.yml")
  )
rescue YAML::ParseError
  errors.add(:file_content, I18n.t("backups.index.meta_invlaid"))
  return false
end

self.version = meta["version"]
self.size = path.size
self.created_at = Time.zone.parse(meta["created_at"])

...
```

MaximilianMeister: > What if it's not a valid yaml file?

Why wouldn't it be valid yaml when it is created by Crowbar? Sure I can add a check, but then we could also check each file in the backup. to some point we have to rely on the user to just take a backup that was created by crowbar, and not put together by hand...but I will add a check

@tboerger testing some invalid yaml i ran into a `Psych::SyntaxError`

vuntz: > Why wouldn't it be valid yaml when it is created by Crowbar?

How do you know the tarball is created by Crowbar and not hand-crafted by the user?

vuntz: > to some point we have to rely on the user to just take a backup that was created by crowbar, and not put together by hand...

Sorry, replied too fast :-) I agree that we can't validate everything (like the content of the chef bits, for instance -- although it might actually be validated already?), but we do need at the very minimum some basic validation. 

TBH, I'm thinking we should validate more than what we do and check that the backup has all the required files, for instance -- to be sure that a restore won't fail because of a missing file.

MaximilianMeister:
> How do you know the tarball is created by Crowbar and not hand-crafted by the user?

do you mean that the user has changed some values in there after creation? is this even supported?

MaximilianMeister:
> do you mean that the user has changed some values in there after creation? is this even supported?

sorry I guess you revoked the question, but it wasn't crossed out so I thought it is still valid.

I was thinking if we could validate the meta.yml with a Backup Model validation, like `validate_meta_data` but I think the content of the backup is not part of the model validation, what do you say @tboerger?

crowbar_framework/app/models/backup.rb
tboerger: So `Psych::SyntaxError` should be fine to catch

MaximilianMeister:
> So Psych::SyntaxError should be fine to catch

@vuntz @tboerger done.
with a meaningful gating it's the same issue, we would need to wait for https://github.com/crowbar/crowbar-core/pull/355 to be merged

tboerger:
:+1:

bmwiedemann:
test failed because e.g. crowbar dns proposal show default
errored with 401 Unauthorized

tboerger:
@bmwiedemann but that is not related to this pr

MaximilianMeister:
@bmwiedemann https://github.com/crowbar/crowbar-core/pull/359 is required for the gating

MaximilianMeister:
backup restore gating succeeded

crowbar_framework/app/models/backup.rb
```diff
@@ -164,14 +164,26 @@ def save_archive
...
+    begin
+      meta = YAML.load_file(meta_file)
+    rescue Psych::SyntaxError
+      errors.add(:file_content, I18n.t("backups.index.invalid_file_content"))
+      return false
     end
 
-    meta = YAML.load_file(data.join("meta.yml"))
-    self.version = meta["version"].to_s
+    self.version = meta["version"]
```
vuntz: Is losing the `to_s` expected?

MaximilianMeister: I think it is expected IIRC because the versions in the model were floats before. but now they are strings. but i will change it back, `to_s` would help when someone (even though unlikely) uploads an old backup which was created with a float version. thanks for catching that

rsalevsky:
+1

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -718,6 +718,8 @@ en:
       enter_name: 'Please enter a name for the image'
       invalid_filename_format: 'Wrong file format, please dont use whitespace or special characters'
       invalid_filename_exists: 'Invalid filename, backup already exists.'
+      invalid_file_content: 'Invalid file content, please make sure the tarball contains valid backup data'
+      missing_meta: 'Metadata meta.yml is missing'
```
vuntz: Maybe change this to "meta.yml metadata file is missing in the backup"

MaximilianMeister:
rebased

MaximilianMeister:
@vuntz can you merge this?

vuntz:
> @vuntz can you merge this?

As for the others, missing +1...

Here's mine: +1.

(btw, gating failed in tempest, so unrelated to this PR)

MaximilianMeister:
> As for the others, missing +1...

I just updated the i18n... :-) 

/end-of-nagging ;-)

rsalevsky:
+1

vuntz:
Gah, the PR has conflicts :/

MaximilianMeister:
@vuntz resolved

MaximilianMeister:
@vuntz I think this is ready to merge now after the rebase

vuntz:
+1; but maybe rebase to retrigger travis.

MaximilianMeister:
> +1; but maybe rebase to retrigger travis.

@vuntz you can retrigger it over travis directly, do you want me to rebase nevertheless?

vuntz:
> @vuntz you can retrigger it over travis directly, do you want me to rebase nevertheless?

Nope, no need for a rebase, just a retrigger was good enough, thanks.

rsalevsky:
Still +1

Gating failed in tempest.

PR #346, opened 2016-02-16
dont require nscd

bmwiedemann:
install-suse-cloud was failing
when nscd was not installed or not running.
And since there is nothing needing it,
we make this part optional

note: also affects Cloud5, so needs backport.

### DIFF
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -403,6 +403,7 @@ execute "reload nscd after dns config change" do
   command "nscd -i hosts"
   action :nothing
   subscribes :run, "template[/etc/bind/db.#{node[:dns][:domain]}]"
+  only_if { File.exist?("/var/run/nscd/nscd.pid") }
 end
 
 node.set[:dns][:zones]=zones


chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -403,6 +403,7 @@ def make_zone(zone)
   command "nscd -i hosts"
   action :nothing
   subscribes :run, "template[/etc/bind/db.#{node[:dns][:domain]}]"
+  only_if { File.exist?("/var/run/nscd/nscd.pid") }
```
vuntz: Are we sure that path is correct for all systems? Or would `only_if "which nscd > /dev/null"` be better?

vuntz:
Can you fix add a "dns: " prefix to the commit message?

chef/cookbooks/bind9/recipes/default.rb
bmwiedemann: I checked that the .pid file exists on SLE11 and SLE12 iff nscd is running. having nscd installed is not enough.

bmwiedemann:
fixed commit msg

chef/cookbooks/bind9/recipes/default.rb
vuntz: Did you check on other distros?

Anyway...

vuntz:
+1

chef/cookbooks/bind9/recipes/default.rb
bmwiedemann: just checked on Debian-7.9 and 8.3 - it also uses /var/run/nscd/nscd.pid so that has pretty good chances of working on most relevant platforms.

rhafer:
+1

bmwiedemann:
btw: gating failure is unrelated

PR #348, opened 2016-02-16
Creating tar only if the backup name is valid (bsc#966424)

rsalevsky:
- create archive only if the name is valid
- run custom validations only in upload case
- don't validate version and size

### DIFF
--- a/crowbar_framework/app/models/backup.rb
+++ b/crowbar_framework/app/models/backup.rb
@@ -20,7 +20,8 @@ require "find"
 class Backup < ActiveRecord::Base
   attr_accessor :file
 
-  before_validation :save_or_create_archive, on: :create
+  before_validation :save_archive, on: :create, if: :upload?
+  before_save :create_archive, on: :create, unless: :upload?
   after_validation :delete_archive, unless: -> { errors.empty? }
   after_destroy :delete_archive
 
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/backup.rb
```diff
@@ -120,12 +116,10 @@ def attribute_names
...
   protected
 
-  def save_or_create_archive
-    if name.blank?
-      save_archive
-    else
-      create_archive
-    end
+  def upload?
+    return false unless file.present?
```
tboerger: for this cases i generally prefer

```
if file.present?
  true
else
  false
end
```

tboerger:
Some NITPICK otherwise :+1:

MaximilianMeister:
+1

crowbar_framework/app/models/backup.rb
```diff
@@ -32,16 +33,11 @@ class Backup < ActiveRecord::Base
       message: "allows only letters and numbers"
     }
 
-  validates :version,
-    presence: true
-
-  validates :size,
-    presence: true
```
vuntz: I don't understand why this gets removed. Can this  be explained in the commit message?

MaximilianMeister: agree that the commit message should explain it. it is not validated, as the values are set by crowbar, so we assume they are valid.

rhafer:
+1

vuntz:
+1, can you backport to 3.0?

PR #349, opened 2016-02-16
batch: Fix looking for unknown aliases

vuntz:
We were looking for quotes around the alias, while when we expand the
alias, we don't require quotes. This means that batch files not using
quotes (which is valid) for aliases will raise an exception like

/opt/dell/bin/barclamp_lib.rb:532:in `eval': (<unknown>): found character that cannot start any token while scanning for the next token at line 20 column 9 (Psych::SyntaxError)

### DIFF
--- a/bin/crowbar_batch
+++ b/bin/crowbar_batch
@@ -44,7 +44,7 @@ require "easy_diff"
 require "chef/mixin/deep_merge"
 require "pp"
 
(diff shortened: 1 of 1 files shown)


dirkmueller:
+1

vmoravec:
+1

vuntz:
cc @aspiers

tboerger:
@vuntz Please open a pull request for the crowbar-client as well.

PR #353, opened 2016-02-17
Fix bulk edit view

rsalevsky:
bulk_edit

Approved by @kwwii


vuntz:
Dropping the license key means that we can't allocate windows nodes from bulk edit anymore. Is that intended?

rsalevsky:
@vuntz Yes, we want to drop everything what is not used often or by the majority of our customers. Otherwise it is not possible to show everything in a way where it is possible to use it. For example the license key field was around 1cm for a long license key.

tboerger:
> @vuntz Yes, we want to drop everything what is not used often or by the majority of our customers. Otherwise it is not possible to show everything in a way where it is possible to use it. For example the license key field was around 1cm for a long license key.

But our customers also want to batch edit windows nodes oO

tboerger:
Maybe we have to do it entirely different as we don't have enough space.

rsalevsky:
I discussed this with @kwwii and @jdsn and we are sure that just some customers need this and not the majority. Sure this is not the final solution this is just a fix for Cloud 6...

tboerger:
Maybe it's even better to put the license instead of the group as the group can be done with drag and drop on the dashboard.

crowbar_framework/app/views/nodes/list.html.haml
```diff
@@ -30,77 +29,42 @@
             %th
               = t(".name")
             %th
-              = t(".hw_description")
```
vuntz: Can you remove the strings you're removing from the locales file?

vuntz:
In general, LGTM. Just one detail to fix.

crowbar_framework/app/views/nodes/list.html.haml
rsalevsky: Done

tboerger:
Do you have an updated screenshot?

rsalevsky:
crowbar list

crowbar_framework/app/views/nodes/list.html.haml
```diff
@@ -30,77 +29,44 @@
...
-                  = text_field_tag "node[#{node.handle}][group]", node.group, :class => "form-control"
-                %td
-                  = t("na")
-                - if crowbar_options[:show].include? :bios
-                  %td
-                    = t("na")
-                - if crowbar_options[:show].include? :raid
-                  %td
-                    = t("na")
+                %td{ class: "text-center" }
```
MaximilianMeister: `%td.text-center` doesn't work? same below

vuntz:
+1

crowbar_framework/app/views/nodes/list.html.haml
```diff
@@ -108,19 +74,26 @@
...
+                    %label
+                      = check_box_tag("node[#{node.handle}][allocate]", value = "1", checked = true, options = { disabled: "disabled" })
                   - else
                     %label
                       %input.allocate{ :type => "checkbox", :value => "1", :name => "node[#{node.handle}][allocate]" }
+          %tr
+            %td{ colspan: 6 }
+            %td.text-center
+              %label
+                %input{ :type => "checkbox", :value => "all", "data-checkall" => "input.allocate" }
```
MaximilianMeister:
```ruby
%input{ type: "checkbox", value: "all", data: { checkall: "input.allocate" } }
```

MaximilianMeister:
+1 (after using the new hash syntax)

MaximilianMeister:
+1 again

mjura:
+1

vuntz:
+1, but travis is red; it's unrelated to the PR, but it'd be nice to fix that first so we can avoid ignoring travis :/

MaximilianMeister:
since https://github.com/bbatsov/rubocop/pull/2931 is merged, which supposedly causes the issue I think we have to wait until the new version is published somehow, and the error disappears

workaround is this in the `Gemfile`
```ruby
group :test do
   # Temporary fix for rubocop rake 11 incompatibility
   # https://github.com/bbatsov/rubocop/pull/2931
   gem "rake", "< 11.0.0"
   ...
```

taken from https://github.com/crowbar/crowbar-client/pull/80

rsalevsky:
I run the tests manually, everything is green.

PR #354, opened 2016-02-17
Refactor dropdown menu for power actions (bsc#955810)

vwallfahrer:
- Dropdown menu collapse after action
- Shows flash message
- Shows blockUI while rebooting

### DIFF
--- a/crowbar_framework/app/views/nodes/_buttons_power.html.haml
+++ b/crowbar_framework/app/views/nodes/_buttons_power.html.haml
@@ -5,26 +5,24 @@
   - if @node[:platform] != 'windows' || @node.bmc_set?
 
     .dropdown
-      %button.btn.btn-default.dropdown-toggle{ "aria-expanded" => "true", "aria-haspopup" => "true",
-        "data-toggle" => "dropdown", :type => "button", role: "menu" }
+      %button.btn.btn-default.dropdown-toggle{ "aria-expanded" => "true", "aria-haspopup" => "true", "data-toggle" => "dropdown", :type => "button", role: "menu" }
         = t(".actions")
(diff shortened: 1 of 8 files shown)


crowbar_framework/app/views/nodes/_buttons_power.html.haml
```diff
@@ -2,23 +2,28 @@
...
+      %ul.dropdown-menu{ "aria-labelledby" => "dropdownMenu1" }
         %li{ role: "presentation" }
-          = link_to t(".reboot"), hit_node_path(@node.handle, "reboot"), role: "menuitem", remote: true
+          %a{ :href => hit_node_path(@node.handle, "reboot"), role: "menuitem",
+            data: { blockui_click: t(".blockui_message") } }
+            = t(".reboot")
         %li{ role: "presentation" }
-          = link_to t(".shutdown"), hit_node_path(@node.handle, "shutdown"), role: "menuitem", remote: true
-        %li.divider{ role: "presentation" }
-      - if @node.bmc_set?
```
jsuchome: I think you've lost this condition. In the old code, Power On button could be shown only `if @node.bmc_set?`

tboerger:
The other power buttons should get also a blockui message. Otherwise :+1:

crowbar_framework/app/views/nodes/_buttons_power.html.haml
tboerger: True, I missed that

rsalevsky:
Commit message is wrong... bsc#955810

tboerger:
Closed in favor of https://github.com/crowbar/crowbar-core/pull/375

PR #355, opened 2016-02-18
backup: better crowbar version version comparing

MaximilianMeister:


MaximilianMeister:
I will try to use the `PlatformVersion` class to compare the versions

tboerger:
:+1:

crowbar_framework/app/models/backup.rb
```diff
@@ -188,9 +190,11 @@ def validate_chef_file_extension
   end
 
   def validate_version
-    if version < 1.9
+    min_version = VersionCompare.new("1.9")
+    backup_version = VersionCompare.new(version)
+    if backup_version < min_version
       errors.add(:base, I18n.t("backups.validation.version_too_low"))
-    elsif version > ENV["CROWBAR_VERSION"].to_f
+    elsif version.to_f > ENV["CROWBAR_VERSION"].to_f
```
vuntz: Why not use VersionCompare here?

crowbar_framework/lib/version_compare.rb
```diff
@@ -1,4 +1,4 @@
-class PlatformVersion
+class VersionCompare
```
vuntz: I'd prefer to avoid the renaming for Cloud 6. It's unclear if this is used anywhere else and so on, and I don't want to risk a regression for this.

vuntz:
+1

vuntz:
@MaximilianMeister gating job still failed. Can you just try to backup/restore on your setup? Will probably be faster to do that to get it right. FWIW, I'll revert the other patch in the package for now.

MaximilianMeister:
> gating job still failed.

yeah it's because somehow the patch is not applied... must have rebuilt with the wrong commit sha... sorry will have a look

MaximilianMeister:
@vuntz after the latest patch, the restore was triggered (manually) lets wait for the next gating

crowbar_framework/app/models/backup.rb
```diff
@@ -188,9 +196,10 @@ def validate_chef_file_extension
   end
 
   def validate_version
-    if version < 1.9
+    min_version = PlatformVersion.new("1.9")
+    if backup_version < min_version
       errors.add(:base, I18n.t("backups.validation.version_too_low"))
-    elsif version > ENV["CROWBAR_VERSION"].to_f
+    elsif backup_version > system_version
```
bmwiedemann: if these are strings, it will say 
"1.9" > "1.10"
=> true

MaximilianMeister: normally yes, but they are [`PlatformVersion`](https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/platform_version.rb) objects:
```
irb(main):041:0> PlatformVersion.new("1.9") > PlatformVersion.new("1.10")
=> false
```

rhafer:
Would you mind squashing the two commits? Otherwise lgtm.

MaximilianMeister:
> Would you mind squashing the two commits? Otherwise lgtm.

@rhafer squashed

rhafer:
+1

tboerger:
:+1:

aplanas:
+1

PR #359, opened 2016-02-18
restore: call chef-client directly

MaximilianMeister:
single_chef_client.sh runs in the background, and though comes back
immediately. It is better to wait here and not go in the background.
The benefit here is that we can be sure that all necessary changes
after restore are applied, and the user can continue to use the cloud
immediately after restore.
This also helps unbreaking our CI where testsetup needs the chef-server
really quickly after restore is done.

### DIFF
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -205,7 +205,13 @@ module Crowbar
         end
 
         Rails.logger.info("Re-running chef-client locally to apply changes from imported proposals")
-        system("sudo", "-i", "/opt/dell/bin/single_chef_client.sh")
+        system(
+          "sudo",
+          "-i",
+          "/usr/bin/chef-client",
+          "-L",
+          "#{ENV["CROWBAR_LOG_DIR"]}/chef-client/#{NodeObject.admin_node.name}.log"
+        )
       end
 
       def restore_files(source, destination)


MaximilianMeister:
is there a way to call chef-client more elegantly by using `Chef::` classes directly? `Chef::Client.new.run` didn't do it somehow... not sure if it is even better

rhafer:
I suggest to have have it log ot a separate file. Something like what is used in ```apply_role``` ```"#{ENV['CROWBAR_LOG_DIR']}/chef-client/#{node.name}.log"```

In the end I am not entirely sure if this is the right solution. At least we ensure that the call will not return before chef-client is finished. Something that ```single_chef_client.sh``` doesn't. And AFAICT chef-client will nowadays take about the locking to avoid concurrent runs by itself.

cc @aspiers @vuntz

MaximilianMeister:
somehow we also would need to unbreak the CI to have some test covergae of backup/restore. without the fixes (this PR and https://github.com/crowbar/crowbar-core/pull/355) it is a bit difficult or unreliable.

does it make sense to add some workarounds in the meantime to mkcloud or a patch to the package?
unfortunately I think we need both patches applied to get gating green. even though https://ci.suse.de/job/openstack-mkcloud/17330 was green, not sure why it didn't fail there

MaximilianMeister:
I could also move the commit of this PR to the other one and do a manual gating, I just thought we should separate the discussion

MaximilianMeister:
how can we make some progress here? all backup/restore related gating for PR's is still blocked by this.

most recent example https://ci.suse.de/job/openstack-mkcloud/17596

rhafer:
@MaximilianMeister I guess we either need to convince @vuntz that this is important enough to be merged currently (which is probably a bit far fetched as it is not a release blocker) or add a workaround to mkcloud.

rhafer:
@MaximilianMeister btw, it's still missing change I suggested for logging (i.e. call chef-client -L logfile)

MaximilianMeister:
> btw, it's still missing change I suggested for logging (i.e. call chef-client -L logfile)

@rhafer yeah, thanks for reminding, but why exactly do we need to log into a separate file?

rhafer:
> but why exactly do we need to log into a separate file?

Without and option it ends up on stdout (i.e. production.out) which not very nice I think. Putting it into ```#{ENV['CROWBAR_LOG_DIR']}/chef-client/#{node.name}.log``` means to have it where it ends up for all the other crowbar-triggered chef-client runs as well.

MaximilianMeister:
> Without and option it ends up on stdout (i.e. production.out) which not very nice I think. Putting it into #{ENV['CROWBAR_LOG_DIR']}/chef-client/#{node.name}.log means to have it where it ends up for all the other crowbar-triggered chef-client runs as well.

thanks for the explanation, I have updated it (but currently still untested)

rhafer:
+1 (But please test this)

MaximilianMeister:
@rhafer just tested it and it works fine. blockui fades away right after the chef-client, that was triggered here, finishes on the admin node.

vuntz:
+1; do you want to fix the hound issue?

MaximilianMeister:
> do you want to fix the hound issue?

@vuntz fixed

MaximilianMeister:
btw I have triggered a backup/restore gating here, now that we have https://github.com/crowbar/crowbar-core/pull/355 merged hopefully it succeeds

vuntz:
In the future, we should probably isolate all chef-client calls in one method (because we have slight variations here and there).

But +1 for now.

vuntz:
@MaximilianMeister hrm. It seems gating failed in the very same way the jenkins job failed :/

MaximilianMeister:
> It seems gating failed in the very same way the jenkins job failed

:-( what is even stranger is that `/var/log/crowbar/chef-client/` is empty

MaximilianMeister:
```
I, [2016-02-23T13:32:28.322421 #15402:0x00000004e154e8]  INFO -- Re-running chef-client locally to apply changes from imported proposals
D, [2016-02-23T13:32:28.527294 #15402:0x00000004e154e8] DEBUG -- Restoring Crowbar database
```

is an important bit here. it seems there was no waiting for the chef-client...

MaximilianMeister:
maybe i was too quick. I think the rpm was not published when I retriggered the gating... :-(

vuntz:
Cool, still +1.

bmwiedemann:
+1

PR #360, opened 2016-02-19
Ensure SW RAID is stopped and device signatures really wiped out.

jsuchome:
Attempt to https://bugzilla.suse.com/show_bug.cgi?id=943623

This seems to help when installing over previous RAID disks.


jsuchome:
So yes, this does not solve the bug completely, because there are situations when installation fails even when there was no RAID previously on that node.

However, it helps partially, when we had RAID before and are reinstalling (or forgetting + descovering + allocating while using same disks)

aplanas:
the other cases are responsibility of yast?

jsuchome:
YaST or some other tool YaST is using

vuntz:
+1

aplanas:
+1

PR #361, opened 2016-02-19
delete backup from filesystem after restore (bsc#966489)

MaximilianMeister:


rsalevsky:
Instead of deleting it we should add it to the database again after restore. This makes IMHO more sense and it's the same amount of work.

rsalevsky:
Just got a other idea, we can just amend the current backup table dump to the yaml_db file in the backup.  Due to we don't backup the backup table we should have no conflict.

tboerger:
Needs rebase but otherwise :+1:

MaximilianMeister:
rebased

jsuchome:
Tested, seems to work well +1

vuntz:
Was there some green gating before we branched?

vuntz:
+1, but need some green gating.

MaximilianMeister:
> but need some green gating.

retriggered

rsalevsky:
+1

vuntz:
Can you backport to 3.0?

PR #363, opened 2016-02-19
network: Don't modify ifs mash when setting node attributes

vuntz:
We need to duplicate the mash because we're changing an attribute of the
variable we set to this mash -- which means, when there's no dup, that
we change the mash too.

Here, it's actually really not intended because it means we move from
IP::IP4 objects to strings. Noone noticed because the rest of the code
can live with both, but it's not intended.

Solves part of https://bugzilla.suse.com/show_bug.cgi?id=967489 (but by
luck): the order of the IP addresses for the interface is not a sorted
one, but the order in which networks are defined. It turns out that
admin is defined before bastion, so it works.


dirkmueller:
+1

tboerger:
Ok... :+1:

rhafer:
+1

PR #364, opened 2016-02-19
provisioner: Force server-identifier option for dhcp subnet (bsc#967489)

vuntz:
This way, we don't rely on dhcpd guessing the right IP address to use,
which might be wrong in case multiple addresses are set on the
interface.

https://bugzilla.suse.com/show_bug.cgi?id=967489


toabctl:
+1

aplanas:
+1

PR #365, opened 2016-02-21
crowbar: Delete client for node when node is in discovering (bsc#967100)

vuntz:
When a node enters discovering state, there's no reason for it to be
able to re-use an old client for chef. So destroy any pre-existing one.

https://bugzilla.suse.com/show_bug.cgi?id=967100


toabctl:
+1

dirkmueller:
+1

PR #369, opened 2016-02-22
crowbar: Do not use @logger in forks where it's not available anymore

vuntz:
This is what is causing these lines in the logs:
  log writing failed. closed stream

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1588,14 +1588,22 @@ class ServiceObject
   private
 
   def wait_for_chef_clients(node_name, options = {})
-    options = if options.fetch(:logger)
+    use_logger = options.fetch(:logger, nil)
+    options = if use_logger
       {logger: @logger}
     else
       {}
     end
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/service_object.rb
```diff
@@ -1588,14 +1588,22 @@ def wait_for_chef_daemons(node_list)
...
+    if use_logger
+      @logger.debug("wait_for_chef_clients: Waiting for already running chef-clients on #{node_name}.")
+    end
     unless RemoteNode.chef_ready?(node_name, 1200, 10, options)
-      @logger.error("Waiting for already running chef-clients on #{node_name} failed.")
+      msg = "Waiting for already running chef-clients on #{node_name} failed."
+      if use_logger
+        @logger.error(msg)
+      else
+        puts(msg)
```
MaximilianMeister: so not even `Rails.logger` is available here?

vuntz: Hrm, I don't know actually, but the intent is to actually log (through stdout) to a different file. Which is what we already do in `RemoteNode.chef_ready?`. That's why I put it that way.

PR #370, opened 2016-02-22
fix broken link to documentation resources

MaximilianMeister:
since the docs_controller is now activated we can link to the view

### DIFF
--- a/crowbar_framework/app/views/installer/installers/show.html.haml
+++ b/crowbar_framework/app/views/installer/installers/show.html.haml
@@ -22,6 +22,6 @@
     = t(".documentation")
   %ul.list-unstyled
     %li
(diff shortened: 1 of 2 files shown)


vuntz:
Hrm, but the docs view has a navigation bar, doesn't it?

MaximilianMeister:
closing for now in favor patching the branding in https://github.com/SUSE-Cloud/crowbar-core/commit/9acea36234807af4ee3cba00897706366ea37a7e

vuntz:
For the record, another issue is that the page redirects to the installer in install mode ;-)

MaximilianMeister:
@vuntz ouh... then we need to skip the filter for the docs controller, i think we already had that once and took it out again... another issue is do you think it makes sense to add a link to the release notes as well in the branding?

tboerger:
Why have this been closed?

PR #371, opened 2016-02-23
Stop dnsmasq to avoid dns queries on admin network during restore and resume later (bsc#966419)

mmnelemane:
Fixing https://bugzilla.suse.com/show_bug.cgi?id=966419 without iptables rules and instead disable and re-enable dnsmasq service letting only the nameserver listen on 127.0.0.1 during restore. An additional state file "disable_dns" is added since the file "restore_steps" waits for completion of chef-client while enabling dns service can be done immediately after the chef-server is running.

### DIFF
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -360,26 +360,9 @@ admin_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin"
 # the nameserver to just the loopback interface while restoring. That however
 # needs coordination with dnsmasq which might be listening there already.
 if node["crowbar"]["admin_node"] && ::File.exist?("/var/lib/crowbar/install/restore_steps")
-  admin_if = node[:crowbar_wall][:network][:nets][:admin].first
-  bash "block incoming DNS request during install/upgrade" do
-    code <<-EOH
-      iptables -I INPUT -i #{admin_if} -p udp -m udp --dport 53 \
-        -j REJECT -m comment --comment "UPGRADE_DNSBLOCK"
-      iptables -I INPUT -i #{admin_if} -p tcp -m tcp --dport 53 \
-        -j REJECT -m comment --comment "UPGRADE_DNSBLOCK"
-    EOH
-    not_if "iptables -L INPUT | grep -q UPGRADE_DNSBLOCK"
-  end
-elsif node["crowbar"]["admin_node"]
-  admin_if = node[:crowbar_wall][:network][:nets][:admin].first
-  bash "remove upgrade dns block" do
-    code <<-EOH
-      iptables -D INPUT -i #{admin_if} -p udp -m udp --dport 53 \
-        -j REJECT -m comment --comment "UPGRADE_DNSBLOCK"
-      iptables -D INPUT -i #{admin_if} -p tcp -m tcp --dport 53 \
-        -j REJECT -m comment --comment "UPGRADE_DNSBLOCK"
-    EOH
-    only_if "iptables -L INPUT | grep -q UPGRADE_DNSBLOCK"
+  admin_addr = "127.0.0.1"
+  service "dnsmasq" do
+    action [:stop, :disable]
   end
 end
 
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -356,12 +356,10 @@ admin_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin"
 # When we're restoring the admin node from backup or upgrade data,
 # reject incoming DNS traffic to avoid sending wrong results to running
 # clients.
-# FIXME: A cleaner approach would be to restrict the listener address of
-# the nameserver to just the loopback interface while restoring. That however
-# needs coordination with dnsmasq which might be listening there already.
-if node["crowbar"]["admin_node"] && ::File.exist?("/var/lib/crowbar/install/restore_steps")
+if node["crowbar"]["admin_node"] && ::File.exist?("/var/lib/crowbar/install/disable_dns")
   admin_addr = "127.0.0.1"
-  service "dnsmasq" do
+  service "stop dnsmasq during restore" do
+    service_name "dnsmasq"
     action [:stop, :disable]
   end
 end
--- a/chef/cookbooks/resolver/recipes/default.rb
+++ b/chef/cookbooks/resolver/recipes/default.rb
@@ -57,6 +57,7 @@ unless node[:platform_family] == "windows"
         # invalidate dnsmasq cache if local zone changes
         subscribes :reload, "template[/etc/bind/db.#{node[:dns][:domain]}]"
       end
+      not_if { ::File.exist?("/var/lib/crowbar/install/restore_steps") }
     end
 
     dns_list = dns_list.insert(0, "127.0.0.1").take(3)
--- a/chef/cookbooks/resolver/recipes/default.rb
+++ b/chef/cookbooks/resolver/recipes/default.rb
@@ -57,7 +57,7 @@ unless node[:platform_family] == "windows"
         # invalidate dnsmasq cache if local zone changes
         subscribes :reload, "template[/etc/bind/db.#{node[:dns][:domain]}]"
       end
-      not_if { ::File.exist?("/var/lib/crowbar/install/restore_steps") }
+      not_if { node["crowbar"]["admin_node"] && ::File.exist?("/var/lib/crowbar/install/disable_dns") }
     end
 
     dns_list = dns_list.insert(0, "127.0.0.1").take(3)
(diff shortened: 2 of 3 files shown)


rhafer:
The title of the commit is somewhat missleading. We're not aiming to stop the dns server here. The goal is to avoid having it answering queries on the admin network.

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -360,26 +360,8 @@ def make_zone(zone)
...
-      iptables -D INPUT -i #{admin_if} -p udp -m udp --dport 53 \
-        -j REJECT -m comment --comment "UPGRADE_DNSBLOCK"
-      iptables -D INPUT -i #{admin_if} -p tcp -m tcp --dport 53 \
-        -j REJECT -m comment --comment "UPGRADE_DNSBLOCK"
-    EOH
-    only_if "iptables -L INPUT | grep -q UPGRADE_DNSBLOCK"
+  service "dnsmasq" do
+    action [:stop, :disable]
   end
 end
```
rhafer: While stopping dnsmasq is correct, you're still missing the part the reconfigures named to only listen on the loopback interface.

I think you can do that by setting "admin_addr" to 127.0.0.1 here. That will cause the named.conf to be reconfigured further down in this recipe.

rhafer: Just to clarify. Yes, the admin server runs both dnsmasq and bind.

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -359,27 +359,10 @@ def make_zone(zone)
...
-    code <<-EOH
-      iptables -D INPUT -i #{admin_if} -p udp -m udp --dport 53 \
-        -j REJECT -m comment --comment "UPGRADE_DNSBLOCK"
-      iptables -D INPUT -i #{admin_if} -p tcp -m tcp --dport 53 \
-        -j REJECT -m comment --comment "UPGRADE_DNSBLOCK"
-    EOH
-    only_if "iptables -L INPUT | grep -q UPGRADE_DNSBLOCK"
+if node["crowbar"]["admin_node"] && ::File.exist?("/var/lib/crowbar/install/disable_dns")
+  admin_addr = "127.0.0.1"
+  service "dnsmasq" do
```
rhafer: Aren't you getting chef warning about duplicate resources names being used here? Because ```service "dnsmasq"``` is already used in resolver/recipes/default.rb ?

I'd suggest to rename this one here to:

```
service "stop dnsmasq during restore" do
    service_name "dnsmasq"
    action [:stop, :disable]
end
```

rhafer:
@MaximilianMeister could you please review the changes to restore.rb in here?

rhafer:
@mmnelemane Please add the bsc reference to the commit's subject (```bsc#966419```). Also some text in the commit message describing what this change is about would be helpful (including a link to the bug report)

mmnelemane:
@rhafer : Thanks. I just updated the title and service names as per your comment.

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -359,27 +359,11 @@ def make_zone(zone)
...
-  admin_if = node[:crowbar_wall][:network][:nets][:admin].first
-  bash "remove upgrade dns block" do
-    code <<-EOH
-      iptables -D INPUT -i #{admin_if} -p udp -m udp --dport 53 \
-        -j REJECT -m comment --comment "UPGRADE_DNSBLOCK"
-      iptables -D INPUT -i #{admin_if} -p tcp -m tcp --dport 53 \
-        -j REJECT -m comment --comment "UPGRADE_DNSBLOCK"
-    EOH
-    only_if "iptables -L INPUT | grep -q UPGRADE_DNSBLOCK"
+if node["crowbar"]["admin_node"] && ::File.exist?("/var/lib/crowbar/install/disable_dns")
```
MaximilianMeister:
```
if node["crowbar"]["admin_node"] &&  Crowbar::Backup::Restore.disable_dns_path.exist?
```

chef/cookbooks/resolver/recipes/default.rb
```diff
@@ -57,6 +57,7 @@
         # invalidate dnsmasq cache if local zone changes
         subscribes :reload, "template[/etc/bind/db.#{node[:dns][:domain]}]"
       end
+      not_if { ::File.exist?("/var/lib/crowbar/install/disable_dns") }
```
MaximilianMeister:
```
not_if { Crowbar::Backup::Restore.disable_dns_path.exist? }
```

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -204,6 +214,8 @@ def restore_chef
           raise "Restore failed"
         end
 
+        # allow enable dns after restore.
+        disable_dns_path.delete if disable_dns_path.exist?
```
MaximilianMeister: can you please add an empty line here?

chef/cookbooks/bind9/recipes/default.rb
rhafer: Is Crowbar::Backup::Restore even available from within a chef recipe? I doubt.

rhafer:
> Thanks. I just updated the title and service names as per your comment.

@mmnelemane Actually I was referring to the individual commits  themselves not the title of the PR.

chef/cookbooks/bind9/recipes/default.rb
MaximilianMeister: right, in chef we don't have the Crowbar libs... :(

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -36,6 +36,11 @@ def restore_background
       def restore
         cleanup if self.class.restore_steps_path.exist?
 
+        # disable/stop dns service before starting restore
+        self.class.disable_dns_path.open("a") do |f|
+          f.write "#{Time.zone.now.iso8601}\n"
```
MaximilianMeister: maybe we could even add the name of the backup from which the restore was triggered

and why do you append instead of just "w"?

mmnelemane: @MaximilianMeister: i will try updating the backup name and yes "w" should do here. Thanks.

crowbar_framework/lib/crowbar/backup/restore.rb
MaximilianMeister: @mmnelemane I was also wondering if we could put this into the cleanup method, or do you think that has to be deleted before restoring the database?

mmnelemane: @MaximilianMeister : The idea was to delete the file before the chef-client is run. Thats why I put it outside cleanup().

MaximilianMeister: > The idea was to delete the file before the chef-client is run. Thats why I put it outside cleanup().

okay thanks for the explanation. with the fixes mentioned above it looks good to me!

and if this fixes a bug can you please reference it in the commit? just add the (bsc#966419) and the link to bugzilla

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -359,27 +359,11 @@ def make_zone(zone)
 # FIXME: A cleaner approach would be to restrict the listener address of
 # the nameserver to just the loopback interface while restoring. That however
 # needs coordination with dnsmasq which might be listening there already.
```
vuntz: Can you remove the FIXME?

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -36,6 +36,11 @@ def restore_background
       def restore
         cleanup if self.class.restore_steps_path.exist?
 
+        # disable/stop dns service before starting restore
```
vuntz: "restrict dns server to not answer requests from other nodes (because...)"

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -204,6 +214,9 @@ def restore_chef
           raise "Restore failed"
         end
 
+        # allow enable dns after restore.
```
vuntz: "now that restore is done, dns server can answer requests fro other nodes"

rhafer:
+1

MaximilianMeister:
Thanks for the changes
+1

chef/cookbooks/resolver/recipes/default.rb
vuntz: This should be the same condition as in the other recipe: `node["crowbar"]["admin_node"] && ::File.exist?("/var/lib/crowbar/install/disable_dns")`

vuntz:
Minor nitpick, +1 after that.

chef/cookbooks/resolver/recipes/default.rb
mmnelemane: @vuntz: Thanks. Done.

mmnelemane:
@vuntz: Fixed it.. thanks.

vuntz:
+1, thanks

rhafer:
Hm, ci failed in "Manila Client tests". Seems to be unrelated to this PR.

MaximilianMeister:
still +1

PR #372, opened 2016-02-23
Fix data type for "version" in Backup

rhafer:
We're treating version as a string since 3e3bc1fa33de7f83f77adcd88b8f2a23011bfd0e.

### DIFF
--- a/crowbar_framework/app/models/backup.rb
+++ b/crowbar_framework/app/models/backup.rb
@@ -83,7 +83,10 @@ class Backup < ActiveRecord::Base
       return false
     end
 
-    upgrade if upgrade?
+    if upgrade? && !upgrade
+      return false
+    end
+
     if background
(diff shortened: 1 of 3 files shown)


MaximilianMeister:
so the commit was after the release I guess, then this makes total sense. thanks

+1

dirkmueller:
string wise comparison of version numbers?

rhafer:
> string wise comparison of version numbers?

For comparing the versions ```PlatformVersion``` is used (https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/platform_version.rb). But we store the versions as strings in the database (which I is the only sane choice I think).

dirkmueller:
@rhafer thanks for the explanation. LGTM

vuntz:
+1

PR #373, opened 2016-02-23
provisioner: Tell yast to use text mode if serial console is used

vuntz:
This enables seeing something in the console.


jsuchome:
+1

dirkmueller:
+1

PR #377, opened 2016-02-24
Fleshed out UCS emulator documentation

jgrassler:
* Documented network settings for emulator
* Added example Hardware Inventory along with instructions on adding it
* Added walkthrough documentation for minimal service profile templates
* Documented limits of emulation (no simulated "physical" hardware running in
  the emulator VM)
* Removed unneccesary 20G disk for emulator VM from documentation and libvirt
  example.
* Some Markdown formatting tweaks
* Moved all material related to the UCS emulator to the ucspe/ subdirectory.


MaximilianMeister:
Could you add the screenshots also inline? with the `alt tag` syntax?
also I think the rebase to master went wrong, you have a merge commit in there.

jgrassler:
> Could you add the screenshots also inline? with the alt tag syntax?

Ok, done.

> also I think the rebase to master went wrong, you have a merge commit in there.

Interesting. That's the first time I used the 'Update branch' button...does it usually do that? (Just for future reference, fixed manually now).

crowbar_framework/public/docs/cisco_ucs.md
```diff
@@ -77,26 +77,145 @@ The UCS Platform Emulator can be [downloaded from Cisco's
...
+`suse-cloud-storage` templates described above govern the machine pools
+machines for a given service template are drawn from and the properties
+machines need to have (i.e. in cases where they are drawn from heterogenous
+machine pools with multiple available configurations). This section will walk
+you through the service creation wizard. Since you will not deploy actual
+machines, most settings in the wizard are not important for the emulator's
+purpose. There are a few identifiers such as MAC addresses and UUIDs that are
+drawn from finite pools that start out empty and must be defined manually.
+
+_Note: On dialog steps that are not mentioned below you can simply hit the
```
MaximilianMeister: the formatting here is still a bit broken

MaximilianMeister:
> Interesting. That's the first time I used the 'Update branch' button...does it usually do that? (Just for future reference, fixed manually now).

yeah when you use the github buttons you will get those commits, usually we do it from the command line and force push to the remote branch to avoid those commits.

MaximilianMeister:
other than that the docs look good to me (although I can't say much about the functionality, as I have never tested it)

but you might want to add it in `config/docs/core.yml` if it should  be accessible over the crowbar UI (not sure if that is wanted)

MaximilianMeister:
> but you might want to add it in config/docs/core.yml if it should be accessible over the crowbar UI (not sure if that is wanted)

btw I just saw that we are patching the links out for the SOC branding. so if you add it here upstream in core.yml, for our branding you would probably have to patch it out again in https://build.suse.de/package/view_file/Devel:Cloud:6:Staging/crowbar-core-branding-SOC/suse-branding.patch

jgrassler:
> the formatting here is still a bit broken

Ok, bad formatting is fixed. For future reference: Markdown syntax highlighting in vim can cope with nested italics, while Github's Markdown converter evidently can't (fixed-width inside italics works fine).



>> but you might want to add it in config/docs/core.yml if it should be accessible over the crowbar UI >> (not sure if that is wanted)

> btw I just saw that we are patching the links out for the SOC branding. so if you add it here upstream in core.yml, for our branding you would probably have to patch it out again in https://build.suse.de/package/view_file/Devel:Cloud:6:Staging/crowbar-core-branding-SOC/suse-branding.patch

I didn't touch config/docs/core.yml in this pull request, but it is already in
there anyway. Not on a cloud built from packages, though. So it looks like it's
getting patched out.

MaximilianMeister:
> but it is already in there anyway.

ouh right, I have overlooked that... sorry for the noise

rsalevsky:
For the screenshots it would be nice if you can use a real rectangle. ;)

jgrassler:
What do you mean by "real rectangle"? All screenshots normalized to be regular and/or the same size?

rsalevsky:
@jgrassler I mean your blue and red markers.

MaximilianMeister:
> For the screenshots it would be nice if you can use a real rectangle. ;)

I would keep them like they are. they add some decent human touch :-)

dirkmueller:
LGTM

dirkmueller:
imho its good enough. I don't see any objections, so merging.

PR #379, opened 2016-02-25
Look for db_dump_location into database proposal

jsuchome:
Needed with https://github.com/crowbar/crowbar-openstack/pull/308


rhafer:
I guess this needs at least a migration to remove the attribute from the crowbar proposal again.

And do we need to handle the case where somebody changed that value in the proposal? I assume not, because anybody who did that will most probably have the upgrade completed already.

PR #380, opened 2016-02-25
Save the location that was used for DB dump to node.

jsuchome:
Alternative solution to https://github.com/crowbar/crowbar-openstack/pull/308

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -56,20 +56,22 @@ when "dump_openstack_database"
 
   include_recipe "crowbar::stop-services-before-upgrade"
 
+  dump_location = node[:crowbar][:upgrade][:db_dump_location]
+
+  # Ensure the dump_location is saved in DB nodes even after we later drop the role
+  node.set[:crowbar][:upgrade][:db_dump_location] = dump_location
+
   # If postgres is not running here, it means we're in the DB cluster and database runs
   # on the other node: let the other node take care of the rest.
   unless ::Kernel.system("service postgresql status")
     if node[:crowbar][:upgrade][:db_dumped_here]
       node[:crowbar][:upgrade].delete :db_dumped_here
-      node.save
     end
+    node.save
     return
   end
 
   # Check the available space before the dump
-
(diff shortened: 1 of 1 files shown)


jsuchome:
I actually like this more than the alternative, as it requires less changes to current code.

Ther might be a problem though, that this information is saved only on one node, and in cluster, db might be started on the other one... but maybe this should not be problem, restore could still be done from the other db node, right?

rhafer:
> restore could still be done from the other db node, right?

No. It can only be done on the node where the db is started (and where the filesystem is mounted)

rhafer:
I am currently -1 on this. I think using a default attribute on the postgresql cookbook is a cleaner solution.

jsuchome:
> No. It can only be done on the node where the db is started

Now it should be saved in all cluster nodes.

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -56,20 +56,26 @@
 
   include_recipe "crowbar::stop-services-before-upgrade"
 
+  dump_location = node[:crowbar][:upgrade][:db_dump_location]
+  save_node = false
+
+  unless node[:crowbar][:upgrade][:db_dump_location]
```
vuntz: Oh, hrm, your check won't work because the attribute is existing through the role. You probably need to check if it's set on the node directly. Or just rename the attribute?

vuntz: Also, can you add a comment explaining why we do that?

jsuchome:
So I've tested with HA and non-HA setups, and with https://github.com/crowbar/crowbar-core/pull/344 included, my database was correctly restored and upgrade roles are gone.

vuntz:
> So I've tested with HA and non-HA setups, and with #344 included, my database was correctly restored and upgrade roles are gone.

But wasn't this just by luck because the database started on the same node it was running before?

vuntz:
@jsuchome clarified that I'm wrong on IRC :-) So +1.

vuntz:
@rhafer what do you think?

rhafer:
+1 (since the solution via default attribute on the postgresql cookbook doesn't work)

PR #381, opened 2016-02-25
enhance restore thread logging

MaximilianMeister:
at least in the case a thread disappears somehow we would have
logged the name, id, and status of the thread

### DIFF
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -29,6 +29,7 @@ module Crowbar
 
       def restore_background
         @thread = Thread.new do
+          Rails.logger.debug("Starting restore in background thread")
           restore
         end
       end
@@ -42,7 +43,11 @@ module Crowbar
           if any_errors?
             cleanup
             @backup.errors.add(:restore, error_messages.join(" - "))
-            Thread.exit if @thread
+            Rails.logger.error("Restore failed: #{@backup.errors.full_messages.first}")
+            if @thread
+              Rails.logger.debug("Exiting restore thread due to failure")
+              Thread.exit
+            end
             return false
           end
           # set_failed is called directly after the fail


crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -28,9 +28,11 @@ def initialize(backup)
       end
 
       def restore_background
+        Rails.logger.debug("Starting restore in the background")
         @thread = Thread.new do
           restore
         end
+        Rails.logger.debug("Started restore thread: #{@thread.inspect}")
```
vuntz: Maybe move one of the debug lines to the thread, before the call to `restore`?

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -28,9 +28,12 @@ def initialize(backup)
       end
 
       def restore_background
+        Rails.logger.debug("Starting restore in the background")
         @thread = Thread.new do
+          Rails.logger.debug("Entering thread: #{Thread.current.inspect}")
```
MaximilianMeister: > Maybe move one of the debug lines to the thread, before the call to restore?

@vuntz more or less like this?

vuntz: Well, that's three log lines -- I'm not sure we need that much. I think just one inside the thread saying "Starting restore in background thread" is enough. And no need to give info about the thread, the prefix in the log gives that already.

The important bit to me is to have one line at the beginning of the thread so that we can identify in the log which thread is responsible for restoring.

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -42,7 +43,15 @@ def restore
           if any_errors?
             cleanup
             @backup.errors.add(:restore, error_messages.join(" - "))
-            Thread.exit if @thread
+            Rails.logger.error("Restore failed: #{@backup.errors.full_messages.first}")
+            if @thread
+              Rails.logger.debug(
+                "Exiting thread: #{@thread.inspect} \
+                Backtrace: #{@thread.backtrace} \
+                Alive: #{@thread.alive?}"
```
MaximilianMeister: @vuntz are you fine with keeping this? or is it too verbose?

crowbar_framework/lib/crowbar/backup/restore.rb
MaximilianMeister: updated

crowbar_framework/lib/crowbar/backup/restore.rb
vuntz: I'm not sure the backtrace and alive bits would bring us something we don't already know (or am I missing something)? I would keep the "exiting thread" bit -- but maybe just have it be "Exiting restore thread due to failure".

vuntz:
Thanks, that will be useful. +1.

rsalevsky:
+1

vuntz:
Can you backport to 3.0?

PR #382, opened 2016-02-29
Allow subsequent restores after failure during upgrade (bsc#967848)

MaximilianMeister:
Instead of a `Retrigger restore` button it was discussed and suggested by @kwwii to simply redirect to the start upgrade page and repeat the process after the conflict is resolved.
After the redirect an error message with details how to solve the conflict is displayed.

**DEPENDS ON** https://github.com/crowbar/crowbar-core/pull/361 where the backup is deleted from the filesystem...

### DIFF
--- a/crowbar_framework/app/controllers/installer/upgrades_controller.rb
+++ b/crowbar_framework/app/controllers/installer/upgrades_controller.rb
@@ -228,6 +228,7 @@ module Installer
         flash[:info] = t("installer.upgrades.restore.multiple_restore")
         true
       else
+        Crowbar::Backup::Restore.purge
         @backup.restore(background: true, from_upgrade: true)
       end
     end
--- a/crowbar_framework/app/controllers/installer/upgrades_controller.rb
+++ b/crowbar_framework/app/controllers/installer/upgrades_controller.rb
@@ -182,9 +182,14 @@ module Installer
     end
 
     def restore_status
+      @status = Crowbar::Backup::Restore.status
+
       respond_to do |format|
         format.json do
-          render json: Crowbar::Backup::Restore.status
+          if @status[:failed]
+            flash[:alert] = t("installer.upgrades.restore.failed")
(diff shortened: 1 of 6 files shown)


vuntz:
@MaximilianMeister have you tried it? I'd hate to assume that this fixes the bug without being sure that we're purging all the files that can create issues..

MaximilianMeister:
> have you tried it? I'd hate to assume that this fixes the bug without being sure that we're purging all the files that can create issues..

with #361 all files should be purged, even after a successful restore I could retrigger the restore (if it is that what you mean)

tboerger:
Not beautiful but seems to work.
:+1:

jsuchome:
+1 (tested)

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -223,6 +228,7 @@ def save_and_restore
         flash[:info] = t("installer.upgrades.restore.multiple_restore")
         true
       else
+        Crowbar::Backup::Restore.purge
```
vuntz: I'm a bit confused here: we are purging the restore steps file with this, but the test in the `if` condition is checking if it exists. Can't this cause a problem? Or is the restore steps file remove on failure anyway?

MaximilianMeister: > Can't this cause a problem?

no as it is double checked if the file exists before deleting

>  Or is the restore steps file remove on failure anyway?

https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/backup/restore.rb#L44

but i see your confusion, I think we could bring together the `cleanup` method with `purge` somehow, but I'd prefer to do that afterwards with a better overview when all related PRs are merged

vuntz:
Can you backport to 3.0?

PR #383, opened 2016-02-29
Add user notification about wiping attached disks during Allocate Node(bsc#845602)

mmnelemane:
Workaround fix for https://bugzilla.suse.com/show_bug.cgi?id=845602
Adds notification to user that the shared disks will be wiped during allocate (hardware and OS installation) of a discovered node.
Link to the UI Screenshot with the popup: 
screenshot_20160229_153628

### DIFF
--- a/crowbar_framework/app/views/nodes/_form.html.haml
+++ b/crowbar_framework/app/views/nodes/_form.html.haml
@@ -9,7 +9,7 @@
           %input.btn.btn-default{ :type => "submit", :name => "submit", :value => t(".save") }
 
           - unless @node.allocated? or @node.admin?
-            %input.btn.btn-default{ :type => "submit", :name => "submit", :value => t(".allocate") }
+            %input.btn.btn-default{ type: "submit", name: "submit", value: t(".allocate"), data: { confirm: t(".confirm_allocate"), title: t(".allocate_tooltip") } }
 
     .panel-body
       - if @node.allocated?
@@ -127,4 +127,4 @@
         %input.btn.btn-default{ :type => "submit", :name => "submit", :value => t(".save") }
 
         - unless @node.allocated? or @node.admin?
-          %input.btn.btn-default{ :type => "submit", :name => "submit", :value => t(".allocate") }
+          %input.btn.btn-default{ type: "submit", name: "submit", value: t(".allocate"), data: { confirm: t(".confirm_allocate"), title: t(".allocate_tooltip") } }
--- a/crowbar_framework/config/locales/crowbar/en.yml
+++ b/crowbar_framework/config/locales/crowbar/en.yml
@@ -441,6 +441,8 @@ en:
       identify: 'Identify'
     form:
       allocate: 'Allocate'
+      confirm_allocate: 'Allocating the node will format all the disks attached to this node, including shared disks. Do you want to proceed and allocate this node?'
+      allocate_tooltip: 'Allocate this node according to the parameters in this form'
       repo_hint: 'To find out why a platform is disabled, check the <a href="%{url}">status of repositories</a>.'
       save: 'Save'
       raid: 'RAID'


rsalevsky:
UI change so needs approval from @kwwii.

kwwii:
While the pop-up is good, the checkbox with "Prevent this page from creating additional dialogs" would ruin the UX as the user can easily turn future warnings off. It would be better to use a bootstrap modal like the rest of crowbar

MaximilianMeister:
> While the pop-up is good, the checkbox with "Prevent this page from creating additional dialogs" would ruin the UX as the user can easily turn future warnings off.

can we prevent that checkbox for now (not sure if it can be configured), and then in a separate PR replace all the `data-confirm` messages with bootstrap modals? then we have a bugfix for now, and not one modal here and the other places still the standard javascript popup.

>  It would be better to use a bootstrap modal like the rest of crowbar

@kwwii the modals that we have are for more complex dialogs, not yet for the simple confirm messages

mmnelemane:
@kwwii, @MaximilianMeister : I dont find a way to disable the check box since it is a browser specific feature. (data-confirm messages use alert() call which forces the checkbox)

MaximilianMeister:
> I dont find a way to disable the check box since it is a browser specific feature. (data-confirm messages use alert() call which forces the checkbox)

@mmnelemane can you make them use`confirm()` somehow? at least a `confirm()` doesn't offer a checkbox

mmnelemane:
@MaximilianMeister : it seems that this feature is common to "alert/confirm/prompt". So the only way is to change is a preference setting from the client (only possible in firefox) and cannot be done by the application. Not sure if i missed something.

MaximilianMeister:
oh right, that is weird, after trying `confirm("foo")` again in the browser console it suddenly shows the checkbox... anyways as the `data-confirm` is rails built in we don't want to change anything.

could you then create a trello card for replacing the remaining occurences of the confirm dialog with a bootstrap modal? I am fine with introducing it here, it's just not to forget to change it in the other places as well

mmnelemane:
@MaximilianMeister : Here is the card for that: https://trello.com/c/jioLvzFi/979-replace-data-confirm-messages-with-modals-in-crowbar-ui

kwwii:
@MaximilianMeister Which other broswer pop-ups do we have in Crowbar? I am only aware of the login.

mmnelemane:
@kwwii: AFAICS, Reinstall, Deallocate, Forget also use similar popups. @MaximilianMeister : Not sure if I missed few more.

crowbar_framework/app/views/nodes/_form.html.haml
```diff
@@ -127,4 +127,4 @@
         %input.btn.btn-default{ :type => "submit", :name => "submit", :value => t(".save") }
 
         - unless @node.allocated? or @node.admin?
-          %input.btn.btn-default{ :type => "submit", :name => "submit", :value => t(".allocate") }
+          %input.btn.btn-default{ :type => "submit", :name => "submit", :value => t(".allocate"), "data-remote" => true, "data-confirm" => t(".confirm_allocate"), :title => t(".allocate_tooltip") }
```
tboerger: Remove the remote and use this instead:

```
%input.btn.btn-default{ type: "submit", name: "submit", value: t(".allocate"), data: { confirm: t(".confirm_allocate"), title: t(".allocate_tooltip") }
```

crowbar_framework/app/views/nodes/_form.html.haml
```diff
@@ -9,7 +9,7 @@
           %input.btn.btn-default{ :type => "submit", :name => "submit", :value => t(".save") }
 
           - unless @node.allocated? or @node.admin?
-            %input.btn.btn-default{ :type => "submit", :name => "submit", :value => t(".allocate") }
+            %input.btn.btn-default{ :type => "submit", :name => "submit", :value => t(".allocate"), "data-remote" => true, "data-confirm" => t(".confirm_allocate"), :title => t(".allocate_tooltip") }
```
tboerger: Remove the remote and use this instead:

```
%input.btn.btn-default{ type: "submit", name: "submit", value: t(".allocate"), data: { confirm: t(".confirm_allocate"), title: t(".allocate_tooltip") }
```

MaximilianMeister: why do we need `data-remote`?

tboerger:
@kwwii those are the confirms accross all barclamps

```
# rgrep confirm * | grep -v cookbooks | grep -v rails.js | grep -v en.yml | grep -v backups_controller | grep -v _nav.html | grep -v filter.rb
core/crowbar_framework/app/views/barclamp/_node_selector.html.haml:83:              %a.unassign{ :href => "#", "data-id" => element, "data-confirm" => t(".unassign_confirm", :role => element) }
core/crowbar_framework/app/views/nodes/_buttons_chef.html.haml:2:  = link_to "Forget", hit_node_path(@node.handle, "delete"), :class => "btn btn-default", "data-remote" => true, "data-confirm" => t(".confirm_forget"), :title => t(".forget_tooltip")
core/crowbar_framework/app/views/nodes/_buttons_chef.html.haml:5:    = link_to "Hardware Update", hit_node_path(@node.handle, "update"), :class => "btn btn-default", "data-remote" => true, "data-confirm" => t(".confirm_hw_update"), :title => t(".hw_update_tooltip")
core/crowbar_framework/app/views/nodes/_buttons_chef.html.haml:8:    = link_to "Reinstall", hit_node_path(@node.handle, "reinstall"), :class => "btn btn-default", "data-remote" => true, "data-confirm" => t(".confirm_reinstall"), :title => t(".reinstall_tooltip")
core/crowbar_framework/app/views/nodes/_buttons_chef.html.haml:9:    = link_to "Deallocate", hit_node_path(@node.handle, "reset"), :class => "btn btn-default", "data-remote" => true, "data-confirm" => t(".confirm_deallocate"), :title => t(".deallocate_tooltip")
core/crowbar_framework/app/views/nodes/_buttons_power.html.haml:19:          = link_to t(".poweron"), hit_node_path(@node.handle, "poweron"), role: "menuitem", remote: true, "data-confirm" => t("are_you_sure")
core/crowbar_framework/app/helpers/proposals_helper.rb:61:      "data-confirm" => t(".destroy_changes")
core/crowbar_framework/app/helpers/proposals_helper.rb:72:      "data-confirm" => t(".delete_changes")
core/crowbar_framework/app/helpers/proposals_helper.rb:98:      "data-confirm" => t(".apply_changes"),
core/crowbar_framework/app/helpers/backups_helper.rb:32:      data: { confirm: t("are_you_sure") }
core/crowbar_framework/app/helpers/backups_helper.rb:42:      data: { confirm: t(".restore_warning"), blockui: t(".blockui_message") }
core/crowbar_framework/vendor/assets/javascripts/jquery/changeDetection.js:62:          if (!confirm(self.options.message)) {
core/crowbar_framework/public/docs/cisco_ucs.md:125:the Radio button to _Yes_ to confirm. Restarting should take 2-3 minutes. After
openstack/crowbar_framework/app/assets/javascripts/barclamps/manila/application.js:31:    if (confirm("All shares in the backend will be made unavailable; do you really want to delete this backend?")) {
openstack/crowbar_framework/app/assets/javascripts/barclamps/cinder/application.js:36:    if (confirm("All volumes in the backend will be made unavailable; do you really want to delete this backend?")) {
```

crowbar_framework/app/views/nodes/_form.html.haml
mmnelemane: @tboerger : Updated accordingly... Thanks.

dirkmueller:
LGTM, the wording is a bit ambiguous (Cleanup vs destroy in the two messages)

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -444,6 +444,8 @@ en:
       identify: 'Identify'
     form:
       allocate: 'Allocate'
+      confirm_allocate: 'WARNING! Allocating the node will wipe all the disks attached to the node including the shared disks. Proceed?'
+      allocate_tooltip: 'Cleanup and Allocate node with the intended role'
```
vuntz: I don't think you should mention "intended role" here as it might not be set. Also, what does "cleanup" mean?

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -444,6 +444,8 @@ en:
       identify: 'Identify'
     form:
       allocate: 'Allocate'
+      confirm_allocate: 'WARNING! Allocating the node will wipe all the disks attached to the node including the shared disks. Proceed?'
```
vuntz: I'm not a UI expert, but I think it's generally wrong to have a whole uppercase word (even for WARNING).

vuntz:
I'm fine with an approach like this for tex/3.0, but I think for master, it'd be better to rework the whole thing: we should make it possible to select in the UI which disk to use as boot disk, as well as which disks to wipe.

But we can still get this in in the meantime, of course.

crowbar_framework/config/locales/crowbar/en.yml
tboerger: I would drop the `WARNING!`.

tboerger:
Beside the `WARNING!` thing I'm fine with it.

mmnelemane:
Removed "WARNING!".

MaximilianMeister:
just a question about the tooltip, is it a bootstrap tooltip?

mmnelemane:
No. Its similar to other buttons.

mmnelemane:
rebased

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -444,6 +444,8 @@ en:
       identify: 'Identify'
     form:
       allocate: 'Allocate'
+      confirm_allocate: 'Allocating the node will wipe all the disks attached to the node including the shared disks. Proceed?'
+      allocate_tooltip: 'Allocate this node according to the parameters in this form'
```
vuntz: It really feels to me we're adding a tooltip just to have a tooltip. Is there any useful information that we can provide in the tooltip? If not, do we really need a tooltip?

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -444,6 +444,8 @@ en:
       identify: 'Identify'
     form:
       allocate: 'Allocate'
+      confirm_allocate: 'Allocating the node will wipe all the disks attached to the node including the shared disks. Proceed?'
```
vuntz: "Allocating the node will format all disks attached to this node, including shared disks. Do you want to proceed and allocate this node?"

mmnelemane: @vuntz: Done.

tboerger:
:+1:

crowbar_framework/config/locales/crowbar/en.yml
MaximilianMeister: I would drop the tooltip, but maybe @kwwii has an opinion here as well?

mmnelemane:
@vuntz , @dirkmueller : Is it a +1 now ?

dirkmueller:
+1

PR #385, opened 2016-03-02
Modal implementation for warning allocate node

mmnelemane:
Modal implementation to warn user on node allocate. This is an alternative to using data-confirm message popups which show a check box allowing the user to disable future popups. The implementation needs to be extended to other warning popups in the crowbar UI. Depends on https://github.com/crowbar/crowbar-core/pull/383
screenshot_20160302_120839

### DIFF
--- a/crowbar_framework/app/views/nodes/_form.html.haml
+++ b/crowbar_framework/app/views/nodes/_form.html.haml
@@ -9,7 +9,7 @@
           %input.btn.btn-default{ :type => "submit", :name => "submit", :value => t(".save") }
 
           - unless @node.allocated? or @node.admin?
-            %input.btn.btn-default{ :type => "submit", :name => "submit", :value => t(".allocate") }
+            %input.btn.btn-default{ :type => "submit", :name => "submit", :value => t(".allocate"), "data-remote" => true, "data-confirm" => t(".confirm_allocate"), :title => t(".allocate_tooltip") }
 
     .panel-body
       - if @node.allocated?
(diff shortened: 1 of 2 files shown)


crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -444,6 +444,8 @@ en:
       identify: 'Identify'
     form:
       allocate: 'Allocate'
+      allocate_node: 'Allocate Node'
+      confirm_allocate: 'WARNING! Allocating the node will wipe all the disks attached to the node including the shared disks. Proceed?'
```
MaximilianMeister: If you ask `Proceed?` you would expect `Ok` and `Cancel` . instead it says `Allocate`, so maybe drop the `Proceed?`

crowbar_framework/app/views/nodes/_form.html.haml
```diff
@@ -127,4 +128,23 @@
...
+          = t(".allocate_node")
+      = form_for :node, :url => update_node_path(@node.handle), class: "btn btn-danger", "data-remote" => true, :html => { :role => "form", "data-type" => "html", "data-method" => "put" } do |f|
+        .modal-body
+          %h3
+            = t(".confirm_allocate")
+        .modal-footer
+          .row
+            .btn.btn-default{ data: { dismiss: "modal" } }
+              = t("cancel")
+            = f.submit(class: "btn btn-danger", value: t(".allocate"))
```
MaximilianMeister: Isn't a green button `btn-success` better here?

crowbar_framework/app/views/nodes/_form.html.haml
```diff
@@ -127,4 +128,23 @@
...
+            = t(".allocate")
+
+.modal.fade{ id: "node_allocate_modal" }
+  .modal-dialog
+    .modal-content
+      .modal-header
+        .btn.close{ data: {dismiss: "modal" } }
+          &times
+        %h2
+          = t(".allocate_node")
```
MaximilianMeister: can you call it just `t(".header")`? we have that in some other places as well

tboerger:
:-1: It's totally overkill to define a specific modal for that. There are enough bootstrap plugins that offer proper dialogs for confirm.

mmnelemane:
UI decision is not made on how to implement such popups in future.. Hence closing the PR for now.

PR #386, opened 2016-03-02
show sanity checks instead of failing at start-time

MaximilianMeister:
the rails app just refused to start when the sanity checks failed.
now the UI redirects to a sanity-check view and lists the conflicts.

for the future the checks can be extended to cover other areas than
just the network and thus make sure crowbar is in a sane state during
runtime.

**DEPENDS ON** https://github.com/crowbar/crowbar-core/pull/409


MaximilianMeister:
still needs a look by @kwwii

vuntz:
Thanks for working on this, this will make things much better; do you have a screenshot maybe?

MaximilianMeister:
> Thanks for working on this, this will make things much better; do you have a screenshot maybe?

yes, though it is still more of a placeholder, because it is not yet discussed with @kwwii 

sanitychecks

crowbar_framework/config/routes.rb
```diff
@@ -184,6 +184,11 @@
     end
   end
 
+  scope :sanity do
```
tboerger: Why a scope for a single route?

crowbar_framework/lib/crowbar/sanity.rb
```diff
@@ -0,0 +1,67 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+module Crowbar
+  class Sanity
+    class << self
+      def check
+        [].tap do |errors|
```
tboerger: Sorry but this structure is totally wired

tboerger:
:-1: I think we don't want to check that on EVERY request.

crowbar_framework/lib/crowbar/sanity.rb
MaximilianMeister: I guess you mean weird? any suggestions for improvement?

tboerger: Why are you putting arrays in arrays and so on? Just use a flat design for that.

MaximilianMeister: for now we have only the network checks, but it could be more sanity checks in the future. so each check below (which itself calls more sub-checks) can return multiple errors, that is why I chose this design.

crowbar_framework/config/routes.rb
MaximilianMeister: would that be better?

```ruby
get "sanity(.:format)", controller: "sanity", action: "index", as: "sanity"
```

MaximilianMeister:
>  I think we don't want to check that on EVERY request.

sure that is not really optimal, but how would you make sure crowbar is sane during the runtime? would you only run the checks periodically like with a cronjob?

vuntz:
> sure that is not really optimal, but how would you make sure crowbar is sane during the runtime? would you only run the checks periodically like with a cronjob?

@tboerger is right, it'll have too much impact if it's on every request; my feeling is that we should only do it on start (but not the way we were doing it before). No idea how to make it happen or even if it's possible, though :/

MaximilianMeister:
> @tboerger is right, it'll have too much impact if it's on every request; my feeling is that we should only do it on start (but not the way we were doing it before). No idea how to make it happen or even if it's possible, though :/

if someone has an idea then, please share :-)

tboerger:
Maybe keep it as a before filter but use a rails cache value to check if it already have been done for the current start of the application.

MaximilianMeister:
> Maybe keep it as a before filter but use a rails cache value to check if it already have been done for the current start of the application.

thanks for the suggestion.
sounds reasonable, and with threading there should be no issue i guess. @vuntz what do you say?

vuntz:
> @vuntz what do you say?

+1

crowbar_framework/config/routes.rb
tboerger: As we are switching more an more to resources I would suggest `resource :sanity, only: [:show]`

MaximilianMeister: thanks, done

MaximilianMeister:
> Maybe keep it as a before filter but use a rails cache value to check if it already have been done for the current start of the application.

done

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -26,7 +26,12 @@ class ApplicationController < ActionController::Base
   rescue_from Crowbar::Error::ChefOffline, with: :chef_is_offline
 
   before_filter :enable_profiler, if: proc { ENV["ENABLE_PROFILER"] == "true" }
-  before_filter :enforce_installer, unless: proc { Crowbar::Installer.successful? || ENV["RAILS_ENV"] == "test" }
+  before_filter :enforce_installer, unless: proc {
+    Crowbar::Installer.successful? || \
+    ENV["RAILS_ENV"] == "test" || \
```
MaximilianMeister: how would you do that? `proc do ... end` doesn't work here unfortunately

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -26,7 +26,12 @@ class ApplicationController < ActionController::Base
   rescue_from Crowbar::Error::ChefOffline, with: :chef_is_offline
 
   before_filter :enable_profiler, if: proc { ENV["ENABLE_PROFILER"] == "true" }
-  before_filter :enforce_installer, unless: proc { Crowbar::Installer.successful? || ENV["RAILS_ENV"] == "test" }
+  before_filter :enforce_installer, unless: proc {
```
MaximilianMeister: how would you do that? `proc do ... end` doesn't work here unfortunately

crowbar_framework/config/environment.rb
```diff
@@ -18,3 +18,4 @@
 require File.expand_path("../application", __FILE__)
 
 Rails.application.initialize!
+Rails.cache.write(:sanity_check_errors, Crowbar::Sanity.check)
```
MaximilianMeister: JFYI: I can't use the initializers to fill the cache, as the i18n is not yet available at this point

MaximilianMeister:
sanity_checks

@kwwii thanks for the input. is that fine with you like it is now? please have a look

MaximilianMeister:
sanity_checks

@kwwii please have another look

MaximilianMeister:
sanity

so that is the latest version approved by @kwwii

vuntz:
Same question as in the other PR: why "conflicts" and not "errors"?

MaximilianMeister:
> Same question as in the other PR: why "conflicts" and not "errors"?

I just thought conflict sounds better. also it implies that it is manageable and not the end of the world.

vuntz:
> > Same question as in the other PR: why "conflicts" and not "errors"?
> 
> I just thought conflict sounds better. also it implies that it is manageable and not the end of the world.

My problem is that "Unable to detect FQDN" is not really a conflict. Trying to find a better term than error... Maybe failure? Not sure it's really better...

MaximilianMeister:
> My problem is that "Unable to detect FQDN" is not really a conflict. 

Why is it not a conflict? I see it as "Conflicts right now with Running Crowbar"

vuntz:
>> My problem is that "Unable to detect FQDN" is not really a conflict.
> 
> Why is it not a conflict? I see it as "Conflicts right now with Running Crowbar"

Well, to me it's not a conflict at all :-) Let's ask a native english speaker, will be easier: @aspiers ?

aspiers:
Not conflicts for me either; sorry @MaximilianMeister ;-)  I see what you're saying about it being a conflict with trying to run crowbar, but that's really not obvious in how it comes across, because that's not a typical use of the word.  I think "errors" is fine.  Not so much "failures" since that implies a concrete point in time when something went wrong, which is not the case with a simple misconfiguration.

MaximilianMeister:
> Not conflicts for me either; sorry @MaximilianMeister ;-)

don't be sorry :-) thanks for the input, I will change it then

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -729,6 +729,18 @@ en:
       non_json_file: 'Unprocessable chef file found'
       invalid_file_extension: 'Filename needs to be a valid tarball with a .tar.gz file extension'
 
+  sanities:
+    show:
+      header: 'Sanity Checks'
```
dirkmueller: Maybe "Sanity Check Failures" or something like that?

MaximilianMeister: yeah, probably makes more sense

MaximilianMeister:
@dirkmueller updated the i18n

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -31,6 +31,7 @@ class ApplicationController < ActionController::Base
     ENV["RAILS_ENV"] == "test" || \
     Rails.cache.read(:sanity_check_errors).any?
   }
+  before_filter :sanity_checks_cache
   before_filter :sanity_checks, unless: proc {
     Rails.cache.read(:sanity_check_errors).empty? || \
```
tboerger: `Rails.cache.exist?(:sanity_check_errors)`

crowbar_framework/app/views/sanities/show.html.haml
```diff
@@ -0,0 +1,4 @@
+.row
+  .col-lg-12
+    .alert.alert-danger
+      = flash_for_sanity_checks(@errors)
```
vuntz: Can we show a green alert when there's no error (since the page is always reachable)?

MaximilianMeister: good point

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -26,7 +26,16 @@ class ApplicationController < ActionController::Base
   rescue_from Crowbar::Error::ChefOffline, with: :chef_is_offline
 
   before_filter :enable_profiler, if: proc { ENV["ENABLE_PROFILER"] == "true" }
-  before_filter :enforce_installer, unless: proc { Crowbar::Installer.successful? || ENV["RAILS_ENV"] == "test" }
+  before_filter :enforce_installer, unless: proc {
+    Crowbar::Installer.successful? || \
+    ENV["RAILS_ENV"] == "test" || \
+    Rails.cache.read(:sanity_check_errors).any?
+  }
+  before_filter :sanity_checks_cache
```
vuntz: Shouldn't this be before the `enforce_installer` filter?

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -26,7 +26,16 @@ class ApplicationController < ActionController::Base
...
+  before_filter :sanity_checks, unless: proc {
```
vuntz: Also, I'm not totally sure why we need two filters here; I'd put `sanity_checks_cache` inside `sanity_checks`.

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -178,4 +187,25 @@ def enforce_installer
       end
     end
   end
+
+  def sanity_checks_cache
+    respond_to do |format|
+      format.html do
```
vuntz: Do we really need a `respond_to` here? If yes, why don't we handle the json case?

crowbar_framework/app/controllers/sanities_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+#
+
+class SanitiesController < ApplicationController
+  skip_before_filter :sanity_checks
+  before_filter :hide_navigation
+
+  def show
+    sanity_check_errors = Rails.cache.read(:sanity_check_errors)
+    @errors = if sanity_check_errors.nil? || sanity_check_errors.empty?
+      Crowbar::Sanity.check
```
vuntz: Hrm, do we need to run the checks again if it's empty? Doesn't empty mean "no error" (and not "nothing in the cache")?

crowbar_framework/app/controllers/sanities_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+        format.html do
+          begin
+            redirect_to :back
+          rescue ActionController::RedirectBackError
+            redirect_to root_url
+          end
+        end
+      else
+        format.html
+      end
```
vuntz: No json? (I actually don't know: should we deal with json for all controllers?)

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -743,6 +743,18 @@ en:
...
+  sanities:
+    show:
+      header: 'Sanity Check Failures'
+      notice: |
+        There are errors that must be resolved. Please restart Crowbar after solving the errors.
+      fqdn_detected: 'Unable to detect fully-qualified hostname.'
+      ip_resolved: 'Could not resolve %{fqdn} to an IPv4 or IPv6 address.'
+      loopback_unresolved: '%{fqdn} resolves to a loopback address.'
+      ip_configured: 'No local interface is configured with an correct ip address.'
+      firewall_disabled: 'Firewall is not completely disabled.Firewall is not completely disabled.'
```
vuntz: Missing space before Firewall.

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -743,6 +743,18 @@ en:
       non_json_file: 'Unprocessable chef file found'
       invalid_file_extension: 'Filename needs to be a valid tarball with a .tar.gz file extension'
 
+  sanities:
+    show:
+      header: 'Sanity Check Failures'
+      notice: |
+        There are errors that must be resolved. Please restart Crowbar after solving the errors.
+      fqdn_detected: 'Unable to detect fully-qualified hostname.'
```
vuntz: s/Unable to/Could not/ (for consistency)

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -743,6 +743,18 @@ en:
...
+      ip_resolved: 'Could not resolve %{fqdn} to an IPv4 or IPv6 address.'
+      loopback_unresolved: '%{fqdn} resolves to a loopback address.'
+      ip_configured: 'No local interface is configured with an correct ip address.'
```
vuntz: s/an/a/ (I think) and s/ip/IP/

Would it be possible to specify what is a correct IP address? :-)

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -26,7 +26,12 @@ class ApplicationController < ActionController::Base
   rescue_from Crowbar::Error::ChefOffline, with: :chef_is_offline
 
   before_filter :enable_profiler, if: proc { ENV["ENABLE_PROFILER"] == "true" }
-  before_filter :enforce_installer, unless: proc { Crowbar::Installer.successful? || ENV["RAILS_ENV"] == "test" }
+  before_filter :enforce_installer, unless: proc {
+    Crowbar::Installer.successful? || \
+    ENV["RAILS_ENV"] == "test" || \
+    Rails.cache.read(:sanity_check_errors).any?
```
tboerger: https://github.com/crowbar/crowbar-core/pull/386/files#r56304610

crowbar_framework/app/controllers/application_controller.rb
tboerger: I have said that we want to display the error page only for HTML

crowbar_framework/app/helpers/sanities_helper.rb
```diff
@@ -0,0 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+module SanitiesHelper
+  def flash_for_sanity_checks(errors)
+    html = "<h3>#{I18n.t("sanities.show.header")}</h3>"
```
tboerger: :-1: for this. If you write content structure within helpers just use the `content_tag` helpers.

crowbar_framework/app/views/sanities/show.html.haml
tboerger: Why have you added `flash_for_sanity_checks` which generates html if it's just called here?

crowbar_framework/app/controllers/application_controller.rb
vuntz: So if the first thing we do for crowbar is a json formatted request, then we don't do the sanity checks?

crowbar_framework/app/views/sanities/show.html.haml
MaximilianMeister: > Can we show a green alert when there's no error (since the page is always reachable)?

actually the site is always reachable, but if there are no errors we will be redirected. nevertheless I will add a check

MaximilianMeister:
sanity

MaximilianMeister:
regarding travis, I have no idea why it fails... locally it is green. I guess there is a dependency issue in travis or sth.

rsalevsky:
+1

rsalevsky:
Still +1

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -178,4 +189,15 @@ def enforce_installer
...
     end
   end
+
+  def sanity_checks
+    respond_to do |format|
+      format.html do
+        redirect_to sanity_path
+      end
+      format.json do
+        render json: { error: I18n.t("error.pre_install") }, status: :unprocessable_entity
```
vuntz: I don't see this string defined anywhere. Am I looking at the wrong place?

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -748,6 +748,20 @@ en:
       non_json_file: 'Unprocessable chef file found'
       invalid_file_extension: 'Filename needs to be a valid tarball with a .tar.gz file extension'
 
+  sanities:
+    show:
+      header: 'Sanity Check Failures'
+      notice: |
+        There are errors that must be resolved. Please restart Crowbar after solving the errors.
```
vuntz: Do we actually really need to restart crowbar? Or is it okay to just do the sanity checks again? (there's a button in the ui for that, if I understand correctly)

crowbar_framework/app/controllers/application_controller.rb
MaximilianMeister: duh, actually it is called `before_install`. thanks

crowbar_framework/config/locales/crowbar/en.yml
MaximilianMeister: I think it depends on what checks we use, and which one fails. It is a bit strange from UX I agree, so maybe we can drop this suggestion for now until we have a check where a restart is really necessary, what do you say?

vuntz: Agreed

vuntz:
+1

MaximilianMeister:
@vuntz do we want to backport this?

vuntz:
> @vuntz do we want to backport this?

If it's not too complicated, yes. The current way of just failing on error is very confusing.

PR #387, opened 2016-03-02
Check if we can ssh to nodes before trying to apply proposal.

jsuchome:
(bsc#968436)

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -69,10 +69,24 @@ class CrowbarService < ServiceObject
     end
   end
 
-  def shutdown_services_at_non_db_nodes
-    db_nodes = []
-    NodeObject.find("state:crowbar_upgrade").each do |node|
+  # Ensure the nodes are reachable before trying to apply the proposals
+  def check_if_nodes_are_available(upgrade_nodes)
+    unavailable_nodes = []
+    upgrade_nodes.each do |node|
+      ssh_status = node.ssh_cmd("")
+      if ssh_status[0] != 200
+        unavailable_nodes.push node.name
+      end
+    end
+    unless unavailable_nodes.empty?
+      raise I18n.t("model.service.nodes_not_available", names: unavailable_nodes.join(", "))
+    end
+  end
 
+  def shutdown_services_at_non_db_nodes
+    upgrade_nodes = NodeObject.find("state:crowbar_upgrade")
+    check_if_nodes_are_available upgrade_nodes
+    upgrade_nodes.each do |node|
       # Find the nodes with Database role.
       # In this step, database nodes need to wait until everything is shut down at the other nodes.
       step = "openstack_shutdown"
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -70,9 +70,18 @@ def commit_and_check_proposal
...
 
   def shutdown_services_at_non_db_nodes
-    db_nodes = []
-    NodeObject.find("state:crowbar_upgrade").each do |node|
+    upgrade_nodes = NodeObject.find("state:crowbar_upgrade")
 
+    # This is the first action on nodes from current version of product:
+    # ensure the nodes are reachable
+    upgrade_nodes.each do |node|
+      ssh_status = node.ssh_cmd("")
```
vuntz: There's no need to pass a command (like /usr/bin/true)?

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -70,9 +70,18 @@ def commit_and_check_proposal
...
+      if ssh_status[0] != 200
+        raise I18n.t("model.service.node_not_available", name: node.name)
```
vuntz: Can we possible collect all nodes for which it's failing and display one message for all of them, instead of one message just for the first one?

crowbar_framework/app/models/crowbar_service.rb
jsuchome: True, no need.

rhafer:
If we do something like this shouldn't it be done for all steps (i.e. ```dump_openstack_database```, ```finalize_openstack_shutdown``` and maybe even ```prepare_nodes_for_os_upgrade```).  Or is ```shutdown_services_at_non_db_nodes``` somehow special?

rhafer:
BTW, to make our tools happy please add the bsc reference to the title of the commit message. And add some more info to the body about why this is needed.

jsuchome:
> Or is shutdown_services_at_non_db_nodes somehow special?

It is special  because it is the first action done to the nodes during upgrade, after admin server upgrade. So I think that if some node outage or something like that can happen somewhere, than it's most likely here, later you're working with the nodes in the same state. ... But maybe yes, it shouldn't hurt to add that check to other places as well.

jsuchome:
I did not add the check before `dump_openstack_database` because that's called directly after `shutdown_services_at_non_db_nodes` from the same controller action

jsuchome:
So I tested to shutdown one node before each step, and it worked as expected.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -69,10 +69,24 @@ def commit_and_check_proposal
...
+  def check_if_nodes_are_available(upgrade_nodes)
+    unavailable_nodes = []
+    upgrade_nodes.each do |node|
+      ssh_status = node.ssh_cmd("")
+      if ssh_status[0] != 200
+        unavailable_nodes.push node.name
+      end
+    end
+    unless unavailable_nodes.empty?
+      raise I18n.t("model.service.nodes_not_available", names: unavailable_nodes.join(","))
```
vuntz: Maybe `join(", ")` (with a space)

vuntz:
+1

rhafer:
> I did not add the check before dump_openstack_database because that's called directly after shutdown_services_at_non_db_nodes from the same controller action

@jsuchome Hm, though especially that shutdown_services_at_non_db_nodes could cause node being offline because they were fenced (for what ever reason). It's doing lots of stuff to the cluster, right?
While the ssh check wouldn't prevent that, it would at least show problems. What do you think?

jsuchome:
Well, I think the issues were really coming from the situation before we start upgrade on cloud6 side. Anyway, I'll add the check before dump_openstack_database as well.

rhafer:
+1

vuntz:
+1

jsuchome:
Gating failed in tempest: can't be related to the PR...

vuntz:
Agreed.

PR #388, opened 2016-03-02
dont run 'ip' on windows

bmwiedemann:
I think we dont really need these routes anyway
and this breaks chef on windows in the QA clouds
because they use bmc-nat mode

### DIFF
--- a/chef/cookbooks/bmc-nat/recipes/client.rb
+++ b/chef/cookbooks/bmc-nat/recipes/client.rb
@@ -16,6 +16,7 @@
 # Note : This script runs on both the admin and compute nodes.
 # It intentionally ignores the bios->enable node data flag.
 
+return if node[:platform_family] == "windows"
 nets = node[:crowbar][:network] || return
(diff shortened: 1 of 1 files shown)


bmwiedemann:
tested manually on QA4 and hyper-v node now got into ready state

aplanas:
+1

toabctl:
+1

chef/cookbooks/bmc-nat/recipes/client.rb
```diff
@@ -27,6 +27,7 @@
 nat_address = nat_node[:crowbar][:network][:admin][:address]
 
 return if admin_subnet == bmc_subnet && admin_netmask == bmc_netmask
+return if node[:platform_family] == "windows"
```
vuntz: We usually put the check at the top of the recipe (see logging::client). Can you do the small change (and backport to 3.0)?

bmwiedemann:
moved test to the top. will backport when the change is merged to master.

PR #390, opened 2016-03-03
[3.0] Check if we can ssh to nodes before trying to apply proposal (bsc#968…

jsuchome:
…436)

When a node went offline during upgrade, the process will fail and partially
commited proposal moves nodes to wrong state. Better check if the nodes
are accessible before applying upgrade related actions.

(backport of https://github.com/crowbar/crowbar-core/pull/387)


jsuchome:
correct PR is https://github.com/crowbar/crowbar-core/pull/399

PR #391, opened 2016-03-03
[3.0] Save the location that was used for DB dump to node.

jsuchome:
Actually, save it to all nodes in DB cluster, we don't know
which one will be running DB once they are back online.

(backport of https://github.com/crowbar/crowbar-core/pull/390)


rsalevsky:
+1

MaximilianMeister:
was that cherry picked? and i think the reference to https://github.com/crowbar/crowbar-core/pull/390 is also wrong? not sure

jsuchome:
Wasn't cherry picked

PR #392, opened 2016-03-04
3.0 | Creating tar only if the name is valid (bsc#966424)

rsalevsky:
- create archive only if the name is valid
- run custom validations only in upload case
- stop validating size and version due to there are anyway set by the
  system and otherwise create_archive after validate will not work

(cherry picked from commit abacbe72389ce59deaa93b893a8c852725d8eab6)

### DIFF
--- a/crowbar_framework/app/models/backup.rb
+++ b/crowbar_framework/app/models/backup.rb
@@ -20,7 +20,8 @@ require "find"
 class Backup < ActiveRecord::Base
   attr_accessor :file
 
-  before_validation :save_or_create_archive, on: :create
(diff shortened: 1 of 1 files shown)


vuntz:
+1

MaximilianMeister:
+1 (but can you please read https://github.com/crowbar/crowbar-core/pull/402#issuecomment-193150539 for reference?)

vuntz:
Gating failed in tempest (manila), so unrelated to backup/restore.

(this is backport of https://github.com/crowbar/crowbar-core/pull/348, btw)

PR #398, opened 2016-03-04
[3.0] Save the location that was used for DB dump to node.

jsuchome:
Actually, save it to all nodes in DB cluster, we don't know
which one will be running DB once they are back online.

(cherry picked from commit 7c32deb212ecbf733a48b1c01a1d7a03b6b3e9bf)


vuntz:
+1 (it's the backport of https://github.com/crowbar/crowbar-core/pull/380)

rsalevsky:
+1

PR #399, opened 2016-03-04
[3.0] Check if we can ssh to nodes before trying to apply proposal (bsc#968436)

jsuchome:
When a node went offline during upgrade, the process will fail and partially
commited proposal moves nodes to wrong state. Better check if the nodes
are accessible before applying upgrade related actions.

(cherry picked from commit c9d522008937d5ded7baf63d2c6fd48fb1c01dd4)


vuntz:
Backport of https://github.com/crowbar/crowbar-core/pull/387.

+1

MaximilianMeister:
+1

PR #400, opened 2016-03-04
[3.0] Claim RAID disks when node is ready, not earlier

jsuchome:
We need to wait for the first chef-client run to gather ohai data

(cherry picked from commit 365dab2fda629fe6754867984f46f1776028bf0c)

This is actually forward-port from tex branch...


vuntz:
@jsuchome just to clarify: do we have a PR for master first?

jsuchome:
Probably no, but we have one for tex ;-)

jsuchome:
If it did not make it to cloud5, it's probably obsolete

PR #404, opened 2016-03-04
Prepare for Cloud 7

dirkmueller:


aplanas:
PR not found!. <sub>Sorry, I can help it :(</sub>

dirkmueller:
@aplanas  :-)

vuntz:
You probably need to change suse-12.1 to suse-12.2 in the file (except for SES?)

vuntz:
Btw, do we really need to switch to SP2 now? I'm all for doing it early, but doing it while we're branching is a bit ambitious as I'm pretty sure it adds to the complexity of getting the gating working.

dirkmueller:
@vuntz we're not switchign to SP2. but the project repository names are already SP2 based (but we actually build against SP1)

dirkmueller:
I can't change suse-12.1 to 12.2 yet since mkcloud isn't prepared for that. This change is needed so that mkcloud does something (otherwise it rejects a media built from D:C:7 as invalid)

vuntz:
Hrm, okay. Then +1, and let's hope it fixes gating for master...

rsalevsky:
+1

PR #406, opened 2016-03-07
fix autoload of crowbar libs

MaximilianMeister:
this hasn't been noticed yet as we are loading all libs in application.rb anyways

@vuntz @aspiers shall we fix it here or shall we drop it completely following up on the discussion on the autoloading?

### DIFF
--- a/crowbar_framework/lib/crowbar.rb
+++ b/crowbar_framework/lib/crowbar.rb
@@ -20,11 +20,11 @@ module Crowbar
     File.expand_path("../crowbar/backup", __FILE__)
 
   autoload :Installer,
-    File.expand_path("../installer.rb", __FILE__)
+    File.expand_path("../crowbar/installer", __FILE__)
 
   autoload :Migrate,
-    File.expand_path("../migrate.rb", __FILE__)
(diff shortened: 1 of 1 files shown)


tboerger:
This should be done for all and not only partial. But there are others who decides against it instead of using core ruby functionality.

vuntz:
> @vuntz @aspiers shall we fix it here or shall we drop it completely following up on the discussion on the autoloading?

If it's broken, let's just fix it -- no need to wait for the other discussion.

> But there are others who decides against it instead of using core ruby functionality.

@tboerger I guess it's about the discussion in https://github.com/crowbar/crowbar-core/pull/367 and I see this PR is closed now?. Let me follow up there.

aspiers:
> If it's broken, let's just fix it -- no need to wait for the other discussion.

Agreed, +1 assuming the gating is broken for a different reason.

tboerger:
Fixing it is not just replacing the paths. It's inconsistent bullshit to do it just for 2, 3 or 4 files...

rsalevsky:
+1

PR #409, opened 2016-03-08
proper handling of json requests before crowbar is installed

MaximilianMeister:
restrict_api

### DIFF
--- a/crowbar_framework/config/locales/crowbar/en.yml
+++ b/crowbar_framework/config/locales/crowbar/en.yml
@@ -811,6 +811,7 @@ en:
     not_found: 'Not found'
     chef_server_down: 'Chef server is not available'
     param_missing: 'A required parameter is missing'
+    before_install: 'This API is not available before installation'
 
   state:
     unknown: 'Unknown'
--- a/crowbar_framework/app/controllers/application_controller.rb
+++ b/crowbar_framework/app/controllers/application_controller.rb
@@ -169,6 +169,13 @@ class ApplicationController < ActionController::Base
   end
 
   def enforce_installer
-    redirect_to installer_root_path
+    respond_to do |format|
+      format.html do
+        redirect_to installer_root_path
+      end
+      format.json do
+        render json: { error: I18n.t("error.before_install") }, status: :unprocessable_entity
+      end
+    end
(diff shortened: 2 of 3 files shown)


crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -811,6 +811,7 @@ en:
     not_found: 'Not found'
     chef_server_down: 'Chef server is not available'
     param_missing: 'A required parameter is missing'
+    pre_install: 'Crowbar is not installed yet, therefore the API is restricted'
```
vuntz: Maybe change "pre_install" to "before_install" (to avoid potential confusion with preinstalling something).

Also I'd change the string to: "API is not available before installation" (or another simple sentence, as it fits better an error message than a full sentence like yours).

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -811,6 +811,7 @@ en:
     not_found: 'Not found'
     chef_server_down: 'Chef server is not available'
     param_missing: 'A required parameter is missing'
+    before_install: 'Some of the API is not available before installation'
```
MaximilianMeister: "API is not available before installation" is certainly better, but not really true, as it is only part of it that is not available.
It shall not mislead the user to think he can not use the `crowbarctl` at all.

would this be ok for you?

vuntz: Well, my proposal was ambiguous (kind of on purpose ;-)) because it wasn't saying if this was "This API" or "API in general". If you want to be more specific, I'd just say "This API".

rsalevsky:
+1

vuntz:
+1

vuntz:
Gating failing in tempest with one swift error, so unrelated, merging.

@MaximilianMeister do you plan to backport to 3.0?

MaximilianMeister:
> do you plan to backport to 3.0?

yes I would like to, if you're fine with it

vuntz:
Sure, sounds good.

MaximilianMeister:
https://github.com/crowbar/crowbar-core/pull/418

PR #410, opened 2016-03-08
inform the user that some upgrade steps are repeatable after an error

MaximilianMeister:
repeatable_upgrade_error

### DIFF
--- a/crowbar_framework/config/locales/crowbar/en.yml
+++ b/crowbar_framework/config/locales/crowbar/en.yml
@@ -557,6 +557,7 @@ en:
     upgrades:
       title: 'Upgrade'
       header: '%{product}: Upgrade from SUSE OpenStack Cloud 5'
+      repeat_after_error: 'Repeat this step after resolving the error(s).'
       nav:
         prepare: 'Prepare Client Nodes'
         download: 'Download Upgrade Data'
--- a/crowbar_framework/app/controllers/installer/upgrades_controller.rb
+++ b/crowbar_framework/app/controllers/installer/upgrades_controller.rb
@@ -109,7 +109,7 @@ module Installer
             end
           rescue => e
             format.html do
-              flash[:alert] = e.message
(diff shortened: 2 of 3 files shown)


MaximilianMeister:
cc @jsuchome @vuntz @kwwii

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -561,6 +561,7 @@ en:
     upgrades:
       title: 'Upgrade'
       header: '%{product}: Upgrade from SUSE OpenStack Cloud 5'
+      repeat_after_conflict: 'Please repeat this step after resolving the conflicts'
```
jsuchome: The dot at the end of sentence is missing. Also, I don't think "Please" is necessary here (but some English speaker should be judge of that).

jsuchome:
In general, I think it is not necessary (as I tried to explain in the bug report). There's nothing else user can do in the UI than repeating the step anyway, so this adds no extra value.

(This is not -1, I just think it is not needed. But if others think extra message helps, I agree that this is the simple way to add it)

MaximilianMeister:
I have updated the display now after talking to @kwwii, @jsuchome he also said it would be good to at least give a hint about it

jsuchome:
This looks good actually. +1

vuntz:
I am a bit confused: why is the string talking about conflicts instead of errors?

MaximilianMeister:
> I am a bit confused: why is the string talking about conflicts instead of errors?

Here Errors is better I agree, will change it. But in https://github.com/crowbar/crowbar-core/pull/386 i would rather stick to conflicts

MaximilianMeister:
changed to `Errors`

vuntz:
+1

jsuchome:
Still +1

rsalevsky:
Gating failed in tempest.

PR #413, opened 2016-03-10
Set rake to version to 10.3.2

rsalevsky:


tboerger:
Travis still fails because of malicious rails version.

tboerger:
Otherwise :+1:

bmwiedemann:
:+1:

rsalevsky:
Gating fails: Could not find gem 'rake (~> 10.3.2) ruby' in the gems available on this machine. (Bundler::GemNotFound)

bmwiedemann:
Actually, it is not failing from rake here, but from the bundle-audit complaining about rails-4.2.6 (which we dont use in the product anyway)

vuntz:
> Actually, it is not failing from rake here, but from the bundle-audit complaining about rails-4.2.6 (which we dont use in the product anyway)

Can we set these to 4.2.2 (which is what we use in our packages)? I'm not sure if "Solution: upgrade to ~> 5.0.0.beta1.1, ~> 4.2.5.1, >= 4.1.14.1, ~> 4.1.14" means a version that matches all of these criterias or one of these, though...

bmwiedemann:
Also, in SLE-12-SP1 we have rake 10.1.0 as part of ruby-2.1 ... I wonder if there is a way to make all those checks happy or if we should drop some of them.

MaximilianMeister:
travis was green here https://github.com/crowbar/crowbar-core/pull/371

maybe we don't need it anymore?

tboerger:
> Can we set these to 4.2.2 (which is what we use in our packages)? I'm not sure if "Solution: upgrade to ~> 5.0.0.beta1.1, ~> 4.2.5.1, >= 4.1.14.1, ~> 4.1.14" means a version that matches all of these criterias or one of these, though...

Oru version is vulnerable, we should upgrade to `4.2.5.1`

rsalevsky:
Seams to work again.

tboerger:
We should revert the merged pull requests for the other repos.

PR #414, opened 2016-03-10
add common glance user+group (bsc#968251)

bmwiedemann:
for making HA glance on shared NFS backend storage work


jsuchome:
This works correctly as expected, new installed nodes get the glance user and group with same UID and GID.

However, why do we need to patch crowbar_register when we set the group and user in autoyast profile?

bhdn:
IIUC, the node can be set up manually, thus not relying on autoyast.

jsuchome:
@bhdn Makes sense. +1 then

rhafer:
+1 (I guess we want to backport this to stable/3.0 ?)

PR #416, opened 2016-03-11
[3.0] Stop dnsmasq to avoid dns queries on admin network during restore and resume later (bsc#966419) 

mmnelemane:
Fixing https://bugzilla.suse.com/show_bug.cgi?id=966419 without iptables rules and instead disable and re-enable dnsmasq service letting only the nameserver listen on 127.0.0.1 during restore. An additional state file "disable_dns" is added since the file "restore_steps" waits for completion of chef-client while enabling dns service can be done immediately after the chef-server is running.

Backport from https://github.com/crowbar/crowbar-core/pull/371 on master


vuntz:
+1

MaximilianMeister:
+1

PR #422, opened 2016-03-14
backup: cleanup with higher permissions

MaximilianMeister:
as there are files belonging to root in the rails tmp dir they
need to be removed with higher permissions

from the backup/restore CI:
```
Errno::EACCES (Permission denied @ dir_initialize - /opt/dell/crowbar_framework/tmp/create_archive-20160314-14811-fyefs2/crowbar/root/.ssh):
  app/models/backup.rb:110:in `cleanup'
  app/controllers/backups_controller.rb:58:in `create'
```

### DIFF
--- a/crowbar_framework/app/models/backup.rb
+++ b/crowbar_framework/app/models/backup.rb
@@ -110,7 +110,14 @@ class Backup < ActiveRecord::Base
   end
 
   def cleanup
-    @data.rmtree if @data
+    return unless @data
+
+    system(
+      "sudo",
+      "rm",
+      "-rf",
+      @data.to_s
+    )
     @data = nil
   end
 
--- a/crowbar_framework/app/models/backup.rb
+++ b/crowbar_framework/app/models/backup.rb
@@ -110,7 +110,7 @@ class Backup < ActiveRecord::Base
   end
 
   def cleanup
-    return unless @data || @data.to_s =~ /\A#{Rails.root.join("tmp").to_s}/
+    return unless @data || @data.to_s =~ /\A#{Dir.tmpdir}/
 
     system(
       "sudo",
--- a/crowbar_framework/app/models/backup.rb
+++ b/crowbar_framework/app/models/backup.rb
@@ -110,7 +110,7 @@ class Backup < ActiveRecord::Base
   end
 
   def cleanup
-    return unless @data
+    return unless @data || @data.to_s =~ /\A#{Rails.root.join("tmp").to_s}/
 
     system(
       "sudo",
--- a/crowbar_framework/config/initializers/tmpdir.rb
+++ b/crowbar_framework/config/initializers/tmpdir.rb
@@ -19,7 +19,7 @@ class Dir
   class << self
     def mktmpdir_with_rails_context(prefix_suffix = nil, tmp_dir = nil, *rest)
       # set tmp dir to application tmp dir
-      tmp_dir = Rails.root.join("tmp").to_s if tmp_dir.nil?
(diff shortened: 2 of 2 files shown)


jsuchome:
LGTM (I did not test it)

crowbar_framework/app/models/backup.rb
```diff
@@ -110,7 +110,15 @@ def upgrade
   end
 
   def cleanup
-    @data.rmtree if @data
+    return unless @data
+    cmd = [
+      "sudo",
+      "-i",
```
vuntz: I think we skip `-i` in the other sudo calls from this file, don't we?

MaximilianMeister: I took it out again. thanks

crowbar_framework/app/models/backup.rb
```diff
@@ -110,7 +110,14 @@ def upgrade
   end
 
   def cleanup
-    @data.rmtree if @data
+    return unless @data
```
jdsn: maybe its worth to do some sanity checking on `@data`.
It should not contain any path that could destroy the system. Maybe it makes sense to have a positive path filter that e.g. only allows paths below `/x/y/z/`

rsalevsky:
+1

crowbar_framework/app/models/backup.rb
```diff
@@ -110,7 +110,14 @@ def upgrade
   end
 
   def cleanup
-    @data.rmtree if @data
+    return unless @data || @data.to_s =~ /#{Rails.root.join("tmp").to_s}/
```
tboerger:
```
return unless @data || @data.to_s =~ /\A#{Rails.root.join("tmp").to_s}/

system(
  "sudo",
  "rm",
  "-rf",
  @data.to_s
)
```

Untested but looks IMHO better

tboerger:
:+1:

crowbar_framework/app/models/backup.rb
```diff
@@ -110,7 +110,14 @@ def upgrade
   end
 
   def cleanup
-    @data.rmtree if @data
+    return unless @data || @data.to_s =~ /\A#{Rails.root.join("tmp").to_s}/
```
vuntz: My concern here is that you're relying on internal knowledge of how we patched `mktmpdir`. Can we somehow add a helper is the same location (`config/initializers/tmpdir.rb`) that gives us the parent directory so we don't rely on this internal knowledge (and don't break in case it changes)?

MaximilianMeister: okay, will add it. thanks for the hint

MaximilianMeister:
@vuntz I have added the helper, but the last change is still untested, lets see what gating says

crowbar_framework/config/initializers/tmpdir.rb
```diff
@@ -31,5 +31,15 @@ def mktmpdir_with_rails_context(prefix_suffix = nil, tmp_dir = nil, *rest)
     end
 
     alias_method_chain :mktmpdir, :rails_context
+
+    def tmpdir_with_rails_context(systemdir = false)
+      if systemdir
+        tmpdir_without_rails_context
```
MaximilianMeister: ah actually this returns a string, so below it should do too

MaximilianMeister:
sorry for stressing but this blocks my work on the 6to7 upgrade a bit

@vuntz @rsalevsky @tboerger can you please review again?

vuntz:
+1

rhafer:
+1

vuntz:
@MaximilianMeister will this be backport to 3.0?

PR #425, opened 2016-03-15
Replace cmd class with execute class

rsalevsky:
This change unifies the way how we call bash or ruby methods as
different user.


crowbar_framework/lib/execute.rb
```diff
@@ -0,0 +1,92 @@
...
+      end
+    end
+
+    protected
+
+    def bash_as(root, *params)
+      stderr = nil
+      wait_thr = nil
+      if root
+        _stdin, _stdout, stderr, wait_thr = Open3.popen3("sudo", "-i", *params)
```
vuntz: My usual question: why do we need `-i`?

vuntz:
I'm a bit confused: the old class wasn't used anywhere? Or why is there no need to change code when removing it.

rsalevsky:
The old class wasn't used at all...

crowbar_framework/lib/execute.rb
```diff
@@ -0,0 +1,92 @@
...
+
+      Rails.logger.warn(stderr.read.strip) unless wait_thr.value.success?
+      wait_thr.value.success?
+    end
+
+    def ruby_as(user, *params)
+      execution_uid = uid(user)
+      return false unless execution_uid
+
+      Process.fork do
```
vuntz: So usually, when we fork, we also need to close additional fd, and more. So I guess we need a bit more attention here.

crowbar_framework/lib/execute.rb
```diff
@@ -0,0 +1,92 @@
...
+        Process.euid = execution_uid
```
vuntz: Maybe set `egid` too.

vuntz:
> Just for information, we want to replace later the direct root calls with a wrapper program (like it is done in hawk) to avoid security issues. So this is just a first step.

Sounds good; do we have code for that? If not, we might want to just reuse the rootwrap bits from openstack.

tboerger:
> Sounds good; do we have code for that? If not, we might want to just reuse the rootwrap bits from openstack.

Take a look at the hawk scripts: https://github.com/ClusterLabs/hawk/blob/master/tools/hawk_invoke.c

crowbar_framework/lib/execute.rb
```diff
@@ -0,0 +1,96 @@
...
+      end
+    end
+
+    def ruby_as(user, *params)
+      execution_uid = uid(user)
+      return false unless execution_uid
+
+      Process.fork do
+        Process.euid = execution_uid
+        eval(*params)
```
tboerger: You should disable this check for this specific line as we know what we are doing here.

crowbar_framework/lib/execute.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    def uid(user)
+      execution_user = begin
+        Etc.getpwnam(user)
+      rescue ArgumentError
+        return false
+      end
+
+      if execution_user.nil?
+        return false
```
tboerger: Useless return

crowbar_framework/lib/execute.rb
```diff
@@ -0,0 +1,92 @@
...
+      else
+        return execution_user.uid
```
tboerger: Useless return

crowbar_framework/lib/execute.rb
```diff
@@ -0,0 +1,92 @@
...
+      end
+
+      _pid, status = Process.wait2
+      status.success?
+    end
+
+    def uid(user)
+      execution_user = begin
+        Etc.getpwnam(user)
+      rescue ArgumentError
```
tboerger: Just put that as a rescue of the method and drop the begin part.

tboerger:
```ruby
def uid(user)
  execution_user = Etc.getpwnam(user)

  .....

rescue ArgumentError
  false
end
```

crowbar_framework/lib/execute.rb
```diff
@@ -0,0 +1,92 @@
...
+      return false unless execution_uid
+
+      Process.fork do
+        Process.egid = execution_gid
+        Process.euid = execution_uid
+        eval(*params)
+      end
+
+      _pid, status = Process.wait2
+      status.success?
```
MaximilianMeister: in general wouldn't it be better to return the exit code instead of `true` or `false`? and evaluate the success in the caller?

rsalevsky: I would say not, when I call a method in ruby then I expect in this case true of false.

For example:
```ruby
if Execute.ruby_as("root", "File.new('test.txt', 'w')")
    ....
end
```

This will not work when we return exit codes.

MaximilianMeister: I understand that this makes the usage more convenient, but I would then write at least the exit code into the log, otherwise we can never trace back why a command failed, be it bash or ruby...

as I see there is no logging at all currently, can you add some?

crowbar_framework/lib/execute.rb
```diff
@@ -0,0 +1,92 @@
...
+    def uid_gid(user)
+      execution_user = Etc.getpwnam(user)
+
+      if execution_user.nil?
+        return false
+      else
+        [execution_user.uid, execution_user.gid]
+      end
+    rescue ArgumentError
+      false
```
MaximilianMeister: here would be a place where we could catch the error message and log it?

crowbar_framework/lib/execute.rb
```diff
@@ -0,0 +1,97 @@
...
+      execution_uid, execution_gid = uid_gid(user)
+      return false unless execution_uid
+
+      Process.fork do
+        Process.egid = execution_gid
+        Process.euid = execution_uid
+        eval(*params)
+      end
+
+      _pid, status = Process.wait2
```
MaximilianMeister: thanks for adding some logging. can you add a log entry here with `status.exitstatus` ?

MaximilianMeister:
I also think for such a library it would be really good to have some tests

crowbar_framework/lib/execute.rb
```diff
@@ -0,0 +1,97 @@
...
+        _stdin, _stdout, stderr, wait_thr = Open3.popen3(*params)
+      end
+
+      Rails.logger.error("Execution failed: #{stderr.read.strip}") unless wait_thr.value.success?
+      wait_thr.value.success?
+    end
+
+    def ruby_as(user, *params)
+      execution_uid, execution_gid = uid_gid(user)
+      return false unless execution_uid
```
vuntz: This will not do what you expect if `user` is root (so `execution_uid` is `0`).

crowbar_framework/lib/execute.rb
```diff
@@ -0,0 +1,97 @@
...
+    def bash(user, *params)
+      Rails.logger.debug("Execute bash \"#{params.join(", ")}\" as #{user}.")
+
+      if user.nil?
+        return bash_as(false, *params)
+      elsif user == "root"
+        return bash_as(true, *params)
+      else
+        execution_uid, execution_gid = uid_gid(user)
+        return false unless execution_uid
```
vuntz: Same comment as for the other similar code.

crowbar_framework/lib/execute.rb
```diff
@@ -0,0 +1,97 @@
...
+
+        Process.fork do
+          Process.egid = execution_gid
+          Process.euid = execution_uid
```
vuntz: I don't understand how this can work: we're running as the `crowbar` user. This user cannot change it's euid (afaik). Don't we need to go through sudo/su/something else for that?

crowbar_framework/lib/execute.rb
```diff
@@ -0,0 +1,51 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "open3"
+
+class Execute
+  class << self
+    def bash(*params)
```
tboerger: Let's call it `bash_as_user`, than it's more obvious that it's not executed with higher privileges :)

crowbar_framework/lib/execute.rb
```diff
@@ -0,0 +1,51 @@
...
+    def bash_as(root, *params)
+      stderr = nil
+      wait_thr = nil
+      if root
+        _stdin, _stdout, stderr, wait_thr = Open3.popen3("sudo", *params)
+      else
+        _stdin, _stdout, stderr, wait_thr = Open3.popen3(*params)
+      end
+
+      Rails.logger.error("Execution failed: #{stderr.read.strip}") unless wait_thr.value.success?
```
tboerger: Maybe an if and else to log both cases?

tboerger:
Needs some basic tests and please drop the dots at the end of the log messages.

toabctl:
Can we for now just remove the unused file ```crowbar_framework/lib/cmd.rb``` and implement the other class when we also start using it?

rsalevsky:
@toabctl See https://github.com/crowbar/crowbar-core/pull/552

Itxaka:
Why all the functions which are just an alias for the bash_as?
Seems to be extra code that does not help clarify that much when we can use named parameters anyway no?

Also, if someone calls the bash_for directly, then we get no debug logging. Wouldnt it be better to move the logging to the bash_for function?

not blocking, just thougths, lgtm +1

rsalevsky:
@Itxaka Yeah, the idea behind this changed totally while the review mostly due to technical limitations. At the moment I'm completely unsure if this is still useful/needed in this shape.

tboerger:
Directly calling a protected method? Evil hacking ;)

PR #427, opened 2016-03-16
Claim RAID disks when node is ready, not earlier

jsuchome:
We need to wait for the first chef-client run to gather ohai data

(cherry picked from commit 365dab2fda629fe6754867984f46f1776028bf0c)
(cherry picked from commit 728844eccf4bf55e971c7ad6c38344ea7c9dba00)

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -190,8 +190,9 @@ class CrowbarService < ServiceObject
         return [404, "Node not found"]
       end
 
-      if state == "readying"
-        transition_to_readying inst, name, state, node
+      if state == "ready" && !node.admin?
+        # we need to claim raids when we already have ohai data, i.e. after first chef-client run
+        process_raid_claims node
       end
 
       if %w(hardware-installing hardware-updating update).include? state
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/node_object.rb
```diff
@@ -1449,7 +1449,7 @@ def disk_release(device, owner)
     if device
       crowbar_wall[:claimed_disks] ||= {}
 
-      unless disk_owner(device) == owner
+      if owner.empty? || disk_owner(device) != owner
```
vuntz: I'm fine with that change, but can you also do it in chef/cookbooks/barclamp/libraries/barclamp_library.rb since we keep the code in sync here and there.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -190,8 +190,9 @@ def transition(inst, name, state)
         return [404, "Node not found"]
       end
 
-      if state == "readying"
-        transition_to_readying inst, name, state, node
+      if state == "ready" && !node.admin?
+        # we need to claim raids when we already have ohai data, i.e. after first chef-client run
+        process_raid_claims node
```
vuntz: I'm really not fan of claiming it in `ready`. The thing I don't get is the following: why don't we have ohai data from the discovery image? Shouldn't we?

jsuchome: Isn't it the same problem why raid jobs are failing currently?

vuntz: I was thinking about that too after writing my comment. I guess we really need to focus on that bug :/

jsuchome:
So I guess this should be on hold until the real problem behind is resolved. It's actually only workaround.

jsuchome:
Closing, looks obsolete

PR #428, opened 2016-03-16
Upgrade additions

MaximilianMeister:
* add json endpoint for triggering the upgrade from the command line without being redirected (see also https://github.com/SUSE-Cloud/automation/pull/915)
* allow 3.0 to 4.0 upgrade

### DIFF
--- a/crowbar_framework/app/controllers/installer/upgrades_controller.rb
+++ b/crowbar_framework/app/controllers/installer/upgrades_controller.rb
@@ -42,10 +42,17 @@ module Installer
             format.html do
               redirect_to restore_upgrade_url
             end
+            format.json do
+              render json: t(".success"), status: :ok
+            end
           else
+            msg = @backup.errors.full_messages.first
             format.html do
-              flash[:alert] = @backup.errors.full_messages.first
+              flash[:alert] = msg
               redirect_to start_upgrade_url
+            end
+            format.json do
+              render json: msg, status: :unprocessable_entity
             end
           end
         end
--- a/crowbar_framework/app/models/backup.rb
+++ b/crowbar_framework/app/models/backup.rb
@@ -185,9 +185,14 @@ class Backup < ActiveRecord::Base
       return false
     end
 
+    path.open("wb") do |f|
+      f.write(file.read)
+    end
+
     meta_file = data.join("meta.yml")
     unless meta_file.exist?
       errors.add(:file_content, I18n.t("backups.index.meta_missing"))
+      path.delete
       return false
     end
 
@@ -195,6 +200,7 @@ class Backup < ActiveRecord::Base
       meta = YAML.load_file(meta_file)
     rescue Psych::SyntaxError
       errors.add(:file_content, I18n.t("backups.index.invalid_file_content"))
+      path.delete
       return false
     end
 
@@ -203,10 +209,6 @@ class Backup < ActiveRecord::Base
     self.created_at = Time.zone.parse(meta["created_at"])
     # 20151222144602_create_backups.rb
     self.migration_level = meta["migration_level"] || 20151222144602
-
-    path.open("wb") do |f|
-      f.write(file.read)
-    end
   end
 
   def delete_archive
--- a/crowbar_framework/app/models/backup.rb
+++ b/crowbar_framework/app/models/backup.rb
@@ -75,9 +75,11 @@ class Backup < ActiveRecord::Base
     from_upgrade = options.fetch(:from_upgrade, false)
 
     if Crowbar::Backup::Restore.restore_steps_path.exist?
+      logger.debug "Restore process is already running"
       errors.add(:base, I18n.t("backups.index.multiple_restore"))
       return false
     elsif !from_upgrade && backup_version != system_version
+      logger.debug "Restoring from different Crowbar version is not allowed"
       errors.add(:base, I18n.t("backups.index.version_conflict"))
       return false
     end
--- a/crowbar_framework/lib/crowbar/upgrade.rb
+++ b/crowbar_framework/lib/crowbar/upgrade.rb
@@ -36,6 +36,8 @@ module Crowbar
         knife_files
         crowbar_files
       end
+
+      @status[:status] == :ok
     end
 
     def supported?
--- a/crowbar_framework/lib/crowbar/upgrade.rb
+++ b/crowbar_framework/lib/crowbar/upgrade.rb
@@ -37,6 +37,9 @@ module Crowbar
         crowbar_files
       end
 
+      if @status[:errors].any?
+        @backup.errors.add(:upgrade, @status[:errors])
+      end
       @status[:status] == :ok
     end
 
--- a/crowbar_framework/lib/crowbar/upgrade.rb
+++ b/crowbar_framework/lib/crowbar/upgrade.rb
@@ -37,7 +37,8 @@ module Crowbar
 
     def supported?
       upgrades = [
-        ["1.9", "3.0"]
+        ["1.9", "3.0"],
+        ["3.0", "4.0"]
       ]
       upgrades.include?([@version, ENV["CROWBAR_VERSION"]])
     end
--- a/crowbar_framework/lib/crowbar/upgrade.rb
+++ b/crowbar_framework/lib/crowbar/upgrade.rb
@@ -31,8 +31,11 @@ module Crowbar
     end
 
     def upgrade
-      knife_files
-      crowbar_files
+      case @version
+      when "1.9"
+        knife_files
+        crowbar_files
+      end
     end
 
     def supported?
--- a/crowbar_framework/config/locales/crowbar/en.yml
+++ b/crowbar_framework/config/locales/crowbar/en.yml
@@ -733,9 +733,9 @@ en:
       meta_missing: 'meta.yml metadata file is missing in the backup'
       invalid_backup: 'Cannot restore from invalid backup'
       missing_backup: 'Backup does not exist'
-      upgrade_not_supported: 'Upgrade from %{backup_version} to %{system_version} not supported'
+      upgrade_not_supported: 'Upgrade from %{backup_version} to %{system_version} is not supported'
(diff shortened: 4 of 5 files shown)


MaximilianMeister:
so this patchset should get us a bit further now, I manually tested until restore, but @vuntz it seems, although I am not 100% sure, that the restore background thread disappeared again :-( at least it went missing at restoring the chef files.

the last log entry was https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/backup/restore.rb#L178

I have triggered an upgrade gating now, lets see if it happens again

vuntz:
> so this patchset should get us a bit further now, I manually tested until restore, but @vuntz it seems, although I am not 100% sure, that the restore background thread disappeared again :-( at least it went missing at restoring the chef files.

Do you have the link to the build?

MaximilianMeister:
> Do you have the link to the build?

I was looking for it as well, but I couldn't find it anymore since I triggered a new build in the morning, do you have any idea how to find it in the jenkins UI?

MaximilianMeister:
@vuntz found it https://ci.suse.de/job/openstack-mkcloud/19258

crowbar_framework/app/models/backup.rb
```diff
@@ -75,14 +75,23 @@ def restore(options = {})
     from_upgrade = options.fetch(:from_upgrade, false)
 
     if Crowbar::Backup::Restore.restore_steps_path.exist?
+      logger.debug I18n.t("backups.index.multiple_restore")
```
vuntz: We usually don't put localized messages in the logs (to avoid having to read french messages ;-))

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -226,11 +233,11 @@ def save_and_restore
       return false unless @backup.save
       if Crowbar::Backup::Restore.restore_steps_path.exist?
         flash[:info] = t("installer.upgrades.restore.multiple_restore")
-        true
       else
         Crowbar::Backup::Restore.purge
         @backup.restore(background: true, from_upgrade: true)
       end
+      true
```
vuntz: I'm not sure I understand that change: don't we want to respect the return value of `@backup.restore`?

MaximilianMeister: no we haven't yet, there is no specific return value as it just opens a thread (not even sure what it returns)

we just respect a false return value if the backup can't be saved

vuntz: So maybe that's what we should fix? I mean, if it returns `false` in some case, then it sounds wrong to just always return `true` here.

MaximilianMeister: yes you're right, I will try to come up with something better here

crowbar_framework/app/models/backup.rb
MaximilianMeister: since when do we have french localization? ;-) but I get the point

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -27,6 +27,41 @@
...
+      done
+    EOF
+    only_if { node[:platform] == "suse" }
+  end
+
+  # Disable crowbar-join
+  service "crowbar_join" do
+    # do not stop it, it would change node's state
+    action :disable
+    only_if { node[:platform] == "suse" }
```
MaximilianMeister: we were wondering if this check is needed as we are already checking the platform_family at the beginning of the recipe.

basically I think it is correct in case a service could have different names in different suse products.
any other opinions?

jsuchome: I would drop it from here. The check for platform_family at the beginning of the recipe was added later and should replace these only_if's. If we want distinguish SLES and Leap, we can do it later.

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -27,6 +27,41 @@
...
+when "crowbar_upgrade"
+
+  # Disable openstack services
+  # We don't know which openstack services are enabled on the node and
+  # collecting that information via the attributes provided by chef is
+  # rather complicated. So instead we fall back to a simple bash hack
+
+  bash "disable_openstack_services" do
+    code <<-EOF
+      for i in /etc/init.d/openstack-* /etc/init.d/openvswitch-switch /etc/init.d/ovs-usurp-config-* /etc/init.d/drbd /etc/init.d/openais;
```
jsuchome: This initscript magic worked on Cloud5 (SLES11 based). You might need to double check if it works on SLE12

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
MaximilianMeister: moved the commit to #430

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -224,13 +231,14 @@ def meta_title
...
+      success = true
       if Crowbar::Backup::Restore.restore_steps_path.exist?
         flash[:info] = t("installer.upgrades.restore.multiple_restore")
-        true
       else
         Crowbar::Backup::Restore.purge
-        @backup.restore(background: true, from_upgrade: true)
+        success = @backup.restore(background: true, from_upgrade: true)
       end
+      success
```
rsalevsky: Please keep it as it is.

crowbar_framework/app/models/backup.rb
```diff
@@ -75,14 +75,19 @@ def restore(options = {})
     from_upgrade = options.fetch(:from_upgrade, false)
 
     if Crowbar::Backup::Restore.restore_steps_path.exist?
+      logger.debug "Restore process is already running"
       errors.add(:base, I18n.t("backups.index.multiple_restore"))
       return false
     elsif !from_upgrade && backup_version != system_version
+      logger.debug "Cannot restore from different Crowbar version"
```
rsalevsky: Maybe: "Restoring from different crowbar version is not allowed."?

crowbar_framework/app/models/backup.rb
```diff
@@ -75,14 +75,19 @@ def restore(options = {})
...
       errors.add(:base, I18n.t("backups.index.version_conflict"))
       return false
     end
 
     if upgrade? && !upgrade
+      logger.debug(
+        "Upgrade from #{version} to #{ENV["CROWBAR_VERSION"]} not supported"
```
rsalevsky: is not

crowbar_framework/app/models/backup.rb
MaximilianMeister: i just took the i18n strings that we are already using, so shall we change them as well?

MaximilianMeister: done

rsalevsky:
+1

crowbar_framework/app/models/backup.rb
```diff
@@ -75,14 +75,19 @@ def restore(options = {})
...
     elsif !from_upgrade && backup_version != system_version
+      logger.debug "Restoring from different Crowbar version is not allowed"
       errors.add(:base, I18n.t("backups.index.version_conflict"))
       return false
     end
 
     if upgrade? && !upgrade
+      logger.debug(
+        "Upgrade from #{version} to #{ENV["CROWBAR_VERSION"]} is not supported"
+      )
```
vuntz: Do we know for sure that this will only fail because of this?

MaximilianMeister: well it could fail in `Crowbar::Upgrade.upgrade` because of a different reason

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -37,6 +37,9 @@ def upgrade
         crowbar_files
       end
 
+      if @status[:errors].any?
+        @backup.errors.add(:upgrade, @status[:errors])
+      end
```
MaximilianMeister: @rsalevsky not sure if that is necessary here, do you think it makes sense to add the errors to the model? and does it make a difference that the model object is cached here?

crowbar_framework/app/models/backup.rb
MaximilianMeister: I guess it would be good to split the upgrade method up, to differentiate the error cases, but I'd prefer to not do it in this PR, would it be good enough for you if I create a card? right now in the worst case we would have a misleading log message

vuntz: I'm fine with that, but then I'd drop this log line from this PR.

MaximilianMeister: okay, dropped

vuntz:
+1

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -31,13 +31,22 @@ def initialize(backup)
     end
 
     def upgrade
-      knife_files
-      crowbar_files
+      case ENV["CROWBAR_VERSION"]
+      when "3.0"
```
rsalevsky: what is with 1.9?

MaximilianMeister: is it ever called on cloud5?

rsalevsky: Yeah, true so you should check the backup version here. Otherwise it's totally confusing...

MaximilianMeister: you're right thanks. changed

rsalevsky:
+1

MaximilianMeister:
merging, as this will bring all the ci upgrade jobs further... especially https://github.com/crowbar/crowbar-core/pull/428/commits/84d01b7102e878810e2abd3260825be1767c41f1
....and all gating green as well

PR #430, opened 2016-03-17
upgrade: add back end part of prepare in cloud 6

MaximilianMeister:
the whole backend part of the prepare step needed to be forward ported
from cloud 5

cc @jsuchome 

this would replace https://github.com/SUSE-Cloud/automation/pull/912

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -37,17 +37,11 @@ when "crowbar_upgrade"
 
   bash "disable_openstack_services" do
     code <<-EOF
-      for i in /etc/init.d/openstack-* \
-               /etc/init.d/openvswitch-switch \
-               /etc/init.d/ovs-usurp-config-* \
-               /etc/init.d/drbd \
-               /etc/init.d/openais;
+      for i in $(systemctl list-units openstack* --no-legend | cut -d" " -f1) \
+               drbd.service \
+               pacemaker.service;
       do
-        if test -e $i
-        then
-          initscript=`basename $i`
-          chkconfig -d $initscript
-        fi
+        systemctl disable $i
       done
     EOF
   end
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -27,6 +27,42 @@ return unless node[:platform_family] == "suse"
 upgrade_step = node["crowbar_wall"]["crowbar_upgrade_step"] || "none"
 
 case upgrade_step
+
+when "crowbar_upgrade"
+
+  # Disable openstack services
+  # We don't know which openstack services are enabled on the node and
+  # collecting that information via the attributes provided by chef is
+  # rather complicated. So instead we fall back to a simple bash hack
+
+  bash "disable_openstack_services" do
+    code <<-EOF
+      for i in /etc/init.d/openstack-* \
+               /etc/init.d/openvswitch-switch \
+               /etc/init.d/ovs-usurp-config-* \
+               /etc/init.d/drbd \
+               /etc/init.d/openais;
+      do
+        if test -e $i
+        then
+          initscript=`basename $i`
+          chkconfig -d $initscript
+        fi
+      done
+    EOF
+  end
+
+  # Disable crowbar-join
+  service "crowbar_join" do
+    # do not stop it, it would change node's state
+    action :disable
+  end
+
+  # Disable chef-client
+  service "chef-client" do
+    action [:disable, :stop]
+  end
+
 when "openstack_shutdown"
 
   include_recipe "crowbar::stop-services-before-upgrade"
--- a/crowbar_framework/app/controllers/installer/upgrades_controller.rb
+++ b/crowbar_framework/app/controllers/installer/upgrades_controller.rb
@@ -23,6 +23,25 @@ module Installer
     before_filter :set_progess_values
     before_filter :set_service_object, only: [:services, :backup, :nodes]
 
+    def prepare
+      status = :ok
+
+      begin
+        service_object = CrowbarService.new(Rails.logger)
+
+        service_object.prepare_nodes_for_crowbar_upgrade
+      rescue => e
+        Rails.logger.error e.message
+        status = :unprocessable_entity
+      end
+
+      respond_to do |format|
+        format.html do
+          head :no_content, status: status
+        end
+      end
+    end
+
     def show
       respond_to do |format|
         format.html do
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -312,6 +312,44 @@ class CrowbarService < ServiceObject
     base
   end
 
+  def prepare_nodes_for_crowbar_upgrade
+    proposal = Proposal.find_by(barclamp: "crowbar", name: "default")
+
+    # To all nodes, add a new role which prepares them for the upgrade
+    nodes_to_upgrade = []
+    not_ready_for_upgrade = []
+    all_nodes = NodeObject.all
+    all_nodes.each do |n|
+      next if n.admin? && ["ready", "crowbar_upgrade"].include?(n.state)
+      not_ready_for_upgrade.push(n.name)
+    end
+
+    unless not_ready_for_upgrade.empty?
+      raise I18n.t(
+        "installer.upgrades.prepare.nodes_not_ready", nodes: not_ready_for_upgrade.join(", ")
+      )
+    end
+
+    all_nodes.each do |node|
+      next if node.admin?
+
+      if node[:platform] == "windows"
+        # for Hyper-V nodes, only change the state, but do not run chef-client
+        node.set_state("crowbar_upgrade")
+      else
+        node["crowbar_wall"]["crowbar_upgrade_step"] = "crowbar_upgrade"
+        node.save
+        nodes_to_upgrade.push node.name
+      end
+    end
+
+    # adapt current proposal, so the nodes get crowbar-upgrade role
+    proposal.raw_data["deployment"]["crowbar"]["elements"]["crowbar-upgrade"] = nodes_to_upgrade
+    proposal.save
+    # commit the proposal so chef recipe get executed
+    proposal_commit("default", false, false)
+  end
+
   def disable_non_core_proposals
     upgrade_nodes = NodeObject.all.reject(&:admin?)
     check_if_nodes_are_available upgrade_nodes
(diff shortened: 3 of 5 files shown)


MaximilianMeister:
this is currently untested

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -27,6 +27,38 @@
...
+when "crowbar_upgrade"
+
+  # Disable openstack services
+  # We don't know which openstack services are enabled on the node and
+  # collecting that information via the attributes provided by chef is
+  # rather complicated. So instead we fall back to a simple bash hack
+
+  bash "disable_openstack_services" do
+    code <<-EOF
+      for i in /etc/init.d/openstack-* /etc/init.d/openvswitch-switch /etc/init.d/ovs-usurp-config-* /etc/init.d/drbd /etc/init.d/openais;
```
jsuchome: Please put those on separate lines for better readability (https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/crowbar/recipes/stop-services-before-upgrade.rb#L55)

jsuchome:
In general the port looks correct; does it work? That would mean, nodes are in new state, chef-client is not running there and services are disabled.

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -23,6 +23,27 @@ class UpgradesController < ApplicationController
...
+      if request.post?
+        begin
+          service_object = CrowbarService.new(Rails.logger)
+
+          service_object.prepare_nodes_for_crowbar_upgrade
+        rescue => e
+          Rails.logger.error e.message
+          status = :unprocessable_entity
+        end
+      end
```
jdsn: no else? what happens if request is not post?

MaximilianMeister: I guess we could drop the `if request.post?` as i didn't define any GET route

jdsn: So a get would fail due to missing route - ok fine with me then.
I guess I would keep the `if request.post?` even though.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -312,6 +312,43 @@ def create_proposal
...
 
+  def prepare_nodes_for_crowbar_upgrade
+    proposal = Proposal.find_by(barclamp: "crowbar", name: "default")
+
+    # To all nodes, add a new role which prepares them for the upgrade
+    nodes_to_upgrade = []
+    not_ready_for_upgrade = []
+    all_nodes = NodeObject.all
+    all_nodes.each do |n|
+      not_ready_for_upgrade.push n.name if !n.admin? && !%w(ready crowbar_upgrade).include?(n.state)
```
rsalevsky: `%w(ready crowbar_upgrade)` -> `["ready", "crowbar_upgrade"]`

MaximilianMeister:
> This initscript magic worked on Cloud5 (SLES11 based). You might need to double check if it works on SLE12

@jsuchome continuing from your [comment](https://github.com/crowbar/crowbar-core/pull/428#discussion_r56627684) I just checked on a sle12sp1 node `/etc/init.d/ovs-usurp-config-*` `/etc/init.d/drbd` `/etc/init.d/openais` are the only ones that don't exist, but I guess I haven't applied the required barclamps for that.

so this code block should work fine

crowbar_framework/app/models/crowbar_service.rb
MaximilianMeister: I just took it from the other commit.
I am happy to change it, but if it's not our style guide, why does @houndci-bot not bitch?

rsalevsky: Not sure I think this style is just not tracked by hound. At least we should fix it now the other commit is from the time where we just merged the stuff.

jsuchome:
Regarding the service names, we'll gonna need to adapt it to systemd... AFAIK there's `pacemaker.service` instead of `init.d/openais`

rsalevsky:
+1

MaximilianMeister:
> Regarding the service names, we'll gonna need to adapt it to systemd... AFAIK there's pacemaker.service instead of init.d/openais

so `/etc/init.d/openais` is not available anymore? then we have to use systemctl for all

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -27,6 +27,41 @@
...
+  # We don't know which openstack services are enabled on the node and
+  # collecting that information via the attributes provided by chef is
+  # rather complicated. So instead we fall back to a simple bash hack
+
+  bash "disable_openstack_services" do
+    code <<-EOF
+      for i in /etc/init.d/openstack-* \
+               /etc/init.d/openvswitch-switch \
+               /etc/init.d/ovs-usurp-config-* \
+               /etc/init.d/drbd /etc/init.d/openais;
```
MaximilianMeister: i even forgot to break this line...

MaximilianMeister:
> Regarding the service names, we'll gonna need to adapt it to systemd... AFAIK there's pacemaker.service instead of init.d/openais

@vuntz @aspiers can you confirm that?

aspiers:
> @vuntz @aspiers can you confirm that?

```
root@d52-54-77-77-77-02:~ # rpm -ql pacemaker | grep systemd
/usr/lib/systemd/system/pacemaker.service
```

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -27,6 +27,46 @@
...
+        if test -e $i
+        then
+          initscript=`basename $i`
+          chkconfig -d $initscript
+        fi
+      done
+
+      if systemctl status pacemaker.service 2>&1 > /dev/null
+      then
+        systemctl disable pacemaker.service
```
MaximilianMeister: @jsuchome @aspiers is it ok like that?

jsuchome: 1. `systemctl status` checks if service is running or enabled?

2. I think that we should replace all services in `disable_openstack_services` block with systemd ones, isn't SLES12 systemd-based system?

vuntz: Just always call `systemctl disable`, imho. No need for testing anything.

> systemctl status checks if service is running or enabled?

If it's running. I think `is-enabled` is the right option for the intent.

MaximilianMeister: @vuntz thanks. done

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -27,6 +27,43 @@
...
+  # We don't know which openstack services are enabled on the node and
+  # collecting that information via the attributes provided by chef is
+  # rather complicated. So instead we fall back to a simple bash hack
+
+  bash "disable_openstack_services" do
+    code <<-EOF
+      for i in /etc/init.d/openstack-* \
+               /etc/init.d/openvswitch-switch \
+               /etc/init.d/ovs-usurp-config-* \
+               /etc/init.d/drbd;
```
vuntz: I guess this can be done in a followup, but ovs-usurp-config-* scripts are gone, openvswitch-switch is now openvswitch (but systemd service, not init script). Not sure about drbd.

vuntz:
This is submitted to master, but the message says it's for Cloud 6. So should it just go to 3.0? Or do we want it in master and then backport?

MaximilianMeister:
> Or do we want it in master and then backport?

that was my thought behind it

vuntz:
Also, aren't we missing more code? Like at least `revert_nodes_from_crowbar_upgrade`.

MaximilianMeister:
> Also, aren't we missing more code? Like at least revert_nodes_from_crowbar_upgrade.

for the back-end as it is now, it is not really needed (no view, no abort button). i left it out, because I wanted to skip all UI related code as it is still not discussed

vuntz:
>> Also, aren't we missing more code? Like at least revert_nodes_from_crowbar_upgrade.
> 
> for the back-end as it is now, it is not really needed (no view, no abort button). i left it out, because I wanted to skip all UI related code as it is still not discussed

But we know for sure we'll want to allow that, don't we? It's not purely frontend code, it's just part of the "cancel upgrade" feature.

MaximilianMeister:
> But we know for sure we'll want to allow that, don't we? It's not purely frontend code, it's just part of the "cancel upgrade" feature.

I understand that, but the code wouldn't be tested/touched yet, so no guarantee that it'll work. that's also why I didn't want to just add it just to have it... but if you prefer I can add it

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -27,6 +27,35 @@
...
+  # We don't know which openstack services are enabled on the node and
+  # collecting that information via the attributes provided by chef is
+  # rather complicated. So instead we fall back to a simple bash hack
+
+  bash "disable_openstack_services" do
+    code <<-EOF
+      for i in $(systemctl list-units openstack* --no-legend | cut -d" " -f1) \
+               drbd.service;
+      do
+        systemctl disable $i
```
MaximilianMeister: @jsuchome I have updated now the loop to use only systemd

jsuchome: I like that!

MaximilianMeister:
@rsalevsky is your +1 still valid? I have changed the bash code to disable services with systemd instead of chkconfig

MaximilianMeister:
>> But we know for sure we'll want to allow that, don't we? It's not purely frontend code, it's just part of the "cancel upgrade" feature.

> I understand that, but the code wouldn't be tested/touched yet, so no guarantee that it'll work. that's also why I didn't want to just add it just to have it... but if you prefer I can add it

@vuntz would it be fine for now if we just add it when we need it? so in a followup?

vuntz:
@MaximilianMeister hrm, okay. My more general point is that I don't know if any other code elsewhere might be missing. Hopefully not :-)

MaximilianMeister:
> hrm, okay. My more general point is that I don't know if any other code elsewhere might be missing. Hopefully not :-)

@vuntz I understand, but as all upgrade related ci is currently broken anyways, this will help a lot to get further if we merge this now, and find out if there is something missing :-)

rsalevsky:
Still +1

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -27,6 +27,35 @@
...
+
+  # Disable openstack services
+  # We don't know which openstack services are enabled on the node and
+  # collecting that information via the attributes provided by chef is
+  # rather complicated. So instead we fall back to a simple bash hack
+
+  bash "disable_openstack_services" do
+    code <<-EOF
+      for i in $(systemctl list-units openstack* --no-legend | cut -d" " -f1) \
+               drbd.service;
```
jsuchome: Aren't we missing pacemaker here?

jsuchome:
+1 codewise (but please confirm it works as exepcted)

PR #432, opened 2016-03-18
Update byebug version to 8.2.2

rsalevsky:


MaximilianMeister:
as there is no packaging needed +1

dirkmueller:
would be nice to have a rationale added to the git commit message on why the update is needed, but I don't demand it. +1

PR #435, opened 2016-03-18
crowbar, provisioner: Change default FS to btrfs

vuntz:
Let's try btrfs on all nodes to see if we hit any issue with it.

Note that this setting is only used for the SUSE platform family.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -344,7 +344,7 @@ if not nodes.nil? and not nodes.empty?
                       architecture: arch,
                       is_ses: storage_available && !cloud_available,
                       crowbar_join: "#{os_url}/crowbar_join.sh",
-                      default_fs: mnode[:crowbar_wall][:default_fs] || "ext4",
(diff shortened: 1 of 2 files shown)


jsuchome:
Yey, let's have fun! +1

rsalevsky:
+1

tboerger:
:+1:

dirkmueller:
@vuntz where is the PR that adds the submounts for nova, glance ?

vuntz:
> @vuntz where is the PR that adds the submounts for nova, glance ?

Not there yet because I don't know which subvolumes we need due to specific needs. So I guess we want:

- /var/lib/nova/instances (no snapshot, no cow) -- or just /var/lib/nova?
- /var/lib/glance/images (you were mentioning that one, is that to avoid snapshot, or also cow?) or just /var/lib/glance?
- do we need something for /var/lib/rabbitmq?
- do we want something for /var/lib/cinder (in case loopback device for lvm is used)?
- do we want something for mongodb?

I guess the right thing to do is to create a card to investigate all this, unless someone already knows all the answer.

vuntz:
Need more info for this; my PR is clearly not enough for now, so closing it.

PR #421, opened 2016-03-11
add specs for the Crowbar::Installer library

MaximilianMeister:

PR #438, opened 2016-03-21
[stable/3.0] backport #421

MaximilianMeister:
(cherry picked from commit 5190529ab16891470263532e7380b80b9cea8b53)

I think it doesn't hurt to backport the specs, if we make some changes it is good to have

#421

### DIFF
--- a/crowbar_framework/spec/fixtures/installer_status.json
+++ b/crowbar_framework/spec/fixtures/installer_status.json
@@ -1,0 +1,14 @@
+{
+  "steps":[],
+  "failed":false,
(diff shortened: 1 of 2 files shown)


rsalevsky:
Not sure but backporting tests makes no sense for me.

MaximilianMeister:
> Not sure but backporting tests makes no sense for me.

more coverage is always good to have, and here it doesn't cost anything extra, so why not add them?

vuntz:
LGTM

dirkmueller:
+1

PR #440, opened 2016-03-22
correct the logic

MaximilianMeister:
it was messed up after style formatting from the original code

see https://github.com/crowbar/barclamp-crowbar/blob/release/tex/master/crowbar_framework/app/models/crowbar_service.rb#L200

```ruby
not_ready_for_upgrade.push n.name if !n.admin? && !%w(ready crowbar_upgrade).include?(n.state)
```

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -320,7 +320,7 @@ class CrowbarService < ServiceObject
     not_ready_for_upgrade = []
     all_nodes = NodeObject.all
     all_nodes.each do |n|
-      next if n.admin? && ["ready", "crowbar_upgrade"].include?(n.state)
+      next if n.admin? || ["ready", "crowbar_upgrade"].include?(n.state)
       not_ready_for_upgrade.push(n.name)
     end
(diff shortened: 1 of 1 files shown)


rsalevsky:
+1

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -320,7 +320,7 @@ def prepare_nodes_for_crowbar_upgrade
     not_ready_for_upgrade = []
     all_nodes = NodeObject.all
     all_nodes.each do |n|
-      next if n.admin? && ["ready", "crowbar_upgrade"].include?(n.state)
+      next unless n.admin? && ["ready", "crowbar_upgrade"].include?(n.state)
```
jsuchome: `next if  n.admin? || ["ready", "crowbar_upgrade"].include?(n.state)`

MaximilianMeister: right... confusing because we changed the logic completely with the introduction of `next`

jsuchome:
+1

rsalevsky:
+1

PR #441, opened 2016-03-22
Chef upload library

rsalevsky:


crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,202 @@
...
+          version_constraints_to_update = Hash.new
+          cookbooks_for_upload = Array.new
+
+          cookbooks_repo = ::Chef::CookbookLoader.new(chef_cookbooks_path)
+          cookbooks_repo.each do |cookbook_name, cookbook|
+            cookbooks_for_upload.push(cookbook)
+            version_constraints_to_update[cookbook_name] = cookbook.version
+          end
+
+          if cookbooks_for_upload.any?
```
vuntz: `return unless cookbooks_for_upload.any` so we can lose one indentation level.

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,202 @@
...
+
+        def check_links(cookbook)
+          broken_files = cookbook.dup.manifest_records_by_path.select do |path, info|
+            info[:checksum].nil? || info[:checksum] !~ /[0-9a-f]{32,}/
+          end
+
+          unless broken_files.empty?
+            broken_filenames = Array(broken_files).map { |path, info| path }
+            logger.error(
+              "The cookbook #{cookbook.name} has the folowing broken files: " \
```
vuntz: following

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,202 @@
...
+            check_server_side_cookbooks(cookbook_name, version) ||
+            check_uploading_cookbooks(cookbook_name, version)
+          end
+
+          unless missing_dependencies.empty?
+            missing_cookbooks = missing_dependencies.map do |cookbook_name, version|
+              "\"#{cookbook_name}\" version \"#{version}\""
+            end
+            logger.error(
+              "Cookbook #{cookbook.name} depends on the folowing cookbooks: " \
```
vuntz: following

vuntz: Also the message should say that the dependency cannot be satisfied.

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,202 @@
...
+            logger.info("Uploading cookbooks... #{cookbook_names.join(", ")}")
+            ::Chef::CookbookUploader.new(
+              cookbooks_for_upload,
+              force: false,
+              concurrency: true
+            ).upload_cookbooks
+          end
+        end
+
+        def data_bags
```
vuntz: To be honest, I think we should stop uploading the data bags, they shouldn't be needed as data bags anymore. I'd work on fixing that instead.

vuntz:
One general concern I have on this is that we seem to upload everything, while when we update crowbar-ceph (for instance), this is not needed and will result in something much slower than what we have today.

rsalevsky:
Uploading everything takes at the moment 10 seconds. I don't think this will make a huge difference the cookbooks get anyway uploaded in parallel. There is just some delay (~ 2s) while uploading the data_bags and roles.

vuntz:
> Uploading everything takes at the moment 10 seconds. I don't think this will make a huge difference the cookbooks get anyway uploaded in parallel.

So when you update all the packages, it takes 50 seconds. Instead of 10 seconds. It does make a difference.

rsalevsky:
The calculation is wrong. :) When we just upload the cookbooks we need then we save around 2-3 seconds per run. For 5 packages this makes maximum 15 seconds. At the end this process is anyway handelt by crowbar and then we can trigger it just once.

Also this PR is still WIP and I think there is a way to just upload the cookbooks which changed.

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,193 @@
...
+          cookbooks
+          data_bags
+          roles
+        end
+
+        def cookbooks
+          version_constraints_to_update = Hash.new
+          cookbooks_for_upload = Array.new
+
+          cookbooks_repo = ::Chef::CookbookLoader.new(chef_cookbooks_path)
```
MaximilianMeister: i would drop this variable

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,193 @@
...
+          cookbooks_repo.each do |cookbook_name, cookbook|
```
MaximilianMeister: ...and instead use `cookbook_loader.each`

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,193 @@
...
+            cookbooks_for_upload.push(cookbook)
+            version_constraints_to_update[cookbook_name] = cookbook.version
+          end
+
+          return unless cookbooks_for_upload.any?
```
MaximilianMeister: maybe return something here. that would make testing much easier

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,193 @@
...
+          ::Chef::CookbookUploader.new(
+            cookbooks_for_upload,
+            force: false,
+            concurrency: cookbooks_for_upload.count
+          ).upload_cookbooks
+        end
+
+        def data_bags
+          loader = ::Chef::Knife::Core::ObjectLoader.new(::Chef::DataBagItem, logger)
+          data_bags = loader.find_all_object_dirs(chef_data_bags_path)
```
MaximilianMeister: can this eventually be `nil` ? or something else that is not iterable?

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,193 @@
...
+          data_bags.each do |data_bag|
+            data_bag_items = loader.find_all_objects(chef_data_bags_path.join(data_bag))
```
MaximilianMeister: here as well, does this always return something useful (an iterator)?

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,193 @@
...
+            data_bag_item_paths = normalize_data_bag_item_paths(data_bag_items)
+            data_bag_item_paths.each do |data_bag_item_path|
+              # Workaround for a strange chef behavior
```
MaximilianMeister: maybe explain the behavior a bit more, to know why we need a workaround

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,193 @@
...
+              relative_path = chef_data_bags_path.relative_path_from(Pathname.new(Dir.pwd))
+              data_bag_item = loader.load_from(relative_path, data_bag, data_bag_item_path)
+              bag = ::Chef::DataBagItem.new
+              bag.data_bag(data_bag)
+              bag.raw_data = data_bag_item
+              logger.info("Uploading data_bag #{data_bag_item_path}")
+              bag.save
```
MaximilianMeister: could `bag.save` raise an exception?

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,193 @@
...
+          end
+        end
+
+        def roles
+          loader = ::Chef::Knife::Core::ObjectLoader.new(::Chef::Role, logger)
+          chef_roles_path.each_child do |file|
+            next unless file.file?
+            role = loader.load_from("roles", file)
+            logger.info("Uploading role #{file.basename}")
+            role.save
```
MaximilianMeister: here as well. not sure if that can raise an exception, and if we should handle it here

crowbar_framework/lib/crowbar/chef/upload.rb
rsalevsky: the line is too long otherwise

crowbar_framework/lib/crowbar/chef/upload.rb
rsalevsky: Didn't check but in case it's also not handled by knife.

crowbar_framework/lib/crowbar/chef/upload.rb
rsalevsky: Same here...

crowbar_framework/lib/crowbar/chef/upload.rb
vuntz: I assume it means you copied code from knife? If yes, the copyright header needs to be reflect that (also not sure what's the license for knife).

vuntz:
> The calculation is wrong. :) When we just upload the cookbooks we need then we save around 2-3 seconds per run. For 5 packages this makes maximum 15 seconds.

This is my concern:

```
$ time for i in ceph core ha hyperv openstack; do barclamp_install.rb --rpm $i; done
Installing component ceph from /opt/dell/barclamps/ceph
Chef components for (ceph) (format v1) uploaded.
Installing component core from /opt/dell/barclamps/core
Chef components for (core) (format v1) uploaded.
Installing component ha from /opt/dell/barclamps/ha
Chef components for (ha) (format v1) uploaded.
Installing component hyperv from /opt/dell/barclamps/hyperv
WARNING: didn't find any data bags in /opt/dell/chef/data_bags
WARNING: didn't find any roles in /opt/dell/chef/roles
Chef components for (hyperv) (format v1) uploaded.
Installing component openstack from /opt/dell/barclamps/openstack
Chef components for (openstack) (format v1) uploaded.

real    2m18.789s
user    1m28.016s
sys     0m11.408s
$
$
$
$
$ time for i in ceph core ha hyperv openstack; do barclamp_install.rb --rpm ceph core ha hyperv openstack; done
Installing component ceph from /opt/dell/barclamps/ceph
Installing component core from /opt/dell/barclamps/core
Installing component ha from /opt/dell/barclamps/ha
Installing component hyperv from /opt/dell/barclamps/hyperv
Installing component openstack from /opt/dell/barclamps/openstack
Chef components for (ceph, core, ha, hyperv, openstack) (format v1) uploaded.
Installing component ceph from /opt/dell/barclamps/ceph
Installing component core from /opt/dell/barclamps/core
Installing component ha from /opt/dell/barclamps/ha
Installing component hyperv from /opt/dell/barclamps/hyperv
Installing component openstack from /opt/dell/barclamps/openstack
Chef components for (ceph, core, ha, hyperv, openstack) (format v1) uploaded.
Installing component ceph from /opt/dell/barclamps/ceph
Installing component core from /opt/dell/barclamps/core
Installing component ha from /opt/dell/barclamps/ha
Installing component hyperv from /opt/dell/barclamps/hyperv
Installing component openstack from /opt/dell/barclamps/openstack
Chef components for (ceph, core, ha, hyperv, openstack) (format v1) uploaded.
Installing component ceph from /opt/dell/barclamps/ceph
Installing component core from /opt/dell/barclamps/core
Installing component ha from /opt/dell/barclamps/ha
Installing component hyperv from /opt/dell/barclamps/hyperv
Installing component openstack from /opt/dell/barclamps/openstack
Chef components for (ceph, core, ha, hyperv, openstack) (format v1) uploaded.
Installing component ceph from /opt/dell/barclamps/ceph
Installing component core from /opt/dell/barclamps/core
Installing component ha from /opt/dell/barclamps/ha
Installing component hyperv from /opt/dell/barclamps/hyperv
Installing component openstack from /opt/dell/barclamps/openstack
Chef components for (ceph, core, ha, hyperv, openstack) (format v1) uploaded.

real    11m49.038s
user    6m51.212s
sys     0m53.380s
```

> At the end this process is anyway handelt by crowbar and then we can trigger it just once.

I'm curious how :-)

rsalevsky:
@vuntz Crowbar checks if there are new cookbooks in the directory when yes crowbar starts the upload, this happens at the start of crowbar and later via a cronjob. We install the barclamps and then we start crowbar or restart crowbar so we do it just once.

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,192 @@
...
+
+        def data_bags
+          loader = ::Chef::Knife::Core::ObjectLoader.new(::Chef::DataBagItem, logger)
+          data_bags = loader.find_all_object_dirs(chef_data_bags_path) || []
+          data_bags.each do |data_bag|
+            data_bag_items = loader.find_all_objects(chef_data_bags_path.join(data_bag))
+            data_bag_item_paths = normalize_data_bag_item_paths(data_bag_items) || []
+            data_bag_item_paths.each do |data_bag_item_path|
+              # Workaround for a strange chef behavior
+              relative_path = chef_data_bags_path.relative_path_from(Pathname.new(Dir.pwd))
```
tboerger: Why the `Pathname`?

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,192 @@
...
+        end
+
+        protected
+
+        def logger
+          return Rails.logger unless caller.grep(/rake/).present?
+          Logger.new(STDOUT)
+        end
+
+        def chef_dir_path
```
tboerger: Just `chef_path`, dir path sounds simply strange.

crowbar_framework/lib/tasks/chef.rake
```diff
@@ -0,0 +1,44 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+namespace :chef do
+  namespace :upload do
+    require "crowbar/chef/upload"
+
+    desc "Upload the complete chef data to the server"
+    task :all do
```
tboerger: Do we need some model or environment access later on? As you are anyway using `Rails` already I would suggest to use `task all: [:environment] do`, than you can drop the require and we've got full access to the rails environment. Should be done for the other tasks as well.

tboerger:
A sync can be done relative easy as the chef server API can spit out checksums for the involved files, as an example the `xfs` cookbook, fetched with the scripts from @MaximilianMeister, you can find it at https://github.com/MaximilianMeister/bin:

### chef_api 192.168.106.10 GET /cookbooks/xfs/1.0.0

```json
{
    "attributes": [],
    "chef_type": "cookbook_version",
    "cookbook_name": "xfs",
    "definitions": [],
    "files": [],
    "frozen?": false,
    "json_class": "Chef::CookbookVersion",
    "libraries": [],
    "metadata": {
        "attributes": {},
        "conflicting": {},
        "dependencies": {},
        "description": "Installs packages for working with XFS",
        "groupings": {},
        "license": "Apache 2.0",
        "long_description": "Description\n====\n\nInstalls packages for working with XFS filesystems.\n\nRequirements\n====\n\nTested on Ubuntu 10.04, CentOS 5.5 and Scientific Linux 6.0.\n\nShould work on any Debian or Red Hat family Linux distributions that\nhave the xfs packages in a default repository.\n\nLicense and Author\n====\n\nAuthor:: Joshua Timberman (<joshua@opscode.com>)\nCopyright:: 2009-2011, Opscode, Inc.\n\nLicensed under the Apache License, Version 2.0 (the \"License\");\nyou may not use this file except in compliance with the License.\nYou may obtain a copy of the License at\n\n    http://www.apache.org/licenses/LICENSE-2.0\n\nUnless required by applicable law or agreed to in writing, software\ndistributed under the License is distributed on an \"AS IS\" BASIS,\nWITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.\nSee the License for the specific language governing permissions and\nlimitations under the License.\n",
        "maintainer": "Opscode, Inc.",
        "maintainer_email": "cookbooks@opscode.com",
        "name": "xfs",
        "platforms": {
            "debian": ">= 0.0.0",
            "ubuntu": ">= 0.0.0"
        },
        "providing": {
            "xfs": ">= 0.0.0"
        },
        "recipes": {
            "xfs": "Installs packages for working with XFS"
        },
        "recommendations": {},
        "replacing": {},
        "suggestions": {},
        "version": "1.0.0"
    },
    "name": "xfs-1.0.0",
    "providers": [],
    "recipes": [
        {
            "checksum": "b10fc1d4f6f03513bcbc89114986fef9",
            "name": "default.rb",
            "path": "recipes/default.rb",
            "specificity": "default",
            "url": "http://localhost:4000/cookbooks/xfs/1.0.0/files/b10fc1d4f6f03513bcbc89114986fef9"
        }
    ],
    "resources": [],
    "root_files": [
        {
            "checksum": "788eac8b7f8dbe970cca46d02e729ace",
            "name": "metadata.rb",
            "path": "metadata.rb",
            "specificity": "default",
            "url": "http://localhost:4000/cookbooks/xfs/1.0.0/files/788eac8b7f8dbe970cca46d02e729ace"
        },
        {
            "checksum": "034120e93c94c660cce401032243ef82",
            "name": "README.md",
            "path": "README.md",
            "specificity": "default",
            "url": "http://localhost:4000/cookbooks/xfs/1.0.0/files/034120e93c94c660cce401032243ef82"
        }
    ],
    "templates": [],
    "version": "1.0.0"
}
```

rsalevsky:
@vuntz I implement yet the suggested cookbook sync from @tboerger. This should obsolete the rpm workaround.

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,277 @@
...
+              paths.concat(Dir.glob(File.join(path, "*.json")))
+            else
+              paths.push(path)
+            end
+          end
+          paths
+        end
+
+        def api
+          @rest ||= ::Chef::REST.new("http://localhost:4000")
```
MaximilianMeister: can we call it `@api` like `@api_cookbooks` below?

crowbar_framework/lib/crowbar/chef/upload.rb
MaximilianMeister: IMHO it should stay as it is, because `chef_path` suggest an upstream chef path

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,277 @@
+#!/usr/bin/env ruby
```
tboerger: That should be placed only within bin scripts.

crowbar_framework/lib/crowbar/chef/upload.rb
rsalevsky: Because `relative_path_from` requires a Pathname.

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,284 @@
...
+
+          cookbook_loader.each do |cookbook_name, cookbook|
+            cookbooks_for_upload.push(cookbook)
+            version_constraints_to_update[cookbook_name] = cookbook.version
+          end
+
+          logger.info("Checking existing cookbooks...")
+          cookbooks_for_upload.reject! do |cookbook|
+            next unless api_coobbooks.include?(cookbook.name.to_s)
+            cookbook_exists?(cookbook)
```
MaximilianMeister: thanks for outsourcing this. only `cookbook_exist?` is more conventional, can you change it?

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,284 @@
...
+          return false unless uploader.upload_cookbooks
+
+          true
+        end
+
+        def data_bags
+          loader = ::Chef::Knife::Core::ObjectLoader.new(::Chef::DataBagItem, logger)
+          data_bags = loader.find_all_object_dirs(chef_data_bags_path) || []
+          data_bags.each do |data_bag|
+            data_bag_items = loader.find_all_objects(chef_data_bags_path.join(data_bag))
```
MaximilianMeister: can this return `nil` eventually?

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,284 @@
...
+            data_bag_item_paths.each do |data_bag_item_path|
+              # Workaround for a strange chef behavior
+              relative_path = chef_data_bags_path.relative_path_from(Pathname.new(Dir.pwd))
+              data_bag_item = loader.load_from(relative_path, data_bag, data_bag_item_path)
+              bag = ::Chef::DataBagItem.new
+              bag.data_bag(data_bag)
+              bag.raw_data = data_bag_item
+              logger.info("Uploading data_bag #{data_bag_item_path}")
+
+              return false unless bag.save
```
MaximilianMeister: I think `bag.save` either returns a `DataBagItem` or raises a `Net::HTTPServerException` (at least in the newer chef) so maybe we have to treat this differently here.
http://www.rubydoc.info/github/opscode/chef/Chef/DataBagItem#save-instance_method

can you please double check?

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,284 @@
...
+        end
+
+        def roles
+          loader = ::Chef::Knife::Core::ObjectLoader.new(::Chef::Role, logger)
+          chef_roles_path.each_child do |file|
+            next unless file.file?
+            role = loader.load_from("roles", file)
+            logger.info("Uploading role #{file.basename}")
+
+            return false unless role.save
```
MaximilianMeister: same thing for the `role.save` like mentioned above

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,284 @@
...
+        end
+
+        protected
+
+        def logger
+          return Rails.logger unless caller.grep(/rake/).present?
+          ::Logger.new(STDOUT)
+        end
+
+        def chef_path
```
MaximilianMeister: I liked it more like you had it before, `chef_path` is a bit misleading I think as it is only a directory of our creation. but I don't mind too much.

crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -0,0 +1,284 @@
...
+
+        def check_dependencies(cookbook)
+          missing_dependencies = cookbook.metadata.dependencies.reject do |cookbook_name, version|
+            check_server_side_cookbooks(cookbook_name, version) ||
+              check_uploading_cookbooks(cookbook_name, version)
+          end
+
+          unless missing_dependencies.empty?
+            missing_cookbooks = missing_dependencies.map do |cookbook_name, version|
+              "\"#{cookbook_name}\" version \"#{version}\""
```
MaximilianMeister: why do you have to escape the quotes here?

crowbar_framework/lib/crowbar/chef/upload.rb
rsalevsky: It returns always an array in case there are no data_bags it is just empty.

crowbar_framework/lib/crowbar/chef/upload.rb
rsalevsky: I like this name more...

crowbar_framework/lib/crowbar/chef/upload.rb
rsalevsky: Because I want to print these quotes whiteout escaping the string would be broken.

MaximilianMeister:
we agreed that once this gets integrated into crowbar, some unit tests will follow

+1 codewise as I haven't tested it yet

jdsn:
:+1:

PR #444, opened 2016-03-23
NTP service needs to be restarted (bsc#972299)

dirkmueller:
On SLE12, the systemd support does not handle reload anymore:

$ systemctl reload ntpd
Failed to reload ntpd.service: Job type reload is not applicable for
unit ntpd.service.

### DIFF
--- a/chef/cookbooks/ntp/recipes/default.rb
+++ b/chef/cookbooks/ntp/recipes/default.rb
@@ -90,7 +90,7 @@ else
   service "ntp" do
     service_name "ntpd" if node[:platform_family] == "rhel"
     service_name "ntpd" if node[:platform] == "opensuse" || (node[:platform] == "suse" && node[:platform_version].to_f >= 12.0)
-    supports restart: true, status: true, reload: true
+    supports restart: true, status: true, reload: false
     action [:enable, :start]
   end
   utils_systemd_service_restart "ntp"


toabctl:
+1 (what about filling a bug against ntp from SLE12, too?)

chef/cookbooks/ntp/recipes/default.rb
```diff
@@ -91,7 +91,7 @@
   service "ntp" do
     service_name "ntpd" if node[:platform_family] == "rhel"
     service_name "ntpd" if node[:platform] == "opensuse" || (node[:platform] == "suse" && node[:platform_version].to_f >= 12.0)
-    supports restart: true, status: true, reload: true
+    supports restart: true, status: true, reload: false
```
vuntz: Not that it's wrong, but I fail to understand how this is causing issues as we don't seem to trigger a reload anywhere?

tboerger:
:+1:

bmwiedemann:
is this PR still needed?

toabctl:
@dirkmueller is it still needed?

chef/cookbooks/ntp/recipes/default.rb
vuntz: FWIW, I'd simply remove that `supports` line. Chef will do the right thing without it.

dirkmueller:
@vuntz agreed, might be cosmetical and not actually fixing anything. I've removed the bug number.

chef/cookbooks/ntp/recipes/default.rb
dirkmueller: ```restart``` defaults to false, so this would be an undesired change (running restart is fine, just reload is the problem)

PR #445, opened 2016-03-23
fix too early cleanup of our extracted and upgraded tmpdir

MaximilianMeister:


MaximilianMeister:
once this is merged I will make a backport of all the commits to 3.0

vuntz:
Why the revert?

MaximilianMeister:
> Why the revert?

@vuntz because we actually rely on the same pathname object here, as we know it has been extracted and upgraded there. if the tmpdir is somehow is gone after the upgrade has been performed on the files, and extracted again, then we lose the upgrade modifications.

but thinking about it again, I think the question is if that could happen? currently not when I look at the codepath, but the main point here is that we have to work with one tmpdir all the time, and that revert makes sure that we (at least try to) use only this tmpdir.

MaximilianMeister:
@rsalevsky can you confirm if what I am trying to explain here makes sense?

MaximilianMeister:
@vuntz basically the supposed fix in https://github.com/crowbar/crowbar-core/commit/84d01b7102e878810e2abd3260825be1767c41f1 was just wrong, it should have been https://github.com/crowbar/crowbar-core/pull/445/commits/f0efa5b5807f6590f25b15fc3dc15bc6f98debd5

vuntz:
@MaximilianMeister can you put that info in the commit message of the revert? :-) (too busy to check if you're correct right now, but I would ask this later anyway)

MaximilianMeister:
> can you put that info in the commit message of the revert?

sure :-) done (hope it is clear enough)

rsalevsky:
+1

MaximilianMeister:
triggered gating to do 5 to 6 upgrade

MaximilianMeister:
see https://github.com/SUSE-Cloud/automation/pull/924 for fixing the gating error

tboerger:
:+1:

crowbar_framework/app/models/backup.rb
```diff
@@ -102,7 +102,14 @@ def upgrade?
   def upgrade
     upgrade = Crowbar::Upgrade.new(self)
     if upgrade.supported?
-      upgrade.upgrade
+      begin
+        upgrade.upgrade
+      rescue StandardError => e
```
vuntz: I would probably have put the `rescue` for the whole method, since the initializer for `Crowbar::Upgrade` already extracts the data.

vuntz:
Minor comment, but +1 overall.

crowbar_framework/app/models/backup.rb
MaximilianMeister: right, will change that

vuntz:
+1, thanks for the change

PR #446, opened 2016-03-24
Checks disk claims after upgrade

sayalilunkad:
Checks existing disk claims against disks in ohai data and
returns list of unclaimed disks.

Co-Authored-By: Rick Salevsky <rsalevsky@suse.com>

### DIFF
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -1451,6 +1451,35 @@ class NodeObject < ChefObject
     disk_release(device, owner) and save
   end
 
+  def verify_claimed_disks
+    unreferenced_disks = Array.new
+    crowbar_wall[:claimed_disks].each do |disk, _claim|
+      path, device_name = disk.split("/")[-2..-1]
+
+      # For raid controllers where disk devices are like /dev/cciss
+      # which have the format cciss![filename] in the node attributes.
+      if path == "cciss"
+        unreferenced_disks.push(disk) if node[:block_device]["#{path}!#{device_name}"].nil?
+        next
+      end
+
+      # For disk devices like /dev/diskname
+      unless disk =~ /^\/dev\/disk\//
+        unreferenced_disks.push(disk) if node[:block_device][device_name].nil?
+        next
+      end
+
+      # For disk devices like /dev/disks/by-something/something-else
+      devices = node[:block_device].map do |device, _attr|
+        next if node[:block_device][device]["disks"].nil? ||
+            node[:block_device][device]["disks"][path].nil?
+        node[:block_device][device]["disks"][path]
+      end.compact.flatten
+      unreferenced_disks.push(disk) unless devices.include?(device_name)
+    end
+    unreferenced_disks
+  end
+
   def boot_device(device)
     if device
       Rails.logger.debug "Set boot device to #{device}"


crowbar_framework/app/models/node_object.rb
```diff
@@ -1451,6 +1451,26 @@ def disk_release!(device, owner)
     disk_release(device, owner) and save
   end
 
+  def check_unclaimed_disks
+    unclamed_disks = Array.new
```
vuntz: s/unclamed/unclaimed/

crowbar_framework/app/models/node_object.rb
```diff
@@ -1451,6 +1451,26 @@ def disk_release!(device, owner)
...
+    crowbar_wall[:claimed_disks].each do |disk, _claim|
+      path, device_name = disk.split("/")[-2..-1]
+
+      unless disk =~ /^\/dev\/disk\//
+        unclamed_disks.push(device_name) unless node[:block_device][path].nil?
+        next
+      end
+
+      devices = node[:block_device].map do |device, _claim|
+        next if node[:block_device][device]["disks"].nil? ||
```
vuntz: I've no idea, so asking: does `next` work in `map`? Or should we just do something like `nil if`? (since the `nil` will be dropped in the `compact`)

crowbar_framework/app/models/node_object.rb
```diff
@@ -1451,6 +1451,26 @@ def disk_release!(device, owner)
...
+    unclamed_disks = Array.new
+    crowbar_wall[:claimed_disks].each do |disk, _claim|
+      path, device_name = disk.split("/")[-2..-1]
+
+      unless disk =~ /^\/dev\/disk\//
+        unclamed_disks.push(device_name) unless node[:block_device][path].nil?
+        next
+      end
+
+      devices = node[:block_device].map do |device, _claim|
```
vuntz: Nearly sure that there's no claim-related bits in `node[:block_device]`, so at least the variables here are wrong. But more importantly, wouldn't it make sense to create the list of devices paths outside of the `each` loop (without checking for `path`), and then just do a `include?` test here?

vuntz:
(Btw, I guess this still WIP as this needs to be called from somewhere)

crowbar_framework/app/models/node_object.rb
rsalevsky: At the end it is the same next in map just returns nil.

sayalilunkad:
@vuntz that's right, but the calling will probably be done by the UI so will be another PR.

crowbar_framework/app/models/node_object.rb
sayalilunkad: Yes, will need to rename the _claim variable to something that makes more sense.
I can try to do it that way but IIRC there was some problem with that. But I can check again.

sayalilunkad:
@vuntz  so the basic output format for node[:block_device] is something like this.
```
vdb
{"size"=>"44040192", "removable"=>"0", "vendor"=>"0x1af4", "rotational"=>"1", "disks"=>{"by-id"=>["virtio-c14-node2-ceph1"]}}
```
But many devices don't have a disks or if they have disks it can be "by-id" or "by-path" or something different and if we have to create a list of the paths then we need to consider all these possibilities and some of it completely may not be necessary because we just need to check against the crowbar_wall[:claimed_disks] so I think this way is better.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1451,6 +1451,26 @@ def disk_release!(device, owner)
     disk_release(device, owner) and save
   end
 
+  def check_unclaimed_disks
+    unclaimed_disks = Array.new
+    crowbar_wall[:claimed_disks].each do |disk, _claim|
+      path, device_name = disk.split("/")[-2..-1]
+
+      unless disk =~ /^\/dev\/disk\//
+        unclaimed_disks.push(device_name) unless node[:block_device][path].nil?
```
rhafer: I guess this is supposed to handle disks claimed by name directly in /dev (i.e. /dev/sda, /dev/vdb and so on)?

I am pretty sure something is wrong with that ```unless``` condition. ```path``` will contain ```dev``` in such a case but you'd need to check for ```device_name``` here (which would be ```sda``` or something like that).

And then you only want to add such a disk to ```unclaimed_disks``` if it does not exist in ```node[:block_device]```.
So I think this should be:

```unclaimed_disks.push(device_name) if node[:block_device][device_name].nil?```

File: crowbar_framework/app/models/node_object.rb [hunk 5]
```diff
@@ -1451,6 +1451,26 @@ def disk_release!(device, owner)
     disk_release(device, owner) and save
   end
 
+  def check_unclaimed_disks
```
rhafer: It think the name is a bit missleading. We're not trying to find unclaimed disks. But we're try to check whether the devices in the existing claims to still exist.

How about calling it verify_disk_claims or something like that.

crowbar_framework/app/models/node_object.rb
rhafer: Why not pushing the full disk name (```disk```) into the result array? That makes it easier to lookup the problematic claims for the calling code.

It would just need look into node[:claimed_disk][element_from_the_result] to figure out what claim is problematic.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1451,6 +1451,26 @@ def disk_release!(device, owner)
...
+        unclaimed_disks.push(device_name) unless node[:block_device][path].nil?
+        next
+      end
+
+      devices = node[:block_device].map do |device, _attr|
+        next if node[:block_device][device]["disks"].nil? ||
+            node[:block_device][device]["disks"][path].nil?
+        node[:block_device][device]["disks"][path]
+      end.compact.flatten
+      unclaimed_disks.push(device_name) unless devices.include?(device_name)
```
rhafer: Same here, I'd suggest to push ```disk``` instead of ```device_name```)

crowbar_framework/app/models/node_object.rb
sayalilunkad: Done.

crowbar_framework/app/models/node_object.rb
sayalilunkad: Done.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1451,6 +1451,26 @@ def disk_release!(device, owner)
     disk_release(device, owner) and save
   end
 
+  def verify_disks_claim
```
rhafer: I'm still not sure about this. How about ```verify_claimed_disks```.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1451,6 +1451,26 @@ def disk_release!(device, owner)
...
+    unclaimed_disks = Array.new
```
rhafer: This should also be renamed. This disks returned are not unclaimed. They are claimed but do not refer to any existing device on the node.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1451,6 +1451,26 @@ def disk_release!(device, owner)
     disk_release(device, owner) and save
   end
 
+  def verify_claimed_disks
+    unreferenced_disks = Array.new
+    crowbar_wall[:claimed_disks].each do |disk, _claim|
+      path, device_name = disk.split("/")[-2..-1]
```
rhafer: Sorry, but I just realized that there is another issue with this.

IIRC there are Raid Controllers out there which make disks devices show up as:
/dev/cciss/<somefilename> (instead of /dev/sdX)

The current code will fails with those type of disks. What will make things more complicated is that such a disk will show up as "cciss![somefilename]" inside the ```node[:block_device]``` attributes :(

So in case ```crowbar_wall[:claimed_disks]``` contains just a disk we'd need to replace the ```/``` after cciss with ```!``` to check the ```node[:block_device]``` attributes.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1451,6 +1451,32 @@ def disk_release!(device, owner)
...
+
+      unless disk =~ /^\/dev\/disk\//
+        unreferenced_disks.push(disk) if node[:block_device][device_name].nil?
+        next
+      end
+
+      unless disk.include?("cciss")
```
rhafer: Didn't you mean ```if``` here instead of ```unless```?

How about ```if path == "cciss"```?

crowbar_framework/app/models/node_object.rb
```diff
@@ -1451,6 +1451,32 @@ def disk_release!(device, owner)
...
+        ex = "!"
+        unreferenced_disks.push(disk) if node[:block_device][path<<ex<<device_name].nil?
```
rhafer: I think I'd prefer ```.... node[:block_device]["#{path}!#{device_name}"].nil?```

crowbar_framework/app/models/node_object.rb
```diff
@@ -1451,6 +1451,31 @@ def disk_release!(device, owner)
...
+    unreferenced_disks = Array.new
+    crowbar_wall[:claimed_disks].each do |disk, _claim|
+      path, device_name = disk.split("/")[-2..-1]
+
+      unless disk =~ /^\/dev\/disk\//
+        unreferenced_disks.push(disk) if node[:block_device][device_name].nil?
+        next
+      end
+
+      if path == "cciss"
```
rhafer: I think this block should move above the ```unless disk =~ /^\/dev\/disk\//``` block. Because that matches ```/dev/cciss/``` disks as well.

sayalilunkad: I don't think so because disk =~ /^\/dev\/disk\// strictly checks for only two parts like /dev/cciss but if there is something more to it  which I think will be the case like /dev/cciss/c0c0p1 it will not be checked above. but again there is not harm in moving it up.

rhafer: What I meant to say is that the ```unless disk =~ /^\/dev\/disk\//``` block is going to be executed for disk with ```/dev/cciss/something``` in their name. I will not do any harm AFAICS, but I think is generally cleaner to execute the more specific ```if path == "cciss"``` block first.

crowbar_framework/app/models/node_object.rb
rhafer: Note to other reviewers: The above issue has been addressed meanwhile. Keeping it for further reference.

rhafer:
+1

crowbar_framework/app/models/node_object.rb
```diff
@@ -1451,6 +1451,31 @@ def disk_release!(device, owner)
     disk_release(device, owner) and save
   end
 
+  def verify_claimed_disks
```
MaximilianMeister: the name of the method is not really clear to me in terms of what it returns, why not just call it `unreferenced_disks`?

crowbar_framework/app/models/node_object.rb
MaximilianMeister: could you add the explanation from @rhafer to the code as a comment? would be helpful for the next one who has to look at this code

crowbar_framework/app/models/node_object.rb
sayalilunkad: @rhafer what do you think about @MaximilianMeister suggestion to change the name of the method?

crowbar_framework/app/models/node_object.rb
sayalilunkad: Done. I added a one liner to all the cases for more clarity. Thanks!

crowbar_framework/app/models/node_object.rb
rhafer: @MaximilianMeister the naming is a bit difficult. We're not searching disks that are unreferenced but for claims that reference disk(name)s that are no longer existing.

MaximilianMeister: I think it depends on the caller and how it handles the returned Array of disks, then maybe `verified_claimed_disks`? I am not sure what's best here but if you think it's ok like it is now I trust you there.

MaximilianMeister:
+1 (haven't tested it though after an upgrade, only in a normal cloud)

PR #448, opened 2016-04-01
crowbar: Make apply_role a bit more solid

vuntz:
We move some code that must always be run before returning in the global
ensure; this helps make sure that code is actually run in all cases.

We also catch all exceptions that aren't caught from this method to set
the status of the proposal to "failed", to avoid people being blocked by
a proposal staying forever in "applying".

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1341,6 +1341,11 @@ class ServiceObject
 
     update_proposal_status(inst, "success", "")
     [200, {}]
+  rescue StandardError => e
+    @logger.fatal("apply_role: Uncaught exception #{e.message} #{e.backtrace.join("\n")}")
+    message = "Failed to apply the proposal: uncaught exception (#{e.message})"
+    update_proposal_status(inst, "failed", message)
+    [405, message]
   ensure
     apply_locks.each(&:release) if apply_locks.any?
     restore_to_ready(applying_nodes) if applying_nodes.any?
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -919,6 +919,10 @@ class ServiceObject
   def apply_role(role, inst, in_queue, bootstrap = false)
     @logger.debug "apply_role(#{role.name}, #{inst}, #{in_queue}, #{bootstrap})"
 
+    # Variables used in the global ensure
+    apply_locks = []
+    applying_nodes = []
+
     # Part I: Looking up data & checks
     #
     # we look up the role in the database (if there is one), the new one is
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/service_object.rb
```diff
@@ -1444,11 +1440,16 @@ def apply_role(role, inst, in_queue)
...
 
     update_proposal_status(inst, "success", "")
-    restore_to_ready(applying_nodes)
-    process_queue unless in_queue
     [200, {}]
+  rescue StandardError => e
+    @logger.fatal("apply_role: Uncaught exception #{e.message} #{e.backtrace.join("\n")}")
+    message = "Failed to apply the proposal: uncaught exception (#{e.message})"
+    update_proposal_status(inst, "failed", message)
+    [405, message]
```
MaximilianMeister: just wondering, here we return a message string (like in other places above), but just above the rescue block, on success we return an empty hash. I also couldn't find any consumer of that return message somehow

aspiers:
@vuntz The commit message suggests that this could maybe be split into two commits - is that possible?

crowbar_framework/app/models/service_object.rb
```diff
@@ -953,7 +957,11 @@ def apply_role(role, inst, in_queue, bootstrap = false)
       # Attempt to queue the proposal.  If delay is empty, then run it.
       deps = proposal_dependencies(role)
       delay, pre_cached_nodes = queue_proposal(inst, element_order, new_elements, deps)
-      return [202, "Queued: waiting for #{delay.join(", ")}"] unless delay.empty?
+      unless delay.empty?
+        # force not processing the queue further
+        in_queue = true
```
aspiers: Is this explained by the commit message?

crowbar_framework/app/models/service_object.rb
aspiers: @MaximilianMeister I think the old `crowbar batch` did, at least.

crowbar_framework/app/models/service_object.rb
vuntz: @aspiers that's part of the "We move some code that must always be run before returning in the global ensure", yes :-)

vuntz:
> @vuntz The commit message suggests that this could maybe be split into two commits - is that possible?

@aspiers done, even though I'm not sure it was worth it. But doing it was faster than debating ;-)

aspiers:
@vuntz Thanks - IMHO it was definitely worth it.  +0.9, although shouldn't at least one if not both of these commit messages reference [bsc#840255 (Fix "bricked" proposal after a crash when applying proposal)](https://bugzilla.suse.com/show_bug.cgi?id=840255) due to serving as a partial fix for it?  Of course they don't defend against e.g. the admin server losing power right in the middle of `apply_role`, but that's far less likely than an exception in `apply_role` being caused by a bug in a barclamp.

vuntz:
> although shouldn't at least one if not both of these commit messages reference bsc#840255 (Fix "bricked" proposal after a crash when applying proposal) due to serving as a partial fix for it?

Done. No code change (not even a rebase), just commit message change, and gating was green before. So we can ignore gating if needed.

Still lacking a +1, though.

aplanas:
+1

MaximilianMeister:
looks good

+1

PR #449, opened 2016-04-05
apply_role: log only when needed and at the right place

MaximilianMeister:
* log message is misleading when node list is empty
* like this it is easier to see where (on which node) chef-client failed


jsuchome:
+1

rsalevsky:
+1

rsalevsky:
+1

tboerger:
:+1:

PR #450, opened 2016-04-07
[UNTESTED]Do OS upgrade only for nodes in crowbar_upgrade state (bsc#972527)

snmohan83:
Fix to upgrade OS only on nodes marked for upgrade through "crowbar_upgrade" state. Other nodes (in "ready" state are not altered). 

Fixes bug: https://bugzilla.suse.com/show_bug.cgi?id=972527

Co-authored by @rhafer


mmnelemane:
opening with the correct id

PR #451, opened 2016-04-07
Do OS upgrade only for nodes in crowbar_upgrade state (bsc#972527)

mmnelemane:
Fix to upgrade OS only on nodes marked for upgrade through "crowbar_upgrade" state. Other nodes (in "ready" state are not altered).

Fixes bug: https://bugzilla.suse.com/show_bug.cgi?id=972527

Co-authored by @rhafer

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -385,7 +385,9 @@ class CrowbarService < ServiceObject
   end
 
   def prepare_nodes_for_os_upgrade
-    upgrade_nodes = NodeObject.all.reject { |node| node.admin? || node[:platform] == "windows" }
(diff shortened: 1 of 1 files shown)


jsuchome:
Looks good so far. Does it really solve the bug?

rhafer:
> Does it really solve the bug?

Simple answer: yes. But it requires some manual step on the Cloud 5 side. I'll update the bug with that info.

rhafer:
@jsuchome You're the lucky winner to do the first review ;)

jsuchome:
This looks correct. So I understand that customer will handle Cloud5 part of the install manually, without us needed to adapt the logic, so it would allow to exclude some nodes from upgrade?

jsuchome:
+1

rsalevsky:
+1

PR #453, opened 2016-04-08
add missing json endpoints for upgrade steps

MaximilianMeister:
this avoids redirections when called from mkcloud and opens the
possibility to add an upgrade subcommand to crowbarctl


tboerger:
:+1:

rsalevsky:
+1

PR #457, opened 2016-04-14
Create data_bag before uploading the items

rsalevsky:
this was missing from the last PR

### DIFF
--- a/crowbar_framework/lib/crowbar/chef/upload.rb
+++ b/crowbar_framework/lib/crowbar/chef/upload.rb
@@ -69,6 +69,18 @@ module Crowbar
           loader = ::Chef::Knife::Core::ObjectLoader.new(::Chef::DataBagItem, logger)
           data_bags = loader.find_all_object_dirs(chef_data_bags_path) || []
           data_bags.each do |data_bag|
+            begin
+              logger.info("Creating data_bag #{data_bag}...")
+              api.post_rest("data", name: data_bag)
+            rescue Net::HTTPServerException => e
+              if e.response.code == "409"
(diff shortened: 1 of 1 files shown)


crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -69,6 +69,15 @@ def data_bags
           loader = ::Chef::Knife::Core::ObjectLoader.new(::Chef::DataBagItem, logger)
           data_bags = loader.find_all_object_dirs(chef_data_bags_path) || []
           data_bags.each do |data_bag|
+            begin
+              api.post_rest("data", name: data_bag)
+            rescue Net::HTTPServerException => e
+              unless e.response.code == 409
```
MaximilianMeister: not sure but shouldn't it be an `if` instead of an `unless`? or why do you skip when you get 409?

rsalevsky: 409 is the response code if the data_bag already exists

MaximilianMeister:
> 409 is the response code if the data_bag already exists

okay, then maybe it wouldn't hurt to have at least a log message in `production.log`, so sth like

```ruby
            begin
              logger.info("Creating data_bag #{data_bag}...")
              api.post_rest("data", name: data_bag)
            rescue Net::HTTPServerException => e
              if e.response.code == "409"
                logger.info("data_bag #{data_bag} already exists: #{e.message}")
              else
                logger.error("Creating data_bag #{data_bag} failed (#{e.response.code})")
                return false
              end
            end
```

MaximilianMeister:
+1

jdsn:
:+1:

PR #462, opened 2016-04-20
Treat NICs with unknown speed as 1g interfaces (bsc#975473)

rhafer:
If ohai can't figure out the supported speed(s) of a NIC (e.g. it currently
lacks support for 40G interfaces), treat it as "normal" 1g interfaces.
This is a workaround to make such interface at least somehow usable.

partially fixes: https://bugzilla.suse.com/show_bug.cgi?id=975473


vuntz:
+1

rsalevsky:
+1

PR #464, opened 2016-04-20
use rails logger instead of stdout

MaximilianMeister:


tboerger:
Just overwrite the logger getter. Currently you should be able to override the logger within tests and so on. You are overriding it hard with the rails logger. Beside that you anyway missed the attr_accessor :logger.

tboerger:
:+1:

rsalevsky:
:+1:

PR #389, opened 2016-03-02
provisioner: Fix alternatives when upgrading to SLES 12 SP1

vuntz:
We can have broken alternatives when upgrading for python-related
packages: the reason is that the alternatives are registered for the
2.6 scripts but never removed since the packages are not uninstalled
but simply updated (and on update, we register the 2.7 scripts for the
alternatives).

It's technically a bug in the python packages, but I doubt it'll get
fixed.

For instance, after upgrade, we have this alternative:
 testr -> /usr/bin/testr-2.6
And what happens:
 - python-os-testr was installed on SLES 11 SP3 and created this
   alternative;
 - on the upgrade, the scriptlet that would remove the alternative is
   not run because the package is not removed (just updated);
 - the updated package created a new alternative (/usr/bin/testr-2.7)
   but doesn't remove the old outdated alternative
 - the link group for testr is in automatic mode, but nothing triggers a
   check whether it's still valid, so testr still points to testr-2.6.

The manual of update-alternatives says that we can run
  yes '' | update-alternatives --force --all
to fix all broken alternatives. So let's do that!

PR #470, opened 2016-04-26
enhancing/fixing #389

MaximilianMeister:
#389 didn't work as expected. I opened this PR to trigger a gating, as @vuntz is likely not available

this hopefully fixes the upgrade CI and would be the better solution for https://github.com/SUSE-Cloud/automation/pull/944

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
@@ -115,8 +115,10 @@
         <source>
           <![CDATA[
           # python-related alternatives are not updated, even though the binary
-          # names changed (2.6 to 2.7); fix that
-          yes '' | update-alternatives --force --all
+          # names changed (2.6 to 2.7); find all broken links and fix/update them
+          for i in $(find /etc/alternatives -type l ! -readable -printf "%f\n") ; do
+              update-alternatives --auto "$i"
+          done
           ]]>
         </source>
       </script>
--- a/chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
@@ -108,6 +108,18 @@
         </source>
       </script>
 <% end -%>
+      <script>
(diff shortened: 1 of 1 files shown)


aplanas:
So we discarded to increase the priority in alternatives inside the package?

aplanas:
+1

MaximilianMeister:
just added quotes around the `"$i"`

vuntz:
@MaximilianMeister I think we just need a fixed update-alternatives for my patch, see https://bugzilla.suse.com/show_bug.cgi?id=969171

aplanas:
+1

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
```diff
@@ -108,6 +108,20 @@
...
 <% end -%>
+      <script>
+        <chrooted config:type="boolean">true</chrooted>
+        <debug config:type="boolean">true</debug>
+        <filename>fix_alternatives</filename>
+        <source>
+          <![CDATA[
+          # python-related alternatives are not updated, even though the binary
+          # names changed (2.6 to 2.7); find all broken links and fix/update them
+          for i in $(find /etc/alternatives/ -type l -! -exec test -e {} \; -print | xargs basename -a) ; do
```
nicolasbock: It's generally safer to use `-print0`:

~~~
find /etc/alternatives/ -type l -! -exec test -e {} \; -print0 | xargs --null basename -a
~~~

nicolasbock: I can't find documentation on `-!`. What does it do? Is it the same as `!`?

nicolasbock: You can also write this without using an external command to test:
~~~
find /tmp/alternatives -type l ! -readable -printf "%f\n"
~~~

vuntz:
Would be nice to update the `find` command as suggested, +1 after this.

MaximilianMeister:
> Would be nice to update the find command as suggested, +1 after this.

updated.

@nicolasbock thanks for the hint on `find`

vuntz:
+1

PR #455, opened 2016-04-14
only skip verify_authenticity_token on json requests

MaximilianMeister:
http://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection.html

PR #474, opened 2016-05-04
Revert "only skip verify_authenticity_token on json requests"

MaximilianMeister:
Reverts crowbar/crowbar-core#455

I have encountered some issues during a manual upgrade. I can't upload the backup without running into `ActionController::InvalidAuthenticityToken (ActionController::InvalidAuthenticityToken)`

This needs to be investigated further, as the authenticity token is not stored in the session, but probably in a cookie.

for now it is better to flip the switch back, and create a card for further manual testing debugging

### DIFF
--- a/crowbar_framework/app/controllers/application_controller.rb
+++ b/crowbar_framework/app/controllers/application_controller.rb
@@ -76,7 +76,8 @@ class ApplicationController < ActionController::Base
 
   protect_from_forgery with: :exception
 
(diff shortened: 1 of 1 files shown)


MaximilianMeister:
I have created a card for investigating this:

https://trello.com/c/8RoOL3RS/1054-investigate-and-manually-test-code-for-csrf-protection

tboerger:
:+1:

jgrassler:
:+1:

PR #475, opened 2016-05-04
delete old backup when uploading the same one again

MaximilianMeister:
the backup data always needs to be wiped, e.g. in case of an upgrade
when the first restore failed and the user tries to upload the backup
again there will be an error that the backup already exists, which is
bad


jsuchome:
Is there a bug report when it was fond out as needed?

MaximilianMeister:
> Is there a bug report when it was fond out as needed?

I am not aware of any bug, but Rick sent me a mail during the summit, where he encountered this issue during an upgrade

jsuchome:
+1

tboerger:
:+1:

PR #476, opened 2016-05-12
wait for the nodes to be really available

MaximilianMeister:
after the upgrade of the nodes has finished, the nodes are still
in boot process.
A quick upgrader who switches to the barclamps view right after
the "Go to your Barclamps now" button is available and applies
keystone will get a flash message saying:
  "Role keystone-server can't be used for suse"

I am not sure if that really helps though, maybe even another check is necessary here to be really sure we can hit the apply button. this was just my initial idea to query the node status. any ideas?

### DIFF
--- a/crowbar_framework/app/controllers/installer/upgrades_controller.rb
+++ b/crowbar_framework/app/controllers/installer/upgrades_controller.rb
@@ -267,9 +267,10 @@ module Installer
       respond_to do |format|
         format.json do
           render json: {
-            total: view_context.total_nodes_count,
+            total: view_context.all_non_admin_nodes.count,
             left: view_context.upgrading_nodes_count,
             failed: view_context.failed_nodes_count,
+            ready: view_context.nodes_ready?,
             error: I18n.t(
               "installer.upgrades.nodes_status.failed",
               nodes: NodeObject.find("state:problem").map(&:name).join(", ")
--- a/crowbar_framework/app/helpers/nodes_helper.rb
+++ b/crowbar_framework/app/helpers/nodes_helper.rb
@@ -16,8 +16,8 @@
 #
 
 module NodesHelper
-  def total_nodes_count
-    NodeObject.find("NOT admin_node:true").count
+  def all_non_admin_nodes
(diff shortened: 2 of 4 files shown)


crowbar_framework/app/helpers/nodes_helper.rb
```diff
@@ -16,8 +16,12 @@
 #
 
 module NodesHelper
+  def total_nodes
```
vuntz: That's a weird name. Maybe rename to `all_non_admin_nodes`

vmoravec:
> I am not sure if that really helps though, 

Was having a similar issue with upgrade job on qa hardware with the upgraded nodes when trying to reapply the barclamps (see https://ci.suse.de/view/Cloud/view/QA/job/cloud-mkphyscloud-qa-scenario-0a/93/console).

That problem was fixed by adding a new step to wait for nodes to be in state `ready` which solved that (see the PR here https://github.com/SUSE-Cloud/automation/pull/984/files#diff-05c4438b5e3357d366863c6c4f1b1663R192)

crowbar_framework/app/helpers/nodes_helper.rb
MaximilianMeister: somehow yes, I have just used split the old name as it was without thinking too much about it. l'll change it

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -263,9 +263,10 @@ def nodes_status
       respond_to do |format|
         format.json do
           render json: {
-            total: view_context.total_nodes_count,
+            total: view_context.all_non_admin_nodes_count,
```
rsalevsky: I would just use all_non_admin_nodes.count here then we can safe some code wich is otherwise not really used.

MaximilianMeister: true that, I just looked at the `xyz_count` methods below and wanted to be consistent. but you're right it is not really needed

rsalevsky:
+1

jsuchome:
+1

PR #477, opened 2016-05-13
some batch model fixes

MaximilianMeister:


tboerger:
And how is that an improvement to replace working symbols with strings?

MaximilianMeister:
> And how is that an improvement to replace working symbols with strings?

it's not an improvement, it's a fix. the code just didn't work

```
barclamps = barclamps.select do |_, x|
  x[:user_managed]
end
```

where `x` is the value and its keys are not symbolized. instead of writing 

```
barclamps = barclamps.select do |_, x|
  x.symbolize_keys[:user_managed]
end
```

I just used strings, because the conversion is just cosmetic and to me makes no real sense

tboerger:
You dropped the recursive symbolize keys, and it makes sense because it's recommended to work with symbols instead of strings

MaximilianMeister:
> You dropped the recursive symbolize keys

As I said it didn't work for me, probably because it is not recursive. I just found out if you want to have recursive symbol keys you need to use `deep_symbolize_keys`. will change it

tboerger:
https://github.com/crowbar/crowbar-core/blob/44d0178cc2bb44b6c9c52c4facca6a439c53bcc1/crowbar_framework/lib/utils/extended_hash.rb#L171

But whatever, do what you want, I'm out of this :)

MaximilianMeister:
> But whatever, do what you want, I'm out of this :)

Thanks for the link, I didn't know this. But before running away do you have an idea why https://github.com/crowbar/crowbar-core/pull/478 doesn't work out of the box? I am always getting an empty batch file without the change in this PR here.

rsalevsky:
+1

crowbar_framework/app/models/batch/export.rb
```diff
@@ -46,7 +46,7 @@ def filename
 
     def persist!
       path.open("w") do |file|
-        file.write content.stringify_keys(true).to_yaml
+        file.write content.stringify_keys.to_yaml
```
nicolasbock: @MaximilianMeister: Just to help me understand this, this will stringify only top level keys. Why drop the `(true)` argument?

nicolasbock:
@tboerger Sorry for the possibly obvious question (I am still getting up to speed with ruby), why overload `stringify_keys`? Isn't that already defined in rails?

crowbar_framework/app/models/batch/export.rb
rsalevsky: See http://apidock.com/rails/v4.2.1/Hash/stringify_keys for this but in short the true was just wrong due to stringify_keys doesn't take arguments.

tboerger:
> @tboerger Sorry for the possibly obvious question (I am still getting up to speed with ruby), why overload stringify_keys? Isn't that already defined in rails?

While we have done this there was no recursive version available in rails.

crowbar_framework/app/models/batch/export.rb
MaximilianMeister: the true is/was only necessary for the overloaded `stringify_keys`.

nicolasbock: And now we are using the standard rails function?

MaximilianMeister: no, it uses the overloaded version. but the overloaded version which requires (true) somehow didn't work in this scenario. first I have dropped the overloading, but then I saw it it still required by our `crowbar_` bin scripts. but we will hopefully drop them soon, as soon as the `batch build` works in crowbarctl. I guess I will work on that in the near future, and then will continue on this PR. so lets not merge it for now.

aspiers:
It looks to me like the batch operation was completely rearchitected, but this is the first I heard about it.  Does it run server-side now?  Please can someone explain to me how it works at a high level?  I can't understand what this PR is about at all, which is a bit worrying considering I originally wrote the operation.

MaximilianMeister:
> It looks to me like the batch operation was completely rearchitected, but this is the first I heard about it. Does it run server-side now? Please can someone explain to me how it works at a high level? I can't understand what this PR is about at all, which is a bit worrying considering I originally wrote the operation.

@aspiers The original logic from the bin script has been "moved" by @tboerger into a model, and made accessible by adding an API for batch. so yes it runs server side now. but it still misses the `build` part. This PR was trying to fix the `export` part, which I was testing to make https://github.com/crowbar/crowbar-core/issues/478 happen. for details you have to ask @tboerger, I don't know much about the bin script, and what has been changed to make it fit into the rails framework

aspiers:
This is a good example of where [the blame feature](https://github.com/facebook/mention-bot#how-does-it-work) of the [suse-mention-bot](https://github.com/suse-mention-bot?tab=activity) can help us, i.e. when developers forget to request reviews from relevant stakeholders.

tboerger:
> This is a good example of where the blame feature of the suse-mention-bot can help us, i.e. when developers forget to request reviews from relevant stakeholders.

Not really, because your CLI code have not been touched. Just crowbar-client started to use the API.

aspiers:
@tboerger Rewriting my code in a fresh codebase does not remove me from the list of stakeholders.

tboerger:
> @tboerger Rewriting my code in a fresh codebase does not remove me from the list of stakeholders.

But this removes the option to get detected by any bot mechanism. Beside that it's not our or even my fault if you are not following pull requests.

aplanas:
+1

PR #478, opened 2016-05-13
save a batch file in the backup for convenience

MaximilianMeister:
this could be used during an upgrade if a user chooses to apply all
barclamps at once

depends on https://github.com/crowbar/crowbar-core/pull/477

### DIFF
--- a/crowbar_framework/app/models/batch/export.rb
+++ b/crowbar_framework/app/models/batch/export.rb
@@ -17,6 +17,8 @@
 
 module Batch
   class Export < Base
+    attr_accessor :path
+
     def path
       @path ||= Rails.root.join(
         "storage",
--- a/crowbar_framework/lib/crowbar/backup/export.rb
+++ b/crowbar_framework/lib/crowbar/backup/export.rb
@@ -34,6 +34,7 @@ module Crowbar
         databags
         db
         crowbar
+        batch
         meta
       end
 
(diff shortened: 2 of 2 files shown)


crowbar_framework/app/models/batch/export.rb
```diff
@@ -17,6 +17,8 @@
 
 module Batch
   class Export < Base
+    attr_accessor :path
```
rsalevsky: Why is this needed?

MaximilianMeister: otherwise I wouldn't be able to set `path` which means I cannot set a custom path for the yml output (the backup path in our case)

rsalevsky: Ah missed this.

rsalevsky:
+1

crowbar_framework/app/models/batch/export.rb
MaximilianMeister: but now that you mention it, shall I also make filename accessible? it is not used yet though...

rsalevsky: If you want to use a custom file name then yes, otherwise it makes not really sense if we don't use it.

crowbar_framework/lib/crowbar/backup/export.rb
```diff
@@ -128,6 +129,15 @@ def crowbar
         end
       end
 
+      def batch
+        logger.debug "Backing up batch export"
```
toabctl: can you add the path to the backup file to the log?

MaximilianMeister: doesn't make sense here as this is just a substep of the backup creation. and logging the path is already done at the beginning of the backup creation in https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/backup.rb#L160

toabctl:
+1

nicolasbock:
+1

MaximilianMeister:
https://github.com/crowbar/crowbar-core/issues/477 is merged now, so we can merge this too

PR #479, opened 2016-05-17
drop some methods in favor of upstream

MaximilianMeister:
not fully sure why we still need to override stringify and symbolize_keys
with a recursive extension, as upstream has deep_stringify_keys and
deep_symbolize_keys which are doing the same thing

also in some tests the old code didn't work as expected


MaximilianMeister:
I see the old CLI still depends on the library and its functions, so maybe this is safe to drop once we drop the old CLI altogether. so lets close this for now

PR #481, opened 2016-05-18
don't log the raw proposal

MaximilianMeister:
it could contain sensitive data which you don't want in the logs

### DIFF
--- a/crowbar_framework/config/initializers/filter_sql_log.rb
+++ b/crowbar_framework/config/initializers/filter_sql_log.rb
@@ -1,6 +1,5 @@
 #
-# Copyright 2011-2013, Dell
-# Copyright 2013-2014, SUSE LINUX Products GmbH
+# Copyright 2016, SUSE Linux GmbH
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -15,12 +14,19 @@
 # limitations under the License.
 #
 
-Rails.application.config.tap do |config|
-  config.filter_parameters += [
-    :password,
-    :password_confirmation,
-    :properties,
-    :proposal_attributes,
-    :secret
-  ]
+# filter sensitive data from our proposal table
+module ActiveRecord
+  class LogSubscriber
+    alias :old_sql :sql
+
+    def sql(event)
+      if event.payload[:sql] =~ /^(INSERT INTO|UPDATE) "proposals".*/
+        Rails.logger.debug(
+          "SQL: INSERT/UPDATE transaction: filtering query to not expose potentially sensitive data"
+        )
+      else
+        old_sql(event)
+      end
+    end
+  end
 end


rsalevsky:
+1

vuntz:
One question first: does it mean it removes the full logging related to SQL for proposals, or just the INSERT/UPDATE line?

That being said, I don't believe this fully fixes the issue:

```
I, [2016-02-23T15:13:48.096465 #19977:0x000000092182a8]  INFO -- Started POST "/crowbar/keystone/1.0/proposals/default" for 127.0.0.1 at 2016-02-23 15:13:48 +0000
I, [2016-02-23T15:13:48.104526 #19977:0x000000092182a8]  INFO -- Processing by KeystoneController#proposal_update as JSON
I, [2016-02-23T15:13:48.105076 #19977:0x000000092182a8]  INFO --   Parameters: {"id"=>"default", "description"=>"Centralized authentication and authorization service for OpenStack", "attributes"=>{"keystone"=>{"debug"=>false, "frontend"=>"native", "verbose"=>true, "use_syslog"=>false, "policy_file"=>"policy.json", "database_instance"=>"default", "rabbitmq_instance"=>"default", "domain_specific_drivers"=>false, "domain_config_dir"=>"/etc/keystone/domains", "token_expiration"=>14400, "db"=>{"database"=>"keystone", "user"=>"keystone", "password"=>"[FILTERED]"}, "sql"=>{"idle_timeout"=>30}, "ssl"=>{"certfile"=>"/etc/keystone/ssl/certs/signing_cert.pem", "keyfile"=>"/etc/keystone/ssl/private/signing_key.pem", "generate_certs"=>false, "insecure"=>false, "cert_required"=>false, "ca_certs"=>"/etc/keystone/ssl/certs/ca.pem"}, "api"=>{"protocol"=>"http", "service_port"=>5000, "api_host"=>"0.0.0.0", "admin_port"=>35357, "admin_host"=>"0.0.0.0", "version"=>"3", "region"=>"CustomRegion"}, "admin"=>{"tenant"=>"admin", "username"=>"admin", "password"=>"[FILTERED]"}, "service"=>{"tenant"=>"service", "token"=>"1m8ZV6Gcfcl1"}, "default"=>{"tenant"=>"openstack", "username"=>"crowbar", "password"=>"[FILTERED]"}, "identity"=>{"driver"=>"sql"}, "assignment"=>{"driver"=>"sql"}, "signing"=>{"token_format"=>"UUID", "certfile"=>"/etc/keystone/ssl/certs/signing_cert.pem", "keyfile"=>"/etc/keystone/ssl/private/signing_key.pem", "ca_certs"=>"/etc/keystone/ssl/certs/ca.pem"}, "ldap"=>{"url"=>"ldap://localhost", "user"=>"dc=Manager,dc=example,dc=com", "password"=>"[FILTERED]", "suffix"=>"cn=example,cn=com", "use_dumb_member"=>false, "allow_subtree_delete"=>false, "dumb_member"=>"cn=dumb,dc=nonexistent", "page_size"=>0, "alias_dereferencing"=>"default", "query_scope"=>"one", "user_tree_dn"=>"", "user_filter"=>"", "user_objectclass"=>"inetOrgPerson", "user_id_attribute"=>"cn", "user_name_attribute"=>"sn", "user_mail_attribute"=>"mail", "user_pass_attribute"=>"userPassword", "user_enabled_attribute"=>"enabled", "user_enabled_mask"=>0, "user_enabled_default"=>"True", "user_attribute_ignore"=>"default_project_id", "user_default_project_id_attribute"=>"", "user_allow_create"=>true, "user_allow_update"=>true, "user_allow_delete"=>true, "user_enabled_invert"=>false, "user_enabled_emulation"=>false, "user_enabled_emulation_dn"=>"", "project_tree_dn"=>"", "project_filter"=>"", "project_objectclass"=>"groupOfNames", "project_domain_id_attribute"=>"businessCategory", "project_id_attribute"=>"cn", "project_member_attribute"=>"member", "project_name_attribute"=>"ou", "project_desc_attribute"=>"description", "project_enabled_attribute"=>"enabled", "project_attribute_ignore"=>"", "project_allow_create"=>true, "project_allow_update"=>true, "project_allow_delete"=>true, "project_enabled_emulation"=>false, "project_enabled_emulation_dn"=>"", "role_tree_dn"=>"", "role_filter"=>"", "role_objectclass"=>"organizationalRole", "role_id_attribute"=>"cn", "role_name_attribute"=>"ou", "role_member_attribute"=>"roleOccupant", "role_attribute_ignore"=>"", "role_allow_create"=>true, "role_allow_update"=>true, "role_allow_delete"=>true, "group_tree_dn"=>"", "group_filter"=>"", "group_objectclass"=>"groupOfNames", "group_id_attribute"=>"cn", "group_name_attribute"=>"ou", "group_member_attribute"=>"member", "group_attribute_ignore"=>"", "group_allow_create"=>true, "group_allow_update"=>true, "group_allow_delete"=>true, "tls_cacertfile"=>"", "tls_cacertdir"=>"", "use_tls"=>false, "tls_req_cert"=>"demand", "use_pool"=>false}}}, "deployment"=>{"keystone"=>{"crowbar-revision"=>0, "crowbar-applied"=>false, "schema-revision"=>35, "element_states"=>{"keystone-server"=>["readying", "ready", "applying"]}, "elements"=>{"keystone-server"=>["cluster:services"]}, "element_order"=>[["keystone-server"]], "config"=>{"environment"=>"keystone-config-default", "mode"=>"full", "transitions"=>false, "transition_list"=>[]}}}}
```

Sure, I see some `FILTERED` in there, but the magic that does the filtering does filter everything correctly for all proposals (just stuff that includes "password", it seems, while some attributes are using "secret" instead).

MaximilianMeister:
> One question first: does it mean it removes the full logging related to SQL for proposals, or just the INSERT/UPDATE line?

if you only want to filter `INSERT/UPDATE` you would have to write sth like:
```
event.payload[:sql].match(/(.*(INSERT|UPDATE).*proposals.*)/)
```
do we only want to filter INSERT and UPDATE statements?


> Sure, I see some FILTERED in there, but the magic that does the filtering does filter everything correctly for all proposals (just stuff that includes "password", it seems, while some attributes are using "secret" instead).

not fully understand the issue, shall I add another key `:secret` to the filter list?

vuntz:
>> One question first: does it mean it removes the full logging related to SQL for proposals, or just the INSERT/UPDATE line?
> 
> if you only want to filter `INSERT/UPDATE` you would have to write sth like:
> 
> `event.payload[:sql].match(/(.*(INSERT|UPDATE).*proposals.*)/)`
> 
> do we only want to filter INSERT and UPDATE statements?

I don't have a strong feeling. Maybe it's valuable to somehow still see that there's some SQL work happening in the log, though?

> >    Sure, I see some FILTERED in there, but the magic that does the filtering does filter everything correctly for all proposals (just stuff that includes "password", it seems, while some attributes are using "secret" instead).
> 
> not fully understand the issue, shall I add another key :secret to the filter list?

My point is that I don't think we can create an exhaustive list of all attributes that should be filtered and keep that up-to-date over years, so I'd simply prefer to not have the json of the proposal appearing there. But if it's too hacky, then we should check out what needs to be filtered in addition, based on the current schemas that we have.

crowbar_framework/config/initializers/filter.rb
```diff
@@ -18,6 +18,9 @@
 Rails.application.config.tap do |config|
   config.filter_parameters += [
     :password,
-    :password_confirmation
+    :password_confirmation,
+    :properties,
+    :proposal_attributes,
+    :secret
```
vuntz: Unfortunately:

> Parameters: {"utf8"=>"✓", "authenticity_token"=>"Ltw91V6&#x2F;7c8i43rctf+mhfcnZdlLY3UT04Dx3jvD2byPXK5nsNcPzymJNU4js2C1OlqOwZwRNxI3pjfllXhNlw==", "barclamp"=>"nova", "name"=>"default", "submit"=>"Save", "proposal_attributes"=>"[FILTERED]", "scheduler_ram_allocation_ratio"=>"1.0", "scheduler_cpu_allocation_ratio"=>"16.0", "use_migration"=>"true", "setup_shared_instance_storage"=>"false", "use_shared_instance_storage"=>"false", "kvm_ksm_enabled"=>"true", "vcenter_host"=>"", "vcenter_user"=>"", "vcenter_password"=>"[FILTERED]", "vcenter_clusters"=>"", "vcenter_datastore"=>"", "vcenter_interface"=>"vmnic0", "vcenter_ca_file"=>"", "vcenter_insecure"=>"false", "zvm_zvm_xcat_server"=>"", "zvm_zvm_xcat_username"=>"", "zvm_zvm_xcat_password"=>"[FILTERED]", "zvm_zvm_diskpool"=>"", "zvm_zvm_diskpool_type"=>"", "zvm_zvm_host"=>"", "zvm_zvm_user_profile"=>"", "zvm_zvm_scsi_pool"=>"", "zvm_zvm_xcat_master"=>"", "zvm_zvm_xcat_ssh_key"=>"", "ssl_enabled"=>"false", "ssl_generate_certs"=>"false", "ssl_certfile"=>"/etc/nova/ssl/certs/signing_cert.pem", "ssl_keyfile"=>"/etc/nova/ssl/private/signing_key.pem", "ssl_insecure"=>"false", "ssl_cert_required"=>"false", "vnc_keymap"=>"en-us", "novnc_ssl_enabled"=>"false", "novnc_ssl_certfile"=>"", "novnc_ssl_keyfile"=>"", "verbose"=>"true", "proposal_deployment"=>"{\"crowbar-revision\":0,\"crowbar-applied\":false,\"schema-revision\":41,\"element_states\":{\"nova-controller\":[\"readying\",\"ready\",\"applying\"],\"nova-compute-docker\":[\"readying\",\"ready\",\"applying\"],\"nova-compute-hyperv\":[\"readying\",\"ready\",\"applying\"],\"nova-compute-kvm\":[\"readying\",\"ready\",\"applying\"],\"nova-compute-qemu\":[\"readying\",\"ready\",\"applying\"],\"nova-compute-vmware\":[\"readying\",\"ready\",\"applying\"],\"nova-compute-xen\":[\"readying\",\"ready\",\"applying\"],\"nova-compute-zvm\":[\"readying\",\"ready\",\"applying\"],\"nova-ha-compute\":[\"readying\",\"ready\",\"applying\"]},\"elements\":{\"nova-controller\":[\"cluster:services\"],\"nova-compute-hyperv\":[],\"nova-compute-kvm\":[\"remotes:services\"],\"nova-compute-qemu\":[],\"nova-compute-xen\":[]},\"element_order\":[[\"nova-controller\"],[\"nova-compute-docker\",\"nova-compute-hyperv\",\"nova-compute-kvm\",\"nova-compute-qemu\",\"nova-compute-vmware\",\"nova-compute-xen\",\"nova-compute-zvm\"]],\"element_run_list_order\":{\"nova-controller\":95,\"nova-compute-docker\":97,\"nova-compute-hyperv\":97,\"nova-compute-kvm\":97,\"nova-compute-qemu\":97,\"nova-compute-vmware\":97,\"nova-compute-xen\":97,\"nova-compute-zvm\":97,\"nova-ha-compute\":97},\"config\":{\"environment\":\"nova-config-default\",\"mode\":\"full\",\"transitions\":false,\"transition_list\":[]},\"crowbar-committing\":false,\"crowbar-status\":\"success\",\"crowbar-failed\":\"\",\"crowbar-queued\":false}", "id"=>"default"}

That's because all entries are part of the form in the barclamp, I guess. So we can filter out the proposal, but the attributes are still there. We can't blindly remove all the parameter logs for all methods, so I see two options: either we remove them just for the actions that save a proposal, or we really build a blacklist of terms based on what the current schemas have (I'm not sure that just password and secret are enough, but I don't know for sure).

rsalevsky:
Just got another idea, what is if we change the type in the schema's from `str` to `password` and filter on this base? I could imagine this could also be useful later...

MaximilianMeister:
>> do we only want to filter INSERT and UPDATE statements?

> I don't have a strong feeling. Maybe it's valuable to somehow still see that there's some SQL work happening in the log, though?

I could add a line in the log saying something like:
`SQL: INSERT/UPDATE transaction: skipping sensitive proposal data`

MaximilianMeister:
@vuntz when you have some time, please have another look at the latest change

crowbar_framework/config/initializers/filter_sql_log.rb
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+#
+
+# filter sensitive data from our proposal table
+module ActiveRecord
+  class LogSubscriber
+    alias :old_sql :sql
+
+    def sql(event)
+      if event.payload[:sql] =~ /(.*(INSERT|UPDATE).*proposals.*)/
```
vuntz: Can we make the regexp a bit more restrictive? Right now, it can match many more things than `INSERT INTO "proposals"` or `UPDATE "proposals"`.

crowbar_framework/config/initializers/filter_sql_log.rb
```diff
@@ -0,0 +1,30 @@
...
+        Rails.logger.debug("SQL: INSERT/UPDATE transaction: skipping sensitive proposal data")
```
vuntz: "SQL: INSERT/UPDATE transaction: filtering query to not expose potentially sensitive data"

vuntz:
Some minor comments; I think we should go ahead and merge this afterwards as this is already an improvement compared to status quo.

MaximilianMeister:
> Some minor comments

@vuntz addressed the comments, and tested it once again. thanks

vuntz:
Thanks, +1

rsalevsky:
+1

PR #483, opened 2016-05-20
Cloud7 to use SES3.0 repos

abel-navarro:


jsuchome:
Looks correct. +1

aplanas:
+1

crowbar_framework/config/repos-cloud.yml
```diff
@@ -73,25 +73,25 @@ suse-12.1:
       url:
       smt_path: "SUSE/Updates/SLE-HA/12-SP1/x86_64/update"
       ask_on_error: false
-    suse-enterprise-storage-2_1-pool:
-      name: "SUSE-Enterprise-Storage-2.1-Pool"
+    suse-enterprise-storage-3-pool:
+      name: "SUSE-Enterprise-Storage-3.0-Pool"
```
vuntz: I guess you should do s/3.0/3/ (here and in other places)

abel-navarro:
Rebuilding. The errors were unrelated to this change.

crowbar_framework/config/repos-cloud.yml
abel-navarro: I cannot find the example now, but for SES 2.0 the strings contained "2.0", that's why I used here "3.0". I think also allows for increased versions easily.

jsuchome: But the paths and labels are inconsistent, you have both 3 and 3.0. (No idea what is correct, though)

vuntz: It should really be 3; SES team moved from 1.0, to 2.0, then 2.1 and now 3. There's no 3.0 :-)

abel-navarro: addressed that in the last commit.

jsuchome:
+1 after latest changes (I think you can squash the commits)

abel-navarro:
squashed commits.

vuntz:
+1, but shouldn't be merged until the other PRs are ready to be merged.

PR #484, opened 2016-05-20
Drop dependencies for SP1 and initial support for SP2

aplanas:


jsuchome:
Looks correct, not sure if anything else is not missing, but I don't have idea about anything.

+1

chef/data_bags/crowbar/template-provisioner.json
```diff
@@ -172,7 +189,7 @@
     "provisioner": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 35,
+      "schema-revision": 36,
```
rsalevsky: We should skip some revisions here to have some free space for Cloud 6.

aplanas:
* Jump migration to 50

jsuchome:
+1 after change of migration number

toabctl:
+1

rsalevsky:
+1

aplanas:
gating fails because manila (unrelated)

PR #487, opened 2016-05-25
Enable SLES12 SP2

rsalevsky:
for:
- support_software_raid
- support_default_fs
- pretty_target_platform


vuntz:
Isn't that already part of what was done (and merged) in https://github.com/crowbar/crowbar-core/pull/484 ?

jsuchome:
Yes, this is already done

PR #488, opened 2016-05-26
Remove SP1 based repositories that are not needed for Cloud7

jsuchome:


crowbar_framework/config/repos-cloud.yml
```diff
@@ -21,115 +21,6 @@ suse-12.1:
...
-      name: "SLE12-SP1-HA-Updates"
-      required: "optional"
-      features: ["ha"]
-      repomd:
-        tag: "obsrepository://build.suse.de/SUSE:Updates:SLE-HA:12-SP1:x86_64/update"
-        key_md5: "9a62177e1c6852d48453ed3909956d6b"
-      url:
-      smt_path: "SUSE/Updates/SLE-HA/12-SP1/x86_64/update"
-      ask_on_error: false
-    suse-enterprise-storage-2_1-pool:
```
vuntz: You want to keep SES ;-)

jsuchome: But it's gonna by 3, right? So I am removing 2.1 here, and let relevant PR care about adding new SES

jsuchome: And relevant PR is https://github.com/crowbar/crowbar-core/pull/483 (we'll need to rebase one of these obviously)

vuntz:
Needs a rebase, but it's not urgent.

jsuchome:
Hm, now I'm not sure, but maybe we'll still need those SP1 for Storage...

vuntz:
> maybe we'll still need those SP1 for Storage...

This should not be the case, but I guess this should be double-checked: we won't support setting up a HA cluster on SP1 anymore (after the move to SP2). There might be some packages from the Cloud repo that are not in the Storage repo when allocating nodes, though (supportconfig, for instance).

jsuchome:
> There might be some packages from the Cloud repo that are not in the Storage repo when allocating nodes, though (supportconfig, for instance).

True, now I remember! It's required in the autoyast profile...

abel-navarro:
What about removing completely the 12.1 [section](https://github.com/crowbar/crowbar-core/pull/488/files#diff-0e5ade134dcc291db8527e0ff2fe4188R2)?

jsuchome:
@abel-navarro not possible, we need SP1 for Storage

abel-navarro:
+1

jsuchome:
Rebased, and restarted gating with want_sles12sp2=1

crowbar_framework/config/repos-cloud.yml
```diff
@@ -25,59 +25,6 @@ suse-12.1:
       url:
       smt_path: "SUSE/Updates/SLE-SERVER/12-SP1/aarch64/update"
       ask_on_error: false
-    cloud:
```
rsalevsky: Why do we remove the cloud repo?

rsalevsky:
+1

rhafer:
+1

PR #489, opened 2016-05-31
Add "role recipes" with the goal of simplifying the roles

jsuchome:
We want the roles to only reference one recipe, so that we can put some
intelligence in that recipe:

 - in a first step, we will stop changing the run list of a node
   depending on its state, and instead have the role recipe "decide"
   whether something should be done or not, depending on the state. This
   will solve the issue that the search results for nodes are not
   reliable when a node is in some state that is not "readying",
   "applying" or "ready".

   Very concretely: many recipes are using searches to find attributes
   of another barclamp. The recipes do not expect the search to return
   an empty result (and that's okay, because we know that there must be
   a glance server when we deploy nova, due to barclamp dependencies),
   but that could happen in the past when the node was in "problem" or
   rebooting.

 - in a later step, we might want to have a new attribute that will tell
   us which role to apply instead of applying everything.

More generally speaking, it's actually wrong to pretend that a node
stops having a role because it's, say, rebooting. A glance server is
always a glance server.

(cherry picked from commit 268b5a668a4dfaf64a3c18ca920aca5c5447942b)

Fix hound issues

(cherry picked from commit 481ce1e0332808aab5797473be85b6e390229b3c)


jsuchome:
@vuntz so, these are just 2 of your original commits - is it simple enough or do you actually still consider this as too much at this stage? Should we remove checking `states_for_role` and just include the original recipes always?

jsuchome:
Hm, reading the AC again, yes, we want to include the recipes without any condition, OK.

vuntz:
+1

aplanas:
+1

PR #490, opened 2016-06-01
Test if admin node has dns_server role for upgrade

nkrinner:
The upgrade process will fail if the admin node does not have the
dns_server role applied. Abort upgrade with a warning, if this is
not the case. See also
https://bugzilla.suse.com/show_bug.cgi?id=978060

(cherry picked from commit 3e9955ff99ed4bc2be38b89c16d28ce3145402ad)

### DIFF
--- a/crowbar_framework/config/locales/crowbar/en.yml
+++ b/crowbar_framework/config/locales/crowbar/en.yml
@@ -584,6 +584,10 @@ en:
         nodes_not_ready: |
           Some nodes are not ready: %{nodes}.
           Fix the state of nodes before proceeding with the upgrade.
+        admin_missing_dns_server: |
+          The Administration Server does not have the dns-server role while this is mandatory for
+          the upgrade process. Reapply the DNS proposal with this change before proceeding with
+          the upgrade.
       start:
         cancel: 'Abort Upgrade'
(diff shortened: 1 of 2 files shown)


nkrinner:
I consider this being ready after https://github.com/crowbar/barclamp-crowbar/pull/1395 got merged.

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -489,6 +489,16 @@ en:
         raid10: 'RAID 10'
       raid_disks_selected: 'Failed to save %{node}: Not enough disks selected for this RAID level'
 
+  upgrade:
+    upgrade:
+      nodes_not_ready: |
```
jsuchome: If you want the text to be here, remove the one from `installer` part, and adapt the usage (https://github.com/crowbar/crowbar-core/pull/490/files#diff-006ea8943f6241893e1af682a8f42073L329)

nkrinner:
@jsuchome updated

jsuchome:
+1

jsuchome:
Please also add Cloud6 version

nicolasbock:
+1

PR #486, opened 2016-05-25
crowbar: Fix upgrade from 3.0

vuntz:
The proposals in 3.0 were already in the database (and not chef
anymore), but still need to be migrated with the scema migrations.

PR #491, opened 2016-06-01
crowbar: Fix upgrade from 3.0

nicolasbock:
Small fix for https://github.com/crowbar/crowbar-core/pull/486.

### DIFF
--- a/crowbar_framework/lib/schema_migration.rb
+++ b/crowbar_framework/lib/schema_migration.rb
@@ -25,7 +25,11 @@ module SchemaMigration
   end
 
   def self.run_for_bc bc_name
-    template = Proposal.new(barclamp: bc_name)
+    begin
+      template = Proposal.new(barclamp: bc_name)
+    rescue TemplateMissing
+      return
+    end
 
     return if template.nil?
     return if template["deployment"].nil?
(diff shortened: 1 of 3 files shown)


nicolasbock:
@vuntz The cloudupgrade migrates the barclamps now. However, the upgrade still fails for some reason. The error message is `Schema migration failed: Template for openstack does not exist`. I don't understand why `openstack` is tried here at all. I am investigating...

nicolasbock:
For some reason, `BarclampCatalog.barclamps` contains `openstack`, which is what leads to the failure.

crowbar_framework/lib/schema_migration.rb
```diff
@@ -25,7 +25,11 @@ def self.run
   end
 
   def self.run_for_bc bc_name
-    template = Proposal.new(barclamp: bc_name)
+    begin
+      template = Proposal.new(barclamp: bc_name)
+    rescue StandardError => e
```
nicolasbock: I'll fix it...

crowbar_framework/lib/schema_migration.rb
```diff
@@ -25,7 +25,11 @@ def self.run
   end
 
   def self.run_for_bc bc_name
-    template = Proposal.new(barclamp: bc_name)
+    begin
+      template = Proposal.new(barclamp: bc_name)
+    rescue StandardError
```
vuntz: Don't catch `StandardError`, but `TemplateMissing` :-)

nicolasbock: Ah, and I thought I was being smart to avoid `Exception` :smile:

vuntz:
+1

aplanas:
+1

nicolasbock:
I ran through a full upgrade on mkcloud1. It's working.

PR #492, opened 2016-06-02
Make Cloud7 SP1 repos optional

jsuchome:
This is needed so we can still install SP1 nodes when cloud
was installed with primary SP2 support.

We need this temporary, when we allow SP2 and SP1 Cloud nodes.
Later, SP1 Cloud repos could be deleted completely, because
we will need SP1 nodes only for Storage. For that, we'll have https://github.com/crowbar/crowbar-core/pull/488


bmwiedemann:
+1 from looking at the diff.

vuntz:
+1

PR #493, opened 2016-06-02
Test if admin node has dns_server role for upgrade

nkrinner:
The upgrade process will fail if the admin node does not have the
dns_server role applied. Abort upgrade with a warning, if this is
not the case. See also
https://bugzilla.suse.com/show_bug.cgi?id=978060

(cherry picked from commit 3e9955ff99ed4bc2be38b89c16d28ce3145402ad)
(cherry picked from commit 9481e5081c181a591adca689abf5610a7b827a86)


nkrinner:
See also https://github.com/crowbar/crowbar-core/pull/490
https://github.com/crowbar/barclamp-crowbar/pull/1395

nicolasbock:
+1

jsuchome:
+1

nkrinner:
Removed a commit reference from the commit message.

PR #495, opened 2016-06-05
crowbar: Stop uploading proposal templates as chef data bags

vuntz:
We should not need them: the templates are only ever used when creating
proposals, and the rails app has direct access to the json file.

This should go with https://github.com/crowbar/crowbar/pull/2230

### DIFF
--- a/crowbar_framework/app/models/network_service.rb
+++ b/crowbar_framework/app/models/network_service.rb
@@ -256,8 +256,19 @@ class NetworkService < ServiceObject
       db = Chef::DataBag.load("crowbar/#{k}_network") rescue nil
       if db.nil?
         @logger.debug("Network: creating #{k} in the network")
+
+        # ensure that crowbar data bag exists
+        databag_name = "crowbar"
+        begin
+          Chef::DataBag.load(databag_name)
+        rescue Net::HTTPServerException
+          crowbar_bag = Chef::DataBag.new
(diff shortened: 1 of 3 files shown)


nicolasbock:
Looks good, but waiting for gating tests.

crowbar_framework/app/models/network_service.rb
```diff
@@ -256,8 +256,19 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
       db = Chef::DataBag.load("crowbar/#{k}_network") rescue nil
       if db.nil?
         @logger.debug("Network: creating #{k} in the network")
+
+        # ensure that crowbar data bag exists
+        databag_name = "crowbar"
+        begin
+          Chef::DataBag.load(databag_name)
+        rescue Net::HTTPServerException
+          crowbar_bar = Chef::DataBag.new
```
MaximilianMeister: i guess you wanted to call it `crowbar_bag` ? looks like a typo

nicolasbock: It's a double crowbar :smile:

nicolasbock:
+1

crowbar_framework/app/models/network_service.rb
vuntz: @MaximilianMeister fixed

rsalevsky:
+1

PR #498, opened 2016-06-10
Remove closure-compiler

rsalevsky:
Since we switched to SLE12 we can use NodeJS. This prepares also the
removal of the old sprockets-standalone rubygem.


MaximilianMeister:
looks good, lets wait for the gating and maybe have a quick look at the UI if the assets are displayed correctly

MaximilianMeister:
seems to work
+1

nicolasbock:
+1

PR #499, opened 2016-06-10
Remove schema for glance, nova and swift

toabctl:
Theses schemas are in the crowbar-openstack projects.


jsuchome:
+1

rsalevsky:
I think we can also remove schema.tgz and template-hadoop.schema.

dirkmueller:
+1

PR #502, opened 2016-06-19
workaround bootloader installed to wrong disk (bsc#985422)

bmwiedemann:
https://bugzilla.suse.com/show_bug.cgi?id=985422

There seems to have been a behaviour change of fdisk
in that it does not write out a newly created partition table
if there are no other changes
so we switch to parted to create the partition table unique ID instead

### DIFF
--- a/updates/control.sh
+++ b/updates/control.sh
@@ -201,7 +201,7 @@ nuke_everything() {
         # write new unique MBR signature
         # This initializes a random 32bit Disk signature used to
         # distinguish disks, which helps installing boot loader properly
-        echo w | fdisk /dev/$name
+        parted --script /dev/$name mktable msdos
     done < <(tac /proc/partitions)
 }
 


bmwiedemann:
quick test shows that IDs are written to sda, but bootloader is still not placed there.

vuntz:
Isn't there another tool we could use?

(+1 if no)

updates/control.sh
```diff
@@ -201,7 +201,8 @@ nuke_everything() {
         # write new unique MBR signature
         # This initializes a random 32bit Disk signature used to
         # distinguish disks, which helps installing boot loader properly
-        echo w | fdisk /dev/$name
+        # create+delete a partition for workarounding bsc#985422
```
aspiers: "working around" not "workarounding" ;-)

aspiers:
@vuntz parted maybe?

aspiers:
+0.5

bmwiedemann:
updated it to use parted instead.

rsalevsky:
https://github.com/crowbar/crowbar-core/pull/1106 got merged some time back

PR #503, opened 2016-06-21
Honor MTU setting for os_sdn network

nicolasbock:


rhafer:
In general this looks good to me. Did you already have a chance testing it?

nicolasbock:
I have it deployed in my v5 cloud. Since the default networks used a `path_mtu = 1400` I am trying to create a new network and attach a nova instance to it. So in other words, not really. I have verified though that the deployed cookbook runs and updates the configuration files correctly.

rhafer:
> I have verified though that the deployed cookbook runs and updates the configuration files correctly

That means the interface(s) used for os_sdn get the right mtu set, if configuring non standard mtu for it  in network.json?

chef/cookbooks/network/recipes/default.rb
```diff
@@ -301,7 +301,7 @@ def kill_nic(nic)
     net_ifs << our_iface.name
   end
   if network["mtu"]
-    if name == "admin" or name == "storage"
+    if %w(admin, storage, os_sdn).include? name
```
rsalevsky: Can you please use the normal array syntax? `["admin,", "storage,", "os_sdn"]`

nicolasbock: Is a regular array syntax preferred? Coming from Python it looks more 'normal' to me too :) But we are using the `%w(` syntax over 30 times in this repository and over 300 times in crowbar-openstack, which is why I figured Rubyists prefer it that way.

rsalevsky: Currently it's mixed but I created #506 to make the situation clear.

nicolasbock: Thanks! I think a consistent syntax is definitely helpful here.

vuntz:
LGTM

rhafer:
@nicolasbock Is there a reason for this still being marked "wip"?

nicolasbock:
Please see my for https://github.com/crowbar/crowbar-openstack/pull/411: The MTU value needs to be set manually in the `os_sdn_network` item in the `crowbar` data bag, but will then correctly be used in the neutron barclamp.

rhafer:
+1

vuntz:
Stupid question, but don't we need a migration to add the mtu attribute to existing deployments?

chef/data_bags/crowbar/migrate/network/021_add_mtu_os_sdn.rb
```diff
@@ -0,0 +1,15 @@
+def upgrade(ta, td, a, d)
+  unless a["networks"]["os_sdn"].key? "mtu"
+    a["networks"]["os_sdn"]["mtu"] = ta["networks"]["os_sdn"]["mtu"]
+  end
+
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  unless a["networks"]["os_sdn"].key? "mtu"
```
vuntz: `unless ta` and not `unless a`

nicolasbock: Isn't `ta` coming from the template which now includes the `mtu` key?

vuntz: When you downgrade, ta comes from the old template. Which shouldn't include the key (except if for some reason, it's still there).

TBH, the downgrade bits are probably not very well working ;-) But the intent here is definitely to check the template.

nicolasbock: Sorry, I replied to fast. I was thinking of the upgrade, not the downgrade. I see your point.

rhafer:
Unassigning myself here. As I will be on FTO for the next weeks. Volunteering @vuntz for further review :)

nicolasbock:
I set the MTU of the `os_sdn` network to 2000. The fixed network was using the `gre` driver and had automatically set its MTU to 1958). I was able to verify that from the VM I can `git clone https://github.com`. Associating a floating IP (the floating network's MTU was `unknown`), I was also able to verify that I could `git clone ssh://<floating_ip>` from external. Is there anything else you think I should test to make sure that this PR is sufficient?

vuntz:
@nicolasbock is it possible to test with various MTU values for the floating network: something like 1500 and 2500?

nicolasbock:
@vuntz To clarify, keep `fixed` at MTU 2000, but set `floating` to 1500 and 2500?

nicolasbock:
Another test:

1. Apply this PR and https://github.com/crowbar/crowbar-openstack/pull/411
1. Set `os_sdn_network:mtu = 2000` in network data bag.
1. Deploy `neutron` barclamp
1. Launch two VMs with `ml2` driver `gre` (42 bytes encapsulation overhead). The VMs `dhcp` configures the internal NIC to MTU 1958.
    - `ping -c 1 -M do -s 2000 192.168.123.4` (MTU between VMs): 1958
    - `ping -c 1 -M do -s 2000 github.com` (MTU to github.com): 1958
1. Set the VM's internal NIC to MTU 1500 (`ip link set dev eth0 mtu 1500`)
    - MTU between VMs: 1958
    - MTU to github.com: 1500

Note, this does not involve the `floating` network. However, from my reading of [bsc#984128](https://bugzilla.suse.com/show_bug.cgi?id=984128) this scenario matches the customer's. If the customer sets `os_sdn_network:mtu >= 1550` before deploying neutron, the VMs will support an internal MTU of 1500.

aplanas:
+1 for me

nicolasbock:
Rebased.

vuntz:
+1. As mentioned by @rhafer, we can actually drop the migration as this is optional, but I don't mind keeping it.

toabctl:
FYI: PR for master ist https://github.com/crowbar/crowbar-core/pull/548

vuntz:
See comments in https://github.com/crowbar/crowbar-core/pull/548 ; if possible, it'd be better to use `cherry-pick -x` in this PR to reference the right commits from master (even if you did a forward port in this case ;-))

chef/data_bags/crowbar/template-network.json
```diff
@@ -257,7 +258,7 @@
     "network": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 20,
+      "schema-revision": 100,
```
vuntz: I guess I should have been careful about asking for cherry-pick: we want 21 here. Apologies, my bad.

nicolasbock: No problem, I updated the patches.

vuntz:
+1

nicolasbock:
@vuntz I have cherry-picked your commit into the PR too.

vuntz:
+1, but gating is red (might just need a re-run?)

nicolasbock:
@vuntz Gating is green now (finally :smile:).

vuntz:
@nicolasbock so we just miss the second +1 :-)

PR #504, opened 2016-06-21
Fix EFI bootloader installation

dirkmueller:


crowbar_framework/config/repos-cloud.yml
```diff
@@ -6,18 +6,20 @@ suse-12.1:
...
+        tag: ["obsproduct://build.suse.de/SUSE:SLE-12-SP1:GA/SLES/12.1/POOL/aarch64",
+              "obsproduct://build.suse.de/Devel:ARM:SLE-12-SP1:Update/SLES/12.1/POOL/aarch64"]
+        key_md5: ["9a62177e1c6852d48453ed3909956d6b", "c4a27581cec772c6dc641825d41c63d2"]
       url:
       smt_path: "SUSE/Products/SLE-SERVER/12-SP1/aarch64/product"
       ask_on_error: false
     sles12-sp1-updates:
       name: "SLES12-SP1-Updates"
-      required: "mandatory"
+#       required: "mandatory"
```
vuntz: Maybe just fully comment out the updates for now? This might be simpler.

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
```diff
@@ -120,6 +120,14 @@
...
           <partition>
             <create config:type="boolean">true</create>
+            <format config:type="boolean">true</format>
+           <filesystem config:type="symbol">vfat</filesystem>
+           <partition_id config:type="integer">259</partition_id>
+            <mount>/boot/efi</mount>
+            <size>128M</size>
+          </partition>
+          <partition>
+            <create config:type="boolean">true</create>
```
vuntz: Shouldn't we check if the node is going to boot with EFI and only create the partition in this case?

dirkmueller: It doesn't harm to create the partition unless you don't want to waste the disk space (I think we can reduce the 128M to much smaller values, 16 or 32MB should also work). 

I agree with you that the partition is not needed in non-EFI cases but I'm not sure how to detect that case. checking for arch==aarch64 is not enough since this is not a aarch64 specific issue. you have the same issue on x86 on systems that only boot via UEFI. (There are couple of bug reports about lack of UEFI support).

vuntz: What about a check like `if node["uefi"] && File.exist?("/sys/firmware/efi")`?

aplanas: ah true, we used this trick here: https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/crowbar-hacks/recipes/default.rb#L81

crowbar_framework/config/repos-cloud.yml
```diff
@@ -6,18 +6,20 @@ suse-12.1:
...
       features: ["os"]
       repomd:
-        tag: "obsrepository://build.suse.de/SUSE:Updates:SLE-SERVER:12-SP1:aarch64/update"
-        key_md5: "9a62177e1c6852d48453ed3909956d6b"
+        tag: ["obsproduct://build.suse.de/SUSE:SLE-12-SP1:GA/SLES/12.1/POOL/aarch64",
```
vuntz: That line should probably be the old tag.

crowbar_framework/config/repos-cloud.yml
dirkmueller: I believe there was something checking for the update repo to be available..

bmwiedemann:
+1 on the 2nd commit
+0.5 on the 1st commit (dont know that part of the code so well)

vuntz:
Let's go ahead.

PR #505, opened 2016-06-22
disable cloud 6 repositories after upgrade to cloud 7

MaximilianMeister:


aplanas:
For my personal interest, can you explain now in what moment of the upgrade process is disabled? and where?

Also `disable_by` is a name that indicate to me "by who". Maybe we can find a better name, like `disable_previous`, and avoid to use the parameter

Can this issue be present from 5->6? If so I will recommend not to hard code `version: 6`, I can see that this will be difficult to find when 7->8, because the method is not going to fail if there is not 6 repo there

Note: this will fix the Jenkins job in C7

MaximilianMeister:
> For my personal interest, can you explain now in what moment of the upgrade process is disabled? and where?

@aplanas right now first step during preparation, but thinking about this, it is better to make it an independent step...

> Also disable_by is a name that indicate to me "by who". Maybe we can find a better name, like disable_previous, and avoid to use the parameter

the thought behind it was this is from the ActiveRecord world, e.g. you have a `find_by` method which requires an option hash. and this method can easily be extended, see e.g. the `where` which takes other options as well

> Can this issue be present from 5->6? If so I will recommend not to hard code version: 6, I can see that this will be difficult to find when 7->8, because the method is not going to fail if there is not 6 repo there

From 5 to 6 we do not need this, and I just realized that I have to submit this to `stable/3.0` ... will reopen another PR

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -31,6 +31,8 @@ def prepare
         service_object = CrowbarService.new(Rails.logger)
 
         service_object.prepare_nodes_for_crowbar_upgrade
+        ret, message = Crowbar::Repository.disable_by(version: 6)
+        raise message unless ret == 200
```
vuntz: But that won't work for disabling SP1 repos later on (when we will need that).

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -105,6 +105,26 @@ def where(options = {})
         result
       end
 
+      def disable_by(options = {})
+        version = options.fetch(:version, nil)
+        return if version.nil?
+
+        all_platforms.each do |platform|
+          arches(platform).each do |arch|
+            repositories(platform, arch).each do |repo|
```
vuntz: I'm not sure it will work here: the issue is that Cloud 6 repos are not even returned in the list here; they're just in the data bag, but not in the yaml file.

crowbar_framework/app/controllers/installer/upgrades_controller.rb
MaximilianMeister: I could extend the options hash to fetch service packs (now or later on when we need it)

crowbar_framework/lib/crowbar/repository.rb
MaximilianMeister: right, thanks for the hint

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -300,6 +301,18 @@ def save_and_restore
       end
     end
 
+    def post_cleanup_repos
+      [:pool, :updates].each do |repo|
+        ret, message = ProvisionerService.new(Rails.logger).disable_repository(
+          "suse-12.1",
```
rsalevsky: I would not hardcode this, I think the better way would be to deactivate everything what is not suse-12.2 or cloud 7 in this case.

vuntz: But for the time being, we still need SP1 for Storage 3.

vuntz:
It's a hack, but I can't think of a good way to do this at the moment, so +1.

MaximilianMeister:
> It's a hack, but I can't think of a good way to do this at the momen

there is still a problem, because `ProvisionerService.disable_repository` uses `Crowbar::Repository.where()` which cannot find the databag as it just looks into the yml file... I guess we need to use `Chef::DataBag.chef_server_rest.delete_rest("data/#{name}")` here.

vuntz:
s/helpfult/helpful/ in a commit message

But +1

MaximilianMeister:
> s/helpfult/helpful/ in a commit message

fixed

vuntz:
Still +1

rsalevsky:
+1

MaximilianMeister:
rebased

vuntz:
https://ci.suse.de/job/openstack-mkcloud/26074/consoleFull shows that the PR fixes the current issue. Next issue is when re-deploying keystone after the upgrade, so unrelated.

PR #506, opened 2016-06-22
Force brackets style for SymbolArrays

rsalevsky:


aspiers:
I'm not necessarily against this change, but :-1: until the commit message explains exactly what the change is, and provides some justification for it.  I don't think it makes sense to unilaterally impose a decision about coding style on the whole team without any explanation.

nicolasbock:
The main argument to me is that this syntax needlessly deviates from what many other languages do to express a list. I say `needless` here since it doesn't really save you all that much typing and I don't think the meaning of it is any clearer compared to a traditional `[]` syntax. `%w()` is syntactic sugar for a string split on whitespace [[1]](https://simpleror.wordpress.com/2009/03/15/q-q-w-w-x-r-s/#sw) which might not be obvious for someone not well versed in Rubyese. And once you start to escape whitespaces inside values, the syntactic appeal goes completely out the window in my opinion :smile:.

nicolasbock:
+1

rsalevsky:
This rule is not really new, we had it for a long time but then it got removed by accident and never reintroduced. Hopefully the commit message is now clear.

nicolasbock:
+1

aspiers:
I was about to give a +1, but now the updated commit is also modifying `RegexpLiteral` without any explanation of what that means or why it was done.  So still :-1: sorry - the commit message has to explain and justify the exact changes being made!  If there is a different justification for changing `RegexpLiteral` then it probably belongs in a separate commit.

rsalevsky:
@aspiers Updated the commit message again. Just for clarification I introduce nothing new here it was for nearly a year our standard until rubocop changed there syntax.

aspiers:
Thanks. I am fine with avoiding `%w()` (even though with my Perl background of using `qw()` it's quite natural to me personally, I understand it can be weird for other people). But I am not at all convinced by the need to avoid `%r()`.  Sometimes it can make regexps far more readable by avoiding [leaning toothpick syndrome](https://en.wikipedia.org/wiki/Leaning_toothpick_syndrome), e.g.

    %r(^/etc/)

is much easier to read IMHO than:

    /^\/etc\//

IIUC, this PR prohibits the former style, so I am still :-1:, sorry :-(  However if you split the `RegexpLiteral` change into a separate PR then I am very happy to :+1: the other two changes.

(BTW this is a good example of why it's worth keeping independent changes separate, because it avoids controversial changes blocking the merge of uncontroversial changes.)

rsalevsky:
@aspiers See https://github.com/SUSE/style-guides/pull/15#issuecomment-228017402. ;)

MaximilianMeister:
+1

rsalevsky:
Let's merge it, all the comments should be adapted or answered.

PR #508, opened 2016-06-23
Fix suse-manager barclamp to actually work with SP1 clients

jsuchome:
Fix suse-manager barclamp to actually work with SP1 clients and prepare it for SP2

### DIFF
--- a/chef/cookbooks/suse-manager-client/recipes/default.rb
+++ b/chef/cookbooks/suse-manager-client/recipes/default.rb
@@ -19,7 +19,8 @@
 
 return if node[:crowbar_wall][:suse_manager_client_registered] || false
 
-manager_server = node[:suse_manager_client][:manager_server]
+bootstrap_script_url = node[:suse_manager_client][:bootstrap_script_url]
+
 
 temp_pkg = Mixlib::ShellOut.new("mktemp /tmp/ssl-cert-XXXX.rpm").run_command.stdout.strip
 
@@ -33,9 +34,8 @@ execute "update-ca-certificates" do
   command "update-ca-certificates"
 end
 
-bootstap_script = "bootstrap-sles12.#{node[:platform_version].split(".").last}"
 execute "bootstrap SUMA client" do
-  command "curl https://#{manager_server}/pub/bootstrap/#{bootstap_script}.sh | sh"
+  command "curl #{bootstrap_script_url} | sh"
 end
 
 node.set[:crowbar_wall][:suse_manager_client_registered] = true
--- a/chef/cookbooks/suse-manager-client/recipes/default.rb
+++ b/chef/cookbooks/suse-manager-client/recipes/default.rb
@@ -29,12 +29,8 @@ end
 
 package(temp_pkg)
 
-org_cert = "/usr/share/rhn/RHN-ORG-TRUSTED-SSL-CERT"
-bash "install SSL certificate" do
-  code <<-EOH
-  cp #{org_cert} \
-     /etc/ssl/certs/`openssl x509 -noout -hash -in #{org_cert}`.0
-  EOH
+execute "update-ca-certificates" do
+  command "update-ca-certificates"
 end
 
 bootstap_script = "bootstrap-sles12.#{node[:platform_version].split(".").last}"
--- a/chef/cookbooks/suse-manager-client/recipes/default.rb
+++ b/chef/cookbooks/suse-manager-client/recipes/default.rb
@@ -17,6 +17,8 @@
 # limitations under the License.
 #
 
+return if node[:crowbar_wall][:suse_manager_client_registered] || false
+
 manager_server = node[:suse_manager_client][:manager_server]
 activation_key = node[:suse_manager_client][:activation_key]
 
@@ -36,11 +38,10 @@ bash "install SSL certificate" do
   EOH
 end
 
-# XXX requires chef-client with CHEF-4090 fixed otherwise the package
-# provider can't handle the URL
-package "https://#{manager_server}/pub/bootstrap/sm-client-tools.rpm"
-
-execute "sm-client" do
-  command "sm-client --hostname #{manager_server} --activation-keys #{activation_key}"
+bootstap_script = "bootstrap-sles12.#{node[:platform_version].split(".").last}"
+execute "bootstrap SUMA client" do
+  command "curl https://#{manager_server}/pub/bootstrap/#{bootstap_script}.sh | sh"
 end
 
+node.set[:crowbar_wall][:suse_manager_client_registered] = true
+node.save
--- a/chef/cookbooks/suse-manager-client/recipes/default.rb
+++ b/chef/cookbooks/suse-manager-client/recipes/default.rb
@@ -20,7 +20,6 @@
 return if node[:crowbar_wall][:suse_manager_client_registered] || false
 
 manager_server = node[:suse_manager_client][:manager_server]
-activation_key = node[:suse_manager_client][:activation_key]
 
 temp_pkg = Mixlib::ShellOut.new("mktemp /tmp/ssl-cert-XXXX.rpm").run_command.stdout.strip
 
--- a/crowbar_framework/app/helpers/barclamp/suse_manager_client_helper.rb
+++ b/crowbar_framework/app/helpers/barclamp/suse_manager_client_helper.rb
@@ -25,7 +25,7 @@ module Barclamp
     end
 
     def suse_manager_client_install
-      "/opt/dell/bin/barclamp_install.rb --rpm suse-manager-client"
+      "knife cookbook upload suse-manager-client -o /opt/dell/chef/cookbooks"
     end
   end
 end
--- a/crowbar_framework/config/locales/suse_manager_client/en.yml
+++ b/crowbar_framework/config/locales/suse_manager_client/en.yml
@@ -20,7 +20,6 @@ en:
       edit_attributes:
         manager_server: 'SUSE Manager server hostname'
         manager_server_hint: 'For example susemgr.example.com. You can set a record within the %{link}.'
-        activation_key: 'Activation Key'
         activation_key_hint: 'For example 1-00112233445566778899aabbccddeeff'
 
         manual_steps: |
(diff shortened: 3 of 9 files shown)


chef/cookbooks/suse-manager-client/recipes/default.rb
```diff
@@ -36,11 +38,9 @@
...
 
-# XXX requires chef-client with CHEF-4090 fixed otherwise the package
-# provider can't handle the URL
-package "https://#{manager_server}/pub/bootstrap/sm-client-tools.rpm"
-
-execute "sm-client" do
-  command "sm-client --hostname #{manager_server} --activation-keys #{activation_key}"
+# FIXME we should do it just the first time, otherwise system is registered again
+bootstap_script = "bootstrap-sles12."
+bootstap_script += node[:platform_version].split(".")[1]
```
MaximilianMeister: I would interpolate here as the size is still OK for one line. and use `last` instead of `[1]`

`"bootstrap-sles12.#{node[:platform_version].split(".").last}"`

crowbar_framework/app/helpers/barclamp/suse_manager_client_helper.rb
```diff
@@ -25,7 +25,7 @@ def suse_manager_client_rpm
     end
 
     def suse_manager_client_install
-      "/opt/dell/bin/barclamp_install.rb --rpm suse-manager-client"
+      "knife cookbook upload core -o /opt/dell/chef/cookbooks"
```
MaximilianMeister: would be good to use a crowbar lib like https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/chef/upload.rb, but it seems it cannot upload single cookbooks

chef/cookbooks/suse-manager-client/recipes/default.rb
```diff
@@ -36,11 +38,10 @@
...
 
-# XXX requires chef-client with CHEF-4090 fixed otherwise the package
-# provider can't handle the URL
-package "https://#{manager_server}/pub/bootstrap/sm-client-tools.rpm"
-
-execute "sm-client" do
-  command "sm-client --hostname #{manager_server} --activation-keys #{activation_key}"
+bootstap_script = "bootstrap-sles12.#{node[:platform_version].split(".").last}"
+execute "bootstrap SUMA client" do
+  command "curl https://#{manager_server}/pub/bootstrap/#{bootstap_script}.sh | sh"
```
MaximilianMeister: have you tried the script yet? my question is only if it needs higher permission or anything. difficult to say when you can't see what it does

crowbar_framework/app/helpers/barclamp/suse_manager_client_helper.rb
rsalevsky: The cookbook loader needs to be adapted (https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/chef/upload.rb#L111-L113).

chef/cookbooks/suse-manager-client/recipes/default.rb
jsuchome: I've tried, yes.

crowbar_framework/app/helpers/barclamp/suse_manager_client_helper.rb
jsuchome: BTW, this is never actually executed, it's just a hint to user how to do it.

rsalevsky: Then we should just hint to the rake task `bin/rake chef:upload:cookbooks`.

jsuchome: But we do not need to upload all of them, we only need one file. See how the whole text looks like: http://crowbar.vi5.cloud.suse.de/crowbar/suse_manager_client/1.0/proposals/default

rsalevsky: The lib is designed for such use cases, so it will figure out which cookbook is missing an will just upload this one.

jsuchome: Nice, but still I do not see reall value in proposing it over simple knife cookbook upload. You need to run rake from specific directory for example.

rsalevsky: Ok, let's keep it as it is when we have a button or an API for the lib then we can adapt it later.

chef/cookbooks/suse-manager-client/recipes/default.rb
```diff
@@ -17,8 +17,10 @@
 # limitations under the License.
 #
 
+return unless node[:platform_family] == "suse"
```
vuntz: Doesn't SUMA support RHEL nodes?

crowbar_framework/app/helpers/barclamp/suse_manager_client_helper.rb
vuntz: I doubt the cookbook is named `core` ;-)

chef/cookbooks/suse-manager-client/recipes/default.rb
```diff
@@ -36,11 +38,10 @@
   EOH
```
vuntz: Just wondering: is this still correct? Shouldn't we call `update-ca-certificates` instead?

crowbar_framework/app/helpers/barclamp/suse_manager_client_helper.rb
jsuchome: ouch

chef/cookbooks/suse-manager-client/recipes/default.rb
jsuchome: Good point. How do we test installing RHEL nodes?

chef/cookbooks/suse-manager-client/recipes/default.rb
jsuchome: Well, it worked for me...

chef/cookbooks/suse-manager-client/recipes/default.rb
vuntz: We don't, right now. Maybe just raise an exception for the RHEL case so that people who try it will fix it instead of assuming it works fine.

chef/cookbooks/suse-manager-client/recipes/default.rb
vuntz: Well, it initially worked, but I wonder if it still works after update-ca-certificates is called; it sounds to me like the code was working for SLE11 fine, but is not safe for SLE12.

chef/cookbooks/suse-manager-client/recipes/default.rb
jsuchome: Or we can actually ask user for a path to bootstrap script in crowbar...

jsuchome:
The script can have any name, I'll need to ask user for its exact location.

chef/cookbooks/suse-manager-client/recipes/default.rb
jsuchome: So you think just installing `ssl-cert.rpm` and calling `update-ca-certificates`? Or is some other step needed?

vuntz: Ideally yes. But I don't know if the rpm package is built the correct way for this.

crowbar_framework/config/locales/suse_manager_client/en.yml
```diff
@@ -18,10 +18,12 @@ en:
...
     suse_manager_client:
       edit_attributes:
-        manager_server: 'SUSE Manager server hostname'
-        manager_server_hint: 'For example susemgr.example.com. You can set a record within the %{link}.'
-        activation_key: 'Activation Key'
-        activation_key_hint: 'For example 1-00112233445566778899aabbccddeeff'
+        bootstrap_script_url: 'Full path to the bootstrap script'
+        bootstrap_script_url_hint: |
+          This is the script administrator of SUSE Manager server has
+          prepared to be used for activation of the clients. It its likely located under
```
vuntz: Typo: s/its/is/

crowbar_framework/config/locales/suse_manager_client/en.yml
```diff
@@ -18,10 +18,12 @@ en:
...
+          /pub/bootstrap directory at the server.
```
vuntz: s/at/on/ ?

crowbar_framework/config/locales/suse_manager_client/en.yml
```diff
@@ -18,10 +18,12 @@ en:
   barclamp:
     suse_manager_client:
       edit_attributes:
-        manager_server: 'SUSE Manager server hostname'
-        manager_server_hint: 'For example susemgr.example.com. You can set a record within the %{link}.'
-        activation_key: 'Activation Key'
-        activation_key_hint: 'For example 1-00112233445566778899aabbccddeeff'
+        bootstrap_script_url: 'Full path to the bootstrap script'
```
vuntz: s/Full path/URL/ ?

vuntz:
Minor changes in the string requested, and +1 from me after that.

jsuchome:
@vuntz your comments addressed, and gating finally green again

MaximilianMeister:
+1

crowbar_framework/config/locales/suse_manager_client/en.yml
```diff
@@ -18,10 +18,12 @@ en:
   barclamp:
     suse_manager_client:
       edit_attributes:
-        manager_server: 'SUSE Manager server hostname'
-        manager_server_hint: 'For example susemgr.example.com. You can set a record within the %{link}.'
-        activation_key: 'Activation Key'
-        activation_key_hint: 'For example 1-00112233445566778899aabbccddeeff'
+        bootstrap_script_url: 'URL of the bootstrap script'
+        bootstrap_script_url_hint: |
+          This is the script administrator of SUSE Manager server has
```
vuntz: Sorry, it seems I forgot one: I would add "the" before "administrator" and "SUSE Manager server".

Feel free to ignore and merge, or fix and merge.

vuntz:
+1

PR #510, opened 2016-06-24
Avoid duplicating network definitions in multiple locations

vuntz:
This makes dealing with the network config more difficult, as some bits might get outdated.

Instead, we centralize the definitions in the network role, and have proper API for accessing this easily.

This goes with https://github.com/crowbar/crowbar-ha/pull/119, https://github.com/crowbar/crowbar-hyperv/pull/25 and https://github.com/crowbar/crowbar-openstack/pull/415.

### DIFF
--- a/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+++ b/chef/cookbooks/barclamp/libraries/barclamp_library.rb
@@ -30,19 +30,27 @@ module BarclampLibrary
         answer = []
         intf_to_if_map = Barclamp::Inventory.build_node_map(node)
         node[:crowbar][:network].each do |net, data|
-          intf, interface_list, tm = Barclamp::Inventory.lookup_interface_info(node, data["conduit"], intf_to_if_map)
-          answer << Network.new(net, data, intf, interface_list)
-        end unless node[:crowbar].nil? or node[:crowbar][:network].nil?
+          # network is not valid if we don't have the full definition
+          next unless node[:network][:networks].key?(net)
+          network_def = node[:network][:networks][net].to_hash.merge(data.to_hash)
+          intf, interface_list, tm = Barclamp::Inventory.lookup_interface_info(node, network_def["conduit"], intf_to_if_map)
+          answer << Network.new(net, network_def, intf, interface_list)
+        end unless node[:crowbar].nil? || node[:crowbar][:network].nil? || node[:network][:networks].nil?
         answer
       end
 
       def self.get_network_by_type(node, type)
-        unless node[:crowbar][:network].nil?
+        unless node[:crowbar][:network].nil? || node[:network][:networks].nil?
           [type, "admin"].uniq.each do |usage|
-            if found = node[:crowbar][:network].find { |net, data| net == usage }
+            found = node[:crowbar][:network].find do |net, data|
+              # network is not valid if we don't have the full definition
+              node[:network][:networks].key?(net) && net == usage
+            end
+            unless found.nil?
               net, data = found
-              intf, interface_list, tm = Barclamp::Inventory.lookup_interface_info(node, data["conduit"])
-              return Network.new(net, data, intf, interface_list)
+              network_def = node[:network][:networks][net].to_hash.merge(data.to_hash)
+              intf, interface_list, tm = Barclamp::Inventory.lookup_interface_info(node, network_def["conduit"])
+              return Network.new(net, network_def, intf, interface_list)
             end
           end
           return nil
--- a/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+++ b/chef/cookbooks/barclamp/libraries/barclamp_library.rb
@@ -239,22 +239,25 @@ module BarclampLibrary
       end
 
       class Network
-        attr_reader :name, :address, :mtu, :broadcast, :netmask,
-          :subnet, :router, :vlan, :use_vlan,
-          :add_bridge, :conduit
+        attr_reader :name, :address, :broadcast, :netmask,
+          :subnet, :router, :router_pref, :mtu, :vlan, :use_vlan,
+          :add_bridge, :add_ovs_bridge, :bridge_name, :conduit
         def initialize(node, net, data)
           @node = node
           @name = net
           @address = data["address"]
-          @mtu = (data["mtu"] || 1500).to_i
           @broadcast = data["broadcast"]
           @netmask = data["netmask"]
           @subnet = data["subnet"]
           @router = data["router"]
-          @vlan = data["vlan"]
+          @router_pref = data["router_pref"].to_i
+          @mtu = (data["mtu"] || 1500).to_i
+          @vlan = data["vlan"].to_i
           @use_vlan = data["use_vlan"]
           @conduit = data["conduit"]
           @add_bridge = data["add_bridge"]
+          @add_ovs_bridge = data["add_ovs_bridge"]
+          @bridge_name = data["bridge_name"]
           # let's resolve this only if needed
           @interface = nil
           @interface_list = nil
--- a/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+++ b/chef/cookbooks/barclamp/libraries/barclamp_library.rb
@@ -28,12 +28,15 @@ module BarclampLibrary
 
       def self.list_networks(node)
         answer = []
-        node[:crowbar][:network].each do |net, data|
-          # network is not valid if we don't have the full definition
-          next unless node[:network][:networks].key?(net)
-          network_def = node[:network][:networks][net].to_hash.merge(data.to_hash)
-          answer << Network.new(node, net, network_def)
-        end unless node[:crowbar].nil? || node[:crowbar][:network].nil? || node[:network][:networks].nil?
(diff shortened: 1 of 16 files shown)


chef/cookbooks/network/recipes/default.rb
```diff
@@ -142,55 +142,53 @@ def kill_nic(nic)
...
-  addr = if network["address"]
-           IP.coerce("#{network["address"]}/#{network["netmask"]}")
-         else
-           nil
-         end
-  conduit = network["conduit"]
-  base_ifs = conduit_map[conduit]["if_list"]
+  addr = if network.address
+    IP.coerce("#{network.address}/#{network.netmask}")
+  else
```
vuntz: Some people like to keep the explicit `else` here, so keeping it.

crowbar_framework/app/models/network_service.rb
```diff
@@ -413,8 +412,6 @@ def build_net_info(network, name, db = nil)
     db["network"].each { |k,v|
       net_info[k] = v unless v.nil?
     }
-    net_info["usage"]= network
-    net_info["node"] = name
```
rhafer: Seems the "name" parameter of ```build_net_info``` is no longer used now, shouldn't we remove that as well?

vuntz: I removed it later on, in another commit :-)

rhafer: Really, which one? Looking at https://github.com/crowbar/crowbar-core/pull/510/files#diff-0998300bd8296b99808ceb69e36a90d8R406
it's still there ```def build_net_info(network, name, db = nil)```

rhafer:
Overall I really like this work. Thanks a lot for that. I only did a pretty high level review (and no real testing yet).

Does it make sense to provide an API  also for overriding network attributes via the node object?
AFAIK this currently affects the ```add_ovs_bridge``` and ```use_vlan``` attributes, but as you already noted overriding other attribute (e.g. the conduit) might also make sense.

crowbar_framework/app/models/network_service.rb
vuntz: In https://github.com/crowbar/crowbar-core/pull/511/commits/02f6c3d782f795487518f116ceb8937860e41271, but that's part of the followup PR, indeed. I can move it here if you want, though.

vuntz:
> Overall I really like this work. Thanks a lot for that. I only did a pretty high level review (and no real testing yet).

After fixing a number of regressions/bugs I introduced, I'm running tempest right now, so this seems to work okay-ish so far (testing with ovs and gre+vxlan+vlan, so I guess we should also test linuxbridge and ovs+vlan).

> Does it make sense to provide an API also for overriding network attributes via the node object?
> AFAIK this currently affects the add_ovs_bridge and use_vlan attributes, but as you already noted overriding other attribute (e.g. the conduit) might also make sense.

Yes, I think it makes sense. After giving it some thought, I also want to centralize all the use of `node[:network][:networks]` through some `get_network_definition` API. This will clarify things. I was initially planning to do this in another followup PR, but I can add all that here if you prefer.

vuntz:
> Yes, I think it makes sense. After giving it some thought, I also want to centralize all the use of node[:network][:networks] through some get_network_definition API. This will clarify things. I was initially planning to do this in another followup PR, but I can add all that here if you prefer.

Ah well, I went ahead and did it here.

crowbar_framework/app/models/network_service.rb
vuntz: Done here.

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -197,7 +197,11 @@ def find_node_boot_mac_addresses(node, admin_data_net)
           hostname mnode.name
           macaddress mac_list[i]
           if admin_mac_addresses.include?(mac_list[i])
-            ipaddress admin_data_net.address unless admin_data_net.nil?
+            if admin_data_net.nil?
+              ipaddress mnode[:ipaddress]
+            else
+              ipaddress admin_data_net.address unless admin_data_net.nil?
```
rhafer: The ```unless admin_data_net.nil?```  is not needed here anymore  since we are already in the ```else``` of ```if admin_data_net.nil?```

rhafer:
Apart from the small nitpick (and probably some of the hound complaints) you get my +1 (though I didn't find time to test this)

chef/cookbooks/provisioner/recipes/update_nodes.rb
vuntz: Added a commit for this.

vuntz:
Had to rebase :/

toabctl:
lgtm (but hav not tested it)

aspiers:
Since making #575 I suddenly feel I understand a lot of this code, so this stays on my review TODO list.

crowbar_framework/app/models/node_object.rb
```diff
@@ -830,15 +830,29 @@ def destroy
   end
 
   def networks
-    self.crowbar["crowbar"]["network"] rescue {}
+    networks = {}
+    crowbar["crowbar"]["network"].each do |name, data|
+      # note that node might not be part of network proposal yet (for instance:
+      # if discovered, and IP got allocated by user)
```
rhafer: @vuntz I might be missing something here, but if the node is not part of the network proposal yet, I think we'd not even get here. Because ```crowbar["crowbar"]["network"]``` should also be empty in that case.

rhafer: @vuntz please ignore that. It can actually happen when allocating an IP on the node before allocating the node. (It also happened before you added https://github.com/crowbar/crowbar-core/pull/510/commits/578947a428db527fcc46705dcbd3921d8fb67af6)

rhafer:
Just for the record. I am still +1 on this. @toabctl how about you?

@vuntz: Do we need to provide some cleanup to deduplicate the node.crowbar.network attributes for people upgrading from older releases?

crowbar_framework/app/models/network_service.rb
aspiers: Hmm, the node name will be needed for https://trello.com/c/7W51qfqu - do you have to remove it? I guess it would be easy to put back though...

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -39,7 +39,7 @@ def self.list_networks(node)
       def self.get_network_by_type(node, type)
         unless node[:crowbar][:network].nil?
           [type, "admin"].uniq.each do |usage|
-            if found = node[:crowbar][:network].find { |net, data| data[:usage] == usage }
+            if found = node[:crowbar][:network].find { |net, data| net == usage }
```
toabctl: Unrelated to that line (but to the commit and I can't comment on the commit message): Not sure if that file is ever used for testing but in ```crowbar_framework/spec/fixtures/offline_chef/role_crowbar-admin_crowbar_com.json``` is still ```[:crowbar][:network][:admin][:usage]``` and ```[:crowbar][:network][:bmc][:usage]```

crowbar_framework/app/models/network_service.rb
vuntz: @aspiers pretty sure it's not needed here, as the info is already available somewhere else (which is why I removed it) :-)

vuntz:
> @vuntz: Do we need to provide some cleanup to deduplicate the node.crowbar.network attributes for people upgrading from older releases?

I guess it'd be nice, although not mandatory.

toabctl:
+1 (but I haven't tested it)

crowbar_framework/app/models/network_service.rb
aspiers: @vuntz I think you might be missing my point, unless you mean it's already available somewhere else *in this function*?  My point was not that this is needed *now*, but that it will be needed if in the future we do https://trello.com/c/7W51qfqu.

chef/cookbooks/network/recipes/default.rb
```diff
@@ -142,55 +142,53 @@ def kill_nic(nic)
...
+    nil
+  end
+  base_ifs = conduit_map[network.conduit]["if_list"]
   # Error out if we were handed an invalid conduit mapping.
-  unless base_ifs.all?{ |i|i.is_a?(String) && ::Nic.exists?(i) }
-    raise ::ArgumentError.new("Conduit mapping \"#{conduit}\" for network \"#{name}\" is not sane: #{base_ifs.inspect}")
+  unless base_ifs.all? { |i| i.is_a?(String) && ::Nic.exists?(i) }
+    raise ::ArgumentError.new("Conduit mapping \"#{network.conduit}\" for network \"#{network.name}\" is not sane: #{base_ifs.inspect}")
```
aspiers: Any chance you could split this line?  No need for it to be so long.

crowbar_framework/app/models/network_service.rb
vuntz: Ah, got it: what you need is the parameter to this call. Yes, better to add it back later on if needed, imho. Not a big fan of keeping unused parameters.

PR #511, opened 2016-06-24
Do not put network definition in data bag

vuntz:
~~Contains (and therefore depends on) https://github.com/crowbar/crowbar-core/pull/510 (merged)~~

### DIFF
--- a/crowbar_framework/app/models/network_service.rb
+++ b/crowbar_framework/app/models/network_service.rb
@@ -294,56 +294,58 @@ class NetworkService < ServiceObject
         @logger.error(msg)
         return [400, msg]
       end
-    end
 
-    if state == "hardware-installing"
-      node = NodeObject.find_node_by_name name
-
-      # Allocate required addresses
-      range = node.admin? ? "admin" : "host"
-      @logger.debug("Deployer transition: Allocate admin address for #{name}")
-      result = allocate_ip("default", "admin", range, name)
-      @logger.error("Failed to allocate admin address for: #{node.name}: #{result[0]}") if result[0] != 200
-      if result[0] == 200
-        address = result[1]["address"]
-        boot_ip_hex = sprintf("%08X", address.split(".").inject(0) { |acc, i| (acc << 8) + i.to_i })
-        # reload object after allocating the IP address, to get attribute info about admin network
+      if state == "hardware-installing"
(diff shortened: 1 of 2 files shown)


vuntz:
Rebased

crowbar_framework/app/models/network_service.rb
```diff
@@ -312,9 +317,10 @@ def transition(inst, name, state)
         # If we are the admin node, we may need to add a vlan bmc address.
         # Add the vlan bmc if the bmc network and the admin network are not the same.
         # not great to do it this way, but hey.
-        admin_net = Chef::DataBag.load "crowbar/admin_network" rescue nil
-        bmc_net = Chef::DataBag.load "crowbar/bmc_network" rescue nil
-        if admin_net["network"]["subnet"] != bmc_net["network"]["subnet"]
+        node_networks = node.networks
```
rhafer: @vuntz Somehow the installer fails here for me. AFAICS ```node.networks``` is empty because the node doesn't have the network proposal fully applied yet.

crowbar_framework/app/models/network_service.rb
```diff
@@ -323,7 +329,7 @@ def transition(inst, name, state)
 
         # Allocate the bastion network ip for the admin node if a bastion
         # network is defined in the network proposal
-        bastion_net = Chef::DataBag.load "crowbar/bastion_network" rescue nil
+        bastion_net = node_networks["bastion"]
```
vuntz: This is wrong, we need to look at the network role here as we are trying to decide if we should allocate an address on the bastion network.

vuntz:
@rhafer thanks to your changes, gating seems to be passing.

rhafer:
+1 (though that doesn't fully count as I provided some fixes)

rhafer:
@aspiers Could you have a look at this?

aspiers:
@rhafer I'll try - looks a bit scary though ;-)

rhafer:
> @aspiers  I'll try - looks a bit scary though ;-)

It's mostly @vuntz' work so it can't have any issues ;)

rhafer:
@aspiers, any chance you can look into this soonish? I'd really like to see this in Cloud 7

vuntz:
Rebased... Would welcome a missing +1 :-)

PR #514, opened 2016-06-27
chef: Add CrowbarRoleRecipe.node_state_valid_for_role helper

vuntz:
This is used in role recipes to decide whether the role should be run or
not. This will allow us to stop rebuilding the run_list based on the
state of the nodes, since we move the logic in the role recipes.


chef/cookbooks/utils/libraries/role_recipe.rb
```diff
@@ -0,0 +1,49 @@
...
+      else
+        ["all"]
+      end
+
+      valid = states_for_role.nil? ||
+        states_for_role.include?("all") ||
+        states_for_role.include?(node[:state])
+
+      unless valid
+        Chef::Log.info("Skipping role \"#{role}\" because node is in state \"#{node[:state]}\".")
```
jsuchome: Could we add a list of valid states to this message? It's a detail, but it might make more sense when user finds it in the logs.

vuntz: Not sure if it's really useful, but done.

jsuchome:
+1

chef/cookbooks/utils/libraries/role_recipe.rb
```diff
@@ -0,0 +1,50 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+module CrowbarRoleRecipe
+  # Helper to decide whether a role should be run for a node, depending on the
+  # state of the node.
+  def self.node_state_valid_for_role(node, barclamp, role)
```
MaximilianMeister: why not call it `node_state_valid_for_role?`

chef/cookbooks/utils/libraries/role_recipe.rb
```diff
@@ -0,0 +1,50 @@
...
+    if node.key? barclamp
+      states_for_role = if node[barclamp].key? "element_states"
+        # if nil, then this means all states are valid
+        node[barclamp]["element_states"][role]
+      else
+        ["all"]
+      end
+
```
MaximilianMeister: here we could return early:

```ruby
return true if states_for_role.nil? || states_for_role.include?("all") || states_for_role.include?(node[:state])
```

...

chef/cookbooks/utils/libraries/role_recipe.rb
```diff
@@ -0,0 +1,50 @@
...
+      valid = states_for_role.nil? ||
+        states_for_role.include?("all") ||
+        states_for_role.include?(node[:state])
+
+      unless valid
```
MaximilianMeister: ...and drop the `unless` block ...

chef/cookbooks/utils/libraries/role_recipe.rb
```diff
@@ -0,0 +1,50 @@
...
+        Chef::Log.info("Skipping role \"#{role}\" because node is in state \"#{node[:state]}\". " \
+          "Role \"#{role}\" only applies in the following states: #{states_for_role.join(", ")}.")
+      end
+    else
+      # Generally speaking, if we don't even have attributes related to the
+      # barclamp, then it means the role should not even be there.
+      # There's one exception, though: when bootstrapping the admin server,
+      # this happens for a couple of roles.
```
MaximilianMeister:
```ruby
return true if ["crowbar", "deployer-client"].include? role
```

...

chef/cookbooks/utils/libraries/role_recipe.rb
```diff
@@ -0,0 +1,50 @@
...
+      valid = ["crowbar", "deployer-client"].include? role
+
+      unless valid
```
MaximilianMeister: ... and drop the `unless` block as well ...

chef/cookbooks/utils/libraries/role_recipe.rb
```diff
@@ -0,0 +1,50 @@
...
+        Chef::Log.info("Skipping role \"#{role}\" because node does not have applied proposal for #{role} in its runlist.")
+      end
+    end
+
+    valid
```
MaximilianMeister: ... and finally add a
```ruby
return false
```

vuntz:
@MaximilianMeister integrated all your feedback. To be honest, I'm not convinced by the coding style change, but I won't fight over it :-)

MaximilianMeister:
> integrated all your feedback. To be honest, I'm not convinced by the coding style change, but I won't fight over it :-)

the style changes were just a suggestion (which saves us some useless checks and codepaths), more important was the `?` suffix. but feel free to not do the change if you're not convinced.

anyways, thanks for the update. +1

vuntz:
> but feel free to not do the change if you're not convinced.

I'm too lazy to revert ;-)

jsuchome:
still +1

PR #515, opened 2016-06-27
crowbar: Stop rebuilding node run_list depending on the node state

vuntz:
Whether a role should be handled as part of a chef run will now be
decided in the role recipes.

The benefit of this is that the roles will always be assigned to the
node, even if the roles are not used due to the state. This fixes
several issues, including the fact that the search results for nodes are
not reliable when a node is in some state that is not "readying",
"applying" or "ready".

Depends on ~~https://github.com/crowbar/crowbar-core/pull/514, https://github.com/crowbar/crowbar-ha/pull/121, https://github.com/crowbar/crowbar-ceph/pull/56, https://github.com/crowbar/crowbar-openstack/pull/419~~

### DIFF
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -706,36 +706,32 @@ class NodeObject < ChefObject
   end
 
   def add_to_run_list(rolename, priority, states = nil)
-    states = ["all"] unless states
-    crowbar["run_list_map"] = {} if crowbar["run_list_map"].nil?
-    val = { "states" => states, "priority" => priority }
+    # FIXME: Crowbar 4.0: we keep states parameter for compatibility reason; it
+    # should be removed in Crowbar 5.0
+    crowbar["run_list_map"] ||= {}
+    val = { "priority" => priority }
     crowbar["run_list_map"][rolename] = val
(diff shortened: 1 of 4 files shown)


nicolasbock:
+1

crowbar_framework/app/models/node_object.rb
```diff
@@ -706,58 +706,40 @@ def crowbar_run_list(*args)
...
-    crowbar["run_list_map"] = {} if crowbar["run_list_map"].nil?
+    crowbar["run_list_map"] ||= {}
 
-    # Cull by state
-    map = crowbar["run_list_map"].select do |k, v|
-      v["states"].include?("all") || v["states"].include?(crowbar["state"])
-    end
+    # FIXME: Crowbar 4.0: remove this hack to drop old items, which exists only
+    # to allow upgrade from 3.0 to work; should be removed in Crowbar 5.0
+    crowbar["run_list_map"].delete_if { |k, v| v["priority"] == -1001 }
```
MaximilianMeister: MINOR^2: could use `_k` for the unused block variable. normally hound used to complain about that...

MaximilianMeister:
+1

PR #516, opened 2016-06-29
Rework auto-assignment of roles

vuntz:
This is quite a big PR that reworks auto-assignment of roles:

  - for roles auto-assigned to the admin server, we actually do this in a bootstrap-specific code path. This path used to exist before, but was not really visible, and was abusing things. For instance, apply_role_pre_chef_call was not called, which could have lead to missing attributes on the admin server.

  - for roles that are added to discovered nodes: the transition method of the relevant service objects are simplified, and the addition of roles is now done after discovery, when the node is allocated. This leads to discovered nodes not having roles they don't need (which simplifies things when applying most barclamps because this won't be blocked by non-allocated nodes).

  - we also make sure that on reboot, nodes automatically get again the roles which are auto-assigned, in case they got lost.

  - we also take benefit from role recipes to merge some roles (provisioner-bootdisk-finder merged in provisioner-base, both ipmi-* roles merged into one ipmi role); this matters with this PR, due to the rework of auto-assignment of these roles.

Gating will be red, as there was a slight API change in ServiceObject; this requires ~~https://github.com/crowbar/crowbar-openstack/pull/421 and~~ https://github.com/crowbar/crowbar-ha/pull/123. We also need https://github.com/SUSE-Cloud/automation/pull/1045


vuntz:
Note that https://github.com/SUSE-Cloud/automation/pull/1045 had a gating run with all the required PR (this had to be done against the automation PR because it's not possible to add a mkcloud patch to a non-automation PR).

dirkmueller:
can you split the bootstrap=false addition to a separate PR? then the other api changes related PRs can be gated+merged individually

vuntz:
> can you split the bootstrap=false addition to a separate PR? then the other api changes related PRs can be gated+merged individually

Done in https://github.com/crowbar/crowbar-core/pull/537

toabctl:
@vuntz hound is still not happy

crowbar_framework/app/models/ipmi_service.rb
```diff
@@ -33,6 +33,19 @@ def role_constraints
     end
   end
 
+  def proposal_create_bootstrap(params)
+    if params["deployment"].nil? ||
```
toabctl: DRY. These steps are done in all the services. Can you create a method (i.e. ```set_deployment_element```) ?

toabctl:
Beside https://github.com/crowbar/crowbar-core/pull/516#discussion_r69729377, LGTM

vuntz:
> Beside #516 (comment), LGTM

I added a commit for this.

toabctl:
+1

vuntz:
https://ci.suse.de/job/openstack-mkcloud/26812/ is a gating job with the patch from https://github.com/SUSE-Cloud/automation/pull/1045 (which is required for this to pass).

vuntz:
Rebased (again).

toabctl:
still +1

vuntz:
Yet another rebase...

MaximilianMeister:
looks good to me (but haven't actually tested it)

+1

PR #517, opened 2016-06-29
Do not install openvswitch-kmp-default package for SLES12SP2

jsuchome:
Supposedly it should work well directly with SP2 kernel. See also https://build.suse.de/package/view_file/SUSE:SLE-12-SP2:GA/openvswitch/openvswitch.changes?expand=1

### DIFF
--- a/chef/cookbooks/network/attributes/default.rb
+++ b/chef/cookbooks/network/attributes/default.rb
@@ -23,8 +23,8 @@ when "suse"
     "openvswitch",
     "openvswitch-switch"
   ]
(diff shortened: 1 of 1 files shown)


jsuchome:
And I believe we do not have to care about SP1 nodes, as they are for ceph only

vuntz:
+1; to make it easier to merge (when we still use SP1), you might want to add a check on the version.

Note also that in the nova cookbook, we do something similar for the xen flavor.

jsuchome:
ok, I'll add a version check

aplanas:
+1

PR #518, opened 2016-06-29
crowbar: Fix visibility of license key input in bulk edit

vuntz:
We were not referring to the right name for the tag, and therefore the
license key was always visible.


toabctl:
assuming it works, +1

rsalevsky:
+1

PR #521, opened 2016-06-29
Fix crowbar batch export

vuntz:
It got broken when we stopped uploading the proposal templates as data
bags.


vuntz:
Re-triggered gating with https://github.com/crowbar/crowbar-openstack/pull/423

rsalevsky:
+1

toabctl:
+1

PR #522, opened 2016-06-30
increase crowbar version for API enpoint /crowbar/:barclamp

MaximilianMeister:


vuntz:
Can you explain why in the commit message?

MaximilianMeister:
not sure now, maybe this stands for the API version? then the would be wrong. I thought this refers to the crowbar version. do you have an idea @vuntz ?

vuntz:
I think it's the API version.

MaximilianMeister:
closing, as this doesn't refer to the crowbar version

PR #524, opened 2016-06-30
Routes

rsalevsky:


rsalevsky:
Needs to be spited into smaller PR's, some parts of crowbar acting strange without error messages.

PR #525, opened 2016-06-30
barclamp: Fix potential infinite loop in Nic.dependent

vuntz:
This is actually triggered on SLE 12 SP2.


jsuchome:
+1

toabctl:
Can this be tested with a spectest?

rhafer:
+1 

> Can this be tested with a spectest?

Yeah, having some tests for this would be cool. Especially as there might be more changes for the network related bits coming soon. But as this is bug is currently blocking the SP2 switch I'd prefer this going in even without a testcase. Especially since we don't have any test suite part for crowbar yet. Should we create a card for it?

PR #526, opened 2016-06-30
barclamp: Do not generate parents list for non-vlan ifaces

jsuchome:
It could potentially lead to the infinite loop

This is technically not needed when we have https://github.com/crowbar/crowbar-core/pull/525, but might still have a logical sense, as the function already suggests it should not be used for other kind of interfaces


rhafer:
+1

rsalevsky:
+1

PR #527, opened 2016-06-30
add some API documentation

MaximilianMeister:
this renders a static site which is currently at `/apidoc` (in this configuration)

https://github.com/Apipie/apipie-rails

screenshot from 2016-06-30 13-45-00

### DIFF
--- a/crowbar_framework/Gemfile
+++ b/crowbar_framework/Gemfile
@@ -35,6 +35,7 @@ gem "sqlite3", "~> 1.3.9"
 gem "syslogger", "~> 1.6.0"
 gem "yaml_db", "~> 0.3.0"
 gem "easy_diff", "~> 0.0.5"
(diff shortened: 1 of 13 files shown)


toabctl:
I think you need to add ```apipie``` to the media to get gating green, right? Failure is ```Could not find gem 'apipie-rails (~> 0.3.6) ruby```.
And is it possible to add a spec test?

MaximilianMeister:
> I think you need to add apipie to the media to get gating green, right? Failure is Could not find gem 'apipie-rails (~> 0.3.6) ruby.

yes, thanks for the reminder. I had to do other work and this got a bit forgotten...

> And is it possible to add a spec test?

not sure what to test here... any idea?

rsalevsky:
+1

nicolasbock:
+1

PR #528, opened 2016-07-01
dns: Run dns-client role only in readying/ready/applying states

vuntz:
This is like nearly all other roles.

The change only really impacts the discovery image, but it turns out
that with the discovery image, we use DHCP, so we get the DNS info from
DHCP in that case.

To be super safe, we should merge https://github.com/crowbar/crowbar-core/pull/529 first, but I doubt it matters in reality.


nicolasbock:
+1

MaximilianMeister:
+1

PR #529, opened 2016-07-01
provisioner: Fix list of DNS servers hinted by DHCP

vuntz:
We were hardcoding the admin IP address there.
But we want to allow people to host the DNS server on other nodes than the admin-node.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/dhcp_update.rb
+++ b/chef/cookbooks/provisioner/recipes/dhcp_update.rb
@@ -1,9 +1,17 @@
+admin_ip = Barclamp::Inventory.get_network_by_type(node, "admin").address
 
+dns_servers = search(:node, "roles:dns-server").map do |n|
+  Barclamp::Inventory.get_network_by_type(n, "admin").address
+end
+dns_servers.sort!
+dns_servers.concat(node[:dns][:nameservers]) unless node[:dns].nil?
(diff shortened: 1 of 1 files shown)


bmwiedemann:
Not sure, what problem you are trying to fix here, but as far as I understood, DHCP is only used in discovery and autoyast install stages, where using the admin IP as DNS is the most appropriate (e.g. because it allows reverse lookups of IP when you ssh into a discovered system)... and HA is not a concern, because they cannot do anything without the admin node anyway.

vuntz:
@bmwiedemann right now, it's possible to run a cloud where the admin server is not the dns server. You can deploy one node, move the dns-server role there, then discover other nodes (some people do that, yes).

This is what this would be fixing.

bmwiedemann:
I think, there might be other problems, because we trigger a chef-client run on the admin-node for certain DNS/DHCP changes to happen.
And did we implement moving or uninstalling services yet?

vuntz:
@bmwiedemann I'm not claiming it's fixing the full thing, just that at least the config file is reflecting the reality. Is there an issue with this change?

bmwiedemann:
+1  (was just making sure to understand why we are doing the change - a question that should ideally be answered by the commit message)

nicolasbock:
+1

PR #530, opened 2016-07-01
crowbar: Stop running chef a second time on apply if things failed, and in a different way for admin server

vuntz:
This was a horrible workaround for various bugs, mostly in the cookbooks. This shouldn't be needed anymore -- if it still is, then we should fix the real bugs.

Also, for the admin server: This was done in the days where chef had no locking to avoid concurrent runs, since the admin server had some other chef runs started by crowbar (on transitions, for instance). This has been fixed since a long time now, so there's no reason to continue this.

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1313,12 +1313,6 @@ class ServiceObject
         non_admin_nodes << node_name
       end
 
-      #
-      # XXX: We used to do this twice - do we really need twice???
-      # Yes! We do!  The system has some transient issues that are hidden
-      # but the double run for failing nodes.  For now, we will do this.
-      # Make this better one day.
-      #
       pids = {}
       unless non_admin_nodes.empty?
         @logger.debug(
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/service_object.rb
```diff
@@ -1300,105 +1300,41 @@ def apply_role(role, inst, in_queue)
...
       end
 
-      #
-      # XXX: We used to do this twice - do we really need twice???
-      # Yes! We do!  The system has some transient issues that are hidden
-      # but the double run for failing nodes.  For now, we will do this.
-      # Make this better one day.
-      #
+      @logger.debug(
+        "AR: Calling chef-client for #{role.name} on non-admin nodes #{batch.join(" ")}"
```
jsuchome: So is this actually correct, now when you call `run_remote_chef_client` for all nodes, not just non-admin ones?

vuntz: Fixed.

jsuchome:
+ 0.5 LGTM

rhafer:
Looks to me in general. There are two other places we chef-client runs might be triggered on the admin node (https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/provisioner_service.rb#L175 and https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/node_object.rb#L418). Should we try to adjust that as well? Or would that better be part of another PR?

vuntz:
> There are two other places we chef-client runs might be triggered on the admin node (https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/provisioner_service.rb#L175 and https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/node_object.rb#L418). Should we try to adjust that as well? Or would that better be part of another PR?

I would do it in another PR, as I believe we actually want to keep the current call there: the current call is non-blocking, which is desired in these cases (in apply_role, we actially want a blocking call).

rhafer:
Ok, makes sense to me.

+1

vuntz:
(considering @jsuchome +0.5 enough for a +1 ;-))

PR #531, opened 2016-07-01
crowbar: Add rake task "schema_migrate_status" for migration overview

toabctl:
To get an overview about the current migration status for the different
proposals, add a new rake task called "schema_migrate_status".
Calling the task with

RAILS_ENV=production rake crowbar:schema_migrate_status

gives a list with the template name, the latest possible revision and a list
of proposals with the currently applied revision. It looks like:

```
*barclamp*           *latest revision*    *proposals revision*
nova                 101                  default:101
keystone             35                   default:35
rabbitmq             11                   default:11
neutron              48                   default:48
```

### DIFF
--- a/crowbar_framework/lib/schema_migration.rb
+++ b/crowbar_framework/lib/schema_migration.rb
@@ -25,15 +25,8 @@ module SchemaMigration
   end
 
   def self.run_for_bc bc_name
-    begin
-      template = Proposal.new(barclamp: bc_name)
-    rescue Proposal::TemplateMissing
-      return
-    end
-
+    template = get_template_for_barclamp bc_name
     return if template.nil?
-    return if template["deployment"].nil?
-    return if template["deployment"][bc_name].nil?
 
     all_scripts = find_scripts_for_bc(bc_name)
     return if all_scripts.empty?
@@ -63,11 +56,8 @@ module SchemaMigration
   end
 
   def self.migrate_proposal_from_json(bc_name, json)
-    template = Proposal.new(barclamp: bc_name)
-
+    template = get_template_for_barclamp bc_name
     return if template.nil?
-    return if template["deployment"].nil?
-    return if template["deployment"][bc_name].nil?
 
     all_scripts = find_scripts_for_bc(bc_name)
     return if all_scripts.empty?
@@ -80,11 +70,8 @@ module SchemaMigration
   end
 
   def self.get_barclamp_current_deployment_revison(bc_name)
-    begin
-      template = Proposal.new(barclamp: bc_name)
-    rescue Proposal::TemplateMissing
-      return [nil, nil]
-    end
+    template = get_template_for_barclamp bc_name
+    return [nil, nil] if template.nil?
 
     proposals = Proposal.where(barclamp: bc_name)
     latest_schema_revision = template["deployment"][bc_name]["schema-revision"]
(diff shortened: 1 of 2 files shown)


toabctl:
Not sure if this is useful. I just wanted to learn a bit about crowbar and the migrations....

crowbar_framework/lib/tasks/crowbar.rake
```diff
@@ -37,4 +37,29 @@ namespace :crowbar do
...
+    end
+
+    printf "%-20s %-20s %s\n", "*barclamp*", "*latest revision*", "*proposals revision*"
+    barclamps.split.each do |bc_name|
+      latest_schema_revision, latest_proposals_revision = \
+                              SchemaMigration.get_barclamp_current_deployment_revison bc_name
+      unless latest_proposals_revision.nil?
+        printf "%-20s %-20s %s\n", bc_name, latest_schema_revision,
+               latest_proposals_revision.collect{
+          |prop| "#{prop[:name]}:#{prop[:revision]}"}.join(" ")
```
vuntz: Interesting that hound doesn't complain. Correct style would be:
```ruby
unless latest_proposals_revision.nil?
  proposals_rev = latest_proposals_revision.collect { |prop| "#{prop[:name]}:#{prop[:revision]}" }
  printf "%-20s %-20s %s\n", bc_name, latest_schema_revision, proposals_rev.join(" ")
end
```

crowbar_framework/lib/tasks/crowbar.rake
```diff
@@ -37,4 +37,29 @@ namespace :crowbar do
...
+    require "barclamp_catalog"
+
+    if args[:barclamps].nil?
+      barclamps = BarclampCatalog.barclamps.keys.join(" ")
+    else
+      barclamps = args[:barclamps]
+    end
+
+    printf "%-20s %-20s %s\n", "*barclamp*", "*latest revision*", "*proposals revision*"
+    barclamps.split.each do |bc_name|
```
vuntz: Can you sort barclamps by name?

crowbar_framework/lib/schema_migration.rb
```diff
@@ -79,6 +79,21 @@ def self.migrate_proposal_from_json(bc_name, json)
...
+      template = Proposal.new(barclamp: bc_name)
+    rescue Proposal::TemplateMissing
+      return [nil, nil]
+    end
+
+    proposals = Proposal.where(barclamp: bc_name)
+    latest_schema_revision = template["deployment"][bc_name]["schema-revision"]
+    latest_proposals_revision = proposals.collect do |prop|
+      { name: prop.name, revision: prop["deployment"][bc_name]["schema-revision"] }
+    end
```
vuntz: Can you sort this by proposal name?

vuntz:
> Not sure if this is useful.

Feels useful to me :-)

crowbar_framework/lib/schema_migration.rb
toabctl: done

crowbar_framework/lib/tasks/crowbar.rake
toabctl: done

crowbar_framework/lib/tasks/crowbar.rake
toabctl: done

toabctl:
@vuntz fixed you comments. and added another cleanup commit.

vuntz:
+1

rsalevsky:
+1

If you want to make the output a little bit better then you can use https://github.com/tj/terminal-table. We use it anyway already in the crowbar-client.

toabctl:
> If you want to make the output a little bit better then you can use https://github.com/tj/terminal-table. We use it anyway already in the crowbar-client.

Imo not needed. And keeping the requirements for the server at a minimum is a good thing.

aspiers:
Nice addition, thanks :)

PR #532, opened 2016-07-01
Refactor routes part 1

rsalevsky:


crowbar_framework/config/routes.rb
```diff
@@ -58,26 +59,38 @@
...
-  get "utils(.:format)", controller: "support", action: "index", as: "utils"
-  get "utils/files/:id(.:format)", controller: "support", action: "destroy", constraints: { id: /[^\/]+/ }, as: "utils_files"
-  get "utils/chef(.:format)", controller: "support", action: "export_chef", as: "export_chef"
-  get "utils/supportconfig(.:format)", controller: "support", action: "export_supportconfig", as: "export_supportconfig"
-  get "utils/:controller/1.0/export(.:format)", action: "export", as: "utils_export"
-  get "utils/:controller/1.0(.:format)", action: "utils", as: "utils_barclamp"
-  get "utils/import/:id(.:format)", controller: "support", action: "import", constraints: { id: /[^\/]+/ }, as: "utils_import"
-  get "utils/upload/:id(.:format)", controller: "support", action: "upload", constraints: { id: /[^\/]+/ }, as: "utils_upload"
-  get "utils/repositories(.:format)", controller: "repositories", action: "index", as: "repositories"
-  post "utils/repositories/sync(.:format)", controller: "repositories", action: "sync", as: "sync_repositories"
```
vuntz: I see we use `sync_repositories_path` in the code; shouldn't this be changed?

vuntz:
I'm really no rails expert, so I don't know if it's good or bad :-) As long as all current URLs keep working, I'm good.

crowbar_framework/config/routes.rb
MaximilianMeister: shouldn't be a problem as it is still the same alias after this change

MaximilianMeister:
overall it looks good to me

+1

MaximilianMeister:
wouldn't it have been better to rebase and run gating again? merging one after the other and gate each time

rsalevsky:
Doesn't matter the parts are independent.

PR #535, opened 2016-07-02
Clean up what we do in deployer-client role

vuntz:
Some bits are moved to provisioner-base, some are included only in case of discovery.

Rationale for the change is that deployer-client will always run in all cases, even when we only run a subset of roles, so we should minimize the amount of work it does in general.


jsuchome:
+ 0.5 Looks correct, but I do not know the code that much

abel-navarro:
I'm no expert neither on this code but looks correct: +1

PR #536, opened 2016-07-02
crowbar: Fix error when trying to get status of non-existing proposal

vuntz:
```
F, [2016-07-02T10:02:04.196340 #5021:0x000000061d6f90] FATAL -- Failed to iterate over proposal list due to 'undefined method `barclamp' for nil:NilClass'
/opt/dell/crowbar_framework/app/controllers/barclamp_controller.rb:793:in `block in proposal_status'
/opt/dell/crowbar_framework/app/controllers/barclamp_controller.rb:792:in `each'
/opt/dell/crowbar_framework/app/controllers/barclamp_controller.rb:792:in `proposal_status'
```

### DIFF
--- a/crowbar_framework/app/controllers/barclamp_controller.rb
+++ b/crowbar_framework/app/controllers/barclamp_controller.rb
@@ -788,6 +788,7 @@ class BarclampController < ApplicationController
           ).first
         ]
       end
+      result.compact!
 
       result.each do |prop|
         prop_id = "#{prop.barclamp}_#{prop.name}"


crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -788,6 +788,7 @@ def proposal_status
           ).first
         ]
       end
+      result.compact!
```
rsalevsky: Is it possible that `result` is empty? If yes, compact will return `nil` in this case which breaks the loop in the next line.

vuntz: It will be empty in the case of the bug I fixed, but I don't believe `compact` does what you think:

```
irb(main):001:0> [].compact
=> []
```

toabctl:
+1

rsalevsky:
+1

rsalevsky:
Just hit a bug regarding to this PR:

```
FATAL -- Failed to iterate over proposal list due to 'undefined method `compact!' for #<Proposal::ActiveRecord_Relation:0x000000041a0618>'
```

vuntz:
@rsalevsky full stack trace? I fail to see how `result` cannot be a list in the code...

vuntz:
Ah, got it... Investigating.

rsalevsky:
I hit this issue because of an other bug see #539. But I think it can still happen that `result` is nil.

PR #545, opened 2016-07-11
exclude cloud 6 repos after upgrade

MaximilianMeister:
this is partially related to 

* https://bugzilla.suse.com/show_bug.cgi?id=975384
* https://trello.com/c/1VDjvMFc/1036-bsc-975384-issues-in-crowbar-s-repository-handling
* the upgrade CI (where it fails after the upgrade due to old repos still in the registry)

this extends https://github.com/crowbar/crowbar-core/pull/505 a bit further, which was not the complete fix

### DIFF
--- a/chef/cookbooks/provisioner/recipes/base.rb
+++ b/chef/cookbooks/provisioner/recipes/base.rb
@@ -306,10 +306,8 @@ if node[:platform_family] == "suse" && !node.roles.include?("provisioner-server"
 
       out = `LANG=C zypper --non-interactive repos #{name} 2> /dev/null`
       out.split("\n").each do |line|
-        attribute, value = line.split(":", 2)
+        attribute, value = line.split(":", 2).map(&:strip)
         next if value.nil?
-        attribute.strip!
-        value.strip!
         if attribute == "URI"
           current_url = value
         elsif attribute == "Priority"
--- a/chef/cookbooks/provisioner/recipes/base.rb
+++ b/chef/cookbooks/provisioner/recipes/base.rb
@@ -311,7 +311,7 @@ if node[:platform_family] == "suse" && !node.roles.include?("provisioner-server"
         if attribute == "URI"
           current_url = value
         elsif attribute == "Priority"
-          current_priority = value
+          current_priority = value.to_i
         end
       end
 
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -299,9 +299,15 @@ if not nodes.nil? and not nodes.empty?
 
           target_platform_distro = os.gsub(/-.*$/, "")
           target_platform_version = os.gsub(/^.*-/, "")
-          repos = Provisioner::Repositories.get_repos(target_platform_distro,
-                                                      target_platform_version,
-                                                      arch)
+          repos = Provisioner::Repositories.get_repos(
+            target_platform_distro,
+            target_platform_version,
+            arch
+          )
+          # FIXME: We are just ignoring old repos after the upgrade here
+          # this needs to be improved for the next upgrade by actively removing
+          # the repositories after the upgrade (and on repo deactivation)
+          repos.reject! { |k, _v| k =~ /Cloud-6/ } if mnode[:state] == "os-upgrading"
           Chef::Log.info("repos: #{repos.inspect}")
 
           if node[:provisioner][:suse]


MaximilianMeister:
... I just realized that upgrade gating is useless here, as the cloud 7 patches won't be applied. will run it manually.

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -328,6 +326,16 @@
...
       end
+      # check if attrs contains the :purge key with the databag name as a value
+      if attrs[:purge]
+        repo_bag = attrs[:purge]
+
+        Chef::Log.info("Purging no longer used #{name} zypper repository...")
+        `zypper --non-interactive removerepo #{name}`
+        Chef::Log.info("Purging repository data bag #{repo_bag}...")
+        # dont rescue here because we want to exit the recipe in case of an error
+        Chef::DataBag.chef_server_rest.delete_rest("data/#{repo_bag}")
```
MaximilianMeister: can't do that yet, because another node might still not have run the zypper removerepo...
also it seems to forbidden to delete a databag from a client (non admin) node

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -323,9 +323,11 @@ def post_cleanup_repos
           false
         )
 
-        Crowbar::Repository.chef_data_bag_destroy(
-          "#{repository.data_bag_name}/#{repository.id}"
-        )
+        data_bag = "#{repository.data_bag_name}/#{repository.id}"
```
vuntz: (technically, it's a `data_bag_item` -- but no big deal)

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -323,9 +323,11 @@ def post_cleanup_repos
...
+        # mark the repository as "not needed anymore" for chef to clean it up in the next run
+        db = Chef::DataBag.load(data_bag)
+        db[:purge] = data_bag
+        db.save
```
vuntz: So shouldn't we also do that when disabling a repo from the webui?

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -328,13 +326,25 @@
       if current_priority != attrs[:priority]
         `zypper --non-interactive modifyrepo --priority #{attrs[:priority]} #{name}`
       end
+      # check if attrs contains the :purge key with the databag name as a value
+      if attrs[:purge]
```
vuntz: Every other call to zypper (last two blocks) should probably be in a `else` of this condition.

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -328,13 +326,25 @@
...
+        `zypper --non-interactive removerepo #{name}`
+      end
     end
     # install additional packages
     os = "#{node[:platform]}-#{node[:platform_version]}"
     if node[:provisioner][:packages][os]
       node[:provisioner][:packages][os].each { |p| package p }
     end
   end
+elsif node[:platform_family] == "suse" && \
```
vuntz: You don't need the check on the platform family, imho.

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -328,13 +326,25 @@
...
+    node.roles.include?("provisioner-server") && \
+    Provisioner::Repositories.all_repos_purged?(
+      node[:platform], node[:platform_version], node[:kernel][:machine]
+    )
```
vuntz: So that will only remove purged items related to the OS of the admin server. Don't you want to remove purged items for all OS?

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -328,13 +326,25 @@
...
+  Chef::Log.info("Purging repository data bag #{attrs[:purge]}...")
+  Chef::DataBag.chef_server_rest.delete_rest("data/#{attrs[:purge]}")
```
vuntz: `attrs` is not defined here.

chef/cookbooks/provisioner/libraries/repositories.rb
```diff
@@ -40,12 +40,35 @@ def get_repos(platform, version, arch)
...
+        end
+
+        true
+      end
+
+      def zypper_repo_active?(repo)
+        Chef::Node.list.keys.each do |node|
+          # TODO: this could maybe be done more elegantly
+          return true if `ssh #{node} "zypper lr"`.include?(repo)
+        end
```
vuntz: This is quite wrong, indeed. I would really not want this to go in, as I don't believe we should ssh from the cookbooks. What happens if a node is down, for instance?

vuntz:
So, if we only consider the upgrade CI bit: I would simply disable the repos as part of the upgrade process; right now, that'd be in the autoyast profile.

For the general bug fix, I think the current approach is a bit too hacky -- I really dislike the ssh in the cookbook, for instance. We could either just keep forever a data bag item containing a list of all purged repos, or we could add an attribute on each node to tell that node which repos to purge.

MaximilianMeister:
@vuntz thanks for the review, I kept the first 2 commits and added a new commit for the autoyast file, is that enough/correct like it is?

vuntz:
@MaximilianMeister that looks much cleaner, indeed! Would it be possible to add a FIXME or something so that we remember to remove/update that code for the next upgrade? And actually, should we do Cloud-7 in master, and Cloud-6 in stable/3.0?

MaximilianMeister:
> Would it be possible to add a FIXME or something so that we remember to remove/update that code for the next upgrade? 

will add a comment

> And actually, should we do Cloud-7 in master, and Cloud-6 in stable/3.0?

not sure, is this template rendered on cloud 6 or on cloud 7?

vuntz:
>> And actually, should we do Cloud-7 in master, and Cloud-6 in stable/3.0?
> 
> not sure, is this template rendered on cloud 6 or on cloud 7?

Oh, I think you're right. The admin server runs Cloud 7 when we need to disable the Cloud 6 repos on upgrade.

chef/cookbooks/provisioner/libraries/repositories.rb
MaximilianMeister: > What happens if a node is down, for instance?

jfyi this actually happened to me when I tested the patch, and the check was correct because it evaluated to false (with "Cannot reach node..." on STDOUT). but anyways this is horrible

MaximilianMeister:
@vuntz rebased and added the FIXME comment

vuntz:
It's a workaround, but +1 for the time being. Thanks!

MaximilianMeister:
rebased

rsalevsky:
+1

MaximilianMeister:
upgrade gating not applicable (due to reset of the admin node)

PR #546, opened 2016-07-13
Fix key checksum

dirkmueller:
For some reason or signing key changed. Adjust


bmwiedemann:
+1 as a quick fix
so someone did osc signkey --extend because the expiration time arrived... but what we really should do is match the gpg fingerprint, which does not change when extending the validity period.

rsalevsky:
+1

MaximilianMeister:
+1

aplanas:
There is a fail in C6, and I think that something needs to be backported in https://github.com/crowbar/crowbar/blob/stable/3.0/scripts/install-chef-suse.sh#L521

aplanas:
Also +1

PR #550, opened 2016-07-18
Use repository key fingerprints for verification (bsc#988907)

nicolasbock:


crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -389,19 +389,16 @@ def check_repo_tag
...
       key_path = repodata_media_path.join("repomd.xml.key") unless key_path.file?
 
       if key_path.file?
-        md5 = Digest::MD5.hexdigest(key_path.read)
-        if expected.is_a?(Array)
-          expected.include? md5
-        else
-          md5 == expected
-        end
+        fingerprint = `gpg --with-fingerprint #{key_path} | grep fingerprint | awk -F= '{print $2}'`
```
dirkmueller: LC_ALL=C

aplanas: is the LC_LANG required here for gpg?

crowbar_framework/config/repos-cloud.yml
```diff
@@ -118,6 +119,7 @@ suse-12.1:
       repomd:
         tag: "obsrepository://build.suse.de/SUSE:Updates:SLE-SERVER:12-SP1:x86_64/update"
         key_md5: "9a62177e1c6852d48453ed3909956d6b"
```
dirkmueller: remove the md5 lines?

nicolasbock: Yes, I'll do that.

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -389,19 +389,16 @@ def check_repo_tag
...
       key_path = repodata_media_path.join("repomd.xml.key") unless key_path.file?
 
       if key_path.file?
-        md5 = Digest::MD5.hexdigest(key_path.read)
-        if expected.is_a?(Array)
-          expected.include? md5
-        else
-          md5 == expected
-        end
+        fingerprint = `LC_ALL=C gpg --with-fingerprint #{key_path} | grep fingerprint | awk -F= '{print $2}'`
```
dirkmueller: just thinking out loud, are we sure that "awk" is always installed?

dirkmueller:
@nicolasbock  there are more key_md5 occurrences in repos-ses.yaml as well as many more in repos-cloud.yaml.

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -389,19 +389,15 @@ def check_repo_tag
...
       key_path = repodata_media_path.join("repomd.xml.key") unless key_path.file?
 
       if key_path.file?
-        md5 = Digest::MD5.hexdigest(key_path.read)
-        if expected.is_a?(Array)
-          expected.include? md5
-        else
-          md5 == expected
-        end
+        expected == `LC_ALL=C gpg --with-fingerprint #{key_path} | \
```
dirkmueller: you're removing the support for handling Arrays, imho this might still be needed.

nicolasbock:
Yes, I will have to fill in all of the other values. Can I access the keys somewhere so I don't have to do the filling in by deploying clouds of various versions?

crowbar_framework/lib/crowbar/repository.rb
nicolasbock: I'll add that back in.

dirkmueller:
crowbar_framework/config/locales/crowbar/en.yml:      check_failed_md5: 'Unexpected signature'

maybe we want to update that message (key/value)

dirkmueller:
@nicolasbock in theory "iosc signkey <prjname> | gpg --with-fingerprint -v - " should give you those fingerprints..

crowbar_framework/lib/crowbar/repository.rb
nicolasbock: >> +        fingerprint = `LC_ALL=C gpg --with-fingerprint #{key_path} | grep fingerprint | awk -F= '{print $2}'`
> 
> just thinking out loud, are we sure that "awk" is always installed?
We probably should try to avoid using it. I'll experiment a bit.

dirkmueller:
+1

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -389,18 +389,19 @@ def check_repo_tag
...
+      expected = @config.fetch("repomd", {})["fingerprint"]
       return true if expected.blank?
 
       key_path = repodata_path.join("repomd.xml.key")
       key_path = repodata_media_path.join("repomd.xml.key") unless key_path.file?
 
       if key_path.file?
-        md5 = Digest::MD5.hexdigest(key_path.read)
+        fingerprint = `LC_ALL=C gpg --with-fingerprint #{key_path} | \
+          grep fingerprint | awk -F= '{print $2}'`.split.join(" ")
```
vuntz: I would prefer if we were parsing the output of the gpg command from ruby instead of doing pipes. But not critical.

nicolasbock: I updated the PR

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -389,18 +389,20 @@ def check_repo_tag
...
       return true if expected.blank?
 
       key_path = repodata_path.join("repomd.xml.key")
       key_path = repodata_media_path.join("repomd.xml.key") unless key_path.file?
 
       if key_path.file?
-        md5 = Digest::MD5.hexdigest(key_path.read)
+        fingerprint = `LC_ALL=C gpg --with-fingerprint #{key_path}`.split(/\r?\n/)
+        fingerprint.keep_if { |d| d =~ /fingerprint/ }
+        fingerprint = fingerprint[0].split("=")[1].split.join(" ")
```
vuntz: Thanks, this looks nice. The only thing is that we need to protect against potential errors, like "no line with fingerprint", or "line with fingerprint that has no =", etc.

rsalevsky:
+1

vuntz:
+1

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -389,18 +389,23 @@ def check_repo_tag
...
       key_path = repodata_media_path.join("repomd.xml.key") unless key_path.file?
 
       if key_path.file?
-        md5 = Digest::MD5.hexdigest(key_path.read)
+        fingerprint = `LC_ALL=C gpg --with-fingerprint #{key_path}`.split(/\r?\n/)
+        fingerprint.keep_if { |d| d =~ /fingerprint/ }
+        return false if fingerprint.empty?
+        fingerprint = fingerprint[0].split("=")
+        return false if fingerprint.length != 2
+        fingerprint = fingerprint[1].split.join(" ")
```
rsalevsky: I'm not sure but this block looks like it is possible to do with the ruby array map magic.

PR #551, opened 2016-07-18
[3.0] Use repository key fingerprints for verification (bsc#988907)

nicolasbock:


vuntz:
Can you use `cherry-pick -x`?

nicolasbock:
@vuntz Is it ok if the cherry-pick copies cloud 7 repository information?

vuntz:
Ah, yeah, hrm, maybe mention the other commit even if it's not a cherry-pick then. The key point is that we want to have the link to the change in master.

vuntz:
+1, but gating is red.

nicolasbock:
Gating fails in `testsetup` in manila tests, i.e. past the point during which this piece of code is used.

aspiers:
+1

PR #553, opened 2016-07-19
Make OpenStack database dump name more generic

rsalevsky:


chef/data_bags/crowbar/template-crowbar.json
```diff
@@ -20,7 +20,7 @@
       },
       "simple_proposal_ui": true,
       "upgrade": {
-        "db_dump_location" : "/var/lib/pgsql/db-before-cloud6.dump"
+        "db_dump_location" : "/var/lib/pgsql/upgrade-openstack-database.dump"
```
sayalilunkad: This name doesn't really make sense. Would be better to use something like openstack-db-pre-upgrade.dump

nicolasbock: Or `openstack-db-before-upgrade.dump`?

sayalilunkad:
+1, thanks!

jsuchome:
+1

nicolasbock:
+1

vuntz:
Don't we need a migration to have the new name used in existing proposals?

nicolasbock:
Yes you are right, I missed that :(

PR #554, opened 2016-07-20
Fix openvswitch for non-x86

dirkmueller:
On any other architecture we just use the kernel provided kmp,
so stop trying to install it


toabctl:
+1

rsalevsky:
+1

PR #555, opened 2016-07-20
Create base of Cloud 7 Upgrade API V2 for crowbar-core

MaximilianMeister:


toabctl:
@MaximilianMeister can you document (at least in the commit msg) why the current api is not enough? or can't be extended?

MaximilianMeister:
> can you document (at least in the commit msg) why the current api is not enough? or can't be extended?

it is documented here https://w3.suse.de/~mmeister/screencast/upgrade-api.webm
but I can add it to the commit message as well if needed

crowbar_framework/app/controllers/api/v2/backups_controller.rb
```diff
@@ -0,0 +1,124 @@
+#
+# Copyright 2015, SUSE LINUX GmbH
```
rsalevsky: 2016

rsalevsky:
We need to prepare crowbarctl for this.

toabctl:
@MaximilianMeister my quesation was more about the difference between api v1 and v2 and if it's possible to just extend v1. introducing a new api version needs also more work on the client side and we need to support both versions for some time I think.

rsalevsky:
We had a long discussion before we started to design this API. The problem is that the new angular application needs a well defined restful API to work. An restful API means that we have entities and resources. To do this with the old API we need to change everything and this would also break a lot. Because of this we decided to create a V2 API where we can work on in parallel without breaking everything else and move the stuff step by step to the new API. (Moving means in this case keep the old logic but change the responses.)

MaximilianMeister:
rebased after https://github.com/crowbar/crowbar-core/pull/556 is merged (i had the same commit in here to test if it works)

MaximilianMeister:
stripped the `/v2` from the url. the version will be sent now through the accept header

jsuchome:
Regarding the API that is needed for the angular app. Are we gonna backport whole V2 API to Cloud6, or just parts of it, or will V1 API be sufficient for Cloud6 part of the upgrade?

kirushik:
Guys, I feel really sorry for coming uninvited to your party, but @santiph made me do so.

I have some improvement proposals for your API versioning scheme. We have some experience vendoring and maintaining versioned APIs in SCC team (and not longer than yesterday I had to implement [this ugly hack](https://github.com/SUSE/connect/pull/271/files)) — so I hope my advice will save you some gray cells in the future.

The thing is, I don't like the proposed `Accept: version=2.0` scheme that much. First of all, let's not ruin the MIME-compatibility of our `Accept` values, the `application/vnd.crowbar.v2+json` will be [standards-compliant](https://en.wikipedia.org/wiki/Media_type#Vendor_tree), mean the same and not interfere with the standard way of doing [content-type negotiation](https://en.wikipedia.org/wiki/Content_negotiation).

The alternative way of doing the same can be a custom header (`Accept-API-Version`?), but I personally would go the more compliant way (who knows how picky your next big JS library be about those custom headers?)

All that being said, I really like your `major.minor` versioning scheme — and I think that you can do even better if you add the semver-like meaning to it.
Because, in my experience, APIs are frequently _extended_ with additional methods, and rarely (but still sometimes) _rewritten_ with a conflicting behaviour of an existing endpoint.

So the smart API maintenance strategy would be keeping the minor part of the counter ticking for every non-conflicting API update, and bump the major version when conflict is unavoidable.
This way, your application can request `Accept: application/vnd.crowbar.v2.3+json`, and get every endpoint which it supposedly knows about — from `2.0`, `2.1`, `2.2` and `2.3` generations. Endpoints from `2.4` and `3.0` will be unavailable, as well as endpoints in `1.7`. If those older ones need to be shared in newer versions, you can always just double-mark them (`api_version "1.7", "2.0"`); conflicting URIs with different behaviours should also be easy this way.

In SCC we have implemented that with class inheritance instead of endpoint tagging — and that caused us some serious pain in the past. This "tagging" solution will allow you to keep your code duplication low, while keeping all the benefits of fully-versioned API.

Another important lesson learnt — your server should always specify the API version served; that's especially important when your API consumer hasn't specified a requirement, and you're defaulting to the most recent version on the server. This also allows clients to detect the API version, and even opens up a road to a full version negotiation.

Imagine that you have an old `v2.3` which requires individual calls for every `foo`, and `v3.1` version which allows batching them. Your client can support both, and calling `Accept: application/vnd.crowbar.v3.1+json, application/vnd.crowbar.v2.3+json; q=0.1` it can indicate both supported types, and `Content-Type: application/vnd.crowbar.v2.3+json` will tell it that this is an old server and iterative approach is necessary. Dumber clients can just request `Accept: application/vnd.crowbar.v3.1+json` and get appropriate `406: Not acceptable` http error back.


What do you think? Will this future-proof solution worth its cost?

MaximilianMeister:
thanks @kirushik for the review, and of course you are more than welcome to the party!

> In SCC we have implemented that with class inheritance instead of endpoint tagging

can you provide an example/link to the endpoint tagging? and do you think we need to tag all the old endpoints (which are not versioned via the header)?

> Another important lesson learnt — your server should always specify the API version served; that's especially important when your API consumer hasn't specified a requirement, and you're defaulting to the most recent version on the server. This also allows clients to detect the API version, and even opens up a road to a full version negotiation.

good point we still need to figure out how we can make our [client](https://github.com/crowbar/crowbar-client) smart enough to handle the new API

crowbar_framework/app/constraints/api_constraint.rb
```diff
@@ -0,0 +1,27 @@
...
+
+class ApiConstraint
+  attr_reader :version
+
+  def initialize(options)
+    @version = options.fetch(:version)
+  end
+
+  def matches?(request)
+    request.headers.fetch(:accept) == "application/vnd.crowbar.v#{version}+json"
```
MaximilianMeister: @kirushik updated

kirushik:
> good point we still need to figure out how we can make our client smart enough to handle the new API

The simpliest solution would be:

 - Every time you implement your client, just rely on the latest minor version available in `master` branch of your server. (Or in the latest branch you have actually shipped/packaged, depending on the speed of your delivery loop.)
 - Make that client output something meaningful when it will receive `406: Not acceptable` from the older server.
 - Your server should extract the version from `Accept:` header and match it against the version it implements. So, server version 2.3 will serve all the requests for 2.1 and 2.3, fallback to some older implementation if 1.7 is requested, and throw `406` if 2.4 or 4.1 are asked for by the client. Just make sure that `Content-Type:` your server returns corresponds to the API version served.

Sometimes later you can extend this approach, teaching your client and server to handle multiple MIME types in `Accept:` and behave correspondingly. This will not break any existing system, just will add an additional resilience layer (so your client will fall back to correctly treating API as `1.7` version if no `2.4` is acceptable). This can not be in your short-term plan, just keep it as an option (maybe just teach your server to treat multi-typed `Accept:` in advance, as servers typically more complicated to update/maintain than the clients).

kirushik:
I like your approach with `ApiConstraint`, actually. I would just extend it a little bit.
As I've said earlier, you should properly treat the major+minor versions in the API, not just do the stupid matching of strings.

Other than that — keep your `v1` APIs under one constraint, keep your `v2` APIs under another. And **don't forget to bump minor version** of each **endpoint addition** to the corresponding block.

So your
```ruby
namespace :api_v2, constraints: ApiConstraint.new(version: 2.5) do
  …
  resources :backups, only: [:index]
  …
end
```
should become
```ruby
namespace :api_v2, constraints: ApiConstraint.new(version: 2.6) do
```
when you add `:show` endpoint to your backups.

And yes, you'll have to start from scratch every time you need to change the format/behavior of your existing ones in some incompatible way. It's not a huge deal, actually — as your 2.x endpoints will still be there, and no existing 2.x client will notice anything.

kirushik:
Please keep in mind that you should be very accurate documenting endpoints to the appropriate versions in your Apipie annotations.

The messier, but easier to maintain option is to trust your apipie annotations to be an ultimate source of the truth of the versioning constraints — and just read that info from your Constraint code, hooking to the `_apipie_dsl_data[:api_versions]` (have a look at [its definition](https://github.com/Apipie/apipie-rails/blob/master/lib/apipie/dsl_definition.rb#L144)).

Btw, why are you specifying your API paths manually every time? `api!` method should use [route integration](https://github.com/Apipie/apipie-rails#rails-routes-integration) to extract that info for you, and make your API definitions less error-prone.

MaximilianMeister:
> Btw, why are you specifying your API paths manually every time? api! method should use route integration to extract that info for you, and make your API definitions less error-prone.

I saw that option too, but when we define a `:GET` and a `:POST` for one route I cannot use it, because the description gets lost and so I tried to be consistent.

crowbar_framework/app/constraints/api_constraint.rb
```diff
@@ -15,13 +15,14 @@
 #
 
 class ApiConstraint
-  attr_reader :version
+  attr_reader :major, :minor
 
   def initialize(options)
```
kirushik: Maybe let's use kwargs here? Also it will allow us to state 1.0 as an obvious default:

```ruby
def initialize(major: 1, minor: 0)
  @major = major
  @minor = minor
end
```

crowbar_framework/app/constraints/api_constraint.rb
```diff
@@ -15,13 +15,14 @@
...
-    @version = options.fetch(:version)
+    @major = options.fetch(:major)
+    @minor = options.fetch(:minor)
   end
 
   def matches?(request)
-    request.headers.fetch(:accept) == "application/vnd.crowbar.v#{version}+json"
+    request.headers.fetch(:accept) == "application/vnd.crowbar.v#{major}.#{minor}+json"
```
kirushik: Still not exactly it.
I would propose 

```ruby
VERSION_MIME = /^application\/vnd\.crowbar\.v(?<major>\d+)\.(?<minor>\d+)\+json$/

def matches?(request)
  versions = VERSION_MIME.match request.headers.fetch(:accept, "")
  return false unless versions

  request_major = versions[:major].to_i
  request_minor = versions[:minor].to_i
  return request_major == major && request_minor <= minor
end
```

crowbar_framework/config/routes.rb
```diff
@@ -238,7 +238,7 @@
     via: [:get, :post, :put, :patch, :delete]
 
   namespace :api,
-    constraints: ApiConstraint.new(version: 2.0) do
+    constraints: ApiConstraint.new(major: 2, minor: 0) do
```
kirushik: BTW, are you tagging your `v1` APIs? Should they be a part of `v2` API namespace as well?

Maybe `ApiConstraint.new(1=>4, 2=>0)` will be a better API? It will allow you to tag an endpoint/namespace as a part of several major versions...

kirushik:
> I saw that option too, but when we define a :GET and a :POST for one route I cannot use it, because the description gets lost and so I tried to be consistent.
OK, makes sense.

crowbar_framework/config/routes.rb
MaximilianMeister: > BTW, are you tagging your v1 APIs? Should they be a part of v2 API namespace as well?

that was one of my previous questions, maybe you have missed it. In fact we would not like to namespace them to not break anything. We haven't had any versioning yet unfortunately... not sure what is the best practice in this case

> Maybe ApiConstraint.new(1=>4, 2=>0) will be a better API? It will allow you to tag an endpoint/namespace as a part of several major versions...

goog idea. I have solved that now with being able to provide multiple versions via an array

crowbar_framework/app/constraints/api_constraint.rb
MaximilianMeister: `VERSION_MIME = /^application\/vnd\.crowbar\.v(?<major>\d+)\.(?<minor>\d+)\+json$/`

In ruby you can't use variables in constants :-( but with lower case it works.
I have adapted the code a bit with many parts of your proposal, thanks for the hints!

crowbar_framework/app/constraints/api_constraint.rb
```diff
@@ -15,13 +15,25 @@
...
-    @version = options.fetch(:version)
+  def initialize(*versions)
+    @versions = versions
   end
 
   def matches?(request)
-    request.headers.fetch(:accept) == "application/vnd.crowbar.v#{version}+json"
+    allowed = false
+    major, minor = nil
+    versions.each do |version|
```
kirushik: `versions.any?` is slightly better — no explicit return variable, and also won't check next `version` values once the match is there.

crowbar_framework/app/constraints/api_constraint.rb
```diff
@@ -15,13 +15,25 @@
...
+      major, minor = version.to_s.split(".").map(&:to_i)
```
kirushik: This splitting should really go into the `initialize`.

```ruby
@versions = versions.map{|v| v.to_s.split(".").map(&:to_i)}
```

and then in `matches?`:

```ruby
versions.any? do |major, minor|
  versions_requested = version_mime.match(request.accept)
  versions_requested[:major].to_i == major && versions_requested[:minor].to_i <= minor
end
```

crowbar_framework/app/constraints/api_constraint.rb
```diff
@@ -15,13 +15,25 @@
...
+      version_mime = %r(^application/vnd\.crowbar\.v(?<major>\d+).(?<minor>\d+)\+json$)
+
+      versions_requested = version_mime.match(request.accept)
+      allowed = true if versions_requested &&
+          versions_requested[:major].to_i == major &&
+          versions_requested[:minor].to_i == minor
```
kirushik: Hey, having `==` here is exactly **not** the point!
I expect API v2.4 to be _superset_ of API v2.3, so when I ask for 2.3-compatible response, 2.4 will do!

(Also, don't forget to indicate the resulting version in the response headers.)

crowbar_framework/app/constraints/api_constraint.rb
kirushik: I'm not sure what you mean here. There are no variables in this regex, only named captures.
So it should be pretty constant-friendly.

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -15,6 +15,8 @@
 #
 
 class Api::UpgradeController < ApplicationController
+  before_action :set_content_type
```
kirushik: Maybe it's a proper time to abstract this thing away in `ApiController < ApplicationController` somehow, and subclass from that `ApiContoller` everywhere?

This `before_action` everywhere seems to be very easy to forget at some point...

kirushik:
> We haven't had any versioning yet unfortunately...
Than any request without version header should be treated as 1.0 by your matcher...

This is the point after which it gets really messy without the tests. Do you need any help in setting them up for this PR?

crowbar_framework/app/constraints/api_constraint.rb
MaximilianMeister: :+1:  much more elegant

crowbar_framework/app/constraints/api_constraint.rb
MaximilianMeister: the ruby compiler complains somehow

crowbar_framework/app/constraints/api_constraint.rb
MaximilianMeister: oops typo. thanks

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: yes, makes sense

crowbar_framework/app/constraints/api_constraint.rb
kirushik: Oh, just `(?<minor>)` doesn't use any external information. It's just a way to tell `match` to extract this capture value into the `:minor` named group. It's not equivalent of `#{minor}` of some sort, that would indeed be bad.

I've just tried, and this worked for me in `pry`:
```ruby
class Foo
  VERSION_MIME= %r(^application/vnd\.crowbar\.v(?<major>\d+).(?<minor>\d+)\+json$)
end

pry(main)> Foo::VERSION_MIME
=> /^application\/vnd\.crowbar\.v(?<major>\d+).(?<minor>\d+)\+json$/
```

MaximilianMeister:
> This is the point after which it gets really messy without the tests. Do you need any help in setting them up for this PR?

you mean API tests with Rspec?

> Than any request without version header should be treated as 1.0 by your matcher...

remember we don't have any constraint on the old (1.0) routes, so is this really a problem now?

kirushik:
> you mean API tests with Rspec?
Yes, exactly. I think [request specs](https://www.relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec) will give you a right level of abstraction here — they're fast enough, but allow you to test routing, controllers logic and request headers stuff simultaneously.

kirushik:
> remember we don't have any constraint on the old (1.0) routes, so is this really a problem now?
It is not a problem right now, but how are you going to _not expose_ those endpoints for the API v3 if needed?
You cannot just default and say that those endpoints are going to be included in every future API version, no matter what — it would be unwise.

That's why I advocate to just sit down, try to figure out what are your API constraints should be, and write some request specs to cover all the cases.

Otherwise you might end up with an API scheme which will do for now, but will be completely unextensible in the short-range future.

MaximilianMeister:
> That's why I advocate to just sit down, try to figure out what are your API constraints should be, and write some request specs to cover all the cases.

I would say sure, right now would be a good place to start with TDD, as we havent implemented it yet. after checking with the team, we are not sure if we can make it, as we still have lots to do for the upgrade., and the time is running out a bit. can you estimate how much time that would actually delay the initial work here? we definitely would invest the time

we also have some [old specs](https://github.com/crowbar/crowbar-core/tree/master/crowbar_framework/spec/controllers) for the 1.0 API, not sure if those would need to be adapted as well, or if we write a complete new testsuite

kirushik:
> can you estimate how much time that would actually delay the initial work here?

Well, I haven't even cloned the repo — but it looks like you have Travis already set up and running on every PR. So I wouldn't expect much troubles on this front.

If you're still concerned, I can push a commit to this branch, it will just create one stubbed request spec for your specific case — so you'll be able to concentrate on just writing the cases themselves.

kirushik:
@MaximilianMeister So, I've just created this file called `crowbar_framework/spec/requests/api_versioning_spec.rb`:

```ruby
require 'spec_helper'

describe 'API versioning', type: :request do
  context 'client version 1.0' do
    let(:headers) {{ ACCEPT: 'application/vnd.crowbar.v1.0+json' }}

    it "shouldn't do a thing" do
      expect { get '/api/crowbar', {}, headers }.to raise_error ActionController::RoutingError
    end
  end

  context 'client version 2.0' do
    let(:headers) {{ ACCEPT: 'application/vnd.crowbar.v2.0+json' }}

    it 'should do something' do
      get '/api/crowbar', {}, headers
      expect(response).to have_http_status(:not_implemented)
      expect(response.content_type).to eq('application/vnd.crowbar.v2.0+json')
    end
  end
end
```

It works exactly as it should, no additional configuration required.

aspiers:
@kirushik Thanks a LOT for your awesome advice and help here! I'm not a REST API expert, but it looks like this is moving along very nicely.

@MaximilianMeister you wrote:
> I would say sure, right now would be a good place to start with TDD, as we havent implemented it yet. after checking with the team, we are not sure if we can make it, as we still have lots to do for the upgrade., and the time is running out a bit. can you estimate how much time that would actually delay the initial work here? we definitely would invest the time

I totally sympathise with the concern about lack of time. If it's any comfort my experience is that TDD makes development go *quicker* not slower. I know it's counter-intuitive but the break-even point is almost always far sooner than you would expect. So developing without tests might save you a few hours in the first day or two, but after that you lose bigger and bigger by not having them. One reason is that the longer bugs go undiscovered, the harder they are to fix. Tests help discover bugs extremely early, and of course prevent them ever reoccurring too.

crowbar_framework/config/routes.rb
```diff
@@ -240,4 +236,48 @@
   # TODO(must): Get rid of this wildcard route
   match "/:controller/:action/*(:.format)",
     via: [:get, :post, :put, :patch, :delete]
+
+  namespace :api,
```
aspiers: Is it really necessary to put the API in a separate namespace? Won't the format and/or `Accept` header be enough to distinguish API requests from UI requests? What does the rest of the Rails community typically do here? AFAIK they don't tend to use `/api/` but ICBW.

MaximilianMeister: one problem that we have is the `backups_controller`.
the old one handles UI requests, and the new one json API requests. and if we want to use both, we somehow have to namespace them.

but i think overall you see this `/api` namespacing a lot when you google for it.

not sure how difficult would be to add an `api.crowbar.com` subdomain? that would be nice too...

MaximilianMeister: from http://guides.rubyonrails.org/routing.html#request-based-constraints

```ruby
get 'photos', to: 'photos#index', constraints: { subdomain: 'admin' }
```

kirushik: @aspiers I think `:api` namespace is totally OK; at least that's what we [do in SCC](https://github.com/SUSE/happy-customer/blob/master/glue/config/routes/api.rb) as well.

If your API mostly follows your UI, reusing the same routing scheme helps. If it's some completely separate beast with a different usage patters (which is the case for SCC's `:connect` and `"import` APIs, and I think for Crowbar APIs too) — it doesn't help that much, just bloats your controllers.

@MaximilianMeister There's a tiny subtlety with the subdomain approach — if you're doing HTTPS on your application (and you probably should), this will require a second certificate. Moreover, in some circumstances that certificate can be unobtainable at all (because it's one level deeper, and some issuers'/clients' policies don't allow that).

crowbar_framework/app/constraints/api_constraint.rb
```diff
@@ -0,0 +1,32 @@
...
+
+class ApiConstraint
+  attr_reader :versions
+
+  def initialize(*versions)
+    @versions = versions.map { |v| v.to_s.split(".").map(&:to_i) }
+  end
+
+  def matches?(request)
+    versions.any? do |major, minor|
```
kirushik: Ruby trivia question: what will happen here if `versions` array is empty?

That's one of those places where tests truly shine.

crowbar_framework/config/routes.rb
aspiers: @kirushik OK thanks, that makes perfect sense and aligns with what I was thinking regarding controller reuse. @MaximilianMeister in this context, how much commonality is there between the UI and API regarding usage patterns and controller code?

MaximilianMeister: @aspiers The idea behind the split is to completely get rid of the `ui/backups_controller` once we switch to the Angular UI, and from then on only use the `api/backups_controller`. If we wouldn't use the Rails views anymore we could simply drop it

crowbar_framework/app/constraints/api_constraint.rb
MaximilianMeister: when the versions array is empty, the matcher returns false

kirushik: Yes, indeed.
And for me it's not the most obvious thing.
What is even less obvious is that `[].all?(&:foo?)` returns `true`.

So I don't rely on myself being able to reason about those dark corners of the code, and use testing instead. That's where I'm pointing your attention here.

crowbar_framework/spec/controllers/api/backups_controller_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+    let(:headers) { { ACCEPT: "application/vnd.crowbar.v2.0+json" } }
+
+    it "lists all crowbar backups" do
+      get "/api/crowbar/backups", {}, headers
+      expect(response).to have_http_status(:ok)
+      expect(response.body).to eq("[]")
+    end
+
+    it "creates a backup" do
+      stub_methods
```
kirushik: Do you really need to call it here explicitly here if you're using the _eager_ `let!` variant? (`let` is lazy, evaluated when first used; `let!` is eager, evaluated right when defined.)

Also, why do you need that `let!` at all, why not to use `before(:each)` block? If you want opt-in behavior, just wrap your examples subset in `context` block, and use `before` inside that context.

The reasoning here goes like this: it's much easier to reason about your tests when all the boilerplate is abstracted away, and only the actual moving parts are changing.

Do you want me to continue nitpicking your specs, or you're fine with how it is now, and I'm only an obstacle for you to close the current task and move on?

crowbar_framework/app/constraints/api_constraint.rb
MaximilianMeister: I think the example is not the best, or why is it not obvious to you? ´any?´ returns either ´true´ or ´false´ like any method with a  ´?´ at the end - at least by ruby convention.

crowbar_framework/spec/controllers/api/backups_controller_spec.rb
MaximilianMeister: > why not to use before(:each) block? If you want opt-in behavior, just wrap your examples subset in context block, and use before inside that context.

in this case a `before(:each)` makes more sense, agreed...

> Do you want me to continue nitpicking your specs, or you're fine with how it is now, and I'm only an obstacle for you to close the current task and move on?

depends if you have the time, and you are not an obstacle at all, on the contrary.
The tests will anyways change in the actual implementation phase of the API, right now this is "just" the base. and I'd definitely prefer to have a good base to start with.

crowbar_framework/app/constraints/api_constraint.rb
kirushik: OK, that might be just me and my mental model.
Sorry for starting this thread, it's completely irrelevant to the PR at question.

MaximilianMeister: > OK, that might be just me and my mental model.
> Sorry for starting this thread, it's completely irrelevant to the PR at question.

@kirushik no problem, I got the point though, which was to cover all areas of the code if possible.

are you OK with the changes now, or is there anything critical you would do differently?

kirushik: Yep, it's so much better than before, good job!
The duplication between the apipie versions annotations and routing version constraints (and request-level routing tests) still bothers me — but that's more of a long-term pain for you to think about, not something which can be easily fixed in a short term.

LGTM from my side.

MaximilianMeister: thanks again for all of the input!

mjura:
I think, we should add some authentication mechanism, but this can be done in another PR
+1

rsalevsky:
+1

rsalevsky:
Needs to go with https://github.com/crowbar/crowbar-client/pull/92

kirushik:
@mjura While we are at the topic of Auth, can I recommend a [JWT-based approach](https://www.sitepoint.com/introduction-to-using-jwt-in-rails/)?

It's stateless, fast, and easy to implement on both your [client](https://github.com/auth0/angular-jwt) and [server](https://github.com/eigenbart/rack-jwt).

You can issue the tokens from your existing Rails API (manually with plain [ruby-jwt gem](https://github.com/jwt/ruby-jwt), or [Doorkeeper](https://github.com/chriswarren/doorkeeper-jwt) Rails extension), or choose from [several](https://github.com/Diasporism/jwt_rails) [existing](https://github.com/robmclarty/jwt-api-server) [servers](https://github.com/lucianw/ldap-jwt) just for auth.

MaximilianMeister:
> Your client can support both, and calling Accept: application/vnd.crowbar.v3.1+json, application/vnd.crowbar.v2.3+json; q=0.1 it can indicate both supported types, and Content-Type: application/vnd.crowbar.v2.3+json will tell it that this is an old server and iterative approach is necessary

@kirushik 

how can I best prevent creating a `500` error on the server when I use an iterative approach? or is it wanted to create an error?

from the client: first I try to request the old route `utils/backups`, and when that fails with `500` I use `api/crowbar/backups`.
can I catch this on the server side somehow to prevent logging the `500` errors during thes iterations? like sending a header which indicates the iteration phase, or what would you do here?

kirushik:
@MaximilianMeister I'm afraid I didn't understand your problem fully.

If your API consumer is newer than your API-providing server, then you should be expecting to get an error — because you're obviously relying in your logic on some features that are not there on server. You can use those errors to indicate to your customers that they need to update the server.

And after that maybe try asking for an older API version as a fallback, if it already supports what you want to do.

If we make our API Constraint a little bit smarter (multiple `q`-weighted options in the `Accept` headers of the request), it will allow you to _negotiate_ the API version with server in a single roundtrip, without error detection and re-querying on client side. But that will take additional effort, and I don't know if you can afford that right now.

So the solution "Older clients work with newer servers, but newer clients ask politely to update your server" seems to be good enough for me, for now at least. (Once again, I don't know the full list of crowbar's requirements and constraints.)

jsuchome:
Gating failse because of missing https://github.com/crowbar/crowbar-openstack/pull/499, I think

rsalevsky:
+1

PR #556, opened 2016-07-21
Disable haml rubocops for now

rsalevsky:
Hound's haml linter does not support to change the default rubocop
config. This means that we will have always false positives e.g. single
quotes instead of double quotes.

See: https://github.com/houndci/hound/issues/1121

### DIFF
--- a/.hound.haml.yml
+++ b/.hound.haml.yml
@@ -1,5 +1,6 @@
 linters:
   LineLength:
     max: 150
-  UnnecessaryInterpolation:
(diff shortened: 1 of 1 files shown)


MaximilianMeister:
+1

skazi0:
+1, in theory we can do some more fine-grained disabling of the RuboCop checks but since we don't know the exact config used on production Hound service and this change is supposed to be only temporary, we can go with this for now.

PR #557, opened 2016-07-21
crowbar: Do not use non-existing attributes in bulk edit (bsc#989994)

vuntz:
We removed some attributes from the bulk edit page, but the controller
was still using them. This resulted in the bulk edit page resetting the
groups of the nodes.

Bug was introduced in https://github.com/crowbar/crowbar-core/pull/353 /
e4d575e8.

https://bugzilla.suse.com/show_bug.cgi?id=989994


rsalevsky:
+1

dirkmueller:
+1

PR #560, opened 2016-07-21
crowbar: Fix queue message when applying a proposal (bsc#989958)

vuntz:
There was an attempt to fix the internal API so that
ServiceObject.apply_role always returns [int, str], but this actually
breaks the UI-only path in the controller.

So revert this and document with a FIXME that this should eventually be
fixed by moving everything to [int, hash] where we will have more
freedom.

https://bugzilla.suse.com/show_bug.cgi?id=989958

### DIFF
--- a/crowbar_framework/app/controllers/barclamp_controller.rb
+++ b/crowbar_framework/app/controllers/barclamp_controller.rb
@@ -517,6 +517,8 @@ class BarclampController < ApplicationController
           "crowbar-deep-merge-template"
         )
       )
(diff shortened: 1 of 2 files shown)


dirkmueller:
if thats a revert, can we add the "This reverts commit XXXX" message ?

dirkmueller:
+1

toabctl:
+1 (but gating fails)

dirkmueller:
gating failure seems unrelated and is everywhere broken. merging..

PR #561, opened 2016-07-22
crowbar: On restore, only restore roles that are for proposals and nodes

vuntz:
We do not want to restore roles such as provisioner-base, as the role
should be considered as code and not data, and it might be outdated in
the backup.

### DIFF
--- a/crowbar_framework/lib/crowbar/backup/base.rb
+++ b/crowbar_framework/lib/crowbar/backup/base.rb
@@ -100,6 +100,31 @@ module Crowbar
         def export_files
           restore_files.concat(restore_files_after_install).map(&:reverse)
         end
+
+        def filter_chef_node(name)
+          false
+        end
+        alias_method :filter_chef_nodes, :filter_chef_node
+
+        def filter_chef_role(name)
+          # Filter roles which are not crowbar roles (proposal + node roles)
+          # as they are code, not data
+          # The regexp for node roles is not great, but we know we have a _
+          # due to the domain, so it should be good enough
+          name !~ /(.+-config-.+)|(^crowbar-.+_.+)/
+        end
+        alias_method :filter_chef_roles, :filter_chef_role
+
+        def filter_chef_client(name)
+          # Filter client "crowbar" as we need the new one on restore anyway
+          name =~ /^crowbar$/
+        end
+        alias_method :filter_chef_clients, :filter_chef_client
+
+        def filter_chef_databag(db, name)
+          false
+        end
+        alias_method :filter_chef_databags, :filter_chef_databag
       end
     end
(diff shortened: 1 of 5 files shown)


vuntz:
So I believe we should also have similar code when creating the backup...

crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -191,6 +191,11 @@ def restore_chef
               file = Pathname.new(file)
               # skip client "crowbar"
               next if type == :clients && file.basename.to_s =~ /^crowbar.json$/
+              # skip roles which are not crowbar roles (proposal + node roles)
+              # the regexp for node roles is not great, but we know we have a _
+              # due to the domain, so it should be good enough
+              next if type == :roles &&
+                  file.basename.to_s !~ /.+-config-.+/ && file.basename.to_s !~ /^crowbar-.+_.+/
```
MaximilianMeister: I think you could put it into one regex
```ruby
file.basename.to_s !~ /(.+-config-.+)|(^crowbar-.+_.+)/
```

vuntz: Good point, done

vuntz:
Updated to filter on backup creation too.

crowbar_framework/lib/crowbar/backup/base.rb
```diff
@@ -100,6 +100,29 @@ def restore_files_after_install
...
+
+        def filter_chef_client(name)
+          # Filter client "crowbar" as we need the new one on restore anyway
+          name =~ /^crowbar$/
+        end
+
+        def filter_chef_databag(db, name)
+          # Filter items in crowbar data bag that are not network databag items
+          return false if db != "crowbar"
+          item !~ /.+_network$/
```
MaximilianMeister: s/item/name

MaximilianMeister:
codewise +1 (triggered an upgrade gating too)

vuntz:
I added the patch from https://github.com/crowbar/crowbar-core/pull/562 to help gating.

jsuchome:
+1 (tested)

MaximilianMeister:
@vuntz once gating is unbroken and this is rebased, can you run another upgrade gating here?

MaximilianMeister:
@vuntz I triggered the gating for the upgrade, I hope it works :) even though the last commit hash starts with "bad" :-(  https://github.com/crowbar/crowbar-core/pull/561/commits/bad8795df655872f37844167649b97be07324d09

MaximilianMeister:
+1 we cannot gate the upgrade, because the admin gets reinstalled...

PR #562, opened 2016-07-22
ipmi, provisioner: Fix migrations for new roles

vuntz:
We also need to update the node roles for nodes that are using these
roles.


jsuchome:
+1 after testing

rsalevsky:
The gating failure looks related.

vuntz:
@rsalevsky we also need the patch from https://github.com/crowbar/crowbar-core/pull/561, so we're doing a combined gating there.

vuntz:
Patch was merged as part of https://github.com/crowbar/crowbar-core/pull/561

PR #563, opened 2016-07-22
set maintenance mode or stop corosync/pacemaker on update

Itxaka:
As per all the docs by pacemaker, the services should be stopped for a software update of the cluster stack.  If other packages are being updated but not the cluster stack, then it's enough to set maintenance mode.

**FIXME: however if the update is happening on a remote node, then the `crm` command is unavailable so we need to set maintenance mode another way.** **UPDATE:** something like https://github.com/crowbar/crowbar-ha/pull/187 might fix this.

Depends on ~~crowbar/crowbar-ha#183~~

### DIFF
--- a/chef/cookbooks/updater/recipes/default.rb
+++ b/chef/cookbooks/updater/recipes/default.rb
@@ -17,6 +17,9 @@
 # limitations under the License.
 #
 
+::Chef::Recipe.include CrowbarPacemaker::MaintenanceModeHelpers
+::Chef::Resource.include CrowbarPacemaker::MaintenanceModeHelpers
+
 if !node[:updater].key?(:one_shot_run) || !node[:updater][:one_shot_run]
 
   node[:updater][:one_shot_run] = true
@@ -43,6 +46,65 @@ if !node[:updater].key?(:one_shot_run) || !node[:updater][:one_shot_run]
     execute "refresh PTF repository" do
       command "zypper --non-interactive --gpg-auto-import-keys refresh -fr PTF"
       ignore_failure true
+    end
+
+    ruby_block "check for updates" do
+      block do
+        case node[:updater][:zypper][:method]
+        when "patch"
+          command = "list-patches"
+        when "update"
+          command = "list-updates"
+        when "dist-upgrade"
+          command = "list-updates"
+        end
+
+        node.run_state["needs_update"] = `zypper -q #{command}|wc -l`.chomp.to_i > 0
+
+        command += '|egrep -q "corosync|pacemaker"'
+        system("zypper #{command}")
+        # exit 0: found, 1 not found
+        node.run_state["found_ha_packages"] = $?.exitstatus ? true : false
+      end
+    end
+
+    ["corosync", "pacemaker"].each do |s|
+      service s do
+        action :stop
+        only_if { node.run_state["found_ha_packages"] }
+        not_if { node[:pacemaker] && node[:pacemaker][:is_remote] }
+      end
+    end
+
+    # set cluster to maintenance if
+    # HA packages are NOT gonna be updated
+    # And Node is part of a cluster
+    # And there is packages to update
+    execute "crm --wait node maintenance" do
+      action :nothing
+      notifies :create, "ruby_block[set maintenance mode via this chef run]", :immediately
+    end
+
+    ruby_block "set maintenance mode via this chef run" do
+      action :nothing
+      block do
+        set_maintenance_mode_via_this_chef_run
+      end
+    end
+
+    ruby_block "set cluster maintenance" do
+      block do
+        Chef::Log.info("Triggering maintenance mode for this node")
+        true
+      end
+      only_if do
+        is_cluster = node.role? "pacemaker-cluster-member"
+        !node.run_state["found_ha_packages"] && is_cluster && node.run_state["needs_update"]
+      end
+      not_if do
+        maintenance_mode_set_via_this_chef_run? && maintenance_mode?
+      end
+      notifies :run, "execute[crm --wait node maintenance]", :immediately
     end
 
     # Butt-ugly, enhance Chef::Provider::Package::Zypper later on...
@@ -94,7 +156,13 @@ if !node[:updater].key?(:one_shot_run) || !node[:updater][:one_shot_run]
           end # case
         end # while
       end # block
+      only_if { node.run_state["needs_update"] }
     end # ruby_block
+
+    service "pacemaker" do
+      action :start
+      not_if { node[:pacemaker] && node[:pacemaker][:is_remote] }
+    end
 
   end # platform_family suse block
 
--- a/chef/cookbooks/updater/metadata.rb
+++ b/chef/cookbooks/updater/metadata.rb
@@ -22,5 +22,6 @@ long_description ""
 version "0.0.1"
 
 depends "utils"
+depends "crowbar-pacemaker"
 
 recipe "updater", "System Package Updater"


chef/cookbooks/updater/recipes/default.rb
```diff
@@ -40,6 +40,31 @@
...
+        case node[:updater][:zypper][:method]
+        when "patch"
+          command = "zypper list-patches"
+        when "update"
+          command = "zypper list-updates"
+        when "dist-upgrades"
+          # TODO: check the correct one for here
+          command = "zypper list-updates"
+        end
+        command += '|egrep "corosync|pacemaker" -q'
```
aspiers: According to the `grep` man page, `-q` should come before the pattern, not after.  Even if it's not necessarily, it's more legible because it's what people expect.

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -40,6 +40,31 @@
...
+        `#{command}`
+        node.run_state["found_ha_packages"] = $?.exitstatus
+      end
+      not_if { search(:node, "roles:pacemaker-cluster-member").empty? }
+    end
+
+    ["corosync", "pacemaker"].each do |s|
```
aspiers: Even though it seems necessary to do this at the moment, it shouldn't be.  I've filed [bsc#991989 (pacemaker.service should have BindsTo=corosync.service)](https://bugzilla.suse.com/show_bug.cgi?id=991989) for this.

aspiers: No, I was wrong.  It's enough to stop corosync, [as you already noted](https://trello.com/c/13FOsgtH/13-bug-983617-failed-to-apply-updater-barclamp-when-stopping-pacemaker#comment-578f9d8897f261607988498c).

aspiers:
Let's move the remaining development discussion from Trello to here.  So @Itxaka [wrote](https://trello.com/c/13FOsgtH/13-bug-983617-failed-to-apply-updater-barclamp-when-stopping-pacemaker#comment-5795dc5ced25dd23c835018c):

> Stuck in here. Updater recipe has been adapted to take into account cluster and stop the services but there is some need for sync between nodes. If not, it could happen that corosync/pacemaker is stopped in the founder while another node needs to check against the founder node, resulting in a chef failure.

What kind of failure exactly?  Other nodes synchronise against the founder using Chef, not Pacemaker, so I don't understand the problem.

The other thing missing from this PR is that if no HA packages are being updated, [we need to place the cluster in maintenance mode](https://www.suse.com/documentation/sle-ha-12/book_sleha/data/sec_ha_migration_update.html), or at least the current node.

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -40,6 +40,35 @@
...
     end
 
+    ha_enabled = search(:node, "roles:pacemaker-cluster-member")
+    if ha_enabled
+      case node[:updater][:zypper][:method]
+        when "patch"
+          command = "zypper list-patches"
+        when "update"
+          command = "zypper list-updates"
+        when "dist-upgrades"
```
aspiers: Why is this warning happening? It looks fine to me!

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -40,6 +40,31 @@
...
+          # TODO: check the correct one for here
+          command = "zypper list-updates"
+        end
+        command += '|egrep "corosync|pacemaker" -q'
+        `#{command}`
```
aspiers: I'm not keen on this approach to executing the command.  Backquotes are supposed to be used when the output of the command needs to be captured, but here we are just throwing it away, so `system` would be more appropriate.

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -91,6 +116,12 @@
       end # block
     end # ruby_block
 
+    ["corosync", "pacemaker"].each do |s|
```
aspiers: It's enough to start just `pacemaker` here.

chef/cookbooks/updater/recipes/default.rb
Itxaka: done

chef/cookbooks/updater/recipes/default.rb
Itxaka: done

chef/cookbooks/updater/recipes/default.rb
Itxaka: If we stop corosync on line 61, then we should start corosync in here as well no?

Itxaka:
> What kind of failure exactly? Other nodes synchronise against the founder using Chef, not Pacemaker, so I don't understand the problem.

Unfortunately I havent been able to reproduce it again, but once during the update when the serviecs for HA were stopped, another node did something with its pacemaker config and failed because it could not contact the founder.

node 1: founder, updating, corosync+pacemaker stopped
node 2: not updating, tried to contact the founder, failed.

It may be that I misinterpreted the error or that it had nothing to do with the update, but I cannot confirm it as I cannot reproduce it again.

Itxaka:
> The other thing missing from this PR is that if no HA packages are being updated, we need to place the cluster in maintenance mode, or at least the current node.

Good point, Im guessing that this is due to services potentially restarting. Will update the pr.

chef/cookbooks/updater/recipes/default.rb
aspiers: It will be a no-op, because of the `Requires=corosync.service` dependency. But if you prefer to be safe (or support other distros which might be missing it), you can keep it, in which case I think there should be a comment explaining that it should be unnecessary and we're just playing safe.

aspiers:
> another node did something with its pacemaker config and failed because it could not contact the founder.

I can definitely believe that happened, but it would not be directly caused by Pacemaker/corosync being stopped. The founder synchronization code relies on synchronization marks (revision numbers) which are set by chef-client according to the proposal revision, and it checks them via the Chef server. So unless another proposal was changed, it should already be synced cluster-wide and wouldn't require the founder chef-client to complete.

aspiers:
(See slides 69+ of [the HA training](https://w3.suse.de/~aspiers/cloud/HA-training/HA-training-all-slides.pdf) for more on this.)

Itxaka:
Pushing what I have as Im moving to work on the https://github.com/crowbar/crowbar-ha/pull/135 instead.

aspiers:
@Itxaka Any chance you could fix the hound violations, and also reference [bsc#983617 (Failed to apply Updater barclamp when stopping pacemaker)](https://bugzilla.suse.com/show_bug.cgi?id=983617) in the commit message?  The correct style is to add

    (bsc#983617)

to the end of the first line of the commit message, and then add the full URL as a separate line within the body of the commit message. Thanks!

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -17,6 +17,21 @@
...
 
+include CrowbarPacemaker::MaintenanceModeHelpers
+
+def set_maintenance_mode
+  if maintenance_mode_set_via_this_chef_run?
+    Chef::Log.info("chef-client run pid #$$ already placed this node in Pacemaker maintenance mode")
+  elsif maintenance_mode?
+    Chef::Log.info("Something else already placed this node in Pacemaker maintenance mode")
+  else
+    execute "crm --wait node maintenance" do
```
vuntz: @aspiers isn't this working only for corosync nodes and not for remote nodes?

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -40,6 +55,51 @@
...
     end
 
+    ruby_block "check for updates" do
+      block do
+        case node[:updater][:zypper][:method]
+        when "patch"
+          command = "list-patches"
+        when "update"
+          command = "list-updates"
+        when "dist-upgrades"
```
vuntz: `dist-upgrade` with no s

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -40,6 +55,51 @@
...
+          # TODO: check the correct one for here
+          command = "list-updates"
```
vuntz: I think you can remove the TODO, as it's the right command, I believe.

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -89,8 +149,15 @@
           end # case
         end # while
       end # block
+      only_if { node.run_state["needs_update"] }
     end # ruby_block
 
+    ["corosync", "pacemaker"].each do |s|
```
vuntz: Nearly sure we're not supposed to manually start corosync nowadays.

chef/cookbooks/updater/recipes/default.rb
Itxaka: done

chef/cookbooks/updater/recipes/default.rb
Itxaka: done

Itxaka:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;13FOsgtH&#x2F;13-p2-bug-983617-failed-to-apply-updater-barclamp-when-stopping-pacemaker"> P2 Bug 983617 - Failed to apply Updater barclamp when stopping pacemaker</a>

Itxaka:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;13FOsgtH&#x2F;13-p2-bug-983617-failed-to-apply-updater-barclamp-when-stopping-pacemaker"> P2 Bug 983617 - Failed to apply Updater barclamp when stopping pacemaker</a>

chef/cookbooks/updater/recipes/default.rb
aspiers:
```
2016-02-02 22:32:48     vuntz   ygao: one other thing I'm wondering: I'm not able to put the remote nodes in maintenance; do you know of a way to do that?
2016-02-02 22:37:59     vuntz   hrm, that works in hawk
2016-02-02 22:38:08     vuntz   there must be a way to make it work from the cli
2016-02-02 22:42:20     vuntz   oh
2016-02-02 22:42:22     vuntz   it works now
2016-02-02 22:42:28     vuntz   I guess your fixes helped with that too
```

I'm not sure what CLI you did it with though :)

aspiers: @Itxaka No, I think it means that @vuntz needs to remember how he did it :) Or we can simply ask on `#sleha` IRC, which would be quicker.

aspiers: Indeed it was quicker :-)  Yan says it's done in the same way as for the corosync nodes, i.e. via `crm node maintenance $remote_node`.

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -17,6 +17,21 @@
 # limitations under the License.
 #
 
+include CrowbarPacemaker::MaintenanceModeHelpers
+
+def set_maintenance_mode
```
aspiers: I just noticed this function is copy-n-pasted from https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/providers/service.rb#L221 and IMHO that's not right because

1. it's a DRY violation, and
2. Pacemaker-specific code must live in `crowbar-ha` not this repo.

Itxaka:
Going with Adam comment, I need to expose that function to other cookbooks first: https://github.com/crowbar/crowbar-ha/pull/183

chef/cookbooks/updater/recipes/default.rb
aspiers: Hmm, but that only works when run from the corosync nodes; the remote nodes don't even have `crm` installed (and even if they did, I'm not sure if `crm node maintenance` without the extra node name argument would work).

aspiers:
@Itxaka I think something like https://github.com/crowbar/crowbar-ha/pull/187 should fix the remote case. Any chance you could test it?

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -40,6 +42,50 @@
...
+    # And Node is part of a cluster
+    # And there is packages to update
+    ruby_block "set cluster maintenance" do
+      block do
+        set_maintenance_mode
+      end
+      only_if do
+        ha = node.run_state["found_ha_packages"]
+        is_cluster = !search(:node, "roles:pacemaker-cluster-member").empty?
+        return !ha && is_cluster && node.run_state["needs_update"]
```
Itxaka: dont return here

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -89,8 +135,13 @@
           end # case
         end # while
       end # block
+      only_if { node.run_state["needs_update"] }
     end # ruby_block
 
+    service "pacemaker" do
+      action :start
+    end
```
Itxaka: not if is a remote node

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -40,6 +42,50 @@
...
+        # exit 0: found, 1 not found
+        node.run_state["found_ha_packages"] = $?.exitstatus ? true : false
+      end
+      not_if { search(:node, "roles:pacemaker-cluster-member").empty? }
+    end
+
+    ["corosync", "pacemaker"].each do |s|
+      service s do
+        action :stop
+        only_if { node.run_state["found_ha_packages"] }
```
Itxaka: dont stop anything if its a remote node

Itxaka:
needs rebase + adam patch + some changes around

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -40,6 +42,50 @@
...
+        when "dist-upgrade"
+          command = "list-updates"
+        end
+        system("zypper -q #{command}")
+        # exit 0: no updates, 1: updates
+        node.run_state["needs_update"] = $?.exitstatus ? false : true
+
+        command += '|egrep -q "corosync|pacemaker"'
+        system("zypper #{command}")
+        # exit 0: found, 1 not found
```
Itxaka: This is wrooong! Comment AND check is reversed!
```
root@d52-54-77-77-01-01:~ # zypper lu|egrep "corosync|pacemaker"
root@d52-54-77-77-01-01:~ # echo $?
1
root@d52-54-77-77-01-01:~ # zypper lu|egrep "corosync|bind"
v | SLES12-SP2-Updates-test | bind-libs  | 9.9.9P1-56.1    | 9.9.9P1-59.1      | x86_64
v | SLES12-SP2-Updates-test | bind-utils | 9.9.9P1-56.1    | 9.9.9P1-59.1      | x86_64
root@d52-54-77-77-01-01:~ # echo $?
0
```

Itxaka: wait, its not, then something is not working properly in this check, to revise further!

chef/cookbooks/updater/recipes/default.rb
Itxaka: This does not seem to work properly.

The services got stopped in a node which had no ha packages to update

Itxaka: ah, this seems to fail locally as I havent pushed the next commit with the remote node stuff added

Itxaka:
@aspiers ready for re-review.

Spent most of the day testing the different scenarios and it seems to work perfectly.

On a node part of a cluster, non-remote:
When there are package to update but no ha_packages -> sets maintenance mode before update, then chef lifts the maintenance mode at the end of the run
When there are ha packages to update -> stops the services (pacemaker/corosync) then starts pacemaker after the update
Where there are no packages to update -> does nothing, duh!


Sadly, there is some code duplication in regards to https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/libraries/maintenance_mode_helpers.rb#L60

This is because the helper on crowbar-pacemaker is already using an `execute` resource which cannot be called from a `ruby_block`.
Unfortunately, we need to wrap the crm maintenance mode in a chef resource block, as we need the previous steps to have run to obtain the proper variables in order to fire off the maintenance mode, and you cannot wrap an `execute` resource inside a `ruby_block` one (Actually you probably can, but its a huge hack and terrible)

So instead we create several resources that get triggered in the case that maintenance mode is required due to the late evaluation of the variables.


I could not find a better way of doing it while respecting the chef workflow.

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -45,6 +48,67 @@
...
+        node.run_state["found_ha_packages"] = $?.exitstatus ? true : false
+      end
+      not_if { search(:node, "roles:pacemaker-cluster-member").empty? }
+    end
+
+    ["corosync", "pacemaker"].each do |s|
+      service s do
+        action :stop
+        only_if { node.run_state["found_ha_packages"] }
+        not_if { node[:pacemaker][:is_remote] }
```
jsuchome: I think you should make sure that `node[:pacemaker]` exists; If I'm not missing sometthing, this could could be executed at non-pacemaker nodes as well

Itxaka:
Missing exceptions for non-cluster nodes in some of the resources!

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -45,6 +48,67 @@
...
+      end
+    end
+
+    ruby_block "set cluster maintenance" do
+      block do
+        Chef::Log.info("Triggering maintenance mode for this node")
+        true
+      end
+      only_if do
+        ha = node.run_state["found_ha_packages"]
```
jsuchome: `ha` seems bit confusing, could you rename this variable? (Or just not create it at all and directly do a check for run_state at line 104)

Itxaka: :+1:

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -45,6 +48,67 @@
...
+        is_cluster = !search(:node, "roles:pacemaker-cluster-member").empty?
```
Itxaka: wtf, this is wrong. I wanted to check if the node is part of a cluster, this will just search for any nodes that have that role ¬_¬

chef/cookbooks/updater/recipes/default.rb
Itxaka: :+1:

chef/cookbooks/updater/recipes/default.rb
jsuchome: Oh, true. Look also at the same search used earlier.

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -17,6 +17,9 @@
 # limitations under the License.
 #
 
+::Chef::Recipe.send(:include, CrowbarPacemaker::MaintenanceModeHelpers)
```
aspiers: Just curious - how is this different from `::Chef::Recipe.include CrowbarPacemaker::MaintenanceModeHelpers` ?

aspiers:
@Itxaka I'm still a bit confused how this is supposed to handle the remote case?

chef/cookbooks/updater/recipes/default.rb
Itxaka: Chef-ism. http://lists.opscode.com/sympa/arc/chef/2011-05/msg00286.html

Actually, I think that due the restructuring of the resources in my last update this is not required, I'll retry with the usual `include` to see

Itxaka:
@aspiers Lets see, for remote case, there should be no pacemaker/corosync services runnig so it wont stop them.

The question is, for remote nodes, when upgrading them, do we need to set the node into maintenance mode as well? The HA guide does not mention remote nodes so Im a bit lost in there and what to do with them.

chef/cookbooks/updater/recipes/default.rb
Itxaka: Not needed anymore, reverting to the usual include

aspiers:
Revisiting this due to [Bug 1061834 – installing updates causes HA failures](https://bugzilla.suse.com/show_bug.cgi?id=1061834).  @Itxaka Any chance you could amend the commit message to contain the full bsc URL?  Thanks!

Itxaka:
Ummm, seems that the branch for this PR migth have been removed somehow, so I would need to create a different PR with that branch covered somehow I guess?

Itxaka:
Moved to https://github.com/crowbar/crowbar-core/pull/1353 @cmurphy comments are still unresolved on that PR

Itxaka:
doesnt seem to be needed feel free to restore if so

PR #564, opened 2016-07-22
Add code path for 3.0 to 4.0 upgrade

vuntz:
I'm not fully sure this is fully correct, feedback welcome.

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade.rb
+++ b/crowbar_framework/lib/crowbar/upgrade.rb
@@ -33,8 +33,8 @@ module Crowbar
     def upgrade
       case @version
       when "1.9"
-        knife_files
-        crowbar_files
+        knife_files_1_9
+        crowbar_files_1_9
       end
 
       if @status[:errors].any?
@@ -53,7 +53,7 @@ module Crowbar
 
     protected
 
-    def knife_files
+    def knife_files_1_9
       Rails.logger.debug "Upgrading chef backup files"
       @data.join("knife", "databags", "barclamps").rmtree
       knife_path = @data.join("knife")
@@ -131,7 +131,7 @@ module Crowbar
       end
     end
 
-    def crowbar_files
+    def crowbar_files_1_9
       FileUtils.touch(@data.join("crowbar", "database.yml"))
     end
 
--- a/crowbar_framework/lib/crowbar/upgrade.rb
+++ b/crowbar_framework/lib/crowbar/upgrade.rb
@@ -36,6 +36,9 @@ module Crowbar
         knife_files_1_9
         reset_target_platform_for_nodes
         crowbar_files_1_9
+      when "3.0"
+        network_json_3_0
+        reset_target_platform_for_nodes
       end
 
       if @status[:errors].any?
@@ -53,6 +56,12 @@ module Crowbar
     end
 
     protected
+
+    def network_json_3_0
+      Rails.logger.debug "Re-creating network.json from upgrade data"
+      network_role_path = @data.join("knife", "roles", "network-config-default.json")
+      create_network_json(network_role_path)
+    end
 
     def knife_files_1_9
       Rails.logger.debug "Upgrading chef backup files"
@@ -129,6 +138,7 @@ module Crowbar
     end
 
     def reset_target_platform_for_nodes
+      Rails.logger.debug "Resetting platform of nodes due to upcoming OS upgrade"
       # find admin node and update target_platform
       nodes_path = @data.join("knife", "nodes")
       nodes_path.children.each do |file|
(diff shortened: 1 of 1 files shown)


rsalevsky:
+1

jsuchome:
LGTM, I do not know the code much

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -33,8 +33,12 @@ def initialize(backup)
     def upgrade
       case @version
       when "1.9"
-        knife_files
-        crowbar_files
+        knife_files19
```
MaximilianMeister: feel free to ignore the following taste comment:

can we maybe add another underscore in front of the version like
```ruby
knife_files_19
``` 
? maybe its just me, but it looks somehow ugly ;)

MaximilianMeister:
+1

crowbar_framework/lib/crowbar/upgrade.rb
aspiers: Or even `knife_files_1_9`

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -34,6 +34,7 @@ def upgrade
       case @version
       when "1.9"
         knife_files19
+        reset_target_platform_for_nodes
```
aspiers: Is this really covered by the commit message?  ICBW but it looks to me like you are mixing an "extract method" refactoring (which I am almost always in favour of) with some changes in behaviour.

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -54,6 +57,12 @@ def supported?
 
     protected
 
+    def network_json30
```
aspiers: I would prefer this method name to be a verb.  Currently the name doesn't give any indication of what it does, only that it's related to `network.json` from 3.0.

vuntz:
@MaximilianMeister @aspiers your comments have been implemented, thanks for the review!

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -33,11 +33,11 @@ def initialize(backup)
...
       case @version
       when "1.9"
-        knife_files_1_9
+        upgrade_knife_files_1_9
         reset_target_platform_for_nodes
-        crowbar_files_1_9
+        upgrade_crowbar_files_1_9
       when "3.0"
-        network_json_3_0
+        create_network_json_3_0
```
aspiers: Shouldn't this also be "upgrade_"? or at least "recreate"?  Since the JSON file must already be there.

vuntz: The JSON file that is there is not modified by the user; it's really about creating it from what is in the backup, so I feel the name is correct. I could change to recreate_ if it really makes a difference...

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -34,6 +34,7 @@ def upgrade
       case @version
       when "1.9"
         knife_files_1_9
+        reset_target_platform_for_nodes
```
aspiers: I think [my previous comment about mixing refactoring with behavioural changes](https://github.com/crowbar/crowbar-core/pull/564#discussion_r72160108) still applies here, doesn't it? I'm struggling to understand why this addition is in this commit rather than in the "Patch upgrade data for 3.0 to 4.0 upgrades" commit (and even that one seems to be a mix of two different things).

BTW the main reason for being pedantic about how the commits are split is not (just ;-) ) to annoy you, but that I'm pretty ignorant in this area, so without each commit being small and with a clear commit message, I struggle to give a good review.

vuntz: There's no behavior change, it's just moving code around :-)

crowbar_framework/lib/crowbar/upgrade.rb
aspiers: Oh I see.  So `recreate_` would be better, but not a big deal.

crowbar_framework/lib/crowbar/upgrade.rb
aspiers: I can't see how this isn't a behaviour change.  You're calling a method which wasn't called before, and nothing else in this commit relates to that.

vuntz: Well, check the patch :-) The method is a new one, extracted from `knife_files_1_9`.

aspiers: Doh, I missed the creation of the new method despite reading this about 5 times.  OTOH this is a good example of why it's better to do a single refactoring per commit :-)

MaximilianMeister:
still +1

MaximilianMeister:
I guess this is good to merge now, we will likely need it soon for the new upgrade process

PR #566, opened 2016-07-22
Cleanup config.ru

rsalevsky:
After the switch to puma, Rack::Chunked and Rack::ContentLength are
handled by puma.


nicolasbock:
+1

toabctl:
+1

PR #567, opened 2016-07-25
Sync crowbar service with chef template

rsalevsky:


toabctl:
Why do we need a chef template and a .service file? Isn't one enough?

vuntz:
+1

vuntz:
@toabctl the non-template version is used in the package, so that crowbar can be started prior to chef having run.

toabctl:
+1

aspiers:
@vuntz So why do we need the template version at all?  And why don't we need `ExecStop` and `ExecReload`?

PR #568, opened 2016-07-25
Make sure we're running with correct time on installation

dirkmueller:
With machines that don't have proper RTC time, we're running into
nasty issues when trying to install on a system that is thinking it
lives in the 70ties.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -322,6 +322,9 @@ if not nodes.nil? and not nodes.empty?
             cloud_available = true if name.include? "Cloud"
           end
 
+          ntp_servers = search(:node, "roles:ntp-server")
+          ntp_servers_ips = ntp_servers.map { |n| Chef::Recipe::Barclamp::Inventory.get_network_by_type(n, "admin").address }
+
           autoyast_template = mnode[:state] == "os-upgrading" ? "autoyast-upgrade" : "autoyast"
           template "#{node_cfg_dir}/autoyast.xml" do
             mode 0644
@@ -341,6 +344,7 @@ if not nodes.nil? and not nodes.empty?
                       node_ip: mnode[:crowbar][:network][:admin][:address],
                       node_fqdn: mnode[:fqdn],
                       node_hostname: mnode[:hostname],
+                      ntp_servers: ntp_servers_ips,
                       platform: target_platform_distro,
                       target_platform_version: target_platform_version,
                       architecture: arch,
--- a/chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
@@ -12,6 +12,7 @@
       <final_reboot config:type="boolean">false</final_reboot>
       <halt config:type="boolean">false</halt>
       <second_stage config:type="boolean">true</second_stage>
+      <ntp_sync_time_before_installation><%= @ntp_servers.first %></ntp_sync_time_before_installation>
     </mode>
     <mouse>
       <id>none</id>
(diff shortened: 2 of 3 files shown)


chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -322,6 +322,9 @@ def find_node_boot_mac_addresses(node, admin_data_net)
             cloud_available = true if name.include? "Cloud"
           end
 
+          ntp_servers = search(:node, "roles:ntp-server")
```
toabctl: use ```search_env_filtered```

dirkmueller: we don't seem to be very consistent on this. there is no other usage of search_env_filtered in crowbar-core (actually I can find it only in the nova barclamp, which is strange..)

vuntz: `search_env_filtered` (and `get_instance`) should only be used when a proposal references another proposal through its attributes. It's not the case here, as the provisioner proposal doesn't reference the ntp proposal.

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -322,6 +322,9 @@ def find_node_boot_mac_addresses(node, admin_data_net)
...
+          ntp_servers_ips = ntp_servers.map { |n| Chef::Recipe::Barclamp::Inventory.get_network_by_type(n, "admin").address }
```
vuntz: Just a hint: you don't need the `Chef::Recipe::` prefix (but that won't help with the hound warning)

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
```diff
@@ -12,6 +12,7 @@
       <final_reboot config:type="boolean">false</final_reboot>
       <halt config:type="boolean">false</halt>
       <second_stage config:type="boolean">true</second_stage>
+      <ntp_sync_time_before_installation><%= @ntp_servers.first %></ntp_sync_time_before_installation>
```
vuntz: This will possibly fail if there are no ntp servers configured; don't know how autoyast behaves in that case.

nicolasbock: From [AutoYaST](https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html):

> Please keep in mind that you need a reachable NTP server and network connections while running the installation.

which sounds like the installation will fail in case of a missing NTP server.

chef/cookbooks/provisioner/recipes/update_nodes.rb
rsalevsky: Using a multiline do block should do the trick here and hound should be happy.

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
```diff
@@ -12,6 +12,7 @@
       <final_reboot config:type="boolean">false</final_reboot>
       <halt config:type="boolean">false</halt>
       <second_stage config:type="boolean">true</second_stage>
+      <ntp_sync_time_before_installation><%= @ntp_servers.first %></ntp_sync_time_before_installation>
```
vuntz: @dirkmueller can you add something like `<% unless @ntp_servers.empty? -%>` to protect against an empty list of ntp servers?

chef/cookbooks/provisioner/recipes/update_nodes.rb
vuntz: Can you replace this line with:

```ruby
          ntp_instance = CrowbarHelper.get_proposal_instance(node, "ntp", "default")
          ntp_servers = node_search_with_cache("roles:ntp-server", ntp_instance)
```

aspiers: @vuntz Why can't it retrieve the servers from the new config databag?

vuntz: It's possible now; my comment was from before this got merged :-)

PR #569, opened 2016-07-25
ip_local_port_range needs to have different parity for start/end values

dirkmueller:


toabctl:
@dirkmueller I have no idea why that is needed. Can you exlain that a bit?

rhafer:
> I have no idea why that is needed. Can you exlain that a bit?

@toabctl I guess the answer to that is related to this kernel commit: http://kernel.suse.com/cgit/kernel/commit/?id=07f4c90062f8f and this which added a warning: http://www.spinics.net/lists/netdev/msg330895.html

I just stumbled of that while looking for something completely different.

dirkmueller:
@rhafer updated documentation

rhafer:
+1

nicolasbock:
+1

PR #570, opened 2016-07-26
crowbar-init: enable inclusion of sinatra/rails partials

MaximilianMeister:
when crowbar-init is installed there will be a symlink created in
/etc/apache2/conf.d/crowbar/crowbar.conf which first points to:
https://github.com/crowbar/crowbar-init/blob/master/config/crowbar-sinatra.conf.partial
and once the work from crowbar-init is done this symlink will point to:
https://github.com/MaximilianMeister/crowbar-core/blob/apache/conf-sinatra/configs/crowbar-rails.conf.partial

NOTE:
@rsalevsky is working on the packaging side, and once this is done and tested this PR can be merged

### DIFF
--- a/configs/crowbar.conf
+++ b/configs/crowbar.conf
@@ -9,7 +9,5 @@
     ErrorDocument 422 /422.html
     ErrorDocument 500 /500.html
 
-    ProxyPass / http://127.0.0.1:3000/ timeout=1500 Keepalive=On
-    ProxyPassReverse / http://127.0.0.1:3000/
-    SetEnv proxy-initial-not-pooled 1
+    Include /etc/apache2/conf.d/crowbar.conf.partial
 </VirtualHost>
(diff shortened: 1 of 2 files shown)


rsalevsky:
+1

configs/crowbar.conf
```diff
@@ -9,7 +9,5 @@
     ErrorDocument 422 /422.html
     ErrorDocument 500 /500.html
 
-    ProxyPass / http://127.0.0.1:3000/ timeout=1500 Keepalive=On
-    ProxyPassReverse / http://127.0.0.1:3000/
-    SetEnv proxy-initial-not-pooled 1
+    Include /etc/apache2/conf.d/crowbar/crowbar.conf
```
rsalevsky: `Include /etc/apache2/conf.d/crowbar.conf.partial`

rsalevsky:
+1

rsalevsky:
We just move some code and I need to adapt the package in parallel. So I will merge it now, in case something should be not ok we can adapt it later.

PR #573, opened 2016-07-26
Always install dmidecode

dirkmueller:
ohai and crowbars ohai plugins (and crowbar) depend heavily on the
system name and serial number to be always available, so make sure
this is installed all the time. Currently it is only conincidentally
installed on x86 and missing on aarch64.

### DIFF
--- a/chef/cookbooks/ohai/recipes/default.rb
+++ b/chef/cookbooks/ohai/recipes/default.rb
@@ -82,6 +82,11 @@ unless node[:platform_family] == "windows"
     rescue LoadError
       Chef::Log.fatal("Unable to load cstruct module - install of #{pkg} failed?")
     end
+
+    # Install dmidecode - needed for better machine detection and network.json matching
+    if node[:kernel][:machine] =~ /(x86|aarch64)/
+      package("dmidecode").run_action(:install)
+    end
   end
 end
 


nicolasbock:
+1

rsalevsky:
+1

Itxaka:
+1

vuntz:
Why only in the crowbar role? Don't we need it on other nodes (or is it installed through autoyast/crowbar_register there).

Also, shouldn't this be a dep of the package? :-)

dirkmueller:
@vuntz I am a bit confused about how essential deps are installed on all nodes. is there a better place where this should belong?

technically rubygem-ohai survives fine without dmidecode being installed,it will just not return any data. I guess thats okay. I was pondering a bit about where ot add the dependency. Basically we (Cloud) want it everywhere, but that might not be the right thing to do for other use cases. Also, dmidecode doesn#t exist everywhere (e.g. s390x does not have it or any other non-ACPI platform).

vuntz:
@dirkmueller well, this cookbook is only run on the admin server :-) I'd probably do the change in the ohai cookbook.

dirkmueller:
@vuntz done

vuntz:
Thanks, +1

PR #575, opened 2016-07-27
fail with useful error on unmatched interface (bsc#990745)

aspiers:
If we fail to find an admin interface matching a pattern such as `+10g1` for a given node, don't proceed blindly, but instead bail with a helpful error message. This was experienced in a scenario where the Crowbar nodes were a mix of physical machines and VMs with slow `virtio` interfaces, but the `network.json` was incorrectly configured and the admin network conduit failed to match the slow `virtio` interfaces.

Also add a warning whenever a conduit cannot be matched against available interfaces, to help more quickly identify incorrectly configured `network.json` files, and de-obfuscate the code with comments and better variable names.

### DIFF
--- a/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+++ b/chef/cookbooks/barclamp/libraries/barclamp_library.rb
@@ -317,7 +317,7 @@ module BarclampLibrary
 
         conduit_info = intf_to_if_map[conduit]
         interface_list = conduit_info["if_list"]
-        team_mode = conduit_info["team_mode"] rescue nil
+        team_mode = conduit_info["team_mode"]
 
         return [interface_list[0], interface_list, nil] if interface_list.size == 1
 
--- a/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+++ b/chef/cookbooks/barclamp/libraries/barclamp_library.rb
@@ -276,21 +276,21 @@ module BarclampLibrary
         if_cnt = m[3]
         desired = speeds.index(m[2])
         found = nil
-          filter = lambda { |x|
+        filter = lambda do |x|
           found = if_map["#{speeds[x]}#{if_cnt}"] unless found
-        }
+        end
         case m[1]
-          when "+"
-            (desired..speeds.length).each(&filter)
-          when "-"
-            desired.downto(0,&filter)
-          when "?"
-            (desired..speeds.length).each(&filter)
-            desired.downto(0,&filter) unless found
-          else
-            found = if_map[ref]
-          end
-          found
+        when "+"
+          (desired..speeds.length).each(&filter)
+        when "-"
+          desired.downto(0, &filter)
+        when "?"
+          (desired..speeds.length).each(&filter)
+          desired.downto(0, &filter) unless found
+        else
+          found = if_map[ref]
+        end
+        found
       end
 
       # Find the interface(s) and teaming mode to use for the given
--- a/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+++ b/chef/cookbooks/barclamp/libraries/barclamp_library.rb
@@ -308,7 +308,7 @@ module BarclampLibrary
       def self.lookup_interface_info(node, conduit, intf_to_if_map = nil)
         intf_to_if_map = Barclamp::Inventory.build_node_map(node) if intf_to_if_map.nil?
 
-        return [nil, nil] if intf_to_if_map[conduit].nil?
+        return [nil, nil, nil] if intf_to_if_map[conduit].nil?
 
         conduit_info = intf_to_if_map[conduit]
         interface_list = conduit_info["if_list"]
--- a/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+++ b/chef/cookbooks/barclamp/libraries/barclamp_library.rb
@@ -84,6 +84,8 @@ module BarclampLibrary
         999
       end
 
+      # Returns an array of interfaces like ["em1", "em2", "em3", "em4"],
+      # sorted by bus_order
       def self.sort_ifs(map, bus_order)
         answer = map.sort{|a,b|
           aindex = Barclamp::Inventory.bus_index(bus_order, a[1]["path"])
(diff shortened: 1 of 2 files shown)


aspiers:
In particular, I need someone to confirm that there aren't any scenarios where it's OK to have an unmatched interface pattern.  Because if there are, obviously this will break them.

vuntz:
I might be wrong, but I think it's actually okay to have this be `nil` in some cases, as not all conduits are used on all nodes. So I think your patch is too aggressive here. You can try setting something crazy on the storage network (which is not used unless you deploy swift or ceph), and see if the exception is triggered.

aspiers:
@vuntz I think you're probably right. I've drastically improved the PR.

aspiers:
Not convinced that [gating failure](https://ci.suse.de/job/openstack-mkcloud/28946/) was related to the changes - forced a rebuild.

Itxaka:
lgtm +1

aspiers:
@rhafer Since @vuntz isn't around, your review on this would be appreciated!

rhafer:
@aspiers Yes, I have this on my radar. Just didn't find the time yet to give it the attention it deserves :(

aspiers:
@rhafer OK thanks :) Rebased, slightly improved one of the comments, and fixed some broken indentation.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -158,20 +217,39 @@ def self.build_node_map(node)
...
-            if mk == "if_list"
-              hash["if_list"] = v["if_list"].map do |if_ref|
-                map_if_ref(if_remap, if_ref)
+          conduit_data.each do |conduit_key, conduit_val|
+            if conduit_key == "if_list"
+              hash["if_list"] = conduit_data["if_list"].map do |if_ref|
+                r = map_if_ref(if_remap, if_ref)
+                if r.nil?
+                  Chef::Log.warn "Unable to find interface for #{node.fqdn} " \
+                                 "matching #{if_ref} out of #{if_remap}"
```
rhafer: As @vuntz it's perfectly fine to  have this ```nil``` here for some of the conduits. As long as that conduit is not used on the node (i.e. not network using that conduit is allocated/enabled on that node). I such a case this will cause a lot of unnecessary warnings (```build_node_map``` is called quite frequently) in the logs. I fear that it will just confuse people.

Wouldn't be better to throw a more useful error message for cases where the conduit is actually used on a node (e.g. here: https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/network/recipes/default.rb#L161)

Or even when trying to allocate/enable a network already (i.e. in the rails app)

rhafer:
Apart from the issue about the warning (you might be able to convince me that this is not an issue), these changes are ok IMO.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
aspiers: > I such a case this will cause a lot of unnecessary warnings (build_node_map is called quite frequently) in the logs. I fear that it will just confuse people.

Yeah OK, fair point.

> Wouldn't be better to throw a more useful error message for cases where the conduit is actually used on a node

Sure, that's why this PR also adds [this error](https://github.com/aspiers/crowbar-core/blob/12da87dd0bbca677d84739ec831fb31dae5900cf/chef/cookbooks/provisioner/recipes/update_nodes.rb#L24).

> Or even when trying to allocate/enable a network already (i.e. in the rails app)

You mean, somewhere like in `NetworkService#allocate_ip_by_type`?  That seems to get all its network info from the databag.

aspiers:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;uwPEX4Qp&#x2F;64-bsc-990745-installing-new-node-fails-to-start-install"> bsc#990745 - Installing new node; fails to start install</a>

chef/cookbooks/barclamp/libraries/barclamp_library.rb
aspiers: @rhafer Would it be good enough to change this `Chef::Log.warn` into a `.debug`?  Also, I'm still not sure exactly where else you are suggesting that errors should be added if the interface is `nil`?  (See question above.)

rhafer: > Would it be good enough to change this Chef::Log.warn into a .debug? 

Yeah, I guess that's fine.

> You mean, somewhere like in NetworkService#allocate_ip_by_type?

Yes. And ```NetworkService#enable_interface```. The advantage of doing proper checks there would be that we could actually fail with an error much earlier than during the chef-client run. (Note: I am not say that you should be the one implementing that, it just an idea that I had when reviewing your code)

> That seems to get all its network info from the databag.

Yeah, it would need to be enhanced to call into ```NodeObject#build_node_map```.  Which reminds me that might want to add the debug message there as well.

aspiers: OK, it's now `.debug`.

Based on our follow-on IRC discussion about this, I've created [Add network interface validation to NetworkService#build_net_info (Trello)](https://trello.com/c/7W51qfqu/1158-add-network-interface-validation-to-networkservice-build-net-info).

rhafer:
+1

@aspiers I guess you want to remove the "wip" and "do not merge labels"?

aspiers:
Rebased to fix conflict.

aspiers:
Failed gating build has been deleted, so rebased to trigger another one.

aspiers:
Bah, @vuntz completely rewrote this so rebasing is going to be ugly... One nice thing about gerrit is that it warns you when reviews conflict with each other. #githubsucks

aspiers:
@vuntz I had a quick attempt to rebase and gave up as everything looks different. Let's fix this together this week.

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -20,6 +20,10 @@ def find_node_boot_mac_addresses(node, admin_data_net)
   result = []
   admin_interfaces = admin_data_net.interface_list
   admin_interfaces.each do |interface|
+    if interface.nil?
+      raise "Failed to find admin node interface for node #{node.fqdn}! " \
+            "Check your network barclamp proposal."
+    end
```
vuntz: Thinking more about this: no objection about this part, as it's no regression, but I think it's not enough. We do want to avoid a crash in chef-client here, as having a rogue node could result, thanks to this, in a DoS on part of the deployment infrastructure. I'll submit a patch.

vuntz: See https://github.com/crowbar/crowbar-core/pull/803

aspiers:
@vuntz @rhafer Please can you either rebase this or help me rebase it?   I put a *lot* of effort into adding useful comments to make cryptic code readable, but since you rewrote things completely by moving a lot of stuff to `Crowbar::ConduitResolver` I would have to almost start from scratch, and I don't know what other interface changes you introduced during this move.

vuntz:
Obsoleted by https://github.com/crowbar/crowbar-core/pull/803

aspiers:
Looks like I'll have to redo this myself. Resubmitted as #1082.

PR #577, opened 2016-07-28
Ironic network

skazi0:
This change introduces an additional "Ironic" network type to Crowbar. It is a dedicated network to handle network booting of baremetal nodes.

### DIFF
--- a/chef/data_bags/crowbar/migrate/network/101_add_ironic_network.rb
+++ b/chef/data_bags/crowbar/migrate/network/101_add_ironic_network.rb
@@ -1,0 +1,14 @@
+def upgrade(ta, td, a, d)
+  unless a["networks"].key? "ironic"
+    a["networks"]["ironic"] = ta["networks"]["ironic"]
+  end
+  # the "intf3" conduit mappings are not added as there's no easy way to
+  # auto-assign physical interfaces properly.
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  # there's no easy way to ensure that "ironic" and "intf3" entries were not
+  # added by the user so it's safer to not delete anything
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-network.json
+++ b/chef/data_bags/crowbar/template-network.json
@@ -87,6 +87,9 @@
             },
             "intf2": {
               "if_list": [ "1g1", "1g2" ]
+            },
+            "intf3": {
+              "if_list": [ "1g1", "1g2" ]
             }
           }
         },
@@ -101,6 +104,26 @@
             },
             "intf2": {
               "if_list": [ "?1g1" ]
+            },
+            "intf3": {
+              "if_list": [ "?1g2" ]
+            }
+          }
+        },
+        {
+          "pattern": "single/.*/.*ironic.*",
+          "conduit_list": {
+            "intf0": {
+              "if_list": [ "?1g1" ]
+            },
+            "intf1": {
+              "if_list": [ "?1g1" ]
+            },
+            "intf2": {
+              "if_list": [ "?1g1" ]
+            },
+            "intf3": {
+              "if_list": [ "?1g2" ]
             }
           }
         },
@@ -114,6 +137,9 @@
               "if_list": [ "?1g1" ]
             },
             "intf2": {
+              "if_list": [ "?1g1" ]
+            },
+            "intf3": {
               "if_list": [ "?1g1" ]
             }
           }
@@ -129,6 +155,9 @@
             },
             "intf2": {
               "if_list": [ "1g1" ]
+            },
+            "intf3": {
+              "if_list": [ "1g1" ]
             }
           }
         },
@@ -142,6 +171,9 @@
               "if_list": [ "?1g1" ]
             },
             "intf2": {
+              "if_list": [ "?1g1" ]
+            },
+            "intf3": {
               "if_list": [ "?1g1" ]
             }
           }
@@ -242,6 +274,23 @@
             "host": { "start": "192.168.130.10", "end": "192.168.130.254"}
           }
         },
+        "ironic": {
+          "conduit": "intf3",
+          "vlan": 100,
+          "use_vlan": false,
+          "add_bridge": false,
+          "add_ovs_bridge": false,
+          "bridge_name": "br-ironic",
+          "subnet": "192.168.127.0",
+          "netmask": "255.255.255.0",
+          "broadcast": "192.168.127.255",
+          "router": "192.168.127.1",
+          "router_pref": 50,
+          "ranges": {
+            "admin": { "start": "192.168.127.10", "end": "192.168.127.11" },
+            "dhcp": { "start": "192.168.127.21", "end": "192.168.127.254" }
+          }
+        },
         "admin": {
           "conduit": "intf0",
           "vlan": 100,
@@ -267,7 +316,7 @@
     "network": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 100,
+      "schema-revision": 101,
       "element_states": {
         "network": [ "readying", "ready", "applying" ]
       },
--- a/crowbar_framework/config/locales/network/en.yml
+++ b/crowbar_framework/config/locales/network/en.yml
@@ -54,6 +54,7 @@ en:
       public: 'Public'
       private: 'Private'
       storage: 'Storage'
+      ironic: 'Ironic'
     controller:
       virtual: 'Virtual'
 


chef/data_bags/crowbar/migrate/network/101_add_ironic_network.rb
```diff
@@ -0,0 +1,20 @@
+def upgrade(ta, td, a, d)
+  a["networks"]["ironic"] = ta["networks"]["ironic"]
```
vuntz: Can you add a check that the network doesn't already exists? As people can define their own network, it could and we don't want to override their data.

chef/data_bags/crowbar/migrate/network/101_add_ironic_network.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  # copy intf1 as intf3
+  a["conduit_map"].each do |cm|
+    cm["conduit_list"]["intf3"] = cm["conduit_list"]["intf1"]
+  end
```
vuntz: This is similar here, I'm afraid that intf3 could exist already and that you could be overriding it...

chef/data_bags/crowbar/template-network.json
```diff
@@ -87,6 +87,26 @@
...
             "intf2": {
               "if_list": [ "1g1", "1g2" ]
+            },
+            "intf3": {
+              "if_list": [ "1g1", "1g2" ]
+            }
+          }
+        },
+        {
+          "pattern": "quad/.*/.*",
```
vuntz: I'm unclear why we need this mode. Could you explain?

skazi0: The idea is that many/most servers have four network interfaces and with this mode, default Crowbar mechanism will distribute the networks on those interfaces. This way intf0-3 is actually mapped to physical interfaces. It can be changed by hand in the network.json but with this mode it would be only one parameter change with pretty sane default behavior.
For Ironic this is pretty important as it needs it's own physical network to handle the network booting of baremetal hosts.

aspiers: That makes sense to me.

skazi0: I did some more testing and decided that it's easier to set ironic things up without this "quad" mode. Ironic network is needed on a limited set of nodes and in production setup you need to prepare custom network.json anyway. I will remove the "quad" option for now.

chef/data_bags/crowbar/migrate/network/101_add_ironic_network.rb
```diff
@@ -0,0 +1,20 @@
+def upgrade(ta, td, a, d)
+  unless a["networks"].key? "ironic"
+    a["networks"]["ironic"] = ta["networks"]["ironic"]
+  end
+
+  # copy intf1 as intf3
+  a["conduit_map"].each do |cm|
+    unless cm["conduit_list"].key? "intf3"
+      cm["conduit_list"]["intf3"] = cm["conduit_list"]["intf1"]
```
rhafer: Since ```admin``` and ```ironic``` are on a native VLAN (```use_vlan=false```) does it really work to put them on the same conduit? I suppose the ```ironic``` network is used for PXE booting bare-metal instances?

rhafer: I am asking this, because when you put both ```admin``` and ```ironic``` untagged onto the same physical interface(s) and both networks provide dhcp/pxe for the network that is not goint to work out.

One solution to this might be to make ```ironic``` a tagged VLAN on the compute nodes and reconfigure the switch so that that VLAN will be the native VLAN on the ports from which the bare-metal instances boot.
The other option is to require ```ironic``` to be on a separate physical network, I guess.

skazi0: This is exactly why I introduced `intf3` instead of using the existing conduits.

Current approach is to have ironic on a separate physical network (separate NIC on the nodes which require ironic connectivity).

In practice this means that you need some custom configuration in network.json to make it work. The `"pattern": "single/.*/.*ironic.*"` mode is an actual example of this (which I use in my tests) but it may need to be adjusted for real deployments.

rhafer: > This is exactly why I introduced intf3 instead of using the existing conduits.

Yeah, I am just wondering about the schema migration, because there you assign the ```intf3``` interface to exactly the same physical interfaces as ```intf1```. I am not sure if there is a good solution to this, other than not touching the conduit_maps at all in the migration and document it. As you already mentioned, in practice this will most likely need a custom config anyway.

skazi0: This was the best option I could think of... admin is on `intf0` and I want to avoid `intf3` being on the same interface, so making a copy of the second conduit will take the second NIC in case of `dual` mode and with one NIC it's not going to work anyway.

I'm not sure what will happen if there's no entry in the `conduit_map` matching the network some barclamp is trying to allocate an IP from.

rhafer: > I'm not sure what will happen if there's no entry in the conduit_map matching the network some barclamp is trying to allocate an IP from.

It's going to fail. Latest when trying to make the allocation active with the chef-client run.

But then, using the same interfaces as admin is going to fail as well, just in very different (and probabyl hard to debug) way.

skazi0: OK, I removed adding of the `intf3` in the upgrade migration.

skazi0: Hmmm... what about the default conduit mappings in the template? These were based on the same assumption as the migration. Do you think these should also be "undefined" by default?

rhafer:
> Hmmm... what about the default conduit mappings in the template? These were based on the same assumption as the migration. Do you think these should also be "undefined" by default?

Good question. I guess what we would need is some code in the ironic barclamp that verifies that ```ironic``` and ```admin``` endup on separate physical interfaces. 
In the neutron barclamp we have something like that already for the physcial networks of floating and nova_fixed (it's in https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/neutron/libraries/helpers.rb).

BTW, is it a requirement to have the ```ironic``` network being a part of the default network.json?

skazi0:
Some validation would not hurt for sure but at this stage the process of setting up Ironic requires a lot of knowledge of things running behind anyway.

RE BTW, `ironic` network as introduced by this PR is a "first class citizen" and is visible in the UI. I think it has to be defined somehow even if you don't plan to deploy Ironic.

rhafer:
> Hmmm... what about the default conduit mappings in the template? These were based on the same assumption as the migration. Do you think these should also be "undefined" by default?

Hm, somehow I overlooked that. Does it still  work if you just leave the ```"pattern": "single/.*/.*ironic.*",``` section intact and remove ```intf3```from  ```"pattern": "single/.*/.*",```? 

> Some validation would not hurt for sure but at this stage the process of setting up Ironic requires a lot of knowledge of things running behind anyway.

That's one reason why I am wondering if we should add the ironic network at all by default. Or if we should just document how to add it for now.

> RE BTW, ironic network as introduced by this PR is a "first class citizen" and is visible in the UI. I think it has to be defined somehow even if you don't plan to deploy Ironic.

The so called "bastion" network is also a "first class citizen" and it's also not present in the default config. But only mentioned in the deployment guide.

If we manage to find sensible defaults for "ironic" I am all for adding it. It's just not all that easy to figure out those defaults.

rhafer:
> Some validation would not hurt for sure but at this stage the process of setting up Ironic requires a lot of knowledge of things running behind anyway.

So I guess it would sense to add something to the backlog for ironic network validation (unless that happended already).

skazi0:
I think the `"pattern": "single/.*/.*ironic.*"` mode overrides the default `single` mode for the node where ironic roles are installed. I will check if it will still work if I remove `intf3` from the other modes.

Regarding the "bastion" network, it's a matter of semantics, but I wouldn't call it "first class citizen" as it's not visible anywhere in the GUI.

skazi0:
> Hm, somehow I overlooked that. Does it still work if you just leave the "pattern": "single/.*/.*ironic.*", section intact and remove intf3from "pattern": "single/.*/.*",? 

It fails the validation before crowbar install so probably it will not work: `/etc/crowbar/network.json is invalid. Conduit list with pattern 'single/.*/.*ironic.*' does not have the same list of conduits as catch-all conduit list for mode 'single' (with pattern 'single/.*/.*'): 'intf0, intf1, intf2, intf3' instead of 'intf0, intf1, intf2'.`.

rhafer:
> It fails the validation before crowbar install so probably it will not work: /etc/crowbar/network.json is invalid. Conduit list with pattern 'single/.*/.*ironic.*' does not have the same list of conduits as catch-all conduit list for mode 'single' (with pattern 'single/.*/.*'): 'intf0, intf1, intf2, intf3' instead of 'intf0, intf1, intf2'..

Hm, I guess the validator is overly careful here again. Technically it is allowed to have different conduit_lists for the different pattern. But it's also a question of whether we want to allow setups like that. @vuntz Do you have any opinion on this?

mjura:
LGTM :+1: 

@skazi0: Please create also bug for documentation team, which will describe this change.

vuntz:
> Hm, I guess the validator is overly careful here again. Technically it is allowed to have different conduit_lists for the different pattern. But it's also a question of whether we want to allow setups like that. @vuntz Do you have any opinion on this?

You're right, but I think it's better to be overly careful for now until we encourage people to change the conduit lists after installation. Because right now, we would tell people "no luck, reinstall", which is a shame.

vuntz:
I'm +1 on the PR, but my concern is that we're adding a non-core network here while we would like to slowly move networks out of the core definition whenever possible. Somehow, I'd prefer to allow defining the network from the ironic barclamp and have the ironic barclamp talk to the network barclamp. But that's probably something for the long term, so no reason to block this.

rhafer:
> but my concern is that we're adding a non-core network here while we would like to slowly move networks out of the core definition whenever possible.

@vuntz Right, that's also what I mentioned earlier already. So instead of merging this (which is adds something that will not work in many setups anyway), wouldn't it be better to just document how to add the ironic network to the network.json (similar to how we do already for bastion)? I am still not convinced that we should add this by default.

skazi0:
@rhafer & @vuntz this is something you need to decide. I will adjust the code to whatever option you pick.

rhafer:
@vuntz I think it's ultimately your decision to either merge or close this. I raised my concerns (which you seem to share to some extend), but I promise to not get mad at anyone if this this is merged ;) .

vuntz:
After giving it some thoughts, I'd go the documentation way for now. But we need to make sure the ironic barclamp does some validation that the network proposal has the required network defined -- we do that in neutron already for additional external networks, so should be easy to copy & paste.

skazi0:
Can you point me to the documentation where it's described how to configure such networks (e.g. bastion network)?

vuntz:
> Can you point me to the documentation where it's described how to configure such networks (e.g. bastion network)?

https://www.suse.com/documentation/suse-openstack-cloud-6/book_cloud_deploy/data/sec_depl_inst_admserv_post_network.html (for external networks)

https://www.suse.com/documentation/suse-openstack-cloud-6/book_cloud_deploy/data/sec_depl_adm_inst_crowbar_mode.html (for bastion, but it documents it with yast, so not helping)

PR #578, opened 2016-07-28
Evaluate routes from /var/lib/crowbar/routes.d/

rsalevsky:


crowbar_framework/config/routes.rb
```diff
@@ -17,10 +17,15 @@
 
 Rails.application.routes.draw do
   apipie
-  # Install route from each barclamp
-  Rails.root.join("config", "routes.d").children.each do |routes|
-    eval(routes.read, binding) if routes.extname == ".routes"
-  end if Rails.root.join("config", "routes.d").directory?
+
+  route_dirs = [Pathname.new("/etc/crowbar/routes.d/"), Rails.root.join("config", "routes.d")]
```
MaximilianMeister: not sure if the routes would end up in the backup. I guess that would be wrong then

MaximilianMeister:
+1

crowbar_framework/lib/crowbar/backup/base.rb
```diff
@@ -42,8 +42,16 @@ def restore_files
...
+              "data/backup",
+              "/var/lib/crowbar/backup"
+            ],
+            [
+              "data/cache",
+              "/var/lib/crowbar/cache"
+            ],
+            [
+              "data/config",
+              "/var/lib/crowbar/config"
```
Itxaka: not sure if its missing `/var/lib/crowbar/install/` ?

Itxaka:
small question from my side, lokks good otherwise.

crowbar_framework/lib/crowbar/backup/base.rb
rsalevsky: Yes it's missing but wanted in this case, we don't need this files because while restore we delete this files.

Itxaka:
+1

PR #581, opened 2016-08-01
don't fail on non existing databags

MaximilianMeister:


jsuchome:
Wouldn't it be better to add such check to `chef_data_bag_destroy` ?

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -323,9 +323,14 @@ def post_cleanup_repos
...
 
-        Crowbar::Repository.chef_data_bag_destroy(
-          "#{repository.data_bag_name}/#{repository.id}"
-        )
+        begin
+          Crowbar::Repository.chef_data_bag_destroy(
+            "#{repository.data_bag_name}/#{repository.id}"
+          )
+        rescue Net::HTTPServerException
+          # it is fine to pass if the databag doesn't exist
```
rhafer: Does this really work? According to /usr/lib64/ruby/2.1.0/net/http/responses.rb ```Net::HTTPServerError``` corresponds to errors from the 5xx range. AFAICS ```Crowbar::Repository.chef_data_bag_destroy``` returns 404 Not Found when an item does not exist.

In general I think it's better to catch a more specific error anyway. Just ignoring every Exception that ```Crowbar::Repository.chef_data_bag_destroy``` might throw seems wrong to me.

MaximilianMeister:
> Wouldn't it be better to add such check to chef_data_bag_destroy ?

depends if we want to pass each time when we call it. this would be only for our specific upgrade case, which I thought makes more sense

crowbar_framework/app/controllers/installer/upgrades_controller.rb
MaximilianMeister: yes it really works. we are doing it like that in multiple places where we make chef rest calls

MaximilianMeister: by the way, you have to look for `Net::HTTPServerException` rather that `Net::HTTPServerError`

rsalevsky:
+1

MaximilianMeister:
rebased to trigger another upgrade gating, which somehow failed earlier than the upgrade job. old upgrade gating was https://ci.suse.de/job/openstack-mkcloud/29294/console

jsuchome:
+1

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -323,9 +323,14 @@ def post_cleanup_repos
...
         )
 
-        Crowbar::Repository.chef_data_bag_destroy(
-          "#{repository.data_bag_name}/#{repository.id}"
-        )
+        begin
+          Crowbar::Repository.chef_data_bag_destroy(
+            "#{repository.data_bag_name}/#{repository.id}"
+          )
+        rescue Net::HTTPServerException
```
Itxaka: Shouldnt we try to catch an specific error code (404 in this case) rather than caching _any_ error code? That could lead to hiding issues behind a generic catch-all

MaximilianMeister: we should, but we can't. first we have to catch all `Net::HTTPServerException` and then if `404` is not in the error response code, we have to raise an Exception again in the rescue block.

I could change it, would that be fine with you? but then it would not be consistent anymore with all the other places where we handle databags, but we could make a follow up.

Itxaka: Then no, it sounds fine. I did expect we could directly catch a HTTPServerException404 or something similar :disappointed:

Itxaka:
+1

sayalilunkad:
+1

PR #582, opened 2016-08-03
Force reconnect to the database

rsalevsky:
In case the database connection gets cut, ActiveRecord will not recognize
that the connection is broken until the next access attempt. In this case
it will raise the StatementInvalid error.
In our case this happens when we apply a proposal e.g. provisioner on
the admin node.

For more information please check:
https://github.com/cantino/huginn/issues/1116
https://github.com/rails/rails/issues/12867

The full error message is:
`ActiveRecord::StatementInvalid: PG::ConnectionBad: PQconsumeInput()
server closed the connection unexpectedly`

### DIFF
--- a/crowbar_framework/config/initializers/active_record.rb
+++ b/crowbar_framework/config/initializers/active_record.rb
@@ -1,0 +1,33 @@
+#
+# Copyright 2016, SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# FIXME: WORKAROUND
+# In case the database connection gets cut, ActiveRecord will not recognize that the connection is
+# broken until the next access attempt. In this case it will raise the StatementInvalid error.
+# In our case this happens when we apply a proposal e.g. provisioner on the admin node.
+module ActiveRecord
+  class Base
+    def save(*args)
+      super
+    rescue ActiveRecord::StatementInvalid => e
+      raise e unless e.original_exception.class == PG::ConnectionBad
+
+      Rails.logger.warn("Database connection broken, force ActiveRecord database reconnect...")
+      ActiveRecord::Base.connection.reconnect!
+      super
+    end
+  end
+end


MaximilianMeister:
It works, but I think it is a really ugly workaround, because we will always run knowingly into the Exception once we are saving and committing a proposal on the admin... 

What about updating Rails to `> 4.0.1` instead of this hack?

At least we should be 100% clear which service restart, or whatever it is that happens in the chef-client that causes the disconnection. And maybe we can work on that bit rather than here.

rsalevsky:
@MaximilianMeister  We have rails 4.2.2... What I want to say with the links is that the rails developers don't see this as a bug.

MaximilianMeister:
Another thing that worries me is that we don't really know if this will happen only on `save` or if it could happen on `edit` or `delete` as well...

and nevertheless in this case we only should catch only `PG::ConnectionBad`:
```ruby
module ActiveRecord
  class Base
    def save
      super
    rescue ActiveRecord::StatementInvalid => e
      raise e unless e.original_exception.class == PG::ConnectionBad

      Rails.logger.debug("ActiveRecord force database reconnect...")
      ActiveRecord::Base.connection.reconnect!
      super
    end
  end
￼end
```

crowbar_framework/config/initializers/active_record.rb
```diff
@@ -0,0 +1,33 @@
...
+# limitations under the License.
+#
+
+# FIXME: WORKAROUND
+# In case the database connection gets cut, ActiveRecord will not recognize that the connection is
+# broken until the next access attempt. In this case it will raise the StatementInvalid error.
+# In our case this happens when we apply a proposal e.g. provisioner on the admin node.
+module ActiveRecord
+  class Base
+    def save
```
MaximilianMeister: http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-save

rsalevsky:
@rhafer Do you have an idea why the network connection breaks while the initial chef-client run? All services (crowbar and postgresql) are assuming the connection is still alive but it is broken. Seams that the connection get's cuttet without notifications to the services.

rhafer:
> Do you have an idea why the network connection breaks while the initial chef-client run? All services (crowbar and postgresql) are assuming the connection is still alive but it is broken. Seams that the connection get's cuttet without notifications to the services.

Hard to tell without any logs (or a system where this is actually happening). Probably it needed to reconfigure something on the interfaces. Which "initial chef-client run" exactly are you referring to? Is it really the first one that's happening during the installer run (the one with an empty run list) or is it one of the later ones that happen while transitioning the nodes through the different states?

But somehow I think that ActiveRecord (or the App) need to be able to deal with this kind of problems. Network connection can go away at any point. BTW, does this also happen if you just restart postgresql?

rsalevsky:
It happens during the chef-client run which also configures the bind service. But there are no useful logs just the normal chef client log which looks normal.

The ActiveRecord team says the will not fix this because it is an unexpected system behavior which should not happen. Restarting  postgresql work's because postgresql closes the connection properly before stopping. In this bug the connections is still open but broken because of some system changes.

vuntz:
It's surprising that the ActiveRecord team doesn't consider this a bug in ActiveRecord... It's in no way specific to our setup and could happen due to things unrelated to what Crowbar does, so I think we need a workaround like this one.

+1

vuntz:
Ah, commented too fast: I think @MaximilianMeister is right that `save` might not be the only method impacted...

jsuchome:
I've added this PR to https://github.com/crowbar/crowbar-ha/pull/137, which previously failed with https://bugzilla.novell.com/show_bug.cgi?id=995088 . Now, with this PR in UPDATEREPOS, it seems to build correctly: https://ci.suse.de/job/openstack-mkcloud/31941/

MaximilianMeister:
> I've added this PR to crowbar/crowbar-ha#137,

@jsuchome you don't need to add the patch from github, it is in the package currently

rhafer:
> It's surprising that the ActiveRecord team doesn't consider this a bug in ActiveRecord... It's in no way specific to our setup and could happen due to things unrelated to what Crowbar does, so I think we need a workaround like this one.

I am not sure if really happens in many other cases then what is described in https://github.com/crowbar/crowbar-core/issues/616. If the connection are closed by the server or because of some networking issue it seems that ActiveRecord actually does try to reconnect.

But I guess having this workaround doesn't hurt too much. (Though obviouly can hide valid bugs like the one mentioned above).

rsalevsky:
I removed the patch from the package. Let's see if we see new problems if not I would say let's close this PR.

vuntz:
It seems we're good, so let's close.

PR #583, opened 2016-08-04
remove any traces of sqlite for production

MaximilianMeister:


crowbar_framework/config/boot.rb
```diff
@@ -85,9 +85,6 @@
   gem "simple_navigation_renderers", version: "~> 1.0.2"
   require "simple_navigation_renderers"
 
-  gem "sqlite3", version: "~> 1.3.9"
```
rsalevsky: We need to replace this with pg, just removing is not enough.

rsalevsky:
+1

jgrassler:
+1

PR #584, opened 2016-08-04
only run remote chef client once per node

MaximilianMeister:
when applying the provisioner proposal, where base and server are
on the admin node, the list of nodes says:
`[["admin.crowbar.com"], ["admin.crowbar.com"]]`
which actually triggers two chef client runs in parallel on the admin

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1271,6 +1271,8 @@ class ServiceObject
     pre_cached_nodes = {}
 
     # Each batch is a list of nodes that can be done in parallel.
+    # only run remote chef-clent on a node once
+    batches.uniq!
     batches.each do |nodes|
       @logger.debug "Applying batch: #{nodes.inspect}"
 


rsalevsky:
+1

jsuchome:
It might not be relevant, and I'm not reading the code closely, but I've just seen here https://github.com/MaximilianMeister/crowbar-core/blob/e8a6fbe2ab5df2a6998e6f566b50f19844164b9e/crowbar_framework/app/models/service_object.rb#L1245 about the need (?) to run the deployment twice...

MaximilianMeister:
@jsuchome good point, i guess @vuntz could know that instantaneously ... i am also not sure btw

rhafer:
> It might not be relevant, and I'm not reading the code closely, but I've just seen here https://github.com/MaximilianMeister/crowbar-core/blob/e8a6fbe2ab5df2a6998e6f566b50f19844164b9e/crowbar_framework/app/models/service_object.rb#L1245 about the need (?) to run the deployment twice...

AFAIK that is still a relict from the time when crowbar re-ran chef-client as 2nd time after it failed once. I.e. it's from before this got merged: https://github.com/crowbar/crowbar-core/commit/f229d43f3e5be18043876dce95ece26c50f115ca
That part of the comment is wrong now. It should be removed. Could you do that? Separate PR (or at least commit) please.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1271,6 +1271,8 @@ def apply_role(role, inst, in_queue, bootstrap = false)
     pre_cached_nodes = {}
 
     # Each batch is a list of nodes that can be done in parallel.
+    # only run remote chef-clent on a node once
+    batches.uniq!
```
rhafer: AFAICS ```batches``` is a list of a lists. So I don't think this really does what you want it to do. E.g. assuming you have a setup of two nodes where node1 is member of ```provisioner-server``` and ```provisioner-base``` and node2 just ```provisioner-base```. Won't batches then look like this: ```[["node1"], ["node1", "node2"]]```? Calling ```uniq!``` on that list does not have any effect.

MaximilianMeister: yes, correct. in this case it will not have any effect (maybe it even shouldn't).
I only had the case where the list looked like `[["node1"], ["node1"]]`, that's why I sent this patch

MaximilianMeister:
> AFAIK that is still a relict from the time when crowbar re-ran chef-client as 2nd time after it failed once. I.e. it's from before this got merged: f229d43
> That part of the comment is wrong now. It should be removed. Could you do that? Separate PR (or at least commit) please.

okay thanks for the hint

crowbar_framework/app/models/service_object.rb
rsalevsky: I think `flatten` should do the trick. Then we can also remove the outside loop.

rhafer: > I think flatten should do the trick. Then we can also remove the outside loop.

No. There's a reason to do the run in batches per role (using flatten would break that).  E.g. for the neutron barclamp (and most other barclamps featuring multiple roles)  it's essential that the nodes having the neutron-server role have finished their chef-client run before the run for the neutron-agents nodes is started.

vuntz:
I don't believe this is correct; it just happens that we have only one node, and two roles, but sometimes we do depend on this behavior and the second chef run does something really different (in swift, for instance).

The right fix is really to only apply just the role for a given batch, and not all roles. I have branches for that (see my hackweek project).

So overall, -1.

PR #585, opened 2016-08-05
Fix when node roles are empty

abel-navarro:
We found that sometimes node roles are empty in the hardware-installing
phase. This leads to an error that prevents the nodes to be properly
installed and remain just discovered forever.

### DIFF
--- a/crowbar_framework/app/models/deployer_service.rb
+++ b/crowbar_framework/app/models/deployer_service.rb
@@ -77,7 +77,8 @@ class DeployerService < ServiceObject
     if state == "hardware-installing"
       node = NodeObject.find_node_by_name(name)
       # build a list of current and pending roles to check against
-      roles = node.roles.dup
+      roles = []
+      roles = node.roles.dup if node.roles
       unless node.crowbar["crowbar"]["pending"].nil?
         roles.concat(node.crowbar["crowbar"]["pending"].values)
       end


jsuchome:
Note that this is a partial revert of https://github.com/crowbar/crowbar-core/commit/84827646b66d1470d58ef7cd7b81d3e300d0cb49

jsuchome:
+1

jsuchome:
> We found that sometimes node roles are empty 

Actually the problem is that the roles might be Nil at the beginning, which leads to failures.

Itxaka:
+1

crowbar_framework/app/models/deployer_service.rb
```diff
@@ -77,7 +77,8 @@ def transition(inst, name, state)
     if state == "hardware-installing"
       node = NodeObject.find_node_by_name(name)
       # build a list of current and pending roles to check against
-      roles = node.roles.dup
+      roles = []
+      roles = node.roles.dup if node.roles
```
rhafer: Isn't this a classic case for the tenary operator?

```roles =  node.roles ? node.roles.dup : [] ```

I think that would be somewhat cleaner, but if you prefer your code. Feel free to merge.

abel-navarro: i knew there was a more elegant way to do it, but i'm no ruby expert and that's the best I could came up with.

If the change works I'll submit a PR with that syntax.

PR #587, opened 2016-08-08
Return run_list_to_roles method, to prevent empty roles list.

jsuchome:
This function was removed before, but now we are hitting the error that the roles list of a node is sometimes empty during transition to hardware-installing.

UPDATE: this is the error we're trying to fix:
```
D, [2016-08-08T07:02:08.344171 #11272:0x007f248dcd5d08] DEBUG -- Deployer transition: entering: d52-54-77-77-77-05.vc2.cloud.suse.de for hardware-installing
F, [2016-08-08T07:02:08.393637 #11272:0x007f248dcd5d08] FATAL -- json/transition for deployer:default failed: undefined method `[]=' for nil:NilClass
F, [2016-08-08T07:02:08.393966 #11272:0x007f248dcd5d08] FATAL -- /opt/dell/crowbar_framework/app/models/deployer_service.rb:103:in `transition'
...
```
This reverts parts of https://github.com/crowbar/crowbar-core/pull/515. @nicolasbock @MaximilianMeister , you've reviewed original PR, could you also take a look here?

### DIFF
--- a/crowbar_framework/app/models/deployer_service.rb
+++ b/crowbar_framework/app/models/deployer_service.rb
@@ -82,6 +82,8 @@ class DeployerService < ServiceObject
       unless node.crowbar["crowbar"]["pending"].nil?
         roles.concat(node.crowbar["crowbar"]["pending"].values)
       end
+      roles << node.run_list_to_roles
+      roles.flatten!
 
       # Walk map to categorize the node.  Choose first one from the bios map that matches.
       role = RoleObject.find_role_by_name "#{@bc_name}-config-#{inst}"
@@ -99,9 +101,11 @@ class DeployerService < ServiceObject
         break if done
       end
 
-      os_map = role.default_attributes["deployer"]["os_map"]
-      node.crowbar["crowbar"]["hardware"]["os"] = os_map[0]["install_os"]
-      node.save
+      unless node.crowbar["crowbar"]["hardware"].nil?
+        os_map = role.default_attributes["deployer"]["os_map"]
+        node.crowbar["crowbar"]["hardware"]["os"] = os_map[0]["install_os"]
+        node.save
+      end
     end
 
     # The discovery image needs to have clients cleared.
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -759,6 +759,16 @@ class NodeObject < ChefObject
     old_run_list != crowbar_run_list.run_list_items
   end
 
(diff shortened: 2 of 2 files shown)


crowbar_framework/app/models/deployer_service.rb
```diff
@@ -82,6 +82,7 @@ def transition(inst, name, state)
       unless node.crowbar["crowbar"]["pending"].nil?
         roles.concat(node.crowbar["crowbar"]["pending"].values)
       end
+      roles << node.run_list_to_roles
```
abel-navarro: there's a ```roles.flatten!``` line after this one in the original PR. We may need that one (still investigating it).

jsuchome: I've added it meanwhile

abel-navarro:
+1

aspiers:
Ideally the commit message would refer to the commit(s) it's reverting part of; perhaps you could add that in?

jsuchome:
Commit message updated

aspiers:
Thanks!

crowbar_framework/app/models/deployer_service.rb
```diff
@@ -99,9 +101,11 @@ def transition(inst, name, state)
         break if done
       end
 
-      os_map = role.default_attributes["deployer"]["os_map"]
-      node.crowbar["crowbar"]["hardware"]["os"] = os_map[0]["install_os"]
-      node.save
+      unless node.crowbar["crowbar"]["hardware"].nil?
```
MaximilianMeister: do you have a link to the original error?

and not sure but how can you assume that `role.default_attributes["deployer"]["os_map"]` is not nil when node.crowbar["crowbar"]["hardware"] is not nil?

jsuchome: Error is here http://pastebin.nue.suse.com/18621/src

It should be fixed mostly by the return of `roles << node.run_list_to_roles`, this extra check for nil is just a failsafe.

jsuchome: It's `node.crowbar["crowbar"]["hardware"]` that is supposed to be filled on lines 94-96, so I focused on this check

aspiers:
I don't fully understand this yet, but you probably don't need my +1 anyway :)

aspiers:
Oops, and LGTM thinks I just gave a plus one vote ...

nicolasbock:
+1

PR #590, opened 2016-08-09
Fix check if cookbook is already uploaded (bsc#980888)

rsalevsky:


crowbar_framework/lib/crowbar/chef/upload.rb
```diff
@@ -227,15 +227,12 @@ def local_cookbook_md5_checksums(cookbook)
...
-        local_md5sums = local_cookbook_md5_checksums(cookbook)
-        remote_md5sums = api_cookbook_md5_checksums(cookbook.name)
-        combined_md5sums = local_md5sums.deep_merge(remote_md5sums)
-
-        combined_md5sums.reject! { |cb, _| local_md5sums.include? cb }
-        return false unless combined_md5sums.empty?
-
-        logger.info("Cookbook #{cookbook.name} already exists, skipping...")
-        true
+        if local_cookbook_md5_checksums(cookbook) == api_cookbook_md5_checksums(cookbook.name)
```
nicolasbock: Why is this so much simpler now? I mean, did something change in what `*_cookbook_md5_checksums()` return?

rsalevsky: Nop, but when I wrote the code I tough way too complex. I did forgot that it's possible to compare hashs in ruby without caring about the order. I tried to implement something which checks if a hash is the same but `deep_merge` did not work as expected so at the end it was always true.

nicolasbock: Ah, this looks way better now :smile:

nicolasbock:
+1

nicolasbock:
Gating is failing. Could you have a look?

jsuchome:
+1

MaximilianMeister:
+1

PR #592, opened 2016-08-12
Do not create a postgresql db dump when upgrading

rhafer:
It's not needed currently. Note: This commit just removes the chef part of the
database dump. It's just enough to make the upgrade work again. The crowbar
side (i.e. the steps in the upgrade wizard) is not changed.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -86,55 +86,6 @@ when "dump_openstack_database"
 
   include_recipe "crowbar::stop-services-before-upgrade"
 
-  dump_location = node[:crowbar][:upgrade][:db_dump_location]
-
-  # Ensure the dump_location is saved in DB nodes even after we later drop the role
-  node.set[:crowbar][:upgrade][:db_dump_location] = dump_location
-
-  # If postgres is not running here, it means we're in the DB cluster and database runs
-  # on the other node: let the other node take care of the rest.
-  unless ::Kernel.system("service postgresql status")
-    if node[:crowbar][:upgrade][:db_dumped_here]
-      node[:crowbar][:upgrade].delete :db_dumped_here
-    end
-    node.save
-    return
-  end
-
-  # Check the available space before the dump
-  ruby_block "check available space" do
-    block do
-      require "pathname"
-      path = Pathname.new(dump_location).split.first.to_s
-      available = `df -m #{path} | tail -n 1 | sed "s/ \\+/ /g" | cut -d " " -f 4`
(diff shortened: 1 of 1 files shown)


aplanas:
Is crowbar also doing a dump somewhere else?

+1 if this is the case

jsuchome:
+1

rsalevsky:
@jsuchome Didn't we said the we want to dump the database and store it the admin node in case something goes wrong? At least this is what we discussed with the design team and what they plan for the UI.

sayalilunkad:
I also think it's a good idea to keep the dump in case something goes wrong.

jsuchome:
@rsalevsky hm, it's true we have OpenStack database backup in our notes and mentioned in the angular wiki. @rhafer maybe we can still keep the dump and just remove automatic restore from the dump?

rhafer:
> hm, it's true we have OpenStack database backup in our notes and mentioned in the angular wiki.

But wouldn't that database backup implementation look completely different to what's implemented in the old upgrade procedure?

> maybe we can still keep the dump and just remove automatic restore from the dump?

Well, https://github.com/crowbar/crowbar-openstack/pull/495 should work without this change here. I just thought leaving that code lying around basically unused it probably a bad idea.

jsuchome:
> But wouldn't that database backup implementation look completely different to what's
> implemented in the old upgrade procedure?

Maybe, maybe not. So far, I still think we have to use chef recipes to access the clients, so the database dump might look pretty similar.

> Well, crowbar/crowbar-openstack#495 should work without this change here. 

That's what I meant.

rhafer:
> Maybe, maybe not. So far, I still think we have to use chef recipes to access the clients, so the database dump might look pretty similar.

If you want to dump to the admin node as @rsalevsky suggested it might make more sense to just  to that via RemoteNode.ssh_cmd (https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/remote_node.rb#L82). But anyway, feel free to close this if you don't need it. The other PR should be enough to fix the immediate issue.

jsuchome:
Yes, ssh commands might be our option, that's true.

rsalevsky:
As Ralf already said in the other PR it's git so we can restore the change whenever we want. So let's merge the PR for now and add it as reference to the card where we do the actual implementation for the OpenStack DB dump.

+1

rhafer:
@jsuchome  Do you agree with @rsalevsky 's last comment, or would you prefer to close this?

jsuchome:
I agree with Rick, yes. Sorry if it caused delay.
+1

rhafer:
> Sorry if it caused delay.

Don't worry. It didn't cause any delay. (The other PR was the more important part)

PR #593, opened 2016-08-12
Automatically register modules into crowbar

rsalevsky:
This change is required to include modules like crowbar-init into
crowbar without adapting the crowbar source code for this.


MaximilianMeister:
+1

nicolasbock:
+1

PR #594, opened 2016-08-15
Remeove sqlite monkey patch

rsalevsky:
We use now postgresql by default so this patch is not longer required.


MaximilianMeister:
duplicate of https://github.com/crowbar/crowbar-core/pull/583

PR #595, opened 2016-08-15
remove sqlite monkeypatch

MaximilianMeister:
as part of the sqlite3 cleanup while switching to postgres

split from https://github.com/crowbar/crowbar-core/pull/583 because of gating issues when removing a file (it seems the patch command can't handle that)


rsalevsky:
+1

aplanas:
+1

rsalevsky:
Gating failure is unrelated here...

PR #596, opened 2016-08-16
Added action to initiate the upgrade server process

jsuchome:
This is preliminary location. In needs to be bound to the new upgrade API...

Depends on https://github.com/crowbar/crowbar/pull/2258

### DIFF
--- a/crowbar_framework/app/controllers/installer/upgrades_controller.rb
+++ b/crowbar_framework/app/controllers/installer/upgrades_controller.rb
@@ -91,6 +91,29 @@ module Installer
       end
     end
 
+    # This will initiate the upgrade of admin server.
+    # Once this function is called, server packages will be upgraded to latest versions.
+    # Right after all packages are upgraded, server will be rebooted.
+    # FIXME: We're assuming the preliminary checks have been run already, esp. the checks
+    # for presence of new product repositories.
+    def upgrade_admin_server
+      if request.post?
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -91,6 +91,29 @@ def start
       end
     end
 
+    # This will initiate the upgrade of admin server.
+    # Once this function is called, server packages will be upgraded to latest versions.
+    # Right after all packages are upgraded, server will be rebooted.
```
jsuchome: Honestly, I do not see what's wrong here.

MaximilianMeister:
@jsuchome the new API is now available, just add it to:
https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/controllers/api/crowbar_controller.rb#L35

jsuchome:
@MaximilianMeister as mentioned on #irc, it needs the API in 3.0 branch

rsalevsky:
@jsuchome Not sure, but makes it sense to have it first in master and then create a backport? Otherwise we will have for the C8 upgrade the same problem that we no code at all.

MaximilianMeister:
@jsuchome I moved the code to https://github.com/crowbar/crowbar-core/issues/604 
thanks for the initial work

PR #597, opened 2016-08-16
migrate from proposal instead of role

MaximilianMeister:
SchemaMigration.migrate_proposal_from_json needs a proposal hash and
cannot deal with any Chef::Role as the attributes is called
default_attributes (and could be different) and there is no deployment
information either

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade.rb
+++ b/crowbar_framework/lib/crowbar/upgrade.rb
@@ -60,7 +60,7 @@ module Crowbar
     def create_network_json_3_0
       Rails.logger.debug "Re-creating network.json from upgrade data"
       network_role_path = @data.join("knife", "roles", "network-config-default.json")
-      create_network_json(network_role_path)
+      create_network_json(network_role_path, from_role: true)
     end
 
     def upgrade_knife_files_1_9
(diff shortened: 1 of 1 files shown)


rhafer:
This will break ```upgrade_knife_files_1_9``` I guess?

MaximilianMeister:
@rhafer arrg, good catch. I guess that would have gone unnoticed if we don't backport the patch... but I will adapt it, thanks

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -115,8 +115,20 @@ def upgrade_crowbar_files_1_9
...
-      json = JSON.load(json_path.read)
+    def create_network_json(json_path, options = {})
+      from_role = options.fetch(:from_role, false)
+      json = if from_role
+        # get network proposal data from database.yml
+        proposals = YAML.load_file(
+          @data.join("crowbar", "database.yml")
+        )
+        network_proposal = proposals["proposals"]["records"].detect do |p|
+          p[1] == "network" && p[2] == "default"
```
rhafer: Are you sure that the columns are always in the same order in that file? Or would you need to use ```proposals["proposals"]["columns"]``` to find the correct index for the array element?

MaximilianMeister: It is always the same order, like defined in the `schema.rb` database schema. and I just looked at 2 different database.yml files. so I guess it is not necessary to map it to the columns.

rhafer:
+1

rsalevsky:
+1

PR #598, opened 2016-08-16
Susemanager fixes backport

Itxaka:
Backport of https://github.com/crowbar/crowbar-core/pull/508

Unfortunately the changes werent squashed so I had to do 5 cherry-picks. I can squash them if needed.


jsuchome:
I'd say squash them - the oldest ones were later rewritten anyway, so it is confusing. I should have done it myself :-(

jsuchome:
If we release this already for Cloud6, we should remove the migration bits from Cloud7, right?

rhafer:
> If we release this already for Cloud6, we should remove the migration bits from Cloud7, right?

Usually we adjust the migrations in master so that they also work when the migration has already been run on the previous release. E.g. like this:
https://github.com/crowbar/crowbar-openstack/commit/98489a02c7d542a99d1a599592ff67cd7c60cf3b

Itxaka:
@rhafer @jsuchome added migration protection against having it run several times. Hoping is correct?

jsuchome:
Oh, sorry, I probably confused you. This check will need to be added to relevant migration parts of Cloud7...

Itxaka:
@jsuchome indeed its not from here :o

aspiers:
I think I'd disagree - squashing them makes it much harder to see where they came from, and what is the advantage to squashing?  There is none.

nicolasbock:
+1

aplanas:
+1

Itxaka:
I did some testing of this but could not make it work properly. Will try to redo the tests from zero to see if this time I have a better view on how it works.

Itxaka:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;K0h2uAEO&#x2F;116-p2-bsc-996542-l3-suse-manager-3-0-barclamp-for-suse-openstack-cloud-6-sbsa"> P2: bsc#996542 - L3: SUSE Manager 3.0 barclamp for SUSE Openstack Cloud 6 (SBSA)</a>

Itxaka:
Tested on cloud6. It works but after finishing the deploy, the barclamp does not "finish" per se. With this I mean, the proposal webUI doesnt not move to the finished state, instead it keeps on "processing, please wait" indefinitely.

Probably something has changed between cloud6 and 7 on the crowbar UI for the finished signal.

vuntz:
I think we can merge: we have the +1, it was tested, so it's good to go in.

PR #599, opened 2016-08-16
dump schema.rb in production

MaximilianMeister:
this will be used by YamlDB to import the schema.rb and the data.yml
dump into the new postgresql database during the upgrade from 6 to 7


rsalevsky:
+1

nicolasbock:
+1

PR #602, opened 2016-08-17
[stable/3.0] Change permits to be backwards compatible

MaximilianMeister:
since we changed from httparty to rest-client, we require a different
parameter structure for file uploads.
to be in line with the update in crowbar-client, and to avoid having
to patch the content section in the client, we adapt the permits on
the server side


rsalevsky:
+1

jsuchome:
+1

nicolasbock:
+1

PR #604, opened 2016-08-18
implement crowbar admin server upgrade step

MaximilianMeister:
Co-Authored-By: Jiří Suchomel <jsuchome@suse.com>

supersedes: https://github.com/crowbar/crowbar-core/pull/596 (once it is backported to 3.0)


rsalevsky:
+1

MaximilianMeister:
fixed return of `upgrade!` to return true or false
added model specs

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -14,7 +14,72 @@
...
+      else
+        msg = "Could not find #{upgrade_script}"
+        Rails.logger.error(msg)
+        errors.add(:upgrade, msg)
+
+        false
+      end
+    end
+
+    def maintenance_updates_installed?
```
jsuchome: This needs some more work, I assume. Specific return values based on the various exit status of "zypper patch-check", logging, etc.

MaximilianMeister: now that you say, that was not fully correct indeed. I guess we just need to check for those two:

```
       100 - ZYPPER_EXIT_INF_UPDATE_NEEDED
           Returned by the patch-check command if there are patches available for installation.

       101 - ZYPPER_EXIT_INF_SEC_UPDATE_NEEDED
           Returned by the patch-check command if there are security patches available for
           installation.
```

do you have anything else in mind what we need to check for?

jsuchome: I agree that both 100 and 101 return values mean we should not proceed with the upgrade. 

As for other checks that should be done before the upgrade, I think we will need to check if SLES12SP2 and Cloud7 repo are present. But I do not think it needs to be part of this PR (empty function would be nice though).

MaximilianMeister: yes, that was still missing, I agree another PR would be better, as this extends the API with another endpoint

MaximilianMeister: > I agree that both 100 and 101 return values mean we should not proceed with the upgrade.

updated

jsuchome: I think you can add some logging - or maybe different return values for `maintenance_updates_installed?`, we will need to tell user about specific failure

MaximilianMeister: anyways it should be called `maintenance_updates_missing?`

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -14,7 +14,90 @@
...
+      else
+        msg = "Could not find #{upgrade_script}"
+        Rails.logger.error(msg)
+        errors.add(:upgrade, msg)
+
+        false
+      end
+    end
+
+    def maintenance_updates_missing?
```
MaximilianMeister: > I think you can add some logging - or maybe different return values for maintenance_updates_installed?, we will need to tell user about specific failure

Added some specific logging, but for the business logic I think for those two cases it doesn't really make a difference for the user which updates are missing. both need to be installed anyways.

jsuchome:
+1

rsalevsky:
+1

jsuchome:
Please backport to 3.0

MaximilianMeister:
@jsuchome need to merge https://github.com/crowbar/crowbar-core/pull/608 first

MaximilianMeister:
> @jsuchome need to merge #608 first

backported the commits within #608 now

MaximilianMeister:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;xeN0LnPm&#x2F;95-5-s47p9-implement-api-for-preparing-nodes-for-upgrade-upgrade-step-2"> (5) S47P9 Implement API for preparing nodes for upgrade (upgrade step 2)</a>

PR #612, opened 2016-08-23
Simple check if HA clusters report some problems

jsuchome:
We should do such check before we initiate the upgrade

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -296,6 +296,23 @@ class CrowbarService < ServiceObject
     [200, { name: name }]
   end
 
+  # Simple check if HA clusters report some problems
+  # If there are no problems, empty hash is returned.
+  # If this fails, information about failed actions for each cluster founder is
+  # returned in a hash.
+  # User has to manually clean pacemaker resources before proceeding with the upgrade.
+  def check_cluster_health
+    cluster_founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
+    ret = {}
+    return ret if cluster_founders.empty?
+    check_if_nodes_are_available cluster_founders
+    cluster_founders.each do |n|
+      ssh_retval = n.run_ssh_cmd('crm status | grep "^Failed actions:"')
+      ret[n.name] = ssh_retval[:stdout] if ssh_retval[:exit_code] == 0
+    end
+    ret
+  end
+
   def prepare_nodes_for_crowbar_upgrade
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -296,6 +296,24 @@ def transition(inst, name, state)
...
+    return if cluster_founders.empty?
+    check_if_nodes_are_available cluster_founders
+    nodes_with_failed_actions = []
+    cluster_founders.each do |n|
+      ssh_retval = n.ssh_retval('crm status | grep "^Failed actions:"')
+      nodes_with_failed_actions.push n.name if ssh_retval
+    end
+    unless nodes_with_failed_actions.empty?
+      @logger.error("failed actions on clusters nodes: #{nodes_with_failed_actions.join(', ')}
+Clean pacemaker resources using 'crm resource cleanup' before proceeding with the upgrade.")
```
aspiers: I suspect Hound is silently complaining about this, and there's no good reason to break the indentation here. Please can you fix it? Thanks!

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -296,6 +296,24 @@ def transition(inst, name, state)
...
+      raise "'crm status' command reports some failed actions"
```
MaximilianMeister: do we really need to `raise` here? what's the benefit?
why not return a hash containing the nodes, errors, messages. that would be worth much more for the API consumer in the end.

it could look like
```ruby
{
  clusters_healthy: false,
  failed_nodes: [
    "node1.crowbar.com": {
      message: "crm status: Failed actions: This"
    },
    "node2.crowbar.com": {
      message: "crm status: Failed actions: That"
    }
  ]
}
```
we have all that useful information here, so why throw it away with a generic log message?

jsuchome: I just made it same as other upgrade related functions here (see `prepare_nodes_for_crowbar_upgrade` and others)

MaximilianMeister:
@jsuchome thanks for the updates.

can you https://github.com/blog/2224-change-the-base-branch-of-a-pull-request to `stable/3.0` for this PR, and when it is merged, we forward port the `cluster_health_check` to this model? https://github.com/crowbar/crowbar-ha/blob/master/crowbar_framework/app/models/api/cluster.rb

jsuchome:
@MaximilianMeister I think I'll just create new PR from 3.0, so I have it locally prepared for further modifications...

PR #614, opened 2016-08-23
check if HA clusters report some problems

jsuchome:
We should do such check before we initiate the upgrade

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -315,17 +315,38 @@ class CrowbarService < ServiceObject
   # Simple check if HA clusters report some problems
   # If there are no problems, empty hash is returned.
   # If this fails, information about failed actions for each cluster founder is
-  # returned in a hash.
+  # returned in a hash that looks like this:
+  # {
+  #     "crm_failures" => {
+  #             "node1" => "reason for crm status failure"
+  #     },
+  #     "failed_actions" => {
+  #             "node2" => "Failed action on this node"
+  #     }
+  # }
   # User has to manually clean pacemaker resources before proceeding with the upgrade.
   def check_cluster_health
-    cluster_founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
     ret = {}
+    crm_failures = {}
+    failed_actions = {}
+
+    cluster_founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
     return ret if cluster_founders.empty?
     check_if_nodes_are_available cluster_founders
+
     cluster_founders.each do |n|
+      ssh_retval = n.run_ssh_cmd("crm status 2>&1")
+      if ssh_retval[:exit_code] != 0
+        crm_failures[n.name] = ssh_retval[:stdout]
+        next
+      end
       ssh_retval = n.run_ssh_cmd('crm status | grep "^Failed actions:"')
-      ret[n.name] = ssh_retval[:stdout] if ssh_retval[:exit_code] == 0
+      if ssh_retval[:exit_code] == 0
+        failed_actions[n.name] = ssh_retval[:stdout]
+      end
     end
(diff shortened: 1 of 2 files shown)


rsalevsky:
+1 but I can't say if the cluster test makes sensen or not

jsuchome:
This is the test we've used in 4-5 upgrade. We can't fix it for user, so we just have to tell him something looks wrong.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -312,6 +312,23 @@ def create_proposal
...
+  # If this fails, information about failed actions for each cluster founder is
+  # returned in a hash.
+  # User has to manually clean pacemaker resources before proceeding with the upgrade.
+  def check_cluster_health
+    cluster_founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
+    ret = {}
+    return ret if cluster_founders.empty?
+    check_if_nodes_are_available cluster_founders
+    cluster_founders.each do |n|
+      ssh_retval = n.run_ssh_cmd('crm status | grep "^Failed actions:"')
```
aspiers: `grep -q` ?  But you shouldn't assume that `crm status` succeeds.  If the cluster is dead, it won't output any failed actions.

aspiers:
BTW just wanted to make sure you are aware of https://trello.com/c/r09uuhvj/986-crowbar-health-check-api, which would be a better approach to this in the long-term.  I guess we probably won't have time to code it for SOC7 though :-/

crowbar_framework/app/models/crowbar_service.rb
jsuchome: Hm, I than I should probably do some basic check for crm status retval before.

jsuchome: `grep -q` was there originally, but than we thought it might be good to actually show user the real Failed action for given node (see `:stdout` later)

nicolasbock:
+1

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -312,6 +312,44 @@ def create_proposal
...
+  # {
+  #     "crm_failures" => {
+  #             "node1" => "reason for crm status failure"
+  #     },
+  #     "failed_actions" => {
+  #             "node2" => "Failed action on this node"
+  #     }
+  # }
+  # User has to manually clean pacemaker resources before proceeding with the upgrade.
+  def check_cluster_health
```
vuntz: Could we have that code in a different file? This really doesn't belong to the crowbar service :-) (should probably be in a lib)

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -312,6 +312,44 @@ def create_proposal
...
+    ret = {}
+    crm_failures = {}
+    failed_actions = {}
+
+    cluster_founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
+    return ret if cluster_founders.empty?
+    check_if_nodes_are_available cluster_founders
```
vuntz: Ah, this call means we'd need to move more code out of this file... Don't know if it's easy or not.

(in general, the whole upgrade code in this file is really something that should live somewhere else)

jsuchome: Yes, it might have sense to move everything upgrade related out of this, not just this method

vuntz:
From a quick look, LGTM. I just have the concern that we're adding stuff to crowbar_service.rb again that doesn't belong there. But maybe this should be fixed in a later PR.

rsalevsky:
@vuntz @jsuchome It's in this case just for C6, the code which is forwarded to C7 should go into the new [HA API](https://github.com/crowbar/crowbar-ha/blob/master/crowbar_framework/app/models/api/cluster.rb).

jsuchome:
Right, good point

PR #616, opened 2016-08-23
Run chef-client in a thread instead of a process (bnc#995027)

toabctl:
When the forked process (used to run the chef-client) exit, the database
connection is closed which leads to errors like:

PG::ConnectionBad: PQconsumeInput() server closed the connection unexpectedly

We currently catch this exception but we ran recently into something similar:

PG::UnableToSend: server closed the connection unexpectedly

https://bugzilla.suse.com/show_bug.cgi?id=995027

Hopefully this will be fixed when we use threads instead of processes.

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -21,6 +21,7 @@ require "json"
 require "hash_only_merge"
 require "securerandom"
 require "timeout"
+require "thwait"
 
 class ServiceObject
   include CrowbarPacemakerProxy
@@ -1273,24 +1274,33 @@ class ServiceObject
     batches.each do |nodes|
       @logger.debug "Applying batch: #{nodes.inspect}"
 
-      pids = {}
+      threads = {}
       nodes.each do |node|
         pre_cached_nodes[node] ||= NodeObject.find_node_by_name(node)
         next if pre_cached_nodes[node][:platform_family] == "windows"
         ran_admin = true if pre_cached_nodes[node].admin?
 
         filename = "#{ENV["CROWBAR_LOG_DIR"]}/chef-client/#{node}.log"
-        pid = run_remote_chef_client(node, "chef-client", filename)
-        pids[pid] = node
+        thread = run_remote_chef_client(node, "chef-client", filename)
+        threads[thread] = node
       end
-      status = Process.waitall
-      badones = status.select { |x| x[1].exitstatus != 0 }
 
-      next if badones.empty?
+      # wait for all running threads and collect the ones with a non-zero return value
+      bad_nodes = []
+      logger.debug "Waiting for #{threads.keys.length} threads to finish..."
+      ThreadsWait.all_waits(threads.keys) do |t|
+        logger.debug "thread #{t} for node #{threads[t]} finished (return value '#{t.value}')"
+        unless t.value == 0
+          bad_nodes << threads[t]
+        end
+      end
+
+      next if bad_nodes.empty?
 
       message = "Failed to apply the proposal to: "
-      badones.each do |baddie|
-        message = message + "#{pids[baddie[0]]}\n" + get_log_lines(pids[baddie[0]])
+      bad_nodes.each do |node|
+        message += "#{node}\n"
+        message += get_log_lines(node)
       end
       update_proposal_status(inst, "failed", message)
       restore_to_ready(applying_nodes)
@@ -1408,55 +1418,32 @@ class ServiceObject
     true
   end
 
-  #
-  # fork and exec ssh call to node and return pid.
-  #
+  # run the given command in a thread. the thread returns 0
+  # if the run was successfull
   def run_remote_chef_client(node, command, logfile_name)
-    Kernel::fork {
-      # Make sure all file descriptors are closed
-      ObjectSpace.each_object(IO) do |io|
-        unless [STDIN, STDOUT, STDERR].include?(io)
-          begin
-            unless io.closed?
-              io.close
-            end
-          rescue ::Exception
-          end
-        end
-      end
-
-      # Fix the normal file descriptors.
-      begin; STDIN.reopen "/dev/null"; rescue ::Exception; end
-      if logfile_name
-        begin
-          STDOUT.reopen logfile_name, "a+"
-          File.chmod(0644, logfile_name)
-          STDOUT.sync = true
-        rescue ::Exception
-          begin; STDOUT.reopen "/dev/null"; rescue ::Exception; end
-        end
-      else
-        begin; STDOUT.reopen "/dev/null"; rescue ::Exception; end
-      end
-      begin; STDERR.reopen STDOUT; rescue ::Exception; end
-      STDERR.sync = true
-
+    Thread.new do
       # Exec command
       # the -- tells sudo to stop interpreting options
 
-      ssh = ["sudo", "-u", "root", "--", "ssh", "-o", "TCPKeepAlive=no", "-o", "ServerAliveInterval=15", "root@#{node}"]
-      ssh_cmd = ssh.dup << command
+      ssh_cmd = ["sudo", "-u", "root", "--", "ssh", "-o", "TCPKeepAlive=no",
+                 "-o", "ServerAliveInterval=15", "root@#{node}"]
+      ssh_cmd << command
 
       # check if there are currently other chef-client runs on the node
       wait_for_chef_clients(node, logger: false)
       # check if the node is currently rebooting
       wait_for_reboot(node)
 
-      exit(1) unless system(*ssh_cmd)
-
-      # check if we need to wait for a node reboot
-      wait_for_reboot(node)
-    }
+      ret = 0
+      open(logfile_name, "a") do |f|
+        if system(*ssh_cmd, out: f, err: f)
+          wait_for_reboot(node)
+        else
+          ret = 1
+        end
+      end
+      ret
+    end
   end
 
   def wait_for_chef_daemons(node_list)


aplanas:
And use a connection pool? A helper that store N connections, and if the pool is empty, returns a new one.

MaximilianMeister:
@toabctl good catch! thanks a lot for working on this, this might well cause the bug

maybe have a look at https://github.com/crowbar/crowbar-core/blob/291213be69ed3023274522c80d80334713416554/crowbar_framework/lib/crowbar/backup/restore.rb#L32 where we use the builtin rails threading, and see if that would work too.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1408,39 +1419,11 @@ def add_role_to_instance_and_node(barclamp, instance, name, prop, role, newrole)
...
-        rescue ::Exception
-          begin; STDOUT.reopen "/dev/null"; rescue ::Exception; end
-        end
-      else
-        begin; STDOUT.reopen "/dev/null"; rescue ::Exception; end
-      end
-      begin; STDERR.reopen STDOUT; rescue ::Exception; end
-      STDERR.sync = true
-
+    Thread.new {
```
MaximilianMeister: ah i see you actually already use it here... my bad

aspiers:
Sorry, I find the commit message a bit cryptic but maybe I'm just being stupid.  Which database connection is closed, in which process, and why?  And why would switching to threads help?

nicolasbock:
Can you think of a way to test this hypothesis before we merge this? I don't understand why using threads as opposed to forking is helping.

aplanas:
@nicolasbock @aspiers I can understand that if the process owner of the connection dies, the fd for the socket is closed. So if we use a single process with multiple threads, this fd is not closed when a thread ends.

But what I do not understand is why the owner of the connection is not the main process, but one of the child. This IMHO is a design problem. Hence the pool suggestion.

Also I concur with @nicolasbock to test it before : )

nicolasbock:
Thanks @aplanas, now it's a little clearer. After thinking about this a bit I have this mental model now:
The main process is the owner of the file descriptor. It creates the DB connection before the `fork()`s and the children all get a duplicated `fd`. That `fd` is closed by a dying child which then of course affects the other children and the main process as well. Is that what you were thinking of?

MaximilianMeister:
@aplanas @nicolasbock that's what we have gating for, just look at the logs for https://github.com/crowbar/crowbar-core/pull/582/files#diff-aa14186bf08406349d780bba0d7cfb72R28
or any other PG:: Error

I just found some links by googling "rails postgres process fork" that may help understanding the problem:
https://devcenter.heroku.com/articles/forked-pg-connections
http://stackoverflow.com/questions/27892975/what-causes-activerecord-breaking-postgres-connection-after-forking

aplanas:
@nicolasbock I think that you nailed. Is the dup of the fd, that once is closed, break the connection for the rest. This explain the ownership problem.

@MaximilianMeister did you try the solution from heroku that you pointed? : disconnect before the fork and reconnect after that. Looks like that doing that the children will discover that there is no database connection, and will create a new one for this process.

MaximilianMeister:
> did you try the solution from heroku that you pointed? : disconnect before the fork and reconnect after that. Looks like that doing that the children will discover that there is no database connection, and will create a new one for this process.

@aplanas no, I haven't tried it yet. but it seems that it doesn't create a new connection otherwise we wouldn't run into the errors... maybe it's a bug in https://github.com/rails/rails/blob/v4.2.0/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L623-625

and regarding the pools, we are already using 5 pools. see https://github.com/crowbar/crowbar-init/blob/master/chef/cookbooks/postgresql/templates/default/database.yml.erb#L22

btw. gating is green, and there are no errors or reconnects. maybe we can run a HA gating or sth similar with more parallel chef client runs?

toabctl:
> And use a connection pool? A helper that store N connections, and if the pool is empty, returns a new one.

@aplanas I think Rails already uses a connection pool

toabctl:
> Can you think of a way to test this hypothesis before we merge this?

@nicolasbock Yes. There's a easy reproducer for the first error message (found together with @rhafer yesterday). Do:

1) Remove the patch https://build.suse.de/package/view_file/Devel:Cloud:7:Staging/crowbar-core/582.patch?expand=1 from the package (or remove the exception handler in your env manually)
2) Call ```crowbar cinder proposal commit default
3) watch ```/var/log/crowbar/production.log``` .

This was 100% reproducable with the environment from https://ci.suse.de/view/Cloud/view/Cloud7/job/cloud-mkcloud7-job-magnum-x86_64/

For the second error message (```PG::UnableToSend: server closed the connection unexpectedly```) we haven't found a reproducer yet.

toabctl:
Bw. this is another approach for #582 (which we have in our current package in D:C:7)

toabctl:
There is some good documentation here: https://devcenter.heroku.com/articles/concurrency-and-database-connections#threaded-servers .

toabctl:
> maybe have a look at https://github.com/crowbar/crowbar-core/blob/291213be69ed3023274522c80d80334713416554/crowbar_framework/lib/crowbar/backup/restore.rb#L32 where we use the builtin rails threading, and see if that would work too.

@MaximilianMeister What's the difference the code I use here? Both are using ```Thread.new```. No?

MaximilianMeister:
> What's the difference the code I use here? Both are using Thread.new. No?

nevermind. I just thought on the first glance that you include another library for threading, but it seems that is just for the `wait for all threads`, which the `Thread` implementation doesn't provide
see https://github.com/crowbar/crowbar-core/pull/616#discussion_r75926943

toabctl:
@MaximilianMeister yes. I'm using ```ThreadsWait``` which in the ruby standard libary.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1273,24 +1274,34 @@ def apply_role(role, inst, in_queue, bootstrap = false)
...
+        thread = run_remote_chef_client(node, "chef-client", filename)
+        threads[thread] = node
       end
-      status = Process.waitall
-      badones = status.select { |x| x[1].exitstatus != 0 }
 
-      next if badones.empty?
+      # wait for all running threads and collect the ones with a non-zero return value
+      bad_nodes = []
+      logger.debug "Waiting for #{threads.keys.length} threads to finish ..."
```
vuntz: Detail, but no space before "..." :-)

crowbar_framework/app/models/service_object.rb
rsalevsky: @toabctl  Can you fix the hound issue? Just replace `Thread.new { ... }` with `Thread.new do ... end`.

crowbar_framework/app/models/service_object.rb
toabctl: Fixed

nicolasbock:
+1

rhafer:
> @MaximilianMeister did you try the solution from heroku that you pointed? : disconnect before the fork and reconnect after that. Looks like that doing that the children will discover that there is no database connection, and will create a new one for this process.

@aplanas We briefly tried that but as puma is multithreaded there are some issues with that approach. Closing all connections before the fork might close connections that currently have database transactions in-flight, which will cause other issues of it's own. Additionally closing and reopening all database connections of a pool (puma's pool is per process, i.e. all threads of a process share the same pool) for every fork (i.e. every chef-client run we trigger, plus some other forks) is pretty ugly as well.

rhafer:
> For the second error message (PG::UnableToSend: server closed the connection unexpectedly) we haven't found a reproducer yet.

Actually I was able to see that error a few times now as well. It seems to be a timing issue of whether you get PG::UnableToSend or the other error. The underlying cause is the same (the child terminated the connection of the parent)

aplanas:
@rhafer thanks for the details!!

Is amusing that the threaded approach worked in the gating. Moving from a non-shared variable space to a shared one is usually mode complex : )))))

If works: +1

jsuchome:
I've met some problems with this approach. Did not look closely, but I've seen that the client chef log files that are supposed to be under /var/log/crowbar/chef-client were missing...

MaximilianMeister:
> I've met some problems with this approach. Did not look closely, but I've seen that the client chef log files that are supposed to be under /var/log/crowbar/chef-client were missing...

I remember we had some issue with disappearing logs because of the forking. not sure if that applies here too. I just found 1 old commit now b534135ecee8896d90e6a93e1105c5ab3c581672 but I think @vuntz could remember more details

rhafer:
> I remember we had some issue with disappearing logs because of the forking. not sure if that applies here too.

No, this is actually a problem of the patch. We're working on that.

toabctl:
> I've met some problems with this approach. Did not look closely, but I've seen that the client chef log files that are supposed to be under /var/log/crowbar/chef-client were missing...

@jsuchome Fixed with the latest version...

rhafer:
@aspiers  @toabctl My suggestion for the commit message (in the hope that it clears up some things):

When the forked process (used to execute the chef-client) terminates, it closes all its file descriptors. As it inherits all file descriptors of the parent process, that also causes the ActiveRecord's Database Connections to postgresql to be terminated. This also causes the parent process to loose the connection to the database.
The problems is visible in the log through errors like this:
"PG::ConnectionBad: PQconsumeInput() server closed the connection unexpectedly" or "PG::UnableToSend: server closed the connection unexpectedly" after the chef-client run finishes, when applying a proposal.

This patch tries to address the issue by avoiding the ```fork()``` call and using separate threads to launch chef-client. As a thread doesn't close any file descriptors when terminating. (Note we're still using the ```system``` call to launch chef-client, which obviously also forks another process, but as that is using some call of the ```exec``` family to run the command it does not mess around with the parent's file descriptors/sockets.

toabctl:
@rhafer I updated the commit message. Thanks a lot!

rhafer:
While I think the patch is good to address the immediate issue, it seems obvious to me that we need something more sophisticated in crowbar for running tasks in the background. @vuntz mentioned sucker_punch (https://github.com/brandonhilkert/sucker_punch) as one possible choice for that. (Apart from the name ;) )

vuntz:
+1

rsalevsky:
+1

PR #618, opened 2016-08-25
[stable/3.0] Remove target_platform and default_os valus from admin node

jsuchome:
We need it to be filled by new values after the OS upgrade.

`admin_node["provisioner"]["default_os"]` is used to set the new OS for nodes after the server is upgraded

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -47,6 +47,15 @@ module Api
       end
 
       if upgrade_script_path.exist?
+
+        # admin node will have different OS after the upgrade
+        admin_node = NodeObject.admin_node
+        admin_node["target_platform"] = ""
+        if admin_node["provisioner"] && admin_node["provisioner"]["default_os"]
+          admin_node["provisioner"].delete "default_os"
+        end
+        admin_node.save
+
         pid = spawn("sudo #{upgrade_script_path}")
         Process.detach(pid)
         Rails.logger.info("#{upgrade_script_path} executed with pid: #{pid}")
--- a/crowbar_framework/spec/controllers/api/crowbar_controller_spec.rb
+++ b/crowbar_framework/spec/controllers/api/crowbar_controller_spec.rb
@@ -2,6 +2,7 @@ require "spec_helper"
 
 describe Api::CrowbarController, type: :request do
   context "with a successful crowbar API request" do
+    let(:admin_node) { NodeObject.find_node_by_name("admin") }
     let(:headers) { { ACCEPT: "application/vnd.crowbar.v2.0+json" } }
     let(:pid) { rand(20000..30000) }
     let!(:crowbar_upgrade_status) do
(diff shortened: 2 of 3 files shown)


jsuchome:
Surprisingly it seems to work.

jsuchome:
This would rather belong to Cloud7 part of process, to some initial code run by cloud-init. Probably together with schema migration, like proposed https://github.com/SUSE-Cloud/automation/pull/1156#discussion_r76263439

@rsalevsky what do you think?

rsalevsky:
@jsuchome It's definitely a C7 part, but we can't do it in crowbar-init because we can't access this data.

jsuchome:
Hm, but where should we invoke it on Cloud7 side? Here we at least know we are starting the upgrade.

MaximilianMeister:
I think it makes most sense for now at this stage (although I agree it should be in cloud 7, but that's not really possible now) and in case the upgrade goes wrong I guess the attributes get set back to the old os value again. Also when the user restores from the backup there shouldn't be a problem with this change.

MaximilianMeister:
for me it's a +1

@vuntz what do you think?

vuntz:
I also believe it should be done on the Cloud 7 side of things, and tbh, I'm confused as to why we can't do it there right now. Don't we have some code that runs at the beginning of the upgrade process in the rails app?

We can merge that now if it unblocks things, but please track somewhere the fact that it should be done in another location.

Also, if we really keep it here, then we'll need the same code in master for the 7->8 upgrade...

MaximilianMeister:
> Don't we have some code that runs at the beginning of the upgrade process in the rails app?

i guess you mean on the cloud 7 side? I am not aware of anything like that, the first crowbar API consumed will be the node repo checks (after the database setup done by the crowbar-init api)

MaximilianMeister:
but we could also make it part of `prepare` (https://trello.com/c/xeN0LnPm/1856-5-implement-api-for-preparing-nodes-for-upgrade-upgrade-step-2) after the chef-client daemon stop. and the attributes could be rolled back any time when the user cancels the upgrade...

vuntz:
@MaximilianMeister let me rephrase: how will we make sure we only show the upgrade UI until the upgrade is fully completed? Don't we have code to force this?

MaximilianMeister:
> how will we make sure we only show the upgrade UI until the upgrade is fully completed? Don't we have code to force this?

just yesterday I had the same thought with santiago, i wasn't sure so I created https://trello.com/c/MZP6krgU/1915-restrict-api-access-during-upgrade

as we are now decoupling the UI from Rails, it will be slightly different i guess (in the frontend), but I am afraid I still don't understand you. where would you put that piece of code exactly?

vuntz:
> as we are now decoupling the UI from Rails, it will be slightly different i guess (in the frontend), but I am afraid I still don't understand you. where would you put that piece of code exactly?

If we have code that enforces the upgrade mode, then when crowbar starts, we can trigger automatic actions like the one we're discussing here.

vuntz:
And to be clear: let's not block on this discussion forever, we can merge this PR. What matters to me is that we track the fact that we need to clean this up at a later point.

jsuchome:
@vuntz so could you add your +1? I've created https://trello.com/c/8X7elkka/82-cleanup-upgrade-related-mess to find the proper solution later...

PR #619, opened 2016-08-25
Added function to execute command at remote nodes

jsuchome:
ssh to the node and wait until the command exits

This is partial forward-port of https://github.com/crowbar/crowbar-core/pull/614 . The other commits from that PR are here: https://github.com/crowbar/crowbar-ha/pull/138

### DIFF
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -17,6 +17,7 @@
 
 require "chef/mixin/deep_merge"
 require "timeout"
+require "open3"
 
 class NodeObject < ChefObject
   self.chef_type = "node"
@@ -1199,6 +1200,22 @@ class NodeObject < ChefObject
 
   def get_bmc_password
     @node["ipmi"]["bmc_password"] rescue nil
+  end
+
+  # ssh to the node and wait until the command exits
+  def run_ssh_cmd(cmd)
+    args = ["sudo", "-i", "-u", "root", "--", "timeout", "-k", "5s", "15s",
+            "ssh", "-o", "ConnectTimeout=10",
+            "root@#{@node.name}",
+            cmd.to_s
+    ].join(" ")
+    Open3.popen3(args) do |stdin, stdout, stderr, wait_thr|
+      {
+        stdout: stdout.gets(nil),
+        stderr: stderr.gets(nil),
+        exit_code: wait_thr.value.exitstatus
+      }
+    end
   end
 
   def ssh_cmd(cmd)
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -1207,7 +1207,7 @@ class NodeObject < ChefObject
     args = ["sudo", "-i", "-u", "root", "--", "timeout", "-k", "5s", "15s",
             "ssh", "-o", "ConnectTimeout=10",
             "root@#{@node.name}",
-            cmd.to_s
+            %("#{cmd.gsub('"', '\\"')}")
     ].join(" ")
     Open3.popen3(args) do |stdin, stdout, stderr, wait_thr|
       {


crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -296,6 +296,44 @@ def transition(inst, name, state)
...
+  # {
+  #     "crm_failures" => {
+  #             "node1" => "reason for crm status failure"
+  #     },
+  #     "failed_actions" => {
+  #             "node2" => "Failed action on this node"
+  #     }
+  # }
+  # User has to manually clean pacemaker resources before proceeding with the upgrade.
+  def check_cluster_health
```
rsalevsky: Just for the record, this will go to the HA API.

jsuchome:
This is the actual health check https://github.com/crowbar/crowbar-ha/pull/138

nicolasbock:
+1

rsalevsky:
+1

aspiers:
-1, I find the commit message quite confusing:

1. Is this a forward-port from 3.0 to `master`?
2. It's only a partial port of #614, not a full port - right?  Since it only adds a function.
3. Which is the "different place"?  Why does it need to be moved?

Please can you make the commit message and PR description clear enough to answer these uncertainties?  Thanks.

jsuchome:
@aspiers better?

crowbar_framework/app/models/node_object.rb
```diff
@@ -1207,7 +1207,7 @@ def run_ssh_cmd(cmd)
     args = ["sudo", "-i", "-u", "root", "--", "timeout", "-k", "5s", "15s",
             "ssh", "-o", "ConnectTimeout=10",
             "root@#{@node.name}",
-            cmd.to_s
+            "\"#{cmd}\""
```
aspiers: What about if `cmd` contains double-quotes?

aspiers:
@jsuchome Yes thanks, that's a big improvement; don't think you answered my 3rd question yet though?

jsuchome:
@aspiers That  "different place" was to move that `health_report` function from crowbar_service.rb, where it probably does not belong (that said, there are more upgrade related methods there and all of them should ideally be moved elsewhere). I think @rsalevsky suggested it should go to HA API.

TBH, I think this discussion can take longer, as we are not gonna use that functions in Cloud7.

crowbar_framework/app/models/node_object.rb
jsuchome: I would solve it this way: https://github.com/crowbar/crowbar-core/pull/623/commits/6d3014a096ac02ad298f8c3491aefe804e344f04

And this change was introduced to make commands like https://github.com/crowbar/crowbar-core/pull/623/files#diff-de2e101f791dc5ede34b391d8b632034R106 work. Maybe it's all wrong, it's still WIP

aspiers: 6d3014a seems to no longer exist.  If this function cannot correctly handle commands containing double-quotes characters, then even if none of the commands currently contain double-quote characters, Murphy's Law dictates that they will soon ;-)

BTW why does it have to sudo in order to ssh?  I guess only `root` has the required authorized key?

jsuchome: Ah, the commit is https://github.com/crowbar/crowbar-core/pull/623/commits/3bb7e0cec091a823262543531c7d4a23aee5c682 now. (Seems like I fixed up last change incorrectly)

> I guess only root has the required authorized key?

That seems correct. I just based the method on existing `ssh_cmd`

jsuchome: Maybe use the array as argument for run_ssh_cmd...

jsuchome: I just think whoever wants to pass the quotes to the command must do more testing if the command actually works. It should be possible to use `\\\"` for example, or single quotes, or combination.

aspiers: No, it's the responsibility of the function to do its own escaping of double-quotes, because the requirement to escape them is hard-coupled to the decision to implement the function using double-quotes. Otherwise you are forcing the caller to make a privacy violation via assumptions about the function's internal implementation. Another approach would be to document the function's requirement that any double-quote characters are escaped before passing, which would be ugly but acceptable.

But it's pretty easy to fix this properly, so I'm not sure why you wouldn't want to do it.  Something like

    %("#{s.gsub('"', '\\"')}")

although maybe it needs to escape other characters too, like backslash.

aspiers:
> TBH, I think this discussion can take longer, as we are not gonna use that functions in Cloud7.

Sorry, I don't understand what you mean here.

jsuchome:
>> TBH, I think this discussion can take longer, as we are not gonna use that functions in Cloud7.

> Sorry, I don't understand what you mean here.

I just mean that this is forward port of Coud6 function

crowbar_framework/app/models/node_object.rb
jsuchome: I never wrote I don't want to do it; on the contrary, I wrote it's WIP.

jsuchome:
Adapted, based on @aspiers proposal

jsuchome:
@aspiers, do you have any more concerns here?

PR #621, opened 2016-08-25
Simplify if

jdsn:
simplify an if-else statement and two return if statements

### DIFF
--- a/crowbar_framework/app/helpers/installer/upgrades_helper.rb
+++ b/crowbar_framework/app/helpers/installer/upgrades_helper.rb
@@ -58,9 +58,7 @@ module Installer
     end
 
     def upgrade_ha_repo_needed?
-      return true if Proposal.where(barclamp: "pacemaker").first
-
-      false
+      Proposal.where(barclamp: "pacemaker").first ? true : false
     end
 
     def upgrade_ha_repo?
@@ -74,9 +72,7 @@ module Installer
     end
 
     def upgrade_ceph_repo_needed?
-      return true if Proposal.where(barclamp: "ceph").first
-
-      false
+      Proposal.where(barclamp: "ceph").first ? true : false
     end
 
     def upgrade_ceph_repo?
--- a/crowbar_framework/app/models/api/backup.rb
+++ b/crowbar_framework/app/models/api/backup.rb
@@ -151,11 +151,7 @@ class Api::Backup < ActiveRecord::Base
   protected
 
   def upload?
-    if file.present?
-      true
-    else
-      false
-    end
+    file.present?
   end
 
   def create_archive


crowbar_framework/app/helpers/installer/upgrades_helper.rb
```diff
@@ -58,9 +58,7 @@ def upgrade_error_flash(message)
     end
 
     def upgrade_ha_repo_needed?
-      return true if Proposal.where(barclamp: "pacemaker").first
-
-      false
+      !!Proposal.where(barclamp: "pacemaker").first
```
rsalevsky: I don't like this coding style, it's much harder to read. The existing code is the common ruby style.

crowbar_framework/app/models/api/backup.rb
```diff
@@ -151,11 +151,7 @@ def attribute_names
   protected
 
   def upload?
-    if file.present?
-      true
-    else
-      false
-    end
+    file.present?
   end
```
jsuchome: +1 for this part

crowbar_framework/app/helpers/installer/upgrades_helper.rb
nicolasbock: I kind of like it, it's a bit of a brain teaser at first though :)

aplanas: I agree with @rsalevsky. You need to play with type coercion. A more clear option is implement `.exists?` in the `Proposal` object, or whatever returns the `where`

jdsn: Ok, my aim was to make it simpler, really. I will offer an alternative (I still have some) and update the PR.

jdsn: Ok, let me ask for feedback befor I update the PR, otherwise we run through useless iterations.

not liked:
A: `!!Proposal.where(barclamp: "pacemaker").first`

same thing, but a bit longer:
B: `!Proposal.where(barclamp: "pacemaker").first.nil?`

boolean only test (even simpler):
C: `Proposal.where(barclamp: "pacemaker").size > 0`

jdsn: Ok, @rsalevsky @nicolasbock @aplanas , let me ask for feedback before I update the PR, otherwise we run through useless iterations.

not liked:
A: `!!Proposal.where(barclamp: "pacemaker").first`

same thing, but a bit longer:
B: `!Proposal.where(barclamp: "pacemaker").first.nil?`

boolean only test (even simpler):
C: `Proposal.where(barclamp: "pacemaker").size > 0`

jdsn: I will push C now. Feel free to disagree and comment :)

crowbar_framework/app/helpers/installer/upgrades_helper.rb
```diff
@@ -58,9 +58,7 @@ def upgrade_error_flash(message)
     end
 
     def upgrade_ha_repo_needed?
-      return true if Proposal.where(barclamp: "pacemaker").first
-
-      false
+      Proposal.where(barclamp: "pacemaker").size > 0
```
jdsn: Convinced by houndci I will push D now :)

aplanas:
+1

crowbar_framework/app/helpers/installer/upgrades_helper.rb
```diff
@@ -74,9 +72,7 @@ def upgrade_ha_repo?
     end
 
     def upgrade_ceph_repo_needed?
-      return true if Proposal.where(barclamp: "ceph").first
-
-      false
+      !Proposal.where(barclamp: "ceph").empty?
```
MaximilianMeister:
```ruby
Proposal.where(barclamp: "ceph").any?
```

nicolasbock: I prefer `any?` as well. It sounded a little strange to me at first, but it makes perfect sense now after reading what [where](http://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-where) does and what it [returns](http://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-any-3F).

jdsn: > Proposal.where(barclamp: "ceph").any?

Not sure if that fits, as `any?` does something else than `empty?`

nicolasbock: Yes, it does, but the end result is the same. It "returns true if there are any records.".

jdsn: pushing option E now :)

aplanas:
guys, this is why Python is so much better

+1

crowbar_framework/app/helpers/installer/upgrades_helper.rb
```diff
@@ -58,9 +58,7 @@ def upgrade_error_flash(message)
     end
 
     def upgrade_ha_repo_needed?
-      return true if Proposal.where(barclamp: "pacemaker").first
-
-      false
+      Proposal.where(barclamp: "pacemaker").first ? true : false
```
nicolasbock: +1

nicolasbock:
> guys, this is why Python is so much better

There are just too many ways to express the same idea :)

+1

crowbar_framework/app/helpers/installer/upgrades_helper.rb
jdsn: > Yes, it does, but the end result is the same. It "returns true if there are any records.".
@nicolasbock 
Nope the result was different. The original code evaluated the first element of the list and evaluated it boolean. However `any?` checks if any element evaluates to true. So this was a semantic change. `empty` was even more wrong (read 'semantically different').

PR #622, opened 2016-08-26
[stable/3.0] Invoke cluster health checks

jsuchome:
additionally added one commit from https://github.com/crowbar/crowbar-core/pull/619

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -89,6 +89,21 @@ module Api
       end
     end
 
+    def clusters_healthy?
+      service_object = CrowbarService.new(Rails.logger)
+      cluster_health = service_object.check_cluster_health
+      unless cluster_health.empty?
+        Rails.logger.warn("HA clusters report some problems")
+        (cluster_health["crm_failures"] || {}).each do |node, error|
+          Rails.logger.warn("crm status at node #{node} reports error:\n#{error}")
+        end
+        (cluster_health["failed_actions"] || {}).each do |node, error|
+          Rails.logger.warn("crm at node #{node} reports some failed actions:\n#{error}")
+        end
+      end
+      cluster_health.empty?
+    end
+
     protected
 
     def lib_path
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -340,7 +340,7 @@ class CrowbarService < ServiceObject
         crm_failures[n.name] = ssh_retval[:stdout]
         next
       end
-      ssh_retval = n.run_ssh_cmd('crm status | grep "^Failed actions:"')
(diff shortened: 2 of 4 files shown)


jsuchome:
The report should be probably shown differently than just into the log file. Not sure what would match the api conventions though...

rsalevsky:
I think on C6 side this is fine enough.

jsuchome:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;sKTTY1Wp&#x2F;65-3-s46p13-implement-api-for-first-step-of-upgrade-checks"> (3) S46P13: Implement API for first step of upgrade: checks</a>

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -80,6 +80,15 @@ def maintenance_updates_installed?
       end
     end
 
+    def clusters_healthy?
+      service_object = CrowbarService.new(Rails.logger)
+      cluster_health = service_object.check_cluster_health
+      unless cluster_health.empty?
+        Rails.logger.warn("clusters report some problems: #{cluster_health}")
```
vuntz: I think `cluster_health` is not a string, so not sure if we should use it like this in the log.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -340,7 +340,7 @@ def check_cluster_health
         crm_failures[n.name] = ssh_retval[:stdout]
         next
       end
-      ssh_retval = n.run_ssh_cmd('crm status | grep "^Failed actions:"')
+      ssh_retval = n.run_ssh_cmd("crm status | grep -A 2 '^Failed Actions:'")
```
vuntz: Don't we need some `LANG=C` or something like this, just to be safe?

jsuchome: Right. I was assuming it is already part of run_ssh_cmd, but it is not

crowbar_framework/app/models/api/crowbar.rb
jsuchome: It's a hash describing various types of errors. Is anything wrong with logging that?

vuntz: Well, I prefer nice strings in the logs, instead of data structures :-)

jsuchome: See how the hash could look like. 
https://github.com/crowbar/crowbar-core/blob/stable/3.0/crowbar_framework/app/models/crowbar_service.rb#L319
The original idea probably was that the api somehow passes this to UI which can show it the nice way. I'd say logging such hash is good enough, but ok...

jsuchome:
This is actually not invoked anywhere, not sure if I should wait for green gating...

nicolasbock:
+1

MaximilianMeister:
+1

PR #623, opened 2016-08-26
[stable/3.0] Added check for number of compute nodes

jsuchome:


rsalevsky:
@jsuchome Is this really enough? Wasn't the plan that we need the biggest node as twice (once free)?

jsuchome:
It's not enough, it's just a start. Still looking how to do better check...

jsuchome:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;sKTTY1Wp&#x2F;65-3-s46p13-implement-api-for-first-step-of-upgrade-checks"> (3) S46P13: Implement API for first step of upgrade: checks</a>

jsuchome:
What about parsing the output of `nova host-describe $compute_node`

dguitarbite:
@jsuchome This command should work fine.

dguitarbite:
Thanks for updating. I am going to run this patch with the following scenarios:

- 6 Compute nodes.
- 2 VM's to start with. It should pass ... return true.
- Add 5 VM's to the cluster. It should still pass ... return true.
- Add as many VM's as I could possibly add and this should fail ... return false.

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -80,6 +80,59 @@ def maintenance_updates_installed?
...
 
+    def compute_resources_available?
+      compute_nodes = NodeObject.find("roles:nova-compute-kvm")
+      return true if compute_nodes.empty?
+      if compute_nodes.size == 1
+        Rails.logger.warn("Only one compute node found; non-disruptive upgrade is not possible!")
+        return false
+      end
+
+      # Now, find out if we have anough memory for migrating
```
vuntz: s/anough/enough/

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -80,6 +80,59 @@ def maintenance_updates_installed?
...
+
+      free_compute_memory = 0
+      largest_compute_node_total = 0
+      largest_compute_node_free = 0
+
+      compute_nodes.each do |compute_node|
+        # FIXME: are we sure nova uses the short name of node for its hosts?
+        name = compute_node.handle
+
+        total_cmd = ". /root/.openrc; /usr/bin/nova host-describe #{name}| grep '(total)'"
```
vuntz: Can we put a space before `|`?

dguitarbite:
@vuntz Do you agree with the logic? Basically it is the total_free_memory - free_memory_of_largest_compute >= total_memory_of_largest_compute_node for upgrade to be possible.

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -80,6 +80,60 @@ def maintenance_updates_installed?
...
+        used_cmd = ". /root/.openrc; /usr/bin/nova host-describe #{name} | grep '(used_now)'"
+        ssh_retval = controller.run_ssh_cmd(used_cmd)
+        out = ssh_retval[:stdout] || ""
+        used_memory = out.split[7].to_i
+
+        free_memory = total_memory - used_memory
+        free_compute_memory += free_memory
+
+        # find the node with largest total memory
+        if total_memory > largest_compute_node_memory
```
nicolasbock: Should that be `largest_compute_node_total`?

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -80,6 +80,60 @@ def maintenance_updates_installed?
...
+          largest_compute_node_total = total_memory
+          largest_compute_node_free = free_memory
+        end
+      end
+
+      if free_compute_memory - largest_compute_node_free < largest_compute_node_total
```
nicolasbock: Could you explain what you are trying to calculate? You are testing whether the `total free memory` minus the `free memory on the node with the largest total memory` is larger than the `total memory on the node with the largest total memory`. Why? Are you testing whether you can migrate all instances to the largest node during upgrade? In that case I would have expected to test whether the `total used memory` minus the `used memory on the node with the largest free memory` is less than the `free memory on the node with the largest free memory`.

crowbar_framework/app/models/api/crowbar.rb
jsuchome: True

crowbar_framework/app/models/api/crowbar.rb
jsuchome: Hm, it's true that our current check does not make much sense. But also, we probably do not need your check as well, I think we do not need to  migrate all instances to the largest node. That is maybe too limiting. 
I think what we need is to know that there's always an option to free any node (by migrating its workload elsewhere)

dguitarbite: @nicolasbock The motivation behind this logic from my side was to think about a solution which incorporates the cluster at different points in scale in and out, while trying to cater a simple and easy solution especially with limited feedback channels for getting the right data to make these decisions. The major motivation from my side was to consider if the `Free Memory of the Compute Cluster` is more than the given node which needs to be free'd from all the VM's. Now if this given Compute Node is the largest, then we should consider the worst case where this largest compute node is actually 100% provisioned (let's assume that I am more crazier than nova-scheduler), hence we should have this considered: `Free Memory of the Compute Cluster` >= `Memory of the Largest Compute Node`, but the largest compute node is going to be removed from the cluster! Hence: (`Free Memory of the Compute Cluster` - `Free Memory of the Largest Compute Node`) >= `Free Memory of the Largest Compute Node`. We do not have to migrate all the instances on one compute node. We just need to make sure that the largest compute node could be easily emptied. This logic should be applicable for a 2 node cluster and also for a 1000 node cluster, although I do not know if it is as efficient for a 1000 node cluster especially considering the network traffic generated by the queries for the same and all that jazz. I would be really happy if you could disprove me here because I cannot believe it myself that I could cook this up in a few minutes.

I wanted to make this a non-linear equation but my mathematics is rusty or rather, CPU's could be over-provisioned much better so counting their factor should be limiting to zero (for modern decently powerful server CPU farms ... remember? we have a cluster!) and the hard disk should ideally be on shared storage. So this algorithm although not quite exhaustive is still not missing out anything critical. Thanks for trying to understand this piece of code.

dguitarbite:
Another observation while generating required load here is to address the fact that the size of the compute nodes w.r.t. the average flavor of the instances used is a relevant factor in this case. The size of the compute nodes should on average be bigger than the size of the flavors,  just because we cannot really do any sharding is not possible with VM's.

dguitarbite:
Not sure if the table output of the prechecks command should also show false but the logs atleast are showing the truth values. I have attached some images to best showcase my testing here.

With very high load on the Compute Cluster.
High Load Image1

Although I am not sure if this is lying ...
High Load Image2

When there is enough free space ...
Not so high load Image1

Hope this helps. Except for the gray table output ... I am happy with the same.

crowbar_framework/app/models/api/crowbar.rb
nicolasbock: @dguitarbite and @jsuchome, if I understand the motivation correctly then this is about making sure before the upgrade that the upgrade will not fail because of a lack of nova resources. We are assuming that we have to migrate an instance out before the upgrade of a node, and that we can't `virsh save` the instance on the node during the upgrade.

The problem I see with this approach here is that it will fail in certain situations. Let `N` be the number of nodes in the cluster and `m` the average number of free bytes per node. Then the total free memory is `N * m`. If `N` is large, `N * m` will also be large. Suppose that we want to migrate an instance that happens to use `M >> m` bytes. It is likely to not fit anywhere else. But by considering only something like `M < N * m` or similar we will conclude that we can proceed with the upgrade. I don't know whether such an unbalanced load is common, maybe this is really just a corner case. But it might be worth considering.

dguitarbite: @nicolasbock I agree with you, I guess I was trying to solve this with some sort of distribution curve with probabilities involved where my math is rusty ... something similar to `M/(∑Fm(N)/N)` assuming the worst case `M ==> m1.largest_flavor` \forall \!\, `Fm(N) ∈ RAM of Compute Nodes` should ideally be as small as possible. This number although in my head sticks to .1 or lesser is still a speculation and should be derived from real instance data. But in case the size of the largest flavor is at least 10 times smaller than the average size of the compute node cluster, we should ideally be safe. Worst case we suggest to add another compute node to the cluster which is larger than or as large as the largest flavor or the largest compute node. By far, AFAIK, it is safe to assume that the compute nodes on average should be much more bigger/larger w.r.t. its average/most commonly used flavors.

Something like ``virsh save`` is a no-go for the given scope. I still have serious doubts about ``M < N * m`` being invalid in certain cases where rather the distribution curve of the largest free memory m(N) should be large enough to fit M, meanwhile we could have many such M's so ``M1 <= m(N1)``, ``M2<= m(N2)`` ... ``Mn <= m(Nn)``. Assuming number of large instances is equal to number of compute nodes for the lack of appropriate mathematical signs on github comment, how about a coffee with pen and paper today later in the afternoon?

dguitarbite:
But rest apart, from my side the logic seems to be better than nothing, we could add another warning/check where we could potentially compare the average size of the compute nodes ``(∑Fm(N)/N) >> m1.largest_flavor`` ∀ ``N ∈ ℕ, Number compute nodes & Fm(N) ∈ RAM of Compute Nodes``.

dguitarbite:
+1 This logic should be a good starting point. We should reflect back on this at a later point in time.

jsuchome:
@dguitarbite @nicolasbock Thanks for the work you've invested into this, but as it was just discussed on irc with @aspiers and @aplanas  we probably should not try to be too smart here, because we can't solve the problem 100% correctly. Your approximations might be good, but it is probably better and more fair to say the operator, "You have to make sure you have enough compute power" instead of trying to solve NP-complete problem.
Also consider important point: cloud users could still start new workloads after this point in the upgrade process. That makes our attempts even more useless.

nicolasbock:
@jsuchome Very good points. I completely agree.

aspiers:
Exactly. We could waste a LOT of time here trying to get a good solution, but we would never succeed and the cost:reward ratio makes no sense for something which is not an essential feature anyway. Instead we should:

1. ensure the docs and the UI display sufficient warning that headroom is required, and
2. ensure that if `nova host-evacuate-live` fails, that the whole upgrade process does not fail but simply pauses and waits for the cloud operator to make changes manually and then tell the upgrade process to retry the `host-evacuate-live` again. When it finally succeeds, the process can resume from where it got to.

PR #625, opened 2016-08-30
list addons in the crowbar object

MaximilianMeister:
this is helpful for the UI to determine which checks to perform during
the upgrade, depending on which addon is installed

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -22,11 +22,13 @@ module Api
 
     def initialize
       @version = ENV["CROWBAR_VERSION"]
+      @addons = addons
     end
 
     def status
       {
-        version: @version
+        version: @version,
+        addons: @addons
       }
     end
 
@@ -80,6 +82,14 @@ module Api
       end
     end
 
+    def addons
+      [].tap do |list|
+        ["ceph", "ha"].each do |addon|
+          list.push(addon) if addon_installed?(addon)
+        end
+      end
+    end
+
     protected
 
     def lib_path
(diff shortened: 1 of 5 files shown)


jsuchome:
At which point do you think this call should be needed?

MaximilianMeister:
> At which point do you think this call should be needed?

for now, somewhere in the Angular app

jsuchome:
+1 I assume we want it in 3.0 as well

MaximilianMeister:
> I assume we want it in 3.0 as well

yes, will be backported

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -101,5 +111,19 @@ def failed?
...
+    def addon_installed?(addon)
+      case addon
+      when "ceph"
+        CephService
+      when "ha"
+        PacemakerService
+      else
+        return false
+      end
+      true
```
aplanas: sorry for my lack of Rubyness, but isn't it returning true always, instead of the result of `CephService` or `PacemakerService`?

MaximilianMeister: no actually it would return `PacemakerService` or `CephService` which probably evaluates to `true` in an if context, but it wouldn't be correct. and this is more explicit

aplanas: ah, so you try to resolve `CephService`, if is defined you ignore the return value and return true, if is not defined a NameError happends and return false.

aplanas:
+1

crowbar_framework/app/models/api/crowbar.rb
MaximilianMeister: yes exactly, I thought about looking for files or other indicators, but this is quick and efficient enough
thanks for the review

PR #630, opened 2016-08-31
upgrade: add API for admin repocheck

MaximilianMeister:
this endpoint is to make sure that the cloud and sp2 repos are available
before upgrading the admin server

### DIFF
--- a/crowbar_framework/app/controllers/api/crowbar_controller.rb
+++ b/crowbar_framework/app/controllers/api/crowbar_controller.rb
@@ -52,6 +52,12 @@ class Api::CrowbarController < ApiController
     }
   end
 
+  api :GET, "/api/crowbar/repocheck", "Sanity check for Crowbar server repositories"
+  api_version "2.0"
+  def repocheck
+    render json: @crowbar.repocheck
+  end
+
   protected
 
   def set_crowbar
--- a/crowbar_framework/app/controllers/api/crowbar_controller.rb
+++ b/crowbar_framework/app/controllers/api/crowbar_controller.rb
@@ -55,7 +55,15 @@ class Api::CrowbarController < ApiController
   api :GET, "/api/crowbar/repocheck", "Sanity check for Crowbar server repositories"
   api_version "2.0"
   def repocheck
-    render json: @crowbar.repocheck
+    check = @crowbar.repocheck
+
+    if @crowbar.errors.any?
+      render json: {
+        error: @crowbar.errors.full_messages.first
+      }, status: :service_unavailable
+    else
+      render json: check
+    end
   end
 
   protected
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -92,14 +92,28 @@ module Api
 
     def repocheck
       # FIXME: once we start working on 7 to 8 upgrade we have to adapt the sles version
-      {
-        os: {
-          available: repo_version_available?("SLES", "12.2")
-        },
-        cloud: {
-          available: repo_version_available?("suse-openstack-cloud", "8")
+      zypper_stream = Hash.from_xml(
+        `sudo /usr/bin/zypper-retry --xmlout products`
+      )["stream"]
+
+      {}.tap do |ret|
+        if zypper_stream["message"] =~ /^System management is locked/
+          errors.add(
+            :base,
+            I18n.t("api.crowbar.zypper_locked", zypper_locked_message: zypper_stream["message"])
+          )
+          return ret
+        end
+
+        products = zypper_stream["product_list"]["product"]
+
+        ret[:os] = {
+          available: repo_version_available?(products, "SLES", "12.2")
         }
-      }
+        ret[:cloud] = {
+          available: repo_version_available?(products, "suse-openstack-cloud", "8")
+        }
+      end
     end
 
     protected
@@ -138,11 +152,7 @@ module Api
       false
     end
 
-    def repo_version_available?(product, version)
-      products = Hash.from_xml(
-        `zypper --xmlout products`
-      )["stream"]["product_list"]["product"]
-
+    def repo_version_available?(products, product, version)
       products.any? do |p|
         p["version"] == version && p["name"] == product
       end
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -90,6 +90,18 @@ module Api
       end
     end
 
+    def repocheck
+      # FIXME: once we start working on 7 to 8 upgrade we have to adapt the sles version
+      {
+        os: {
+          available: repo_version_available?("SLES", "12.2")
+        },
+        cloud: {
+          available: repo_version_available?("suse-openstack-cloud", "8")
+        }
+      }
+    end
+
     protected
 
     def lib_path
@@ -124,6 +136,16 @@ module Api
       true
     rescue NameError
       false
+    end
+
+    def repo_version_available?(product, version)
+      products = Hash.from_xml(
+        `zypper --xmlout products`
+      )["stream"]["product_list"]["product"]
+
+      products.any? do |p|
+        p["version"] == version && p["name"] == product
+      end
     end
   end
 end
--- a/crowbar_framework/config/locales/crowbar/en.yml
+++ b/crowbar_framework/config/locales/crowbar/en.yml
@@ -771,6 +771,7 @@ en:
       destroy_failed: 'Failed to destroy %{component}'
     crowbar:
       upgrade_ongoing: 'Upgrade is already ongoing. Please wait.'
+      zypper_locked: '%{zypper_locked_message}'
 
   # Global
   overview: 'Overview'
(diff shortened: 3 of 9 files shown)


MaximilianMeister:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;otxWJSSF&#x2F;96-3-s47p10-implement-api-for-repo-check-for-admin-server-upgrade-step-4"> (3) S47P10 Implement API for repo check for admin server (upgrade step 4)</a>

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -90,6 +90,23 @@ def addons
       end
     end
 
+    def repocheck(options = {})
+      platform = options.fetch(:platform)
+      arch = options.fetch(:arch)
+
+      repos = {}
+      wanted_repos = ["cloud", "sles12-sp2-pool", "sles12-sp2-updates"]
```
jsuchome: Do we need a fixed list here, or could we obtain it from a list of repositories currently available (via a repo library)? (We would need to change sp1 to sp2 obviously)

jsuchome: Ah, and this is even the master code. Here we don't have any idea what could be the needed repo for the 7-8 upgrade...

MaximilianMeister:
@jsuchome I briefly discussed with @vuntz that the repository lib is useless here, we agreed to use a zypper check for the availability of the cloud 7 and sles product.

parts of the solution here could be used for the node repochecks though

crowbar_framework/app/controllers/api/crowbar_controller.rb
```diff
@@ -52,6 +52,12 @@ def maintenance
     }
   end
 
+  api :GET, "/api/crowbar/repocheck", "Sanity check Crowbar repositories"
```
jsuchome: _Sanity check for Crowbar server repositories_ ?

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -90,6 +90,17 @@ def addons
       end
     end
 
+    def repocheck
+      {
+        cloud: {
+          available: repo_version_available("suse-openstack-cloud", "7") == 0
```
jsuchome: Again, does this check make sense in master branch? We want to run the check before the upgrade, to check if the _new_ repos are in place. So in master branch the check should be for cloud8

MaximilianMeister: right, actually the backport is important here. but what would be the sles repo version for master? Do we know yet?

jsuchome: No idea. It's true we do not need to know, so maybe leave it there, but add a comment that it needs to be adapted.

MaximilianMeister: oh I saw you posted the same question already, lets change 7 to 8 for master now and the other one we have to adapt when we work on the next upgrade. by then we will likely know the sles version

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -90,6 +90,18 @@ def addons
       end
     end
 
+    def repocheck
+      # FIXME: once we start working on 7 to 8 upgrade we have to adapt the sles version
+      {
+        cloud: {
+          available: repo_version_available("suse-openstack-cloud", "8") == 0
+        },
+        sles12sp2: {
```
jsuchome: As this is not the value we are checking for, maybe it could be more generic, like "system" (as opposed to "cloud")

MaximilianMeister: good point. fixed

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -125,5 +137,11 @@ def addon_installed?(addon)
     rescue NameError
       false
     end
+
+    def repo_version_available(product, version)
+      `zypper products`.split("\n").detect do |p|
```
jsuchome: This will pick only the first ocurence of product. But as the old product was still not reinstalled, there will be both shown by zypper output:
```
LANG=C zypper products
S | Repository | Internal Name        | Name                                              | Version | Arch   | Is Base
--+------------+----------------------+---------------------------------------------------+---------+--------+--------
i | @System    | SLES                 | SUSE Linux Enterprise Server 12 SP1               | 12.1-0  | x86_64 | Yes    
i | @System    | suse-openstack-cloud | SUSE OpenStack Cloud 6                            | 6-0     | x86_64 | No     
v | Cloud      | suse-openstack-cloud | SUSE OpenStack Cloud 7                            | 7-0     | x86_64 | No     
  | SDK-SP2    | sle-sdk              | SUSE Linux Enterprise Software Development Kit 12 | 12.2-0  | x86_64 | No     
v | sles12sp2  | SLES                 | SUSE Linux Enterprise Server 12 SP2 RC2           | 12.2-0  | x86_64 | No 
```

Good test case for your unit tests...

jsuchome: So I'd say, just put the version check inside as well.

MaximilianMeister: > This will pick only the first ocurence of product. 
> So I'd say, just put the version check inside as well.

good catch, thanks

> Good test case for your unit tests...

seems so, but unfortunately the tests can't really use zypper, as it may not be available (travis), and the repos wouldn't exist. so I need to skip (mock) it...

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -125,5 +137,11 @@ def addon_installed?(addon)
...
+        p =~ /#{product}.*#{version}-.*/
```
MaximilianMeister: I think the regex is still not ideal... would be good to match the values to columns. unfortunately zypper doesn't provide any easy to parse output format.

jsuchome: What about `p =~ /#{product}/ && p.split("|")[4].strip =~ /^#{version}.*/` ?

crowbar_framework/app/models/api/crowbar.rb
jsuchome: Yes, I meant to mock it.

crowbar_framework/app/models/api/crowbar.rb
MaximilianMeister: actually it has `--xmlout`

jsuchome:
+1

jsuchome:
Although, mocking some zypper outputs would be nice to have ... :-)

MaximilianMeister:
@jsuchome sorry to change again, but I found a nicer way to avoid the horrible xml library. the zypper output is now directly converted to a hash. can you have another look?

jsuchome:
+1 after latest commit

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -125,5 +137,15 @@ def addon_installed?(addon)
     rescue NameError
       false
     end
+
+    def repo_version_available?(product, version)
+      products = Hash.from_xml(
+        `zypper --xmlout products`
+      )["stream"]["product_list"]["product"]
```
vuntz: Should we protect this against some exceptions in case zypper is failing? (which could happen if there's the zypp lock)

aplanas:
LGTM, but I agree with @vuntz comment

MaximilianMeister:
@vuntz @aplanas @jsuchome I have added some protection against zypper locks now

aplanas:
+1

MaximilianMeister:
just fixed the unit test, I had to mock the zypper output with an xml fixture

MaximilianMeister:
> Although, mocking some zypper outputs would be nice to have ... :-)

 @jsuchome done

jsuchome:
Cool, thanks!

jsuchome:
We'll need this in 3.0

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -90,6 +90,32 @@ def addons
...
+        end
+
+        products = zypper_stream["product_list"]["product"]
+
+        ret[:os] = {
+          available: repo_version_available?(products, "SLES", "12.2")
+        }
+        ret[:cloud] = {
+          available: repo_version_available?(products, "suse-openstack-cloud", "8")
+        }
```
vuntz: Late comment, and something to check with @santiph: do we want to somehow provide a human-readable name for the product in the output, or is this handled in the UI?

santiph: If the product name won't be translated into other languages, then it could provided by the API and consumed through the UI.
Otherwise, just a product key should be required. But keep in mind that whenever a new product is added in the backend, its name should also be created/updated on Angular's language file to match the new product key and let the UI know what should be displayed.

Let me know, @vuntz if we need to analyse it more to make a decision.

jsuchome: I think this is exactly what we need to discuss in https://trello.com/c/3TALQwDA/1157-upgrade-synchronize-api-responses-with-the-angular-app#

PR #628, opened 2016-08-30
enable canceling the upgrade

MaximilianMeister:
by adding another api endpoint to reset the nodes to ready
this function has been used in cloud5 as well

PR #631, opened 2016-08-31
Additional actions for reverting the upgrade

jsuchome:
This is the addition to https://github.com/crowbar/crowbar-core/pull/628

Needs to be ported to 3.0

### DIFF
--- a/chef/data_bags/crowbar/migrate/crowbar/100_crowbar-upgrade-states.rb
+++ b/chef/data_bags/crowbar/migrate/crowbar/100_crowbar-upgrade-states.rb
@@ -1,0 +1,13 @@
+def upgrade(ta, td, a, d)
+  d["element_states"] = td["element_states"]
+  d["element_order"] = td["element_order"]
+  d["element_run_list_order"] = td["element_run_list_order"]
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  d["element_states"] = td["element_states"]
+  d["element_order"] = td["element_order"]
+  d["element_run_list_order"] = td["element_run_list_order"]
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-crowbar.json
+++ b/chef/data_bags/crowbar/template-crowbar.json
@@ -28,10 +28,10 @@
     "crowbar": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 11,
+      "schema-revision": 100,
       "element_states": {
         "crowbar": [ "all" ],
-        "crowbar-upgrade": [ "crowbar_upgrade" ]
+        "crowbar-upgrade": [ "readying", "ready", "applying", "crowbar_upgrade" ]
       },
       "elements": {},
       "element_order": [
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -28,6 +28,16 @@ upgrade_step = node["crowbar_wall"]["crowbar_upgrade_step"] || "none"
 
 case upgrade_step
 
+when "revert_to_ready"
+
+  service "crowbar_join" do
+    action :enable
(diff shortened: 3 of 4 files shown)


chef/data_bags/crowbar/template-crowbar.json
```diff
@@ -28,10 +28,10 @@
     "crowbar": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 11,
+      "schema-revision": 100,
       "element_states": {
         "crowbar": [ "all" ],
-        "crowbar-upgrade": [ "crowbar_upgrade" ]
+        "crowbar-upgrade": [ "readying", "ready", "applying", "crowbar_upgrade" ]
```
vuntz: Don't we need just "applying" there?

chef/data_bags/crowbar/migrate/crowbar/100_crowbar-upgrade-states.rb
```diff
@@ -0,0 +1,12 @@
+def upgrade(ta, td, a, d)
+  d["element_states"] = td["element_states"]
+  d["element_order"] = td["element_order"]
+  d["element_run_list_order"] = td["element_run_list_order"]
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  d["element_states"] = td["element_states"]
+  d["element_order"] = td["element_order"]
```
vuntz: No element_run_list_order here?

vuntz:
When we cancel, should we also remove the nodes from the crowbar-upgrade role?

chef/data_bags/crowbar/template-crowbar.json
jsuchome: Probably yes, but I was more confident when I just used the same value as in Cloud5

jsuchome:
> When we cancel, should we also remove the nodes from the crowbar-upgrade role?

That's done in https://github.com/crowbar/crowbar-core/pull/628

chef/data_bags/crowbar/template-crowbar.json
vuntz: Ok. Just so you know: I'd like to remove most of the roles running in readying/ready at some point ;-) But this can be done later on indeed.

vuntz:
+1

vuntz:
> > When we cancel, should we also remove the nodes from the crowbar-upgrade role?
> 
> That's done in #628

Actually, reading that PR: it's done in the proposal, but the nodes will still have the roles since the proposal is not applied after being saved.

chef/data_bags/crowbar/migrate/crowbar/100_crowbar-upgrade-states.rb
jsuchome: Fixed.

chef/data_bags/crowbar/template-crowbar.json
jsuchome: I'll test it with just "applying"

jsuchome:
Hm, we could apply the proposal one more time to get rid of that role completely. It shouldn't hurt though it if is for some time, because the step is "revert to ready"

jsuchome:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;xeN0LnPm&#x2F;95-5-s47p9-implement-api-for-preparing-nodes-for-upgrade-upgrade-step-2"> (5) S47P9 Implement API for preparing nodes for upgrade (upgrade step 2)</a>

chef/data_bags/crowbar/template-crowbar.json
jsuchome: Hm, "applying" does not seem to be enough. I'll leave the whole set here.

vuntz:
> It shouldn't hurt though it if is there for some time, because the step is "revert to ready"

Well, it simply means we will always re-enable the services, which might not be wanted.

MaximilianMeister:
+1

PR #633, opened 2016-09-01
add helper for node repo target platform and enable failed repolist for repochecks

MaximilianMeister:
this will be used by ceph and ha to make sure that the system
repositories, alongside the addon repositories are available
and enabled during the upgrade

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -34,6 +34,19 @@ module Api
       }
     end
 
+    def target_platform(options = {})
+      platform_exception = options.fetch(:platform_exception, nil)
+
+      case Api::Crowbar.new.version
+      when "4.0"
+        if platform_exception == :ceph
+          ::Crowbar::Product.ses_platform
+        else
+          NodeObject.admin_node.target_platform
+        end
+      end
+    end
+
     protected
 
     def crowbar_upgrade_status
--- a/crowbar_framework/lib/crowbar/repository.rb
+++ b/crowbar_framework/lib/crowbar/repository.rb
@@ -183,23 +183,19 @@ module Crowbar
 
       private
 
-      def provided_with_enabled(feature, platform = nil, arch = nil, check_enabled = true)
-        repos = {
-          missing_repos: [],
-          inactive_repos: []
-        }
+      def provided_with_enabled(feature, platform = nil, arch = nil, check_enabled = true, repos = {})
         answer = false
 
         if platform.nil?
           all_platforms.each do |p|
-            if provided_with_enabled(feature, p, arch, check_enabled).first
+            if provided_with_enabled(feature, p, arch, check_enabled, repos).first
               answer = true
               break
             end
           end
         elsif arch.nil?
           arches(platform).each do |a|
-            if provided_with_enabled(feature, platform, a, check_enabled).first
+            if provided_with_enabled(feature, platform, a, check_enabled, repos).first
               answer = true
               break
             end
@@ -215,12 +211,25 @@ module Crowbar
             found = true
 
             r = new(platform, arch, repo)
+
             answer &&= r.available?
-            repos[:missing_repos].push(r.name) unless answer
+            unless r.available?
+              repos[:missing_repos] ||= {}
+              repos[:missing_repos][r.arch.to_sym] ||= []
+              unless repos[:missing_repos][r.arch.to_sym].include?(r.name)
+                repos[:missing_repos][r.arch.to_sym].push(r.name)
+              end
+            end
 
             break unless check_enabled
             answer &&= r.active?
-            repos[:inactive_repos].push(r.name) unless answer
+            next if r.active?
+
+            repos[:inactive_repos] ||= {}
+            repos[:inactive_repos][r.arch.to_sym] ||= []
+            unless repos[:inactive_repos][r.arch.to_sym].include?(r.name)
+              repos[:inactive_repos][r.arch.to_sym].push(r.name)
+            end
           end
 
           answer = false unless found
--- a/crowbar_framework/lib/crowbar/repository.rb
+++ b/crowbar_framework/lib/crowbar/repository.rb
@@ -183,7 +183,11 @@ module Crowbar
 
       private
 
-      def provided_with_enabled(feature, platform = nil, arch = nil, check_enabled = true, repos = {})
+      def provided_with_enabled(feature,
+                                platform = nil,
+                                arch = nil,
+                                check_enabled = true,
+                                repos = {})
         answer = false
 
         if platform.nil?
--- a/crowbar_framework/lib/crowbar/repository.rb
+++ b/crowbar_framework/lib/crowbar/repository.rb
@@ -130,11 +130,19 @@ module Crowbar
       end
 
       def provided_and_enabled?(feature, platform = nil, arch = nil)
-        provided_with_enabled?(feature, platform, arch, true)
+        provided_with_enabled(feature, platform, arch, true).first
+      end
+
+      def provided_and_enabled_with_repolist(feature, platform = nil, arch = nil)
+        provided_with_enabled(feature, platform, arch, true)
       end
 
       def provided?(feature, platform = nil, arch = nil)
-        provided_with_enabled?(feature, platform, arch, false)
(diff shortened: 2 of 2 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -57,5 +81,12 @@ def compute_resources_available?
       # FIXME: to be implemented
       true
     end
+
+    def target_platform
+      case Api::Crowbar.new.version
+      when "4.0"
+        "suse-12.2"
```
jsuchome: Target platform for ceph nodes is currently different than for the rest...

MaximilianMeister:
@vuntz I tried to address the reponame from metadata that you pointed out, you think that is correct?

crowbar_framework/app/models/api/upgrade.rb
MaximilianMeister: okay, thanks for the hint, so I guess it is "suse-12.1" ?

jsuchome: Wouldn't it be possible to just gather target_os value(s) from the nodes?

MaximilianMeister:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;ncdOGOzP&#x2F;97-3-s47p11-implement-api-for-repo-checks-during-upgrade-upgrade-step-8"> (3) S47P11 Implement API for repo checks during upgrade (upgrade step 8)</a>

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -34,6 +34,56 @@ def check
...
+    end
+
+    def target_platform(options = {})
+      # FIXME: get rid of platform_exception when SES upgraded to SP2
+      platform_exception = options.fetch(:platform_exception, nil)
+
+      case Api::Crowbar.new.version
+      when "4.0"
+        if platform_exception == :ceph
+          "suse-12.1"
```
vuntz: You should use the method from https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/product.rb instead of hard-coding.

MaximilianMeister:
@vuntz I slightly modified the `provide_with_enabled` to return a repolist of missing/inactive repos, which could be used by the UI

MaximilianMeister:
@vuntz not really sure if it makes sense, but I have restructured the result to include architectures as well.

```json
{
  "os": {
    "available": true,
    "repos": {
      "missing_repos": {
        "aarch64": [
          "SLES12-SP1-Pool",
          "SLES12-SP1-Updates"
        ]
      },
      "inactive_repos": {
        "aarch64": [
          "SLES12-SP1-Pool",
          "SLES12-SP1-Updates"
        ]
      }
    }
  },
  "ceph": {
    "available": false,
    "repos": {
      "missing_repos": {
        "aarch64": [
          "SUSE-Enterprise-Storage-3-Pool",
          "SUSE-Enterprise-Storage-3-Updates"
        ],
        "x86_64": [
          "SUSE-Enterprise-Storage-3-Pool",
          "SUSE-Enterprise-Storage-3-Updates"
        ]
      },
      "inactive_repos": {
        "aarch64": [
          "SUSE-Enterprise-Storage-3-Pool",
          "SUSE-Enterprise-Storage-3-Updates"
        ],
        "x86_64": [
          "SUSE-Enterprise-Storage-3-Pool",
          "SUSE-Enterprise-Storage-3-Updates"
        ]
      }
    }
  }
}
```

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -34,6 +34,20 @@ def check
       }
     end
 
+    def target_platform(options = {})
+      # FIXME: get rid of platform_exception when SES upgraded to SP2
```
vuntz: Actually, I think it's fine to keep it. The current code is generic, so it will keep working in that case.

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -130,11 +130,19 @@ def check_all_repos
       end
 
       def provided_and_enabled?(feature, platform = nil, arch = nil)
-        provided_with_enabled?(feature, platform, arch, true)
+        provided_with_enabled(feature, platform, arch, true).first
+      end
+
+      def provided_and_enabled_with_repolist(feature, platform = nil, arch = nil)
```
vuntz: Shouldn't we have a question mark at the end of the method name here?

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -203,18 +215,32 @@ def provided_with_enabled?(feature, platform = nil, arch = nil, check_enabled =
             found = true
 
             r = new(platform, arch, repo)
+
             answer &&= r.available?
+            unless answer
```
vuntz: Isn't the check wrong? Shouldn't we just test on `r.available?` here (as if a previous repo is available, we won't fill the missing repos at all).

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -203,18 +215,32 @@ def provided_with_enabled?(feature, platform = nil, arch = nil, check_enabled =
...
+              repos[:missing_repos] ||= {}
+              repos[:missing_repos][r.arch.to_sym] ||= []
+              unless repos[:missing_repos][r.arch.to_sym].include?(r.name)
+                repos[:missing_repos][r.arch.to_sym].push(r.name)
+              end
+            end
 
             break unless check_enabled
             answer &&= r.active?
+            next if answer
```
vuntz: Same here, with `r.active?`

vuntz:
> @vuntz not really sure if it makes sense, but I have restructured the result to include architectures as well.

Seems to make sense to me. The one thing that is slightly misleading in our API is that we would tell that the feature is available including when only the repos for aarch64 are there and x86_64 is needed.

I somehow wonder if we should check for the arches of all existing nodes to define if the repos are available?

crowbar_framework/lib/crowbar/repository.rb
MaximilianMeister: I guess not (by convention), as it doesn't return a bool only. that's why i took out the `?`

crowbar_framework/lib/crowbar/repository.rb
MaximilianMeister: ah, okay. now I know what the operator is doing... I just assumed `answer = r.available?` here... my bad

MaximilianMeister:
> I somehow wonder if we should check for the arches of all existing nodes to define if the repos are available?

seems to align with the question that i posted in tthe other pr https://github.com/crowbar/crowbar-ceph/pull/68#issuecomment-244299855

MaximilianMeister:
>> I somehow wonder if we should check for the arches of all existing nodes to define if the repos are available?

> seems to align with the question that i posted in tthe other pr crowbar/crowbar-ceph#68 (comment)

this is now handled by specifying an architecture to `Crowbar::Repository.provided_and_enabled?`

MaximilianMeister:
@vuntz updated, regarding your comments

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -203,18 +215,32 @@ def provided_with_enabled?(feature, platform = nil, arch = nil, check_enabled =
...
+              repos[:missing_repos] ||= {}
+              repos[:missing_repos][r.arch.to_sym] ||= []
+              unless repos[:missing_repos][r.arch.to_sym].include?(r.name)
+                repos[:missing_repos][r.arch.to_sym].push(r.name)
+              end
+            end
 
             break unless check_enabled
             answer &&= r.active?
+            next if r.active?
```
vuntz: For the record, I would have gone with the `unless r.active?` way of coding it, for consistency with the code earlier. But matter of taste, I guess.

vuntz:
+1; coding style nitpick, but it's a matter of taste, so doesn't matter.

jsuchome:
+1

crowbar_framework/lib/crowbar/repository.rb
MaximilianMeister: yeah me too, but hound was complaining, so I changed it...

PR #635, opened 2016-09-02
[stable/3.0] Make ceph cluster health check part of upgrade pre-checks

jsuchome:
Needs rebase after https://github.com/crowbar/crowbar-core/pull/622 is merged

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -89,6 +89,18 @@ module Api
       end
     end
 
+    def ceph_healthy?
+      ceph_node = NodeObject.find("roles:ceph-mon AND ceph_config_environment:*").first
+      return true if ceph_node.nil?
+
+      ssh_retval = ceph_node.run_ssh_cmd("ceph health 2>&1")
+      if ssh_retval[:exit_code] != 0
+        Rails.logger.warn("ceph cluster health check failed with #{ssh_retval[:stdout]}")
+        return false
+      end
+      true
+    end
+
     protected
 
     def lib_path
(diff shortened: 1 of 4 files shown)


jsuchome:

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -89,6 +89,18 @@ def maintenance_updates_installed?
       end
     end
 
+    def ceph_healthy?
+      ceph_node = NodeObject.find("roles:ceph-mon AND ceph_config_environment:*").first
```
MaximilianMeister: can there only be one ceph node?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -57,5 +58,9 @@ def compute_resources_available?
       # FIXME: to be implemented
       true
     end
+
+    def ceph_healthy?
```
MaximilianMeister: can you move this up to `clusters_healthy?`? and to the question above, i am wondering if this should be `storages_healthy?`

MaximilianMeister:
you also need to adapt the fixtures, and add a model test for the new method

crowbar_framework/app/models/api/crowbar.rb
jsuchome: No, but it's fine to check on just one node of the cluster

crowbar_framework/app/models/api/upgrade.rb
jsuchome: It will be rebased after the PR with other "healthy" method is merded.

I'm not sure about `storages`, this is really specific to ceph and we do not have any other kind of check

jsuchome:
We should do it for master first

PR #637, opened 2016-09-02
Make ceph cluster health check part of upgrade pre-checks

jsuchome:


jsuchome:
@MaximilianMeister ok, there's no real testing here. Ideally, I should check mock NodeObject returning empty list, and also ssh command somehow returning non-0 exit value

MaximilianMeister:
> ok, there's no real testing here. Ideally, I should check mock NodeObject returning empty list, and also ssh command somehow returning non-0 exit value

I can send a follow up for the test, no problem

crowbar_framework/spec/models/api/crowbar_spec.rb
```diff
@@ -125,6 +125,21 @@
     end
   end
 
+  context "with ceph cluster healthy" do
+    it "succeeds" do
+      expect(subject.ceph_healthy?).to be true
```
MaximilianMeister: instead of relying, that the test environment doesn't have ceph, i would test it like this:

```ruby
it "succeeds to check ceph cluster health" do
  allow(Node_Object).to receive(:find).with(
    "roles:ceph-mon AND ceph_config_environment:*"
  ).and_return([NodeObject.find_node_by_name("testing.crowbar.com")])
  allow_any_instance_of(NodeObject).to receive(:run_ssh_cmd).with(
    "ceph health 2>&1"
  ).and_return(exit_code: 0, stdout: "Some message", stderr: "")

  expect(subject.ceph_healthy?).to be true
end
```

crowbar_framework/spec/models/api/crowbar_spec.rb
```diff
@@ -125,6 +125,21 @@
...
+    end
+  end
+
+  context "with ceph cluster not healthy" do
+    it "fails" do
```
MaximilianMeister:
```ruby
it "fails to check ceph cluster health" do
  allow(Node_Object).to receive(:find).with(
    "roles:ceph-mon AND ceph_config_environment:*"
  ).and_return([NodeObject.find_node_by_name("testing.crowbar.com")])
  allow_any_instance_of(NodeObject).to receive(:run_ssh_cmd).with(
    "ceph health 2>&1"
  ).and_return(exit_code: 1, stdout: "Some message", stderr: "")

  expect(subject.ceph_healthy?).to be false
end
```

jsuchome:
@MaximilianMeister thanks!

crowbar_framework/spec/models/api/crowbar_spec.rb
```diff
@@ -125,6 +125,36 @@
     end
   end
 
+  context "with ceph cluster healthy" do
+
```
MaximilianMeister: I think you can drop this empty line

crowbar_framework/spec/models/api/crowbar_spec.rb
```diff
@@ -125,6 +125,36 @@
...
+      allow_any_instance_of(NodeObject).to(
+        receive(:run_ssh_cmd).with("ceph health 2>&1").
+        and_return(exit_code: 0, stdout: "Some message", stderr: "")
+      )
+      expect(subject.ceph_healthy?).to be true
+    end
+  end
+
+  context "with ceph cluster not healthy" do
+
```
MaximilianMeister: here too

crowbar_framework/spec/models/api/crowbar_spec.rb
jsuchome: Wow, if Hound is happy, we still have people! ;-)

jsuchome:

MaximilianMeister:
+1 

if `ceph health`only comes back with 0 if everything is healthy, and vice versa. I haven't tested that

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -116,6 +116,17 @@ def repocheck
       end
     end
 
+    def ceph_healthy?
+      ceph_node = NodeObject.find("roles:ceph-mon AND ceph_config_environment:*").first
+      return true if ceph_node.nil?
+      ssh_retval = ceph_node.run_ssh_cmd("ceph health 2>&1")
```
mjura: Maybe we should also check output from ceph health, I would propose something like this

```
ssh_retval = ceph_node.run_ssh_cmd("ceph health | grep -q HEALTH_OK")
```

What do you think?

jsuchome: That's indeed possible, but could it exit with 0 while reporting anything else than HEALTH_OK?

jsuchome: Yep. From @tserong "it'll return 0 even if there's a mon down and it's in HEALTH_WARN"

jsuchome:
@mjura , @MaximilianMeister adapted

crowbar_framework/spec/models/api/crowbar_spec.rb
```diff
@@ -125,6 +125,45 @@
...
+      allow(NodeObject).to(
+        receive(:find).with("roles:ceph-mon AND ceph_config_environment:*").
+        and_return([NodeObject.find_node_by_name("testing.crowbar.com")])
+      )
+      allow_any_instance_of(NodeObject).to(
+        receive(:run_ssh_cmd).with("LANG=C ceph health 2>&1").
+        and_return(exit_code: 1, stdout: "Some message", stderr: "")
+      )
+      expect(subject.ceph_healthy?).to be false
+    end
```
MaximilianMeister: can you add an empty line here?

crowbar_framework/spec/models/api/crowbar_spec.rb
```diff
@@ -125,6 +125,45 @@
...
+
+  context "with ceph cluster not healthy" do
+    it "fails when checking ceph cluster health" do
+      allow(NodeObject).to(
+        receive(:find).with("roles:ceph-mon AND ceph_config_environment:*").
+        and_return([NodeObject.find_node_by_name("testing.crowbar.com")])
+      )
+      allow_any_instance_of(NodeObject).to(
+        receive(:run_ssh_cmd).with("LANG=C ceph health 2>&1").
+        and_return(exit_code: 1, stdout: "Some message", stderr: "")
```
MaximilianMeister: not that it makes a real difference in the test, but

```ruby
and_return(exit_code: 1, stdout: "HEALTH_WARN", stderr: "")
```

is more representative for reality

MaximilianMeister:
thanks! failed gating looks not directly related...?

+1

crowbar_framework/spec/models/api/crowbar_spec.rb
```diff
@@ -125,6 +125,46 @@
...
+        and_return([NodeObject.find_node_by_name("testing.crowbar.com")])
+      )
+      allow_any_instance_of(NodeObject).to(
+        receive(:run_ssh_cmd).with("LANG=C ceph health 2>&1").
+        and_return(exit_code: 1, stdout: "HEALTH_ERR\n", stderr: "")
+      )
+      expect(subject.ceph_healthy?).to be false
+    end
+
+    it "fails when exit value of ceph check is 0 but stdout still not correct" do
```
mjura: I can imagine that there can be situations when exit code is different than `0` or `1` and stdout still different than `HEALTH_ERR` or `HEALTH_WARN`

mjura: I think we should ￼distinguish only two situations, when exit code is 0 and output HEALTH_OK and other rest.

mjura: No need to worry about HEALTH_WARN and HEALTH_ERR

mjura:
Gating looks not related to this change

crowbar_framework/spec/models/api/crowbar_spec.rb
jsuchome: The code itself only checks for HEALTH_OK. This is only additional test that covers the situation mentioned earlier, with exit code 0. (I  know it is not necessary, but it does not hurt)

mjura:
+1

PR #638, opened 2016-09-02
Switch to SES4 repos

tserong:
Note: the repomd piece is commented out, so it won't actually check the
repo properly (this means you can mount early SES4 DVD media instead of
a real repo if necessary).

### DIFF
--- a/crowbar_framework/config/repos-ses.yml
+++ b/crowbar_framework/config/repos-ses.yml
@@ -7,7 +7,7 @@ suse-12.1:
       features: ["os"]
       repomd:
         tag: "obsproduct://build.suse.de/SUSE:SLE-12-SP1:GA/SLES/12.1/POOL/x86_64"
-        key_md5: "9a62177e1c6852d48453ed3909956d6b"
+        fingerprint: "FEAB 5025 39D8 46DB 2C09 61CA 70AF 9E81 39DB 7C82"
       url:
       smt_path: "SUSE/Products/SLE-SERVER/12-SP1/x86_64/product"
       ask_on_error: false
@@ -17,7 +17,7 @@ suse-12.1:
       features: ["os"]
       repomd:
         tag: "obsrepository://build.suse.de/SUSE:Updates:SLE-SERVER:12-SP1:x86_64/update"
-        key_md5: "9a62177e1c6852d48453ed3909956d6b"
+        fingerprint: "FEAB 5025 39D8 46DB 2C09 61CA 70AF 9E81 39DB 7C82"
       url:
       smt_path: "SUSE/Updates/SLE-SERVER/12-SP1/x86_64/update"
       ask_on_error: false
@@ -27,7 +27,7 @@ suse-12.1:
       features: ["ha"]
       repomd:
         tag: "obsproduct://build.suse.de/SUSE:SLE-12-SP1:GA/sle-ha/12.1/POOL/x86_64"
-        key_md5: "9a62177e1c6852d48453ed3909956d6b"
+        fingerprint: "FEAB 5025 39D8 46DB 2C09 61CA 70AF 9E81 39DB 7C82"
       url:
       smt_path: "SUSE/Products/SLE-HA/12-SP1/x86_64/product"
       ask_on_error: false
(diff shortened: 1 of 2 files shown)


tserong:
This (or something like it) will probably be useful soonish.

vuntz:
Note that this requires changes in mkcloud (I assume) as we would want to deploy ceph on SP2 too with SES 4.

crowbar_framework/config/repos-ses.yml
```diff
@@ -41,25 +41,25 @@ suse-12.1:
...
       ask_on_error: false
-    suse-enterprise-storage-3-pool:
-      name: "SUSE-Enterprise-Storage-3-Pool"
+    suse-enterprise-storage-4-pool:
+      name: "SUSE-Enterprise-Storage-4-Pool"
       required: "optional"
       features: ["ceph"]
-      repomd:
-        tag: "obsproduct://build.suse.de/SUSE:SLE-12-SP1:Update:Products:SES3/ses/3/POOL/x86_64"
-        key_md5: "9a62177e1c6852d48453ed3909956d6b"
```
vuntz: Oh, I guess we didn't do the change for this file, but if you look at the repos-cloud.yml file, we moved to fingerprints. So you shouldn't need anymore to remove things like you did.

Maybe move to fingerprints here there?

vuntz:
Btw, is it on purpose that you don't change Cloud to deploy SES 4 (yet)?

tserong:
> Btw, is it on purpose that you don't change Cloud to deploy SES 4 (yet)?

How do you mean?  It's not enough to just change the repo file?

vuntz:
>> Btw, is it on purpose that you don't change Cloud to deploy SES 4 (yet)?
> 
> How do you mean? It's not enough to just change the repo file?

The repo file you're changing is only used by SES and never used by Cloud. The SES repos that Cloud uses to deploy Ceph are defined in the other repo file :-)

tserong:
> The repo file you're changing is only used by SES and never used by Cloud. The SES repos that Cloud uses to deploy Ceph are defined in the other repo file :-)

Ah! I are idiot :-) Thank you, shall address.

tserong:
OK, this should be more like it now (although I can't do a proper test until the SES4 POOL and Updates repos actually come into existence, which I'm sure will be RSN)

crowbar_framework/config/repos-ses.yml
```diff
@@ -37,27 +37,27 @@ suse-12.1:
...
         tag: "obsrepository://build.suse.de/SUSE:Updates:SLE-HA:12-SP1:x86_64/update"
-        key_md5: "9a62177e1c6852d48453ed3909956d6b"
+        fingerprint: "FEAB 5025 39D8 46DB 2C09 61CA 70AF 9E81 39DB 7C82"
       url:
       smt_path: "SUSE/Updates/SLE-HA/12-SP1/x86_64/update"
       ask_on_error: false
     suse-enterprise-storage-3-pool:
       name: "SUSE-Enterprise-Storage-3-Pool"
-      required: "optional"
+      required: "mandatory"
```
toabctl: Jut for clarification: It's now mandatory because it's the ```repos-ses.yml``` file which is used for deploying SES with crowbar, right?

vuntz:
Any reason to keep the SES 3 repos? Is it to not break stuff for the time being?

crowbar_framework/config/repos-ses.yml
tserong: Correct. repos-ses.yml will only be used for storage-only crowbar deploys, and in that environment it makes sense to make it mandatory IMO

tserong:
> Any reason to keep the SES 3 repos? Is it to not break stuff for the time being?

No particular reason. Just that SES3 was already there under suse-12.1, and I've added SES4 under suse-12.2.

dirkmueller:
+1

vuntz:
+1, but once all the PR are merged for the SES 4 switch, can you create a followup PR to remove the SES 3 bits from the files?

tserong:
Sure

vuntz:
Actually, because we're not removing anything, we can merge this one right away.

PR #639, opened 2016-09-02
Use zypper --xmlout for parsing zypper

fschueller:
Less error prone than dissecting the lines of string output

### DIFF
--- a/chef/cookbooks/provisioner/recipes/base.rb
+++ b/chef/cookbooks/provisioner/recipes/base.rb
@@ -304,16 +304,12 @@ if node[:platform_family] == "suse" && !node.roles.include?("provisioner-server"
       current_url = nil
       current_priority = nil
 
-      out = `LANG=C zypper --non-interactive repos #{name} 2> /dev/null`
-      out.split("\n").each do |line|
-        attribute, value = line.split(":", 2).map(&:strip)
-        next if value.nil?
-        if attribute == "URI"
-          current_url = value
-        elsif attribute == "Priority"
-          current_priority = value.to_i
-        end
-      end
+      out = REXML::Document.new(
+        `LANG=C zypper --non-interactive --xmlout repos #{name} 2> /dev/null`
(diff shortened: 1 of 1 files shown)


chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -304,15 +304,16 @@
...
-        if attribute == "URI"
-          current_url = value
-        elsif attribute == "Priority"
-          current_priority = value.to_i
-        end
+      out = REXML::Document.new(
+        `LANG=C zypper --non-interactive --xmlout repos Cloud 2> /dev/null`
+      ).root.elements
+
+      current_priority = unless out["repo-list/repo"].attributes["priority"].nil?
```
MaximilianMeister: I think this check can be dropped, as we can rely on the fact that there is always a url and a priority set for a repository. that would also make hound happy

MaximilianMeister:
+1

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -304,16 +304,12 @@
...
-        next if value.nil?
-        if attribute == "URI"
-          current_url = value
-        elsif attribute == "Priority"
-          current_priority = value.to_i
-        end
-      end
+      out = REXML::Document.new(
+        `LANG=C zypper --non-interactive --xmlout repos #{name} 2> /dev/null`
+      ).root.elements
```
toabctl: What if the repo does not exist? Then I get ```NoMethodError: undefined method `attributes' for nil:NilClass```

MaximilianMeister: did you get the repo from the provisioner library?

toabctl: No. Just asking for the case that this happens.

nicolasbock:
+1

MaximilianMeister:
still +1

PR #640, opened 2016-09-02
Prepare the node OS upgrade

jsuchome:
Create an upgrade script on nodes.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/prepare-os-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-os-upgrade.rb
@@ -1,0 +1,43 @@
+# Cookbook Name:: crowbar
+# Recipe:: prepare-os-upgrade
+#
+# Copyright 2013-2016, SUSE LINUX Products GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# This recipe prepares a script which when executed (at selected time)
+# 1. removes old repositories
+# 2. adds correct new ones
+# 3. runs zypper dup to upgrade the node
+
+old_repos = Provisioner::Repositories.get_repos(
+  node[:platform], node[:platform_version], node[:kernel][:machine]
+)
+
+target_platform, target_platform_version = node[:target_platform].split("-")
+new_repos = Provisioner::Repositories.get_repos(
+  target_platform, target_platform_version, node[:kernel][:machine]
+)
+
+template "/usr/sbin/crowbar-upgrade-os.sh" do
+  source "crowbar-upgrade-os.erb"
+  mode "0770"
+  owner "root"
+  group "root"
+  action :create
+  variables(
+    old_repos: old_repos,
+    new_repos: new_repos,
+    target_platform_version: target_platform_version
+  )
+end
--- a/chef/cookbooks/crowbar/recipes/prepare-os-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-os-upgrade.rb
@@ -20,14 +20,38 @@
 # 2. adds correct new ones
 # 3. runs zypper dup to upgrade the node
 
+arch = node[:kernel][:machine]
+
 old_repos = Provisioner::Repositories.get_repos(
-  node[:platform], node[:platform_version], node[:kernel][:machine]
+  node[:platform], node[:platform_version], arch
 )
 
 target_platform, target_platform_version = node[:target_platform].split("-")
 new_repos = Provisioner::Repositories.get_repos(
-  target_platform, target_platform_version, node[:kernel][:machine]
+  target_platform, target_platform_version, arch
 )
+
+# Find out the location of the base system repository
+provisioner = search(:node, "roles:provisioner-server")[0]
+admin_ip = Barclamp::Inventory.get_network_by_type(provisioner, "admin").address
+
+web_port = provisioner[:provisioner][:web_port]
+provisioner_web = "http://#{admin_ip}:#{web_port}"
+
+web_path = "#{provisioner_web}/#{node[:platform]}-#{node[:platform_version]}/#{arch}"
+old_install_url = "#{web_path}/install"
+
+web_path = "#{provisioner_web}/#{node[:target_platform]}/#{arch}"
+new_install_url = "#{web_path}/install"
+
+# try to create an alias for new base repo from the original base repo
+repo_alias = "SLES12-SP1-12.1-0"
+doc = REXML::Document.new(`zypper --xmlout lr --details`)
+doc.elements.each("stream/repo-list/repo") do |repo|
+  repo_alias = repo.attributes["alias"] if repo.elements["url"].text == old_install_url
+end
+
+new_alias = repo_alias.gsub("SP1", "SP2").gsub(node[:platform_version], target_platform_version)
 
 template "/usr/sbin/crowbar-upgrade-os.sh" do
   source "crowbar-upgrade-os.erb"
@@ -38,6 +62,9 @@ template "/usr/sbin/crowbar-upgrade-os.sh" do
   variables(
     old_repos: old_repos,
     new_repos: new_repos,
-    target_platform_version: target_platform_version
+    target_platform_version: target_platform_version,
+    old_base_repo: old_install_url,
+    new_base_repo: new_install_url,
+    new_alias: new_alias
   )
 end
--- a/chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.erb
@@ -1,0 +1,77 @@
+#!/bin/bash
+#
+# This script starts the upgrade of the node server to the latest version of
+# system and Cloud product.
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+mkdir -p "`dirname "$LOGFILE"`"
+
+# Copy stdout to fd 3
+exec 3>&1
+# Create fd 4 for logfile
+exec 4>> "$LOGFILE"
+# Set fd 1 and 2 to logfile
+exec 1>&4 2>&1
+
+set -x
+
+UPGRADEDIR=/var/lib/crowbar/upgrade
+RUNFILE=/var/run/crowbar/node-upgrading
+
+cleanup()
+{
+    echo "cleaning up after interrupt or exit"
+
+    # cleanup upgrading indication so that the action can be restarted
+    rm -f $RUNFILE
+
+    exit 2
+}
+
+initiate_node_upgrade()
+{
+    mkdir -p $UPGRADEDIR
+
+    if [[ -f $RUNFILE ]] ; then
+        echo "Exit: Upgrade is already running..."
+        exit 1
+    fi
+
+    if [[ -f $UPGRADEDIR/node-upgraded-ok ]] && grep -q "<%= @target_platform_version %>" $UPGRADEDIR/node-upgraded-ok ; then
+        echo "Exit: Node already upgraded"
+        exit 0
+    fi
+
+    # Signalize that the upgrade is running
+    touch $RUNFILE
+
+    trap cleanup INT EXIT
+
+    # Upgrade the distribution non-interactively
+    zypper --no-color --releasever <%= @target_platform_version %> ref -f
+    zypper --no-color --non-interactive dist-upgrade -l --recommends --replacefiles
+    ret=$?
+    if [ $ret != 0 ]; then
+        echo "zypper dist-upgrade has failed with $ret, check zypper logs"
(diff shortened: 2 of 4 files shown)


chef/cookbooks/crowbar/recipes/prepare-os-upgrade.rb
```diff
@@ -0,0 +1,65 @@
...
+web_path = "#{provisioner_web}/#{node[:platform]}-#{node[:platform_version]}/#{arch}"
+old_install_url = "#{web_path}/install"
+
+web_path = "#{provisioner_web}/#{node[:target_platform]}/#{arch}"
+new_install_url = "#{web_path}/install"
+
+# try to create an alias for new base repo
+zypper_out = `zypper lr --details | grep #{old_install_url}`
+repo_alias = zypper_out.split("| ")[1] || "SLES12-SP1-12.1-0"
+new_alias = repo_alias.gsub("SP1", "SP2").gsub(node[:platform_version], target_platform_version)
```
jsuchome: This part looks ugly, but I don't know where else should I get the alias for the base repo. I think normally this is the one passed to the nodes at the boot line during autoinstallation....

jsuchome:

chef/cookbooks/crowbar/recipes/prepare-os-upgrade.rb
MaximilianMeister: why not use `--xmlout` and the REXML lib like in https://github.com/crowbar/crowbar-core/pull/639?

jsuchome: Yeah... although my main concern is actually using zypper for reading that information, I'd like to get it elsewhere if possible

chef/cookbooks/crowbar/templates/default/upgrade_cloud_node.sh.erb
```diff
@@ -0,0 +1,63 @@
...
+# Create fd 4 for logfile
+exec 4>> "$LOGFILE"
+# Set fd 1 and 2 to logfile
+exec 1>&4 2>&1
+
+set -x
+
+initiate_node_upgrade()
+{
+    local installdir=/var/lib/crowbar/install
```
vuntz: Not sure if this path is completely correct; maybe s/install/upgrade/ ?

chef/cookbooks/crowbar/templates/default/upgrade_cloud_node.sh.erb
```diff
@@ -0,0 +1,63 @@
...
+
+    # Upgrade the distribution non-interactively
+    zypper --no-color --releasever <%= @target_platform_version %> ref -f
+    zypper --no-color --non-interactive dist-upgrade -l --recommends --replacefiles
+    ret=$?
+    if [ $ret != 0 ]; then
+        echo "zypper dist-upgrade has failed with $ret, check zypper logs"
+        echo "$ret" > $installdir/node-upgrade-failed
+        # cleanup upgrading indication
+        rm -rf $installdir/node-upgrading
```
vuntz: Not important, but `-r` shouldn't be needed.

chef/cookbooks/crowbar/templates/default/upgrade_cloud_node.sh.erb
```diff
@@ -0,0 +1,63 @@
...
+        return
+    fi
+
+    # Signalize that the upgrade correctly ended
+    touch $installdir/node-upgraded-ok
+
+    # cleanup upgrading indication
+    rm -rf $installdir/node-upgrading
```
vuntz: Same here.

vuntz: Generally speaking: shouldn't we use `trap` to remove the flag file?

chef/cookbooks/crowbar/recipes/prepare-os-upgrade.rb
```diff
@@ -0,0 +1,65 @@
...
+)
+
+target_platform, target_platform_version = node[:target_platform].split("-")
+new_repos = Provisioner::Repositories.get_repos(
+  target_platform, target_platform_version, arch
+)
+
+# Find out the location of the base system repository
+provisioner = search(:node, "roles:provisioner-server")[0]
+admin_ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(provisioner, "admin").address
```
vuntz: For the record: `Chef::Recipe::` could be omitted.

chef/cookbooks/crowbar/recipes/prepare-os-upgrade.rb
```diff
@@ -0,0 +1,65 @@
...
+
+web_path = "#{provisioner_web}/#{node[:platform]}-#{node[:platform_version]}/#{arch}"
+old_install_url = "#{web_path}/install"
+
+web_path = "#{provisioner_web}/#{node[:target_platform]}/#{arch}"
+new_install_url = "#{web_path}/install"
+
+# try to create an alias for new base repo
+zypper_out = `zypper lr --details | grep #{old_install_url}`
+repo_alias = zypper_out.split("| ")[1] || "SLES12-SP1-12.1-0"
```
vuntz: The `[1]` will fail badly and raise an exception if `grep` doesn't return anything. We should protect against this.

chef/cookbooks/crowbar/recipes/prepare-os-upgrade.rb
vuntz: The only other option is to parse the repo config directly, which doesn't sound much better :-)

chef/cookbooks/crowbar/recipes/prepare-os-upgrade.rb
```diff
@@ -0,0 +1,65 @@
...
+new_alias = repo_alias.gsub("SP1", "SP2").gsub(node[:platform_version], target_platform_version)
+
+template "/usr/local/bin/upgrade_cloud_node.sh" do
```
vuntz: I'd put in `/usr/sbin` (not `/usr/local` because it's not created by the user, and not `bin` because it's really very special). Also, can we do s/_/-/ in the filename.

Actually, I'd rename it to /usr/sbin/crowbar-upgrade-os. (No mention of cloud, as it's not cloud specific)

chef/cookbooks/crowbar/templates/default/upgrade_cloud_node.sh.erb
```diff
@@ -0,0 +1,63 @@
...
+    if [ $ret != 0 ]; then
+        echo "zypper dist-upgrade has failed with $ret, check zypper logs"
+        echo "$ret" > $installdir/node-upgrade-failed
+        # cleanup upgrading indication
+        rm -rf $installdir/node-upgrading
+        return
+    fi
+
+    # Signalize that the upgrade correctly ended
+    touch $installdir/node-upgraded-ok
```
vuntz: I just realized:

- we should probably not do anything from this script if that file exists
- we should somehow put a version with this file (either in the content, or in the filename) so that we know it's for the 6->7 upgrade

vuntz:
One thing I wonder: should we remove the script from the node in the next step of the upgrade process, or do we just leave it there forever?

jsuchome:
@vuntz not sure about that trap usage... I think we want to remove `node-upgrading` file when we exit because of zypper failure (so we can restart the step), but we do not want to remove it when the script was accidentally started again ("Exit: Upgrade is already running")

jsuchome:
I'm not sure if removing or keeping the file is the correct. We might want to use it for next upgrade, so I'd lean to keeping.

chef/cookbooks/crowbar/templates/default/upgrade_cloud_node.sh.erb
jsuchome: Adapted.

chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.erb
```diff
@@ -0,0 +1,68 @@
...
+{
+    local upgradedir=/var/lib/crowbar/upgrade
+    mkdir -p $upgradedir
+
+    if [[ -f $upgradedir/node-upgrading ]] ; then
+        echo "Exit: Upgrade is already running..."
+        exit 1
+    fi
+
+    if [[ -f $upgradedir/node-upgraded-ok ]] ; then
```
vuntz: Strictly speaking, I guess you need to check for the content of the file now :-)

vuntz:
> @vuntz not sure about that trap usage... I think we want to remove node-upgrading file when we exit because of zypper failure (so we can restart the step), but we do not want to remove it when the script was accidentally started again ("Exit: Upgrade is already running")

Sure; it's just a matter of calling `trap` after doing the initial check. My point is that if the script is killed by accident (by the user, by the kernel, or anything else), we don't want to be blocked for no reason.

That also makes me think that specific file should live in `/var/cache` so that a reboot will come fresh.

chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.erb
jsuchome: Ha! I think only the next generation of this script needs to check the content, now there can be only one value...

jsuchome:
@vuntz addressed all you issues, except that `/var/cache` one. I think we want to use `upgradedir` also to store the information that persists, so you only mean the `node-upgrading` to have elsewhere. But /var/cache does not have to be cleaned on reboot - or when it is?

vuntz:
> But /var/cache does not have to be cleaned on reboot - or when it is?

It's tmpfs (isn't it?), so it is cleaned on boot :-)

chef/cookbooks/crowbar/recipes/prepare-os-upgrade.rb
```diff
@@ -0,0 +1,72 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# This recipe prepares a script which when executed (at selected time)
+# 1. removes old repositories
+# 2. adds correct new ones
+# 3. runs zypper dup to upgrade the node
+
+require "rexml/document"
```
MaximilianMeister: I think you don't need to require it explicitly, it is built into the stdlib

chef/cookbooks/crowbar/recipes/prepare-os-upgrade.rb
```diff
@@ -0,0 +1,72 @@
+# Cookbook Name:: crowbar
```
jsuchome: What is wrong here?

jsuchome:
@vuntz adapted to use `/var/run`

chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.erb
```diff
@@ -0,0 +1,79 @@
...
+    zypper --no-color --non-interactive dist-upgrade -l --recommends --replacefiles
+    ret=$?
+    if [ $ret != 0 ]; then
+        echo "zypper dist-upgrade has failed with $ret, check zypper logs"
+        echo "$ret" > $UPGRADEDIR/node-upgrade-failed
+        exit $ret
+    fi
+
+    # Signalize that the upgrade correctly ended
+    echo "<%= @target_platform_version %>" > $UPGRADEDIR/node-upgraded-ok
```
vuntz: I wonder if `>>` would be better, but we probably don't know yet.

chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.erb
```diff
@@ -0,0 +1,79 @@
...
+exec 3>&1
+# Create fd 4 for logfile
+exec 4>> "$LOGFILE"
+# Set fd 1 and 2 to logfile
+exec 1>&4 2>&1
+
+set -x
+
+UPGRADEDIR=/var/lib/crowbar/upgrade
+RUNFILE=/var/run/node-upgrading
```
vuntz: I'd have put that in /var/run/crowbar/ or put a crowbar prefix in the basename, but no huge deal.

vuntz:
Thanks; I have minor comments, but feel free to ignore. +1

MaximilianMeister:
LGTM (but haven't actually tested it)

chef/cookbooks/crowbar/recipes/prepare-os-upgrade.rb
abel-navarro: doesn't look important. I think expects something like `#!/usr/bin/env ruby`. http://stackoverflow.com/questions/38325533/disable-frozen-string-literal-comment-checking

chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.erb
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+#
+# This script starts the upgrade of the node server to the latest version of
+# system and Cloud product.
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+mkdir -p "`dirname "$LOGFILE"`"
+
+# Copy stdout to fd 3
+exec 3>&1
```
abel-navarro: This part is a bit confusing. First line redirects fd3 to fd1, which is different from what comment says. Also it is not clear where fd3 is used.

Also, may be it's clearer to redirect fd1 and fd2 to fd4 instead of fd1 to fd4 and fd2 to fd1.

jsuchome: That's the same code we're using elsewhere (like in install-suse-cloud)

abel-navarro:
+1

PR #641, opened 2016-09-06
Search for number of compute nodes.

jsuchome:
Ideally we would like to have a check telling us that it will be
possible to migrate all workloads out from one node.
We can't have such check 100% reliable, so we rather won't do any.

(cherry picked from commit 139fd6904ac582db7d4442a9605e63fb3e2ed72d)

Forward port of https://github.com/crowbar/crowbar-core/pull/623

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -128,9 +128,12 @@ module Api
     end
 
     def compute_resources_available?
-      compute_nodes = NodeObject.find("roles:nova-compute-kvm")
-      if compute_nodes.size == 1
-        Rails.logger.warn("Only one compute node found; non-disruptive upgrade is not possible!")
+      ["kvm", "xen"].each do |virt|
+        compute_nodes = NodeObject.find("roles:nova-compute-#{virt}")
+        next unless compute_nodes.size == 1
+        Rails.logger.warn(
+          "Found only one compute node of #{virt} type; non-disruptive upgrade is not possible"
+        )
         return false
       end
       true
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -127,6 +127,15 @@ module Api
       true
     end
 
+    def compute_resources_available?
+      compute_nodes = NodeObject.find("roles:nova-compute-kvm")
+      if compute_nodes.size == 1
+        Rails.logger.warn("Only one compute node found; non-disruptive upgrade is not possible!")
+        return false
+      end
+      true
+    end
+
     protected
 
     def lib_path
--- a/crowbar_framework/spec/models/api/crowbar_spec.rb
+++ b/crowbar_framework/spec/models/api/crowbar_spec.rb
@@ -33,6 +33,7 @@ describe Api::Crowbar do
       "spec/fixtures/crowbar_repocheck_zypper_locked.xml"
     ).to_s
   end
+  let!(:node) { NodeObject.find_node_by_name("testing") }
 
   before(:each) do
     allow_any_instance_of(Kernel).to(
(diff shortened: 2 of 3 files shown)


jsuchome:

MaximilianMeister:
slightly misleading description, this is a forward port ;-) and I the unit test is still missing, also in the `stable/3.0` branch

crowbar_framework/spec/models/api/crowbar_spec.rb
```diff
@@ -125,6 +125,36 @@
     end
   end
 
+  context "with enough compute resources" do
+    it "succeeds to find enough compute nodes" do
+      allow(NodeObject).to(
+        receive(:find).with("roles:nova-compute-kvm").
+        and_return([NodeObject.find_node_by_name("node1"), NodeObject.find_node_by_name("node2")])
```
MaximilianMeister: thanks for adding the tests, but you have to use `[NodeObject.find_node_by_name("testing.crowbar.com"), NodeObject.find_node_by_name("testing.crowbar.com")]`, as we only have one node and the admin node in the fixtures.

crowbar_framework/spec/models/api/crowbar_spec.rb
```diff
@@ -44,6 +44,7 @@
         with(pid).
         and_return(pid)
     )
+    @node = NodeObject.find_node_by_name("testing")
```
MaximilianMeister: we don't use class variables anymore in the specs, (see http://betterspecs.org/#let) please just define it in a `let` block

```ruby
let!(:node) { NodeObject.find_node_by_name("testing") }
```

jsuchome: Well, we're using it in current code at several places... :-)

crowbar_framework/spec/models/api/crowbar_spec.rb
```diff
@@ -125,6 +126,36 @@
     end
   end
 
+  context "with enough compute resources" do
+    it "succeeds to find enough compute nodes" do
+      allow(NodeObject).to(
+        receive(:find).with("roles:nova-compute-kvm").
+        and_return([:node, :node])
```
MaximilianMeister: does that really return the nodes? I think it should be just `node`

crowbar_framework/spec/models/api/crowbar_spec.rb
MaximilianMeister: yes, old code, but the new code shouldn't use it anymore

crowbar_framework/spec/models/api/crowbar_spec.rb
jsuchome: It passes for all 3 test cases, so I assume it really works this way

MaximilianMeister: it passes, because it doesn't perform anything on the NodeObject, just the size of the Array. so this is wrong

MaximilianMeister:
+1

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -116,6 +116,15 @@ def repocheck
       end
     end
 
+    def compute_resources_available?
+      compute_nodes = NodeObject.find("roles:nova-compute-kvm")
```
rsalevsky: Why only kvm? Is xen not supported in the upgrade case?

jsuchome: Hm, good point. We should probably do also a search for xen compute roles. However, I assume it's not possible to migrate instances from kvm host to xen one (and vice versa) so it should be additional check.
(Correct me if I'm wrong here)

jsuchome:
Once this is merged, we really need to revisit all the check functions and let them return meaninfull message for the UI, not just true/false

rsalevsky:
+1

MaximilianMeister:
still +1

rsalevsky:
Still +1

PR #642, opened 2016-09-06
Don't try to write PXE config files on S390

rhafer:


dirkmueller:
+1

rsalevsky:
+1

dirkmueller:
alternative solution is in #649

rhafer:
The alternative solution was merged.

PR #644, opened 2016-09-07
integrate repocheck for nodes in the upgrade api

MaximilianMeister:
integrate the repochecks for nodes, which can be performed for the system repos and the addon repos.

expose an accumulated check through the upgrade API

some code moved from https://github.com/crowbar/crowbar-ceph/pull/68

### DIFF
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -80,6 +80,12 @@ class Api::UpgradeController < ApiController
     render json: { error: e.message }, status: :unprocessable_entity
   end
 
+  api :GET, "/api/upgrade/repocheck", "Check for missing node repositories"
+  api_version "2.0"
+  def repocheck
+    render json: @upgrade.repocheck
+  end
+
   protected
 
   def set_upgrade
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -24,7 +24,8 @@ module Api
 
       {}.tap do |ret|
         ret[addon] = {
-          "available" => true
+          "available" => true,
+          "repos" => {}
         }
         platform = Api::Upgrade.new.target_platform(platform_exception: addon)
 
@@ -40,7 +41,6 @@ module Api
                 feature, platform, architecture
               )
               ret[addon]["available"] &&= available
-              ret[addon]["repos"] ||= {}
               ret[addon]["repos"].deep_merge!(repolist.deep_stringify_keys)
             end
           else
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -20,6 +20,7 @@ module Api
       addon = options.fetch(:addon, "os")
       features = []
       features.push(addon)
+      architectures = node_architectures
 
       {}.tap do |ret|
         ret[addon] = {
@@ -28,8 +29,8 @@ module Api
         platform = Api::Upgrade.new.target_platform(platform_exception: addon)
 
         features.each do |feature|
-          if node_architectures(addon: addon).any?
-            node_architectures(addon: addon).each do |architecture|
+          if architectures[feature]
+            architectures[feature].each do |architecture|
               unless ::Crowbar::Repository.provided_and_enabled?(feature,
                                                                  platform,
                                                                  architecture)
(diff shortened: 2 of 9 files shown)


crowbar_framework/app/models/api/node.rb
```diff
@@ -15,6 +15,56 @@
...
 
 module Api
-  class Node
+  class Node < Tableless
+    def repocheck(options = {})
+      addon = options.fetch(:addon, "os")
+      features = ["os"]
+      features.push(addon) if addon
+
+      {}.tap do |ret|
```
rsalevsky: the ret variable is somehow meaning less, maybe you can replace it with a better name

crowbar_framework/app/models/api/node.rb
```diff
@@ -15,6 +15,56 @@
...
+        ret[addon] = {
+          "available" => true
+        }
+        platform = Api::Upgrade.new.target_platform(platform_exception: addon)
+
+        features.each do |feature|
+          node_architectures(addon: addon).each do |architecture|
+            unless ::Crowbar::Repository.provided_and_enabled?(feature,
```
jsuchome: Does the `provided_and_enabled` take "os" value for a feature? Was that added in some PR I do not see?
And when do we check for presence of Cloud repository?

MaximilianMeister: > Does the provided_and_enabled take "os" value for a feature? Was that added in some PR I do not see?

it takes any value that is defined in the repos.yml in the features section



> And when do we check for presence of Cloud repository?

this I am not yet sure

crowbar_framework/app/models/api/node.rb
MaximilianMeister: `ret` means `return` in short, I just used it because we are using it in many places in crowbar, not sure what else to call it

crowbar_framework/app/models/api/node.rb
jsuchome: > And when do we check for presence of Cloud repository?

I think it should be part of the same API call

jsuchome: > > Does the provided_and_enabled take "os" value for a feature? Was that added in some PR I do not see?

> it takes any value that is defined in the repos.yml in the features section

Ah, I see!

MaximilianMeister: > I think it should be part of the same API call

we just have to add `openstack` to the feature list I guess

MaximilianMeister:
@vuntz still not sure if everything is correct here, and works as expected

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -34,6 +34,15 @@ def check
       }
     end
 
+    def repocheck
+      response = {}
+      addons = Api::Crowbar.new.addons
+      addons.push("os").push("openstack").each do |addon|
```
jsuchome: What about `addons.concat([ "os", "openstack"]).each` ?

(But I do not really care)

MaximilianMeister: now that you mention i also could call `addons.push("os", "openstack")`

MaximilianMeister:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;ncdOGOzP&#x2F;97-3-s47p11-implement-api-for-repo-checks-during-upgrade-upgrade-step-8"> (3) S47P11 Implement API for repo checks during upgrade (upgrade step 8)</a>

crowbar_framework/app/models/api/node.rb
```diff
@@ -15,6 +15,61 @@
...
+          else
+            ret.delete[addon]
+          end
+        end
+      end
+    end
+
+    protected
+
+    def node_architectures(options = {})
```
MaximilianMeister: @vuntz how can we determine the architectures for the `os` and the `openstack` feature? i just realized, we can't do the same check (by one deployed barclamp) as here in `node_architectures`

crowbar_framework/app/models/api/node.rb
```diff
@@ -15,6 +15,72 @@
...
+          if pacemaker_node?(node)
+            ret["ha"] ||= []
+            ret["ha"].push(arch) unless ret["ha"].include?(arch)
+          end
+        end
+      end
+    end
+
+    def ceph_node?(node)
+      node.roles.include?("ceph-mon")
```
jsuchome: There might be different roles, not just `ceph-mon`, look into https://github.com/crowbar/crowbar-ceph/tree/master/chef/roles

jsuchome: Can we look for `ceph-config-default` ?

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -80,6 +80,12 @@ def cancel
     render json: { error: e.message }, status: :unprocessable_entity
   end
 
+  api :GET, "/api/upgrade/repocheck", "Check for missing node repositories"
```
jsuchome: So, this was raised before by @vuntz , shouldn't we rather call it differently than repocheck for admin server repositories?

jsuchome:
Seems to work well now.

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: we could also adapt the response of `/api/crowbar/repocheck` to look the same. that would be better for using it in `crowbarctl` 

btw https://github.com/crowbar/crowbar-client/pull/104 adapts `crowbarctl` to the new node repocheck

MaximilianMeister: https://github.com/crowbar/crowbar-core/issues/652 adapts the response of the admin repocheck

jsuchome:
Actually, I do not mind if it is renamed or not. +1

rsalevsky:
Didn't test it but codewise +1

jsuchome:
+1

crowbar_framework/app/models/api/node.rb
```diff
@@ -15,6 +15,72 @@
...
+        end
+      end
+    end
+
+    def ceph_node?(node)
+      node.roles.include?("ceph-config-default")
+    end
+
+    def pacemaker_node?(node)
+      node.roles.grep(/^pacemaker-config-.*/).any?
```
jsuchome: We could check for `node["run_list_map"].key? "pacemaker-cluster-member"`. But I think your check is correct as well.

MaximilianMeister: if it doesn't make a real difference, than lets keep it as it is

PR #647, opened 2016-09-07
Initiate the upgrade of controller nodes

jsuchome:
This is just a dumb function now. 

But once we merge https://github.com/crowbar/crowbar-core/pull/640 , nodes won't be upgraded for some time which will break the upgrade CI job


abel-navarro:
+1

PR #648, opened 2016-09-08
[stable/3.0] upgrade: Move the cluster health check from crowbar_service

jsuchome:
In master, we have this function under api/cluster (which is not present in 3.0), see https://github.com/crowbar/crowbar-ha/pull/138

But here in 3.0, it probably belongs to the same place where the other checks are (ceph_healthy? etc.)

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -114,18 +114,49 @@ module Api
       end
     end
 
+
+    # Simple check if HA clusters report some problems
+    # If there are no problems, empty hash is returned.
+    # If this fails, information about failed actions for each cluster founder is
+    # returned in a hash that looks like this:
+    # {
+    #     "crm_failures" => {
+    #             "node1" => "reason for crm status failure"
+    #     },
+    #     "failed_actions" => {
+    #             "node2" => "Failed action on this node"
+    #     }
+    # }
+    # User has to manually clean pacemaker resources before proceeding with the upgrade.
     def clusters_healthy?
-      service_object = CrowbarService.new(Rails.logger)
-      cluster_health = service_object.check_cluster_health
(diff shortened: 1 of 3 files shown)


crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -114,18 +114,49 @@ def repocheck
...
-      unless cluster_health.empty?
-        Rails.logger.warn("HA clusters report some problems")
-        (cluster_health["crm_failures"] || {}).each do |node, error|
-          Rails.logger.warn("crm status at node #{node} reports error:\n#{error}")
+      cluster_health = {}
+      crm_failures = {}
+      failed_actions = {}
+
+      founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
+      return true
```
jsuchome: Oh :-)

jsuchome:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;8X7elkka&#x2F;82-cleanup-upgrade-related-mess"> Cleanup upgrade related mess</a>

rsalevsky:
+1 But we should try to get the gating green.

jsuchome:
@rsalevsky do you have idea what's actually wrong? I did not add any new file, and local rake spec is working (as well as travis check)

rsalevsky:
Retriggert the gating, the change and the error looks unrelated.

MaximilianMeister:
a bit of confusion as well in the test. `context "with no ha cluster" ... expect(subject.clusters_healthy?).to be true` (same with ceph btw) but this is minor, as there is no other way to handle it with a boolean

so overall this looks good, although I haven't really tested it

+1

jsuchome:
We'll need to work on the return values when we do that synchronization work. For now, this is really mostly for moving one function to correct place

PR #649, opened 2016-09-08
Restrict PXE boot to those architectures that are supported

dirkmueller:


chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -88,32 +86,29 @@ def find_node_boot_mac_addresses(node, admin_data_net)
...
-      pxefile = "#{discovery_dir}/#{arch}/#{pxecfg_subdir}/#{boot_ip_hex}"
-      uefi_dir = "#{discovery_dir}/#{arch}/#{uefi_subdir}"
-      if use_elilo
-        uefifile = "#{uefi_dir}/#{boot_ip_hex}.conf"
-        grubdir = nil
-        grubcfgfile = nil
-        grubfile = nil
-      else
+    pxefile = nil
+    uefifile = nil
```
rhafer: AFAICS ```uefifile``` is now always nil and not really used anymore. Seems it can be removed.

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -431,7 +426,7 @@ def find_node_boot_mac_addresses(node, admin_data_net)
       [{ file: pxefile, src: "default.erb" },
        { file: uefifile, src: "default.elilo.erb" }].each do |t|
```
rhafer: ```uefifile``` is always nil. Which also means we can remove the ```default.elilo.erb``` template.

rhafer:
+1

nicolasbock:
+1

PR #651, opened 2016-09-08
Temporary fix to enable upgrade of the nodes

jsuchome:
After merging https://github.com/crowbar/crowbar-core/pull/640 we are currently not upgrading the nodes, only preparing them for upgrade

Upgrade of the nodes needs to be correctly orchestrated, to upgrade nodes one by one
Now we're adding this only only to imitate original autoyast-based upgrad and to
have current upgrade CI job working.

Remove this code once we have correct orchestration in place.

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -389,6 +389,14 @@ class CrowbarService < ServiceObject
     end
 
     commit_and_check_proposal
+
+    # FIXME: upgrade of the nodes needs to be correctly orchestrated, to upgrade nodes one by one
+    # Now we're adding this only to imitate original autoyast-based upgrade and to
+    # have current upgrade CI job working
+    upgrade_nodes.each do |node|
+      Rails.logger.info("initiating upgrade of node #{node.name}")
+      node.ssh_cmd("/usr/sbin/crowbar-upgrade-os.sh")
+    end
   end
 
   def revert_nodes_from_crowbar_upgrade


jsuchome:

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -389,6 +389,14 @@ def prepare_nodes_for_os_upgrade
     end
 
     commit_and_check_proposal
+
+    # FIXME: upgrade of the nodes needs to be correctly orchestrated, to upgrade nodes one by one
+    # Now we're adding this only only to imitate original autoyast-based upgrad and to
```
MaximilianMeister: `only only` and `s/upgrad/upgrade`

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -389,6 +389,14 @@ def prepare_nodes_for_os_upgrade
...
+    # have current upgrade CI job working
+    upgrade_nodes.each do |node|
+      Rails.logger.info("initiating upgrade of node #{node.name}")
+      node.ssh_cmd("/usr/sbin/crowbar-upgrade-os.sh")
```
MaximilianMeister: so this script is actually doing nothing, just imitating? or do we have it in git somewhere?

jsuchome: No, the script is already there, on all nodes, created by the proposal commit at line 391 (see https://github.com/crowbar/crowbar-core/pull/640 for details). 
But now it is not executed. 
By imitation, I mean: when we execute the script on every node at once, we do not care about non-disruptive upgrade, we just upgrade all nodes to SP2, like we did with the old upgrade path.

MaximilianMeister: okay thanks

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -389,6 +389,14 @@ def prepare_nodes_for_os_upgrade
     end
 
     commit_and_check_proposal
+
+    # FIXME: upgrade of the nodes needs to be correctly orchestrated, to upgrade nodes one by one
+    # Now we're adding this only to imitate original autoyast-based upgrade and to
+    # have current upgrade CI job working
+    upgrade_nodes.each do |node|
+      Rails.logger.info("initiating upgrade of node #{node.name}")
+      node.ssh_cmd("/usr/sbin/crowbar-upgrade-os.sh")
```
MaximilianMeister: you think it makes sense to check for the exit code here?

jsuchome: No, because it executes zypper dup in background. Later, when we use it properly, we have to use some checks if upgrade goes well, but this is really just a temporary hotfix

MaximilianMeister:
+1

jsuchome:
This would not be needed if we decide to go with https://github.com/SUSE-Cloud/automation/pull/1223

rsalevsky:
+1

jsuchome:
Hm, well, this is still not enough. The nodes are not back in ready state after reinstallation....

jsuchome:
Closing in favor of https://github.com/crowbar/crowbar-core/pull/662

PR #652, opened 2016-09-08
adapt admin repocheck response to the nodes repocheck response

MaximilianMeister:
when using the repocheck API for nodes and for crowbar the result
should have the same structure

@vuntz @jsuchome that's just one proposal, the other solution would be to rename one of `/api/crowbar/repocheck` or `/api/upgrade/repocheck` to clear a bit the confusion that the check is performed in a different way. (see https://github.com/crowbar/crowbar-core/pull/644)

I am not sure what is best here, but IMO it is nicer to have `crowbarctl` return the same json structure for the `repocheck` subcommand (see https://github.com/crowbar/crowbar-client/pull/104)

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -107,12 +107,23 @@ module Api
 
         products = zypper_stream["product_list"]["product"]
 
+        os_available = repo_version_available?(products, "SLES", "12.2")
         ret[:os] = {
-          available: repo_version_available?(products, "SLES", "12.2")
+          available: os_available,
+          repos: {}
         }
+        ret[:os][:repos][admin_architecture.to_sym] = {
+          missing_repos: ["SUSE Linux Enterprise Server 12 SP2"]
+        } unless os_available
+
+        cloud_available = repo_version_available?(products, "suse-openstack-cloud", "8")
         ret[:cloud] = {
-          available: repo_version_available?(products, "suse-openstack-cloud", "8")
+          available: cloud_available,
+          repos: {}
         }
+        ret[:cloud][:repos][admin_architecture.to_sym] = {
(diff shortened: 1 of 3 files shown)


jsuchome:
I agree with unifying the output of those different repository checks, that seems like a good idea.

I does not need to be alternative to renaming the API though, we could do both.

jsuchome:
+1

MaximilianMeister:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;otxWJSSF&#x2F;96-3-s47p10-implement-api-for-repo-check-for-admin-server-upgrade-step-4"> (3) S47P10 Implement API for repo check for admin server (upgrade step 4)</a>

rsalevsky:
+1

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -107,12 +107,23 @@ def repocheck
...
         products = zypper_stream["product_list"]["product"]
 
+        os_available = repo_version_available?(products, "SLES", "12.2")
         ret[:os] = {
-          available: repo_version_available?(products, "SLES", "12.2")
+          available: os_available,
+          repos: {}
         }
+        ret[:os][:repos][admin_architecture.to_sym] = {
+          missing_repos: ["SUSE Linux Enterprise Server 12 SP2"]
```
rsalevsky: For C8 it will be mostly SP3.

jsuchome: Ha! So we need a backport for 3.0 as well.

MaximilianMeister: > For C8 it will be mostly SP3.

fixed

rsalevsky:
+1

MaximilianMeister:
backport will follow

PR #653, opened 2016-09-09
Stop openstack services before upgrade

abel-navarro:
Skip keystone and neutron so L3 agents can migrate.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -75,67 +75,14 @@ when "openstack_shutdown"
 
   include_recipe "crowbar::stop-services-before-upgrade"
 
-  bash "delete pacemaker resources in non-DB cluster" do
-    code <<-EOF
-      if /etc/init.d/openais status ; then
-        cibadmin -E -f
-      fi
-    EOF
-    only_if { ::File.exist?("/usr/sbin/cibadmin") }
-  end
-
-  # Stop DRBD and corosync.
-  # (Note that this node is not running database)
-  service "drbd" do
-    action :stop
-  end
-  service "openais" do
-    action :stop
-  end
-  service "openais-shutdown" do
-    action :stop
-  end
-
 when "dump_openstack_database"
 
   include_recipe "crowbar::stop-services-before-upgrade"
 
 when "db_shutdown"
 
-  bash "stop remaining pacemaker resources" do
-    code <<-EOF
-      for type in clone ms primitive; do
-        for resource in $(crm configure show | grep ^$type | cut -d " " -f2);
-        do
-          crm --force --wait resource stop $resource
-        done
-      done
-    EOF
-    only_if { ::File.exist?("/usr/sbin/crm") }
-  end
+  # nothing to see here, move along
 
-  bash "delete pacemaker resources in DB cluster" do
-    code <<-EOF
-      if /etc/init.d/openais status ; then
-        cibadmin -E -f
-      fi
-    EOF
-    only_if { ::File.exist?("/usr/sbin/cibadmin") }
-  end
-
-  # Stop the database and corosync
-  service "drbd" do
-    action :stop
-  end
-  service "openais" do
-    action :stop
-  end
-  service "postgresql" do
-    action :stop
-  end
-  service "openais-shutdown" do
-    action :stop
-  end
 when "done_openstack_shutdown", "wait_for_openstack_shutdown"
   Chef::Log.debug("Nothing to do on this node, waiting for others to finish their work...")
 else
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -54,7 +54,6 @@ when "crowbar_upgrade"
         systemctl disable $i
       done
     EOF
-    not_if { ::File.exist?("/usr/sbin/crm") }
   end
 
   # Disable crowbar-join
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -47,7 +47,7 @@ when "crowbar_upgrade"
 
   bash "disable_openstack_services" do
     code <<-EOF
-      for i in $(systemctl list-units openstack* --no-legend | cut -d" " -f1 | grep -v keystone | grep -v neutron) \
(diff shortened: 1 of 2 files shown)


chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -47,13 +47,14 @@
 
   bash "disable_openstack_services" do
     code <<-EOF
-      for i in $(systemctl list-units openstack* --no-legend | cut -d" " -f1) \
+      for i in $(systemctl list-units openstack* --no-legend | cut -d" " -f1 | grep -v keystone | grep -v neutron) \
```
aplanas: Below you use `grep -Ev A|N`, can be used here?

abel-navarro: sure, there you go.

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -47,13 +47,14 @@
...
                drbd.service \
                pacemaker.service;
       do
         systemctl disable $i
       done
     EOF
+    not_if { ::File.exist?("/usr/sbin/crm") }
```
jsuchome: Are you sure we don't want to disable services if we have HA clusters? I would assume we should disable it and later start/enable it explicitly in some orderly fashion...

abel-navarro: I'm actually not sure, but I think pacemaker should not be happy by something external disabling services.

I could revert this later on, but would like to try it this way.

jsuchome: Well, pacemaker would be disabled as well (line 52)

abel-navarro: right, i'll leave it the old way then.

jsuchome: I did not say it was wrong, I'm just trying to understand why would it be needed. But I guess you can bring it back again if need.

aspiers: > pacemaker should not be happy by something external disabling services.

Huh?  The opposite is true.  Services controlled by pacemaker must not be enabled, because they must not be started on boot by systemd, but instead get started by Pacemaker.  And indeed [we already make Chef do this](https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/providers/service.rb#L41).  So in the HA context, this code should be a no-op, because the services will already be disabled.  Maybe you are confusing disable with stop?

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -47,7 +47,7 @@
 
   bash "disable_openstack_services" do
     code <<-EOF
-      for i in $(systemctl list-units openstack* --no-legend | cut -d" " -f1) \
+      for i in $(systemctl list-units openstack* --no-legend | cut -d" " -f1 | grep -Ev 'keystone|neutron') \
```
aspiers: As mentioned elsewhere, in the HA case this should be a no-op because [they're already disabled](https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/providers/service.rb#L41). However it's worth doing this anyway just in case.

But why aren't you disabling `keystone` and `neutron`?

chef/cookbooks/crowbar/recipes/stop-services-before-upgrade.rb
```diff
@@ -63,6 +63,7 @@
       fi
     done
   EOF
+  not_if { ::File.exist?("/usr/sbin/crm") }
```
aspiers: I don't like this check.  You can't safely assume that the services are controlled by Pacemaker just because the `crmsh` package is installed.  It might be installed on non-HA environments.

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
abel-navarro: we are going to keep `keystone` and `neutron` up. It was to keep in line semantically __service disabled__ with __service_stopped__. I guess it may be a problem after the node reboots and tries to bring up these services while we upgrade and `chef-client` is down. So I'm removing those exceptions.

chef/cookbooks/crowbar/recipes/stop-services-before-upgrade.rb
abel-navarro: These are the checks we used before to differentiate HA and non-HA nodes. Do you have a better suggestion?

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
aspiers: Yes exactly, it doesn't make sense to keep them in line for this reason.  As with the others, they should be disabled anyway, but it's good to make sure.

chef/cookbooks/crowbar/recipes/stop-services-before-upgrade.rb
```diff
@@ -33,26 +34,11 @@
...
-  EOF
-  only_if { ::File.exist?("/usr/sbin/crm") }
-end
-
 # Stop openstack services on this node.
 # Note that for HA setup, they should be already stopped by pacemaker.
 bash "stop OpenStack services" do
   code <<-EOF
-    for i in /etc/init.d/openstack-* \
+    for i in `ls /etc/init.d/openstack-* | grep -Ev 'keystone|neutron'` \
```
jsuchome: Why are we excluding keystone/neutron for non-HA case?

abel-navarro: The idea was to have the L3 agents live as long as possible in the non-HA case. But now that you mention it, during my investigations you could stop all services and the agents were still alive. I'll check that again and see if we can remove this here.

abel-navarro: i verified (again) that shutting down all openstack services, including the l3 agent, does keep the router namespace up. The VM does not loose connectivity while doing so. I'll remove those exceptions.

jsuchome:
+1 (although, likely to be replaced by something proposed in https://github.com/crowbar/crowbar-core/pull/655)

abel-navarro:
@jsuchome i think we can close this PR. #655 is overlapping and more complete than this one.

jsuchome:
Ah, true. It was actually different PR which included stopping of the services.

PR #655, opened 2016-09-09
Orchestrate the upgrade of first controller node

jsuchome:
Execute the upgrade action via
 `curl -s -X POST -H "Accept: application/vnd.crowbar.v2.0+json" localhost:3000/api/upgrade/nodes`

(after the initialization done in `services` step)

### DIFF
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -88,3 +88,30 @@ template "/usr/sbin/crowbar-shutdown-services-before-upgrade.sh" do
     use_ha: use_ha
   )
 end
+
+# Following script executes all actions that are needed directly on the node
+# directly before the OS upgrade is initiated.
+
+# Find out v2 auth URL that is needed for neutron-ha-tool
+os_auth_url_v2 = ""
+if use_ha && node["run_list_map"].key?("neutron-network")
+  keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+  os_auth_url_v2 = KeystoneHelper.versioned_service_URL(
+    keystone_settings["protocol"],
+    keystone_settings["internal_url_host"],
+    keystone_settings["service_port"],
+    "2.0"
+  )
+end
+
+template "/usr/sbin/crowbar-pre-upgrade.sh" do
+  source "crowbar-pre-upgrade.sh.erb"
+  mode "0770"
+  owner "root"
+  group "root"
+  action :create
+  variables(
+    use_ha: use_ha,
+    os_auth_url_v2: os_auth_url_v2
+  )
+end
--- a/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
@@ -1,0 +1,28 @@
+#!/bin/bash
+#
+# This does what is necessary directly before the start of OS upgrade
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+mkdir -p "`dirname "$LOGFILE"`"
+exec >>"$LOGFILE" 2>&1
+
+set -x
+
+<% if @use_ha %>
+
+echo "Evacuating l3 agents out of this node..."
+
+if [[ -f /usr/bin/neutron-ha-tool ]] ; then
+    source /root/.openrc
+    OS_AUTH_URL="<%= @os_auth_url_v2 %>" neutron-ha-tool --l3-agent-evacuate $(hostname)
+fi
+
+# Shutdown pacemaker so the remaining OpenStack services are stopped
+# This is needed so that the zypper dup won't trigger any db migrations on package update
+service pacemaker stop
+
+<% else %>
+
+echo "No HA setup found..."
+
+<% end %>
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -85,6 +85,19 @@ class Api::UpgradeController < ApiController
     end
   end
 
+  api :POST, "/api/upgrade/nodes", "Initiate the upgrade of all nodes"
+  api_version "2.0"
+  # This is gonna initiate the upgrade of all nodes.
+  # The method runs asynchronously, so there's a need to poll for the status and possible errors
+  def nodes
+    # FIXME: implement this as asynchronous call!
+    if Api::Upgrade.nodes
+      head :ok
+    else
+      render json: { error: "Node Upgrade failed" }, status: :unprocessable_entity
+    end
+  end
+
   api :GET, "/api/upgrade/prechecks", "Shows a sanity check in preparation for the upgrade"
   api_version "2.0"
   example '
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -16,6 +16,98 @@
 
 module Api
   class Node < Tableless
+    def initialize(name = nil)
+      @node = NodeObject.find_node_by_name name
+    end
+
+    def upgraded?
+      # FIXME: check this by looking at some upgraded-ok file on the node
+      current_platform = "#{@node[:platform]}-#{@node[:platform_version]}"
+      @upgraded ||= current_platform == @node[:target_platform]
+    end
+
+    def pre_upgrade
+      # TODO: save the global status info about this substep (we started prepare for upgrade)
+
+      # Migrate out the l3 agents and shut down pacemaker
+      script = "/usr/sbin/crowbar-pre-upgrade.sh"
+      out = @node.run_ssh_cmd(script)
+      unless out[:exit_code].zero?
+        Rails.logger.error("Executing of pre upgrade script has failed on node #{@node.name}.")
+        Rails.logger.error("Script location: #{script}")
+        Rails.logger.error("stdout: #{out[:stdout]}, stderr: #{out[:stderr]}")
+        return false
+      end
+      true
+    end
+
+    def os_upgrade
+      # Upgrade one node
+      # TODO: save the global status info about this substep (we started upgrade of the node)
+      ssh_status = @node.ssh_cmd("/usr/sbin/crowbar-upgrade-os.sh")
+      if ssh_status[0] != 200
+        Rails.logger.error("Executing of os upgrade script has failed on node #{@node.name}.")
+        return false
+      end
+      true
+    end
+
+    def post_upgrade
+      # FIXME: so far, we have no post-upgrade script
+      return true
+      # TODO: save the global status info about this substep (we started post upgrade stuff)
+
+      # Join the cluster: start pacemaker and run selected recipes
+      script = "/usr/sbin/crowbar-post-upgrade.sh"
+      out = @node.run_ssh_cmd(script)
+      unless out[:exit_code].zero?
+        Rails.logger.error("Executing of post upgrade script has failed on node #{@node.name}.")
+        Rails.logger.error("Script location: #{script}")
+        Rails.logger.error("stdout: #{out[:stdout]}, stderr: #{out[:stderr]}")
+        return false
+      end
+    end
+
+    # Do the complete upgrade of one node
+    def upgrade
+      # FIXME: check the global status:
+      # if we failed in some previous attempt (pre/os/post), continue from the failed substep
+
+      # this is just a fallback check, we should know by checking the global status that the action
+      # should not be executed on already upgraded node
+      return true if upgraded?
+
+      unless pre_upgrade
+        save_error_state("Error while executing pre upgrade script")
+        return false
+      end
+
+      unless os_upgrade
+        save_error_state("Error while executing upgrade script")
+        return false
+      end
+
+      # FIXME: wait until the upgrade is done
+
+      # FIXME: reboot the node
+
+      unless post_upgrade
+        save_error_state("Error while executing post upgrade script")
+        return false
+      end
+      true
+    end
+
+    def save_node_state(message = "")
+      # FIXME: save the node status to global status
+      Rails.logger.info(message)
+    end
+
+    def save_error_state(message = "")
+      # FIXME: save the error to global status
+      Rails.logger.error(message)
+    end
+
     class << self
       def repocheck(options = {})
         addon = options.fetch(:addon, "os")
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -113,6 +113,15 @@ module Api
 
       return false if upgrade_failure
 
+      # reboot the node after successful upgrade
+      ssh_status = @node.ssh_cmd("/sbin/reboot")
+      if ssh_status[0] != 200
+        save_error_state("Failed to reboot the machine. Could not ssh.")
+        return false
+      end
+
+      # FIXME: wait until node comes back
+
       unless post_upgrade
         save_error_state("Error while executing post upgrade script")
         return false
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -87,9 +87,31 @@ module Api
         return false
       end
 
-      # FIXME: wait until the upgrade is done
+      # wait until the OS upgrade is finished
+      upgrade_failure = false
+      begin
+        Timeout.timeout(600) do
+          loop do
+            out = @node.run_ssh_cmd("test -e /var/lib/crowbar/upgrade/node-upgraded-ok")
+            if out[:exit_code].zero?
+              save_node_state("Package upgrade was successful.")
+              break
+            end
+            out = @node.run_ssh_cmd("test -e /var/lib/crowbar/upgrade/node-upgrade-failed")
+            if out[:exit_code].zero?
+              upgrade_failure = true
+              save_error_state("Installation of node #{@node.name} failed")
+              break
+            end
+            sleep(5)
+          end
+        end
+      rescue Timeout::Error
+        save_error_state("Error during upgrading node. Action did not finish after 10 minutes")
+        return false
+      end
 
-      # FIXME: reboot the node
+      return false if upgrade_failure
 
       unless post_upgrade
         save_error_state("Error while executing post upgrade script")
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -107,7 +107,65 @@ module Api
         }
       end
 
+      # Orchestrate the upgrade of the nodes
+      def nodes
+        # check for current global status
+        # 1. TODO: return if upgrade has finished
+        # 2. TODO: find the next big step
+        next_step = "controllers"
+
+        if next_step == "controllers"
+
+          # TODO: Save the "current_step" to global status
+          if upgrade_controller_nodes
+            # upgrading controller nodes succeeded, we can continue with computes
+            next_step = "computes"
+          else
+            # upgrading controller nodes has failed, exiting
+            # leaving next_step as "controllers", so we continue from correct point on retry
+            return false
+          end
+        end
+
+        if next_step == "computes"
+          # TODO: Save the "current_step" to global status
+          upgrade_compute_nodes
+        end
+        true
+      end
+
       protected
+
+      def upgrade_controller_nodes
+        # TODO: find the controller node that needs to be upgraded now
+        # First node to upgrade is DRBD slave
+        drbd_slave = ""
+        NodeObject.find(
+          "state:crowbar_upgrade AND (roles:database-server OR roles:rabbitmq-server)"
+        ).each do |db_node|
+          cmd = "LANG=C crm resource status ms-drbd-{postgresql,rabbitmq}\
+          | grep \\$(hostname) | grep -q Master"
+          out = db_node.run_ssh_cmd(cmd)
+          unless out[:exit_code].zero?
+            drbd_slave = db_node.name
+          end
+        end
+        # FIXME: prepare for cases with no drbd out there
+        return false if drbd_slave.empty?
+
+        node_api = Api::Node.new drbd_slave
+
+        # FIXME: save the global status information that this node is being upgraded
+        node_api.upgrade
+
+        # FIXME: if upgrade went well, continue with next node(s)
+        true
+      end
+
+      def upgrade_compute_nodes
+        # TODO: not implemented
+        true
+      end
 
       def crowbar_upgrade_status
         Api::Crowbar.upgrade
--- a/crowbar_framework/config/routes.rb
+++ b/crowbar_framework/config/routes.rb
@@ -274,6 +274,7 @@ Rails.application.routes.draw do
       only: [:show, :update] do
       post :prepare
       post :services
+      post :nodes
       get :prechecks
       post :cancel
       get :repocheck
--- a/crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
+++ b/crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
@@ -57,6 +57,22 @@ describe Api::UpgradeController, type: :request do
       expect(response).to have_http_status(:unprocessable_entity)
     end
 
+    it "initiates the upgrade of nodes" do
+      allow(NodeObject).to(
+        receive(:find).and_return([NodeObject.find_node_by_name("testing")])
+      )
+      post "/api/upgrade/nodes", {}, headers
+      expect(response).to have_http_status(:ok)
+    end
+
+    it "initiates the upgrade of nodes and fails" do
+      allow(NodeObject).to(
+        receive(:find).and_return([])
+      )
+      post "/api/upgrade/nodes", {}, headers
+      expect(response).to have_http_status(:unprocessable_entity)
+    end
+
     it "shows a sanity check in preparation for the upgrade" do
       allow(Crowbar::Sanity).to receive(:sane?).and_return(true)
 
--- a/crowbar_framework/spec/models/api/node_spec.rb
+++ b/crowbar_framework/spec/models/api/node_spec.rb
@@ -99,7 +99,7 @@ describe Api::Node do
         k.delete("openstack")
       end
 
-      expect(subject.class.repocheck).to eq(os_repo_fixture)
+      expect(Api::Node.repocheck).to eq(os_repo_fixture)
     end
 
     it "finds the os and addon repositories required to upgrade the nodes" do
@@ -110,7 +110,7 @@ describe Api::Node do
           ).and_return([true, {}])
         )
 
-        expect(subject.class.repocheck(addon: feature)).to eq(
+        expect(Api::Node.repocheck(addon: feature)).to eq(
           feature.to_s => node_repocheck[feature]
         )
       end
@@ -127,7 +127,7 @@ describe Api::Node do
         send("#{feature}_repo_missing".to_sym)
 
         expected[feature] = node_repocheck_missing[feature]
-        got.merge!(subject.class.repocheck(addon: feature))
+        got.merge!(Api::Node.repocheck(addon: feature))
       end
 
       expect(got).to eq(expected)
@@ -148,7 +148,7 @@ describe Api::Node do
         k.delete("openstack")
       end
 
-      expect(subject.class.repocheck(addon: "ceph")).to eq(expected)
+      expect(Api::Node.repocheck(addon: "ceph")).to eq(expected)
     end
 
     it "doesn't find any node with the ceph addon deployed" do
@@ -166,7 +166,7 @@ describe Api::Node do
         k["ceph"]["available"] = false
       end
 
-      expect(subject.class.repocheck(addon: "ceph")).to eq(expected)
+      expect(Api::Node.repocheck(addon: "ceph")).to eq(expected)
     end
   end
 end
--- a/crowbar_framework/spec/models/api/upgrade_spec.rb
+++ b/crowbar_framework/spec/models/api/upgrade_spec.rb
@@ -171,6 +171,23 @@ describe Api::Upgrade do
     end
   end
 
+  context "upgrading the nodes" do
+    it "successfully upgrades the nodes" do
+      allow(NodeObject).to(
+        receive(:find).
+        with("state:crowbar_upgrade AND (roles:database-server OR roles:rabbitmq-server)").
+        and_return([NodeObject.find_node_by_name("testing.crowbar.com")])
+      )
+      expect(subject.class.nodes).to be true
+    end
+    it "fails during the upgrade of controller nodes" do
+      allow(subject.class).to(
+        receive(:upgrade_controller_nodes).and_return(false)
+      )
+      expect(subject.class.nodes).to be false
+    end
+  end
+
   context "canceling the upgrade" do
     it "successfully cancels the upgrade" do
       allow_any_instance_of(CrowbarService).to receive(


crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -59,15 +59,38 @@ def services
...
-      # now, initiate the services shutdown by calling scripts on all nodes
-      # FIXME: this needs to go elsewhere
-      # FIXME: this needs to be orchestrated one by one
+      # FIXME: This is primitive orchestration code that should be written elsewhere
+      # (library/model/external tool)
+
+      # Second, initiate the services shutdown in all clusters
+      # These are the non essential services that could be shut down everywhere
+      # FIXME: non-HA case not being handled
       NodeObject.find("pacemaker_founder:true").each do |node|
```
abel-navarro: probably we don't need this anymore since the loop you introduced below goes through all cluster members, including the founder one. I don't think we do anything special for the founder one.

Maybe later on we want to choose a specific node to start the shutdown and upgrade sequence, i.e. starting with the DRBD slave. But that can be arranged later.

jsuchome: Ah, that was a leftover code!

jsuchome:
@MaximilianMeister @rsalevsky could you please comment on proposed API change in this PR?

jsuchome:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;hdEzsetg&#x2F;102-5-orchestrate-upgrade-of-first-controller-node-in-ha-setup-upgrade-step-12-part-1"> (5) Orchestrate upgrade of first controller node in HA setup (upgrade step 12 part 1)</a>

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -57,7 +57,56 @@ def prepare
   api_version "2.0"
   def services
     if request.post?
-      head :not_implemented
+
```
MaximilianMeister: I would drop this empty line

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -57,7 +57,56 @@ def prepare
...
+      # prepare the scripts for various actions in one chef run
+      service_object = CrowbarService.new(Rails.logger)
+      service_object.prepare_nodes_for_os_upgrade
```
MaximilianMeister: can this fail somehow, and do we need to report/handle a failure here for the caller of the api?

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -57,7 +57,56 @@ def prepare
...
+      # These are the non essential services that could be shut down everywhere
+      # FIXME: non-HA case not being handled
+      NodeObject.find("pacemaker_founder:true").each do |node|
+        node.ssh_cmd("/usr/sbin/crowbar-shutdown-services-before-upgrade.sh")
+      end
+      # FIXME: dump database on database node - currently missing
+
+      head :ok
+    else
+      render json: [], status: :not_implemented
```
MaximilianMeister: can we list the services here that will be stopped? if we can't then we may drop this endpoint

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -57,7 +57,56 @@ def prepare
...
+        node_api.pre_upgrade node
+        node_api.upgrade node
+        node_api.post_upgrade node
+      end
+
+      # Go node by node and upgrade it
+      # FIXME: are we sure this search excludes compute nodes?
+      # FIXME: What about possible unclustered nodes (storage)?
+      NodeObject.find("run_list_map:pacemaker-cluster-member").each do |node|
+
```
MaximilianMeister: i would drop this empty line as well

crowbar_framework/app/controllers/api/upgrade_controller.rb
jsuchome: Hm. Maybe we could, but I actually don't see a point in showing that.

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -57,7 +57,56 @@ def prepare
...
+        node_api = Api::Node.new
+        next if node_api.already_upgraded node
+        node_api.pre_upgrade node
+        node_api.upgrade node
+        node_api.post_upgrade node
+      end
+
+      head :ok
     else
       render json: [], status: :not_implemented
```
MaximilianMeister: the else case has to be dropped i think (as there is no GET route)

crowbar_framework/app/models/api/node.rb
```diff
@@ -50,6 +50,32 @@ def repocheck(options = {})
       end
     end
 
+    def already_upgraded(node)
```
MaximilianMeister: as it returns a bool, just name it `already_upgraded?(node)`

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -57,7 +57,56 @@ def prepare
...
+      # (library/model/external tool)
+
+      # FIXME: we need to go cluster by cluster...?
+      # start with drbd slave (this actually does all drbd slaves in all clusters)
+      NodeObject.find("drbd_rsc_postgresql_master:true").each do |node|
+        node_api = Api::Node.new
+        next if node_api.already_upgraded node
+        node_api.pre_upgrade node
+        node_api.upgrade node
+        node_api.post_upgrade node
```
MaximilianMeister: can we check here if the commands returned with 0 ?

crowbar_framework/config/routes.rb
```diff
@@ -218,6 +218,8 @@
         post :repos
         get :services
         post :services
+        get :controllers
+        post :controllers
```
MaximilianMeister: why do you add the routes to the old upgrades controller?

crowbar_framework/config/routes.rb
```diff
@@ -275,6 +277,8 @@
       post :prepare
       post :services
       get :services
+      post :controllers
+      get :controllers
```
MaximilianMeister: what is the GET route supposed to return? or in other words do we really need it?

crowbar_framework/app/controllers/api/upgrade_controller.rb
jsuchome: Well, there's no execution of code on nodes in this one, but usual applying of crowbar proposal. So may think about similar errors like from the `prepare` action.

MaximilianMeister:
@jsuchome I have seen before you used an initializer and passed a node(name) to it. I think that made sense here. any specific reason why you dropped it again? I think it was a good idea.

we would then just need to put all the non instance methods (like repocheck) into a class method context.

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: maybe in the UI we could say that this and that services will be stopped?

crowbar_framework/app/controllers/api/upgrade_controller.rb
jsuchome: Yes, some of them. But for example `upgrade` command runs in the background.

jsuchome:
@MaximilianMeister yes, I also liked that initialization, but changed it after rebase with `repocheck` I felt it does not fit any more

crowbar_framework/app/controllers/api/upgrade_controller.rb
jsuchome: I'm not sure if it makes sense to show result of https://github.com/crowbar/crowbar-core/pull/653/files#diff-f76725e43d676e1fb0ea7adda50cfc00R28

But maybe yes... have you count with this information for the UI?

MaximilianMeister:
> yes, I also liked that initialization, but changed it after rebase with repocheck I felt it does not fit any more

do you mind to re-add it? there are some places (`s/Api::Node.new.method/Api::Node.method/g`) of course that you would need to adapt, but not many. besides creating a class method context in the model

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -64,6 +64,24 @@ def services
     end
   end
 
+  api :POST, "/api/upgrade/controllers", "Orchestrate the upgrade of controller nodes"
```
jsuchome: As alternative to this API, we could have "/api/upgrade/controller".

Such call would upgrade **next non-upgraded controller node**, and return some special value when there's no one left so the caller could use different API call to continue. This way, each call of  "/api/upgrade/controller" could be reasonably reflected in the UI. 

@MaximilianMeister @rsalevsky what do you think?

jsuchome:
> do you mind to re-add it? 

I'm not sure now which methods should be class ones and which instance methods. Also, I'm but lost of how do I need to adapt the testsuite...

jsuchome:
@MaximilianMeister see last commit about those method changes and tests adaptations

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -64,6 +64,23 @@ def services
...
+  api_version "2.0"
+  def controllers
+    # First node to upgade is DRBD slave
+    # FIXME: this actually does all drbd slaves in all clusters, do we want that?
+    NodeObject.find("drbd_rsc_postgresql_master:true").each do |node|
+      node_api = Api::Node.new node.name
+      next if node_api.upgraded?
+      node_api.pre_upgrade
```
abel-navarro: other API calls do report the status on the execution. These functions (`pre_upgrade`, `upgrade` and `post_upgrade`) would benefit of having a return code. A boolean for success/failure may be enough.

crowbar_framework/app/models/api/node.rb
```diff
@@ -50,37 +50,65 @@ def repocheck(options = {})
...
-    def node_architectures
-      {}.tap do |ret|
-        NodeObject.all.each do |node|
-          arch = node.architecture
-          ret["os"] ||= []
-          ret["os"].push(arch) unless ret["os"].include?(arch)
+    def pre_upgrade(node)
+      # Migrate out the l3 agents and shut down pacemaker
+      # FIXME: should we make thise code part of crowbar-upgrade-os.sh?
+      node.ssh_cmd("/usr/sbin/crowbar-pre-upgrade.sh")
```
abel-navarro: It would be nice to check if the file to execute actually existed (just in case) and return a boolean on command success.

jsuchome: Yep, good idea.

jsuchome: ... or not. The file is on the remote node, so if it is not there, ssh_cmd should fail itself.

abel-navarro: good point. Yes, I guess return code from `node.ssh_cmd` is enough.

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: but then the user can't really choose which controller to upgrade. or do you want to provide some parameter which controller to upgrade? if not, then that would only make sense to implement internally (so not exposed through the api). would be also a question for the UI/UX team?

jsuchome: Well, user can't choose which controller to upgrade anyway.

This change I'm proposing should IMHO make it better for the UI team, because they want to show "upgrading node XY" while this is being done. So if the API would be created the way to only upgrade one controller at the time, UI would know better how to show the progress.

Other option (which is currently proposed in this PR) is to have API call for all controllers.

rsalevsky: Why can't we choose which controller should be upgraded next? Both makes actually no real sense to me... I would expect if I call `/controllers` that all controllers get upgraded automatically or if I call `/controller` that I can choose the controller. Maybe it's not possible to upgrade all controllers but we can provide a selection via a get route. This would make it a little bit more dynamic. Also if we don't expose it via the UI a admin can still do it manual which some customers want to do.

jsuchome: No, it's not possible for user to choose controller to upgrade. The whole thing about upgrading controller nodes is pretty complicated to orchestrate, as you can see at https://etherpad.nue.suse.com/p/cloud-upgrade-6-to-7 , step 12 (starting at line 154). 
Very shortly:
-  the controllers need to be upgraded one-by-one, not in parallel (that's why the /controller action would take quite a lot time to complete if we'd want from it upgrade of everything)
- the order in which the nodes are going to be upgraded depends on their configuration, on the services which are running there. We have to determine this programmaticaly, not letting user to choose.

crowbar_framework/app/models/api/node.rb
```diff
@@ -50,6 +50,32 @@ def repocheck(options = {})
...
+      current_platform = "#{node[:platform]}-#{node[:platform_version]}"
+      return false if current_platform != node[:target_platform]
+
+      Rails.logger.info("node #{node.name} is already upgraded")
+      true
+    end
+
+    def pre_upgrade(node)
+      # Migrate out the l3 agents and shut down pacemaker
+      # FIXME: should we make thise code part of crowbar-upgrade-os.sh?
```
aspiers: I think slightly cleaner to keep it separate, and this might also help later when we make it async and implement an API call for querying the status - then the UI could distinguish between these two steps and show a progress bar or something.

crowbar_framework/app/models/api/node.rb
```diff
@@ -50,6 +50,32 @@ def repocheck(options = {})
...
+
+    def upgrade(node)
+      # Upgrade one node
+      # TODO: signal to the UI that we started upgrade of the node
+      node.ssh_cmd("/usr/sbin/crowbar-upgrade-os.sh")
+    end
+
+    def post_upgrade(node)
+      # Join the cluster: start pacemaker and run selected recipes
+      # Migrate l3 agents to this node
```
aspiers: Migrating L3 agents back here is not necessary - that will happen anyway if necessary when the next node does its pre-upgrade step. When evacuating L3 agents from a node, we don't want to specify a target because we want to rely on the built-in intelligence of neutron to pick the best destination for the migration, in order to keep things nicely balanced.

crowbar_framework/config/routes.rb
```diff
@@ -274,6 +274,7 @@
       only: [:show, :update] do
```
aspiers: Looks like broken indentation here (although I know it's not related to this PR).

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -67,15 +67,14 @@ def services
...
-    # FIXME: we need to go cluster by cluster...?
     # First node to upgade is DRBD slave
     # FIXME: this actually does all drbd slaves in all clusters, do we want that?
     NodeObject.find("drbd_rsc_postgresql_master:true").each do |node|
-      node_api = Api::Node.new
-      next if node_api.already_upgraded? node
-      node_api.pre_upgrade node
-      node_api.upgrade node
-      node_api.post_upgrade node
+      node_api = Api::Node.new node.name
```
aspiers: This is correcting the previous commit in this PR, so it should be squashed to keep the history clean.

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -67,18 +67,37 @@ def services
...
     # First node to upgade is DRBD slave
-    # FIXME: this actually does all drbd slaves in all clusters, do we want that?
-    NodeObject.find("drbd_rsc_postgresql_master:true").each do |node|
-      node_api = Api::Node.new node.name
-      next if node_api.upgraded?
-      node_api.pre_upgrade
-      node_api.upgrade
-      node_api.post_upgrade
-      # FIXME: check return values if it makes sense
+    # FIXME: what about other drbd slaves?
```
aspiers: Well this only looks for the master, not slaves. But see my previous comment on this.

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -67,18 +67,37 @@ def services
...
+    node = NodeObject.find("drbd_rsc_postgresql_master:true").first
```
aspiers: Is this change really related to "Check for return values of ssh commands"? I don't think it belongs in this commit. Should be squashed into the commit it depends on, since it seems to be part of transient development.

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -0,0 +1,28 @@
...
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+mkdir -p "`dirname "$LOGFILE"`"
+
+# Copy stdout to fd 3
+exec 3>&1
+# Create fd 4 for logfile
+exec 4>> "$LOGFILE"
+# Set fd 1 and 2 to logfile
+exec 1>&4 2>&1
```
MaximilianMeister: > exec >>"$LOGFILE" 2>&1

I think this was just copy and pasted from install-chef-suse, where we still cared about stdout on the command line. and as i understand this script is just called from crowbar, and in no case by the user?

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -64,6 +64,24 @@ def services
     end
   end
 
+  api :POST, "/api/upgrade/controllers", "Orchestrate the upgrade of controller nodes"
+  api_version "2.0"
+  def controllers
+    # FIXME: we need to go cluster by cluster...?
+    # First node to upgade is DRBD slave
+    # FIXME: this actually does all drbd slaves in all clusters, do we want that?
```
jsuchome: The message is obsoleted, originally there was a loop over all drbd slaves

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -64,6 +64,24 @@ def services
...
+    NodeObject.find("drbd_rsc_postgresql_master:true").each do |node|
```
jsuchome: Yes, it's looking for master, that's a bug. 
Why is it unreliable to look into chef data? We're doing it all the time for various settings...

crowbar_framework/app/models/api/node.rb
```diff
@@ -50,6 +50,32 @@ def repocheck(options = {})
       end
     end
 
+    def already_upgraded?(node)
+      current_platform = "#{node[:platform]}-#{node[:platform_version]}"
+      return false if current_platform != node[:target_platform]
```
jsuchome: Yes, we could add a check like that.

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -64,6 +64,24 @@ def services
...
+      node_api = Api::Node.new
+      next if node_api.already_upgraded? node
+      node_api.pre_upgrade node
```
jsuchome: Not necessarily. I think the whole method for installing controllers should be asynchronous

crowbar_framework/app/models/api/node.rb
jsuchome: Yes, this is old code, before the latest rounds of discussion what needs to happen on the nodes...

crowbar_framework/app/controllers/api/upgrade_controller.rb
jsuchome: No. Oviously I did not pay attention while doing the fixup

crowbar_framework/app/models/api/node.rb
jsuchome: Obsoleted comment

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -64,6 +64,24 @@ def services
     end
   end
 
+  api :POST, "/api/upgrade/controllers", "Orchestrate the upgrade of controller nodes"
+  api_version "2.0"
+  def controllers
+    # FIXME: we need to go cluster by cluster...?
+    # First node to upgade is DRBD slave
```
jsuchome: obsolete

crowbar_framework/app/controllers/api/upgrade_controller.rb
jsuchome: obsolete

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -85,8 +85,54 @@ def services
...
+      end
+      true
+    end
+
     protected
 
+    def upgrade_controller_nodes
+      # TODO: find the controller node that needs to be upgraded now
+      # First node to upgade is DRBD slave
+      # FIXME: what about other drbd slaves?
```
aspiers: We can only have 1 slave.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -85,8 +85,54 @@ def services
...
+        upgrade_compute_nodes
+      end
+      true
+    end
+
     protected
 
+    def upgrade_controller_nodes
+      # TODO: find the controller node that needs to be upgraded now
+      # First node to upgade is DRBD slave
```
aspiers: "upgrade"

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
aspiers: Yeah - it's wrong to copy and paste that stuff from `install-chef-suse.sh` to here, because that has a totally different approach to output (a user-friendly ASCII-graphics display using `dialog(1)`, spinners etc., plus more verbose logging to files), whereas here there is no pseudo-graphical UI.

crowbar_framework/app/models/api/node.rb
```diff
@@ -50,37 +50,122 @@ def repocheck(options = {})
...
+      # FIXME: check the global status:
+      # if we failed in some previous attempt (pre/os/post), continue from the failed substep
+
+      # this is just a fallback check, we should know by checking the global status that the action
+      # should not be executed on already upgraded node
+      return true if @upgraded
+
+      unless pre_upgrade
+        Rails.logger.error("Error while executing pre upgrade script")
+        # FIXME: save the error to global status and exit
```
aspiers: Rather than adding a FIXME for each time where we need to persist the error and exit, why not just call another `upgrade_error` method which is empty with a single FIXME in it?  Then later on there is only one place which needs fixing.

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
jsuchome: done

crowbar_framework/app/models/api/node.rb
jsuchome: done

jsuchome:
> The I/O redirection needs to be fixed

Done

crowbar_framework/app/models/api/node.rb
```diff
@@ -50,37 +50,126 @@ def repocheck(options = {})
...
+      true
+    end
+
+    def post_upgrade
+      # FIXME: so far, we have no post-upgrade script
+      return true
+      # TODO: save the global status info about this substep (we started post upgrade stuff)
+
+      # Join the cluster: start pacemaker and run selected recipes
+      script = "/usr/sbin/crowbar-post-upgrade.sh"
```
jsuchome: Yes, that's fine, we do not have post_ugrade method yet (that's part of orchestration of more nodes)

crowbar_framework/app/controllers/api/upgrade_controller.rb
jsuchome: For the record, this discussion was resolved elsewhere. We want to use only one endpoint (`/nodes`), run the method asynchronously, and use some global status information to track current state.

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -62,6 +62,16 @@ def services
...
   end
 
+  api :POST, "/api/upgrade/nodes", "Initiate the upgrade of all nodes"
+  api_version "2.0"
+  # This is gonna initiate the upgrade of all nodes.
+  # The method runs asynchronously, so there's a need to poll for the status and possible errors
+  def nodes
+    # FIXME: implement this as asynchronous call!
+    Api::Upgrade.nodes
+    head :ok
```
MaximilianMeister: you can drop that, it is implicit

crowbar_framework/app/models/api/node.rb
```diff
@@ -16,6 +16,129 @@
...
 module Api
   class Node < Tableless
+    def initialize(name = nil)
+      @node = NodeObject.find_node_by_name name
+    end
+
+    def upgraded?
+      # FIXME: check this by looking at some upgraded-ok file on the node
+      current_platform = "#{@node[:platform]}-#{@node[:platform_version]}"
+      @upgraded = current_platform == @node[:target_platform]
```
MaximilianMeister: for what do we need the class variable `@upgrade` ? is it used somewhere later?

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -62,6 +62,16 @@ def services
     end
   end
 
+  api :POST, "/api/upgrade/nodes", "Initiate the upgrade of all nodes"
+  api_version "2.0"
+  # This is gonna initiate the upgrade of all nodes.
+  # The method runs asynchronously, so there's a need to poll for the status and possible errors
+  def nodes
+    # FIXME: implement this as asynchronous call!
```
MaximilianMeister: as it will be asynchronous, i think its ok not to catch any errors here

crowbar_framework/app/models/api/node.rb
MaximilianMeister: ah i just saw the usage, maybe memoize that

```ruby
def upgraded?
  @upgraded ||= current_platform == @node[:target_platform]
end
```

crowbar_framework/app/controllers/api/upgrade_controller.rb
jsuchome: That FIXME is for asynchronous implementation, not for missing error check

crowbar_framework/app/controllers/api/upgrade_controller.rb
jsuchome: addressed

crowbar_framework/app/models/api/node.rb
jsuchome: done

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: then you have to add an error check here

crowbar_framework/config/routes.rb
MaximilianMeister: actually not, as it is aligned correctly with 272

crowbar_framework/app/controllers/api/upgrade_controller.rb
jsuchome: I do not understand, you said before _its ok not to catch any errors here_ 
I do not have any there. But I'm missing asynchronous implementation, but that would be part of some future PR.

MaximilianMeister: ok, so there is no error that can happen in `Api::Upgrade.nodes` ? I mean do we need to check for its return value? `if true | false`

crowbar_framework/app/models/api/node.rb
```diff
@@ -16,6 +16,98 @@
...
+      current_platform = "#{@node[:platform]}-#{@node[:platform_version]}"
+      @upgraded ||= current_platform == @node[:target_platform]
+    end
+
+    def pre_upgrade
+      # TODO: save the global status info about this substep (we started prepare for upgrade)
+
+      # Migrate out the l3 agents and shut down pacemaker
+      script = "/usr/sbin/crowbar-pre-upgrade.sh"
+      out = @node.run_ssh_cmd(script)
```
aspiers: I think it would be worth running it inside `screen(1)`, so that if the ssh pipe fails unexpectedly, at least the script should still stand a good chance of finishing cleanly, which would make it more likely that resuming the upgrade later will succeed. I presume that wrapping inside `screen` should still preserve the exit code status.

crowbar_framework/app/models/api/node.rb
```diff
@@ -16,6 +16,98 @@
...
+        Rails.logger.error("stdout: #{out[:stdout]}, stderr: #{out[:stderr]}")
+        return false
+      end
+      true
+    end
+
+    def os_upgrade
+      # Upgrade one node
+      # TODO: save the global status info about this substep (we started upgrade of the node)
+      ssh_status = @node.ssh_cmd("/usr/sbin/crowbar-upgrade-os.sh")
```
aspiers: Same comment about `screen` here.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -107,8 +107,53 @@ def cancel
...
+        end
+        true
+      end
+
       protected
 
+      def upgrade_controller_nodes
+        # TODO: find the controller node that needs to be upgraded now
+        # First node to upgrade is DRBD slave
+        node = NodeObject.find("drbd_rsc_postgresql_master:false").first
```
aspiers: No, [as I mentioned previously](https://github.com/crowbar/crowbar-core/pull/655#discussion_r79941960), there is no guarantee that this Chef data really reflects the master/slave status.  [Chef actually tracks whether the node is the founder](https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/providers/drbd.rb#L36), not whether it's the master.  You should query Pacemaker to find which is the master.  An XPath query on `cibadmin -Ql` is probably the most correct, although parsing `crm resource status ms-drbd-{postgresql,rabbitmq}` is easier and probably good enough.

crowbar_framework/config/routes.rb
aspiers: OK, got confused by github UI.

crowbar_framework/app/models/api/node.rb
```diff
@@ -113,6 +113,15 @@ def upgrade
 
       return false if upgrade_failure
 
+      # reboot the node after successfull upgrade
```
aspiers: "successful"

crowbar_framework/app/controllers/api/upgrade_controller.rb
jsuchome: I do not think we should when the call is supposed to be asynchronous. We should check for error states by subsequent status calls.

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -88,3 +88,25 @@
     use_ha: use_ha
   )
 end
+
+# Following script executes all actions that are needed directly on the node
+# directly before the OS upgrade is initiated.
+
+# Find out v2 auth URL that is needed for neutron-ha-tool
+keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
```
jsuchome: As I'm using this only to get auth url for neutron-ha-tool, I assume I can limit the KeystoneHelper call to the nodes with neutron role(s) deployed. 
That way I'd prevent those errors I hit yesterday when retrieving keystone_settings failed. (That said, it should not fail anyway and the library should be fixed). @vuntz what do you think?

vuntz: Sounds good.

jsuchome: Done

crowbar_framework/app/models/api/upgrade.rb
jsuchome: I hope it makes sense to look only for e.g. `ms-drbd-postgresql`, not for both resources

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -107,8 +107,65 @@ def cancel
...
       protected
 
+      def upgrade_controller_nodes
+        # TODO: find the controller node that needs to be upgraded now
+        # First node to upgrade is DRBD slave
+        drbd_slave = ""
+        NodeObject.find(
+          "state:crowbar_upgrade AND (roles:database-server OR roles:rabbitmq-server)"
+        ).each do |db_node|
+          cmd = "crm resource status ms-drbd-{postgresql,rabbitmq}| grep \\$(hostname) | grep -q Master"
```
jsuchome: I'd rather keep it this way

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -88,3 +88,28 @@
...
+# directly before the OS upgrade is initiated.
+
+# Find out v2 auth URL that is needed for neutron-ha-tool
+os_auth_url_v2 = ""
+if use_ha && node["run_list_map"].key?("neutron-network")
+  keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+  os_auth_url_v2 = KeystoneHelper.versioned_service_URL(keystone_settings["protocol"],
+    keystone_settings["internal_url_host"],
+    keystone_settings["service_port"],
+    "2.0")
```
MaximilianMeister: not that it matters too much but I think:

```ruby
os_auth_url_v2 = KeystoneHelper.versioned_service_URL(
  keystone_settings["protocol"],
  keystone_settings["internal_url_host"],
  keystone_settings["service_port"],
  "2.0"
)
```

looks better readable

crowbar_framework/app/models/api/node.rb
```diff
@@ -16,6 +16,129 @@
...
+        end
+      rescue Timeout::Error
+        save_error_state("Error during upgrading node. Action did not finish after 10 minutes")
+        return false
+      end
+
+      return false if upgrade_failure
+
+      # reboot the node after successful upgrade
+      ssh_status = @node.ssh_cmd("/sbin/reboot")
```
MaximilianMeister: I guess we don't call `@node.reboot` because that sets the state to rebooting?

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -85,6 +85,15 @@ def services
...
+    render json: Api::Upgrade.nodes
```
MaximilianMeister: I wouldn't render anything here, as I just realized it is a POST request. I'd do more sth like

```ruby
if Api::Upgrade.nodes
  head :ok
else
  render json: { error: "Node Upgrade failed" }, status :unprocessable_entity
end
```

we shouldn't expect the request to return anything else than a http code in case of success, and an error + http code in case of failure

crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
```diff
@@ -57,6 +57,24 @@
       expect(response).to have_http_status(:unprocessable_entity)
     end
 
+    it "initiates the upgrade of nodes" do
+      allow(NodeObject).to(
+        receive(:find).and_return([NodeObject.find_node_by_name("testing")])
+      )
+      post "/api/upgrade/nodes", {}, headers
+      expect(response).to have_http_status(:ok)
+      expect(response.body).to eq("true")
```
MaximilianMeister: as mentioned in the comment above, i think we dont need to check for the body here

crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
```diff
@@ -57,6 +57,24 @@
...
+    end
+
+    it "initiates the upgrade of nodes and fails" do
+      allow(NodeObject).to(
+        receive(:find).and_return([])
+      )
+      post "/api/upgrade/nodes", {}, headers
+      expect(response).to have_http_status(:ok)
+      expect(response.body).to eq("false")
```
MaximilianMeister: and here the response body should ideally contain an error message, not just false

crowbar_framework/app/models/api/node.rb
jsuchome: Yes. At least until we figure out what state we want there

crowbar_framework/app/controllers/api/upgrade_controller.rb
jsuchome: Done

crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
jsuchome: Done

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
jsuchome: Done

jsuchome:
@aspiers I've adapted that drbd search

crowbar_framework/app/controllers/api/upgrade_controller.rb
aspiers: FTR I agree with that :-)

crowbar_framework/app/models/api/upgrade.rb
aspiers: And I'd rather you didn't :-)  There's really no good reason not to split the line.  Just before the `| grep` would be the nicest place to do it.

crowbar_framework/app/controllers/api/upgrade_controller.rb
aspiers: Right - when it's changed to asynchronous, the background job/thread/whatever can simply persist the error (to somewhere which any later status calls can read from) and then abort.  This could be done by calling a helper, or by raising an exception which gets caught and then the handler persists the error and aborts.

crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
```diff
@@ -57,6 +57,22 @@
       expect(response).to have_http_status(:unprocessable_entity)
     end
 
+    it "initiates the upgrade of nodes" do
```
aspiers: It's great that you're adding tests for this, but please can they be in the same commit which introduces the code they're testing?

crowbar_framework/app/models/api/node.rb
aspiers: @jsuchome You didn't reply to this point yet.  If you disagree or want to postpone this change to the cleanup card then I'm OK with that, but I don't want to +1 this review until all my feedback is addressed in some form.

jsuchome: I didn't really think about this. My first instinct was to say that we won't need to 'resume' the script, and we'd rather make them possible to simply run again. That should be more convenient for user.
However, there might be good reason for `screen`: I'd probably leave the discussion for later.

crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
jsuchome: ok, squashed into relevant commit

jsuchome:
@aspiers are you still requesting changes? I think I addressed your issues...

crowbar_framework/app/models/api/node.rb
aspiers: There *is* a good reason for `screen`, which is what I am trying to say :-)

Running the script again is a way of resuming it; they are not mutually exclusive things. I'm totally fine with this approach based on making the script idempotent, but that misses the point.  Even if each step is idempotent, it's not atomic.  For example, if the script gets killed due to a severed `ssh` connection whilst it's in the middle of database migrations, this might cause some consistency errors at the OpenStack level which could cause even an idempotent db_sync to get stuck on second time around.  I'm not sure if that's a good example, but the point is that if we can avoid non-atomic operations from being interrupted by a killed network connection then we should.

aspiers:
OK, but please add my point about `screen` as a TODO on a Trello card, somewhere.

PR #658, opened 2016-09-12
Set SES platform to suse-12.2

tserong:
This is suitable only for SES 4 (and has no impact on SOC deploys
of any version).


vuntz:
+1, but please merge only with https://github.com/crowbar/crowbar-core/pull/638 and https://github.com/crowbar/crowbar-ceph/pull/70

PR #659, opened 2016-09-13
use exception_notification to log all exceptions in the database

MaximilianMeister:
this would be using the Api::Error model to write any exception
into the database.
collecting all exceptions in a single place saves the time to crawl
through the logs.

### DIFF
--- a/crowbar_framework/Gemfile
+++ b/crowbar_framework/Gemfile
@@ -37,6 +37,7 @@ gem "yaml_db", "~> 0.3.0"
 gem "easy_diff", "~> 0.0.5"
 gem "apipie-rails", "~> 0.3.6"
 gem "pg", "~> 0.17.1"
+gem "exception_notification", "~> 4.2.1"
 
 gem "ohai", "~> 6.24.2"
 gem "chef", "~> 10.32.2"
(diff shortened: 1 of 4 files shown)


MaximilianMeister:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;kXfs4mJk&#x2F;109-log-all-exceptions-with-exception-notification-gem"> log all exceptions with exception_notification gem</a>

aspiers:
I'm not necessarily against this *in principle*, but please can you make the commit message say more about the exact use case(s) you expect this to be used for? There are potential downsides, e.g. bloating of the database, the need for extra cleanup of old exceptions, so it should be made clear why the upsides outweigh the downsides.

MaximilianMeister:
> can you make the commit message say more about the exact use case(s) you expect this to be used for?

I don't think that there is any other use case than looking up errors that happened during runtime. I didn't think about any cleanups yet, at least not automatically on a regular base.

This is also just one step, the more dynamic one which just writes all exceptions automatically. The other steps would be to write errors to that table when we fill an errors Hash in a model e.g., or just manually in the code at some critical points.

the original Idea about collecting errors in one place came from @kwwii IIRC, and it was really only about a one stop for all errors (which is still some way to go)

MaximilianMeister:
closing for now, as it's not really needed at this point in time

PR #660, opened 2016-09-13
perform the prepare in the background (bsc#997293)

MaximilianMeister:
this would (**EDIT**: _partially_) fix https://bugzilla.suse.com/show_bug.cgi?id=997293

(**EDIT**: it wouldn't, it would only fix one particular case, i.e. `/api/upgrade/prepare`)

and would obsolete https://github.com/SUSE-Cloud/automation/pull/1229

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -207,6 +207,16 @@ module Api
         true
       end
 
+      def prepare(options = {})
+        background = options.fetch(:background, false)
+
+        if background
+          prepare_nodes_for_crowbar_upgrade_background
+        else
+          prepare_nodes_for_crowbar_upgrade
+        end
+      end
+
       protected
 
       def upgrade_controller_nodes
@@ -376,6 +386,27 @@ module Api
 
       def admin_architecture
         NodeObject.admin_node.architecture
+      end
+
+      def prepare_nodes_for_crowbar_upgrade_background
+        @thread = Thread.new do
+          Rails.logger.debug("Started prepare in a background thread")
+          prepare_nodes_for_crowbar_upgrade
+        end
+
+        @thread.alive?
+      end
+
+      def prepare_nodes_for_crowbar_upgrade
+        service_object = CrowbarService.new(Rails.logger)
+        service_object.prepare_nodes_for_crowbar_upgrade
+
+        true
+      rescue => e
+        message = e.message
+        Rails.logger.error message
+
+        false
       end
     end
   end
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -208,6 +208,8 @@ module Api
       end
 
       def prepare(options = {})
+        ::Crowbar::UpgradeStatus.new.start_step
+
         background = options.fetch(:background, false)
 
         if background
@@ -401,9 +403,11 @@ module Api
         service_object = CrowbarService.new(Rails.logger)
         service_object.prepare_nodes_for_crowbar_upgrade
 
+        ::Crowbar::UpgradeStatus.new.end_step
         true
       rescue => e
         message = e.message
+        ::Crowbar::UpgradeStatus.new.end_step(false, { prepare_nodes_for_crowbar_upgrade: message })
         Rails.logger.error message
 
         false
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -83,21 +83,8 @@ class Api::UpgradeController < ApiController
   api_version "2.0"
   error 422, "Failed to prepare nodes for Crowbar upgrade"
   def prepare
-    status = :ok
-    msg = ""
-
-    begin
-      service_object = CrowbarService.new(Rails.logger)
-
-      service_object.prepare_nodes_for_crowbar_upgrade
-    rescue => e
-      msg = e.message
-      Rails.logger.error msg
-      status = :unprocessable_entity
-    end
-
-    if status == :ok
-      head status
+    if Api::Upgrade.prepare(background: true)
+      head :ok
     else
       render json: {
         errors: {
@@ -106,7 +93,7 @@ class Api::UpgradeController < ApiController
             help: I18n.t("api.upgrade.prepare.help.default")
           }
         }
-      }, status: status
+      }, status: :unprocessable_entity
     end
   end
 
--- a/crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
+++ b/crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
@@ -54,7 +54,7 @@ describe Api::UpgradeController, type: :request do
 
     it "prepares the crowbar upgrade" do
       post "/api/upgrade/prepare", {}, headers
-      expect(response).to have_http_status(:unprocessable_entity)
+      expect(response).to have_http_status(:ok)
     end
 
     it "stops related services on all nodes during upgrade" do
--- a/crowbar_framework/spec/models/api/upgrade_spec.rb
+++ b/crowbar_framework/spec/models/api/upgrade_spec.rb
@@ -338,4 +338,30 @@ describe Api::Upgrade do
       expect(subject.class.best_method).to eq("none")
     end
   end
+
+  context "with preparing the upgrade" do
+    it "succeeds to spawn the prepare in the background" do
+      allow_any_instance_of(CrowbarService).to receive(
+        :prepare_nodes_for_crowbar_upgrade
+      ).and_return(true)
+
+      expect(subject.class.prepare(background: true)).to be true
+    end
+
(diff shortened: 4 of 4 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -83,5 +93,31 @@ def clusters_healthy?
     def compute_resources_available?
       Api::Crowbar.new.compute_resources_available?
     end
+
+    def prepare_nodes_for_crowbar_upgrade_background
+      @thread = Thread.new do
+        Rails.logger.debug(I18n.t("api.upgrade.prepare.started_background"))
+        status = prepare_nodes_for_crowbar_upgrade
```
MaximilianMeister: why is this useless when I use it at the end of the method as a return value? otherwise the thread will be returned

jsuchome:
Are you sure we have to put it to background on this level? It's just applying proposal, that should not timeout executed from controller. I thought it was bug in crowbarctl

MaximilianMeister:
> Are you sure we have to put it to background on this level?

it should be at the API level, as we can't really do that on the proposal commit level as we might cause a lot of other issues.

>  It's just applying proposal, that should not timeout executed from controller.

I think that is the action that takes so long. see https://bugzilla.novell.com/show_bug.cgi?id=988030

jsuchome:
Maybe I'm missing something, but 'prepare nodes for upgrade' is really the same task as commiting proposal. It does not take longer time than other proposal commits. It does not time out when we call the original API directly, it time outs when we are using crowbarctl.

MaximilianMeister:
> It does not time out when we call the original API directly, it time outs when we are using crowbarctl.

I understand the concern, but that would leave increasing the timeout in crowbarctl as the only solution, which is not even possible, because:

1. it is a global timeout value
2. any subcommand could take longer than the increased timeout, we can never really estimate how long a command could take

I think the only good solution here is to spawn all the tasks that could take some time in a background process (when invoked over the API)

jsuchome:
I do not advocate for increasing the timeout, but for using the background solution one level up, in crowbarctl.

MaximilianMeister:
> I do not advocate for increasing the timeout, but for using the background solution one level up, in crowbarctl.

that will not work with `crowbarctl` as it is making only simple http requests and waits for the response, and on the client side where we call `crowbarctl` we have no server to attach a background thread onto

aspiers:
> I understand the concern, but that would leave increasing the timeout in crowbarctl as the only solution, which is not even possible, because:
> 
> 1. it is a global timeout value
> 2. any subcommand could take longer than the increased timeout, we can never really estimate how long a command could take
>
> I think the only good solution here is to spawn all the tasks that could take some time in a background process (when invoked over the API)

Sorry, but I don't agree with this line of reasoning:

1. Just because it's a global timeout value, that doesn't mean it can't be changed per invocation, and AFAIK each invocation will only trigger a single API operation.
2. (More importantly) Switching to asynchronous background execution doesn't solve the problem, because then you still have to poll for completion of the background task, and then you need to decide a timeout on the polling loop, so you are back to the same issue.

I am not saying that switching to async is a bad idea; in fact I think it's probably a much better design, and after all that's how OpenStack does it.  But switching to async will not magically make the need for timeouts disappear.  They are independent problems.

Ultimately there is an unavoidable trade-off between a) the risk of exceeding a timeout just because things executed successfully but took a long time to do so, and b) the risk of having to wait too long for a timeout when things went wrong with execution.  If you think about it this way, it should be clear that a) is a much bigger risk than b), because a) causes incorrect behaviour, i.e. a false positive for a "failure" which wasn't really a failure.  In contrast with b), if your timeout is 15 minutes, and something goes wrong and execution hangs after 3 minutes, then you might have to wait 12 minutes to discover that something went wrong, *but it's still correct behaviour*, just not great performance.  So the challenge is to pick a timeout in the sweet spot, i.e. high enough that it should hardly ever cause a false positive (except maybe in extreme circumstances, e.g. when running on a stupidly slow overloaded `mkcloud` host), and low enough that the caller doesn't have to wait too long to get notified of the failure.

> > I do not advocate for increasing the timeout, but for using the background solution one level up, in crowbarctl.
> 
> that will not work with `crowbarctl` as it is making only simple http requests and waits for the response, and on the client side where we call `crowbarctl` we have no server to attach a background thread onto

I agree; I can't think of any clean way to keep the API synchronous whilst making `crowbarctl` asynchronous.  Additionally that would not address the problem for other API consumers such as the old `crowbar` command.

MaximilianMeister:
> (More importantly) Switching to asynchronous background execution doesn't solve the problem, because then you still have to poll for completion of the background task, and then you need to decide a timeout on the polling loop, so you are back to the same issue.

that makes sense somehow, but a background thread like we do in the installer or the backup-restore has proven to work good for us, and especially during the upgrade where we have several (some long) steps i still think it is better this way:
starting an action -> writing a file once the action is done... and combining the status of the whole upgrade in just one "polling" endpoint

the only inconveniency I think is really at the cli level, where you have to poll manually... :( but in the UI with javascript/ajax it would not be a problem

aspiers:
> that makes sense somehow, but a background thread like we do in the installer or the backup-restore has proven to work good for us

@MaximilianMeister You are preaching to the choir :-)  That's why I already said:

> I am not saying that switching to async is a bad idea; in fact I think it's probably a much better design, and after all that's how OpenStack does it. But switching to async will not magically make the need for timeouts disappear. They are independent problems.

Anyway, regarding the actual implementation:

> and especially during the upgrade where we have several (some long) steps i still think it is better this way:
> starting an action -> writing a file once the action is done... (to indicate failure or success) and combining the status of the whole upgrade in just one "polling" endpoint

Is there a good reason not to use a well-established Rails pattern for background job execution, such as ActiveJob and/or `delayed_job`?

> the only inconveniency I think is really at the cli level, where you have to poll manually... :( but in the UI with javascript/ajax it would not be a problem

Sure, it makes for a bit more work for the API consumer, but that's OK I think.  Eventually we could maybe switch to WebSockets or similar for a push notification instead of polling.  But as @santiph already said recently, it's not worth the extra effort doing this now.

MaximilianMeister:
> Is there a good reason not to use a well-established Rails pattern for background job execution, such as ActiveJob and/or delayed_job?

I must say I heard it the first time now, so there is no reason not to use it if it does the job. I'll have a look

MaximilianMeister:
I had a look at `delayed_job` https://github.com/collectiveidea/delayed_job which looks promising.
@aspiers are you fine with working on it in the context of another card, either research, or replacing all the Thread usages? Then we can test and evaluate it more isolated. here we need a bugfix, and I would prefer to use the Threading here for now, as we already successfully use it in multiple places. The logic would stay mostly the same anyways, and it would be just a small change.
we could possibly even think about using it in context of the deployment queue (no idea how much work that'd be though)

aspiers:
@MaximilianMeister That sounds fine to me, thanks!

MaximilianMeister:
@aspiers jfyi i created https://trello.com/c/x23thuFP/1163-evaluate-delayed-job-for-background-tasks for the evaluation of a better background processing

aspiers:
@MaximilianMeister Great, thanks!

MaximilianMeister:
depends on the state machine https://trello.com/c/OWehKX6K/126-5-implement-a-state-machine-for-the-upgrade

abel-navarro:
@MaximilianMeister looks like the state machine task will take long time. Could we recover https://github.com/SUSE-Cloud/automation/pull/1229 as temporary workaround?

MaximilianMeister:
> looks like the state machine task will take long time. Could we recover SUSE-Cloud/automation#1229 as temporary workaround?

sure, no need for waiting

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -195,6 +195,8 @@ def nodes
       end
 
       def prepare(options = {})
+        ::Crowbar::UpgradeStatus.new.start_step
```
aspiers: It looks like here you are assuming that the upgrade state is `upgrade_prepare`, but how can you be sure?  Same question as https://github.com/crowbar/crowbar-core/pull/736#discussion_r84558612

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -363,9 +365,11 @@ def prepare_nodes_for_crowbar_upgrade
         service_object = CrowbarService.new(Rails.logger)
         service_object.prepare_nodes_for_crowbar_upgrade
 
+        ::Crowbar::UpgradeStatus.new.end_step
         true
       rescue => e
         message = e.message
+        ::Crowbar::UpgradeStatus.new.end_step(false, message)
```
aspiers: The second parameter to `end_step` defaults to `{}`, so if `message` is a `String` then this looks like a bug.  It should not sometimes be a `Hash` and other times a `String`.

MaximilianMeister: good point. we have to fix that in the lib

jsuchome:
Tested with my Cloud6 instance. Backgrounding works well, but current state of status library is broken (to be addressed by another PR)

PR #662, opened 2016-09-14
Bring back the old nodes upgrade method

jsuchome:
New methods do not provide full functional replacement yet.
Let's use the old one until we have new solution ready. 

And yes, I've renamed the old method, because it really was misleading, it was not preparation, but really starting of the upgrade....

I'm bringing this back to enable green gating of upgrade CI job. It partially reverts https://github.com/crowbar/crowbar-core/pull/640/commits/b4e2946cda7e4e5753f294d787b7e5095512a829

### DIFF
--- a/crowbar_framework/app/controllers/installer/upgrades_controller.rb
+++ b/crowbar_framework/app/controllers/installer/upgrades_controller.rb
@@ -222,7 +222,7 @@ module Installer
           begin
             post_cleanup_repos
             @service_object.disable_non_core_proposals
-            @service_object.prepare_nodes_for_os_upgrade
+            @service_object.initiate_nodes_upgrade
 
             format.json do
               head status
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -391,6 +391,60 @@ class CrowbarService < ServiceObject
     commit_and_check_proposal
   end
 
+  # Prepare the upgrade AutoYaST profile for nodes and let them use it for the upgrade
+  # This is the method from 5-6 upgrade. To be removed once we have full replacement
+  def initiate_nodes_upgrade
+    upgrade_nodes = NodeObject.all.reject do |node|
+      node.admin? || node[:platform] == "windows" || node.state != "crowbar_upgrade"
+    end
+    check_if_nodes_are_available upgrade_nodes
+    admin_node = NodeObject.admin_node
+    upgrade_nodes_failed = []
+
+    upgrade_nodes.each do |node|
+      node["target_platform"] = admin_node["provisioner"]["default_os"]
+      node.save
+      node.set_state("os-upgrading")
+    end
+
+    # wait for the pxe_config to be updated, then reboot the nodes
+    discovery_dir = "#{NodeObject.admin_node[:provisioner][:root]}/discovery/"
+    pxecfg_subdir = "bios/pxelinux.cfg"
+
+    upgrade_nodes.each do |node|
+      boot_ip_hex = node["crowbar"]["boot_ip_hex"]
+      node_arch = node["kernel"]["machine"]
+      pxe_conf = "#{discovery_dir}/#{node_arch}/#{pxecfg_subdir}/#{boot_ip_hex}"
+      ready_for_reboot = false
+
+      while Time.now.to_i < Time.now.to_i + 120 && !ready_for_reboot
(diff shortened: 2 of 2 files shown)


crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -222,7 +222,7 @@ def nodes
           begin
             post_cleanup_repos
             @service_object.disable_non_core_proposals
-            @service_object.prepare_nodes_for_os_upgrade
+            @service_object.initiate_nodes_upgrade
```
MaximilianMeister: do we care about the return value (failed node list) here ?

jsuchome: That would be the exception, catched later.

This is really the old code from 5-6 upgrade, no need to make it better. It's using the old controller action that won't be used in real 6-7 path.

abel-navarro:
OK for me, but I have the feeling we are committing two much intermediate code.

+1

jsuchome:

jsuchome:
Succesfully manually tested.

rsalevsky:
-1

Sorry for the -1 but I lost the complete overview. We merge PRs, revert them partly change something here and there. I have the feeling it's just try and error and this should not be our goal. I have no idea anymore which code is final or which parts are just temporary to fix something else till we run in the next problem.

MaximilianMeister:
@jsuchome maybe we could make a feature branch (if all is crowbar-core)?

the we'd have always a gating for all changes, and it wouldn't mess around in master too much

jsuchome:
My point is to make https://ci.suse.de/view/Cloud/view/Cloud7/job/cloud-mkcloud7-job-upgrade-x86_64/ working until we have a functional replacement.

That job uses the old technique for upgrading the nodes, that is not non-desruptive in a way we plan 6-7 upgrade to work. But it is able to upgrade the nodes, if we return this piece that was recently removed.

>  I have no idea anymore which code is final or which parts are just temporary

I think this is mentioned right in this PR: https://github.com/crowbar/crowbar-core/pull/662/files#diff-006ea8943f6241893e1af682a8f42073R395

@rsalevsky if you refuse this way, do you have an alternative? Do you think we should not care about https://ci.suse.de/view/Cloud/view/Cloud7/job/cloud-mkcloud7-job-upgrade-x86_64/ now?

jsuchome:
My proposal for the upgrade CI jobs is here (at it was relevant to other PR's as well): https://github.com/SUSE-Cloud/automation/pull/1223#issuecomment-246921616

jsuchome:
Also, the revert I mentioned: it's not that I'm removing any new code. I'm trying to leave both API's working.

rsalevsky:
@jsuchome Yes that would work for me. I mean this is test driven development at the beginning something is red and you make it green. So from my point of view as long as we make progress I don't care if the job is red or green.

jsuchome:
@rsalevsky AC at this card
https://trello.com/c/BfMFXB3r/64-5-s47p5-s48p19-add-ci-jobs-for-new-upgrade-process-with-and-without-ha says
"Jenkins jobs exist for both old and new upgrade processes"

That's why I'm trying to make both work (the old one could easily, the new one would be red obviously)

vuntz:
Is this enough to get the old job passing? What we wanted is to not go red so we would keep making sure that the openstack side of things for the upgrade was fine; but if it's hopelessly red, it might be too late and not worth the effort fixing it.

jsuchome:
I tested manually by starting with `plain` step for Cloud6, upgrading admin server, using this patch and continuing with rest of `cloudupgrade`. This has worked and AFAIK is what the current upgrade job is doing.

vuntz:
Ok, so I think we should merge this. This will matter when we move to Newton, as it will help find issues in the L->N upgrade path earlier on.

aspiers:
> I mean this is test driven development at the beginning something is red and you make it green. So from my point of view as long as we make progress I don't care if the job is red or green.

IIUC you should care, because in TDD the tests are never supposed to be red in the mainline repository, only in developers' working copies and temporary branches.

jsuchome:
> @jsuchome maybe we could make a feature branch (if all is crowbar-core)?

@MaximilianMeister might be good idea, but not for this PR, which tries to bring current master to the state where the upgrade job is building

MaximilianMeister:
+1

PR #664, opened 2016-09-14
Add the upgrade menu entry

rsalevsky:


vuntz:
Thanks. Let's only merge this once we're ready to push the maintenance update that allows upgrading.

rsalevsky:
Next update will include the Upgrade so let's merge it.

crowbar_framework/config/navigation.rb
```diff
@@ -38,6 +38,7 @@
       level2.item :queue, t("nav.utils.queue"), deployment_queue_path
       level2.item :repositories, t("nav.utils.repositories"), repositories_path
       level2.item :backup, t("nav.utils.backup"), backups_path
+      level2.item :backup, t("nav.utils.upgrade"), "/upgrade"
```
skazi0: Just noticed. Shouldn't it be `:upgrade` instead of `:backup` here?

skazi0:
[(2) S58P5 Add the "Upgrade" link from Crowbar UI](https://trello.com/c/cdtWLx6h/391-2-s58p5-add-the-upgrade-link-from-crowbar-ui)

PR #661, opened 2016-09-13
Shutdown non-essential services before the upgrade

jsuchome:
Uses research from https://github.com/crowbar/crowbar-core/pull/653  and https://github.com/crowbar/crowbar-core/pull/655

PR #666, opened 2016-09-15
api cleanup: move business logic into the model

MaximilianMeister:


jsuchome:
@MaximilianMeister Do you want me to adapt https://github.com/crowbar/crowbar-core/pull/661 this way, or should we rather merge that one and do the cleanup as part of your PR?

Some other calls should be adapted as well (`prepare`), and for some model methods we'll change the return values, I think (`prechecks`)

MaximilianMeister:
> Do you want me to adapt #661 this way, or should we rather merge that one and do the cleanup as part of your PR?

I can add it here after the merge, no problem

jsuchome:
What about `prepare` ?

MaximilianMeister:
> What about prepare ?

lets wait for the outcome of https://github.com/crowbar/crowbar-core/pull/660

abel-navarro:
+1

jsuchome:
+1

PR #669, opened 2016-09-19
[stable/3.0] Simple check if HA setup is deployed

jsuchome:
Before the start of the upgrade, we also should check if there's HA configuration, otherwise we could not guarantee non-disruptive upgrade.

I'm not sure if this is good enough, or if we should rather check for specific services (`neutron-server` ?) being in pacemaker cluster... ?

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -125,6 +125,15 @@ module Api
       end
     end
 
+    def ha_deployed?
+      founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
+
+      if founders.empty?
+        Rails.logger.warn("No pacemaker cluster is configured.")
+        return false
+      end
+      true
+    end
 
     # Simple check if HA clusters report some problems
     # If there are no problems, empty hash is returned.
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -125,6 +125,15 @@ def repocheck
       end
     end
 
+    def ha_deployed?
+      founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
+
+      if founders.empty?
+        Rails.logger.warn("No pacemaker cluster is configured.")
```
MaximilianMeister: I would just return `founders.any?`, as I think the log entry is not really necessary

jsuchome: It might be necessary now. But we definitely need to show the errors/warnings somehow to user, not just return true/false values from methods. 
I didn't change it here though, as we must do it correctly for all upgrade related methods.

jsuchome:
@MaximilianMeister  Api::Crowbar.new.addons does not seem to be available in 3.0, or is it?

jsuchome:
ok, let's start with master first...

aspiers:
@jsuchome You don't need to close the PR to change the base branch; please read https://github.com/blog/2224-change-the-base-branch-of-a-pull-request

PR #670, opened 2016-09-19
Simple check for presence of HA-based configuration

jsuchome:
Before the start of the upgrade, we also should check if there's HA configuration, otherwise we could not guarantee non-disruptive upgrade.

I'm not sure if this is good enough, or if we should rather check for specific services (neutron-server ?) being in pacemaker cluster... ?

(Needs a backport to 3.0)

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -156,6 +156,22 @@ module Api
         true
       end
 
+      # Check for presence of HA setup, which is a requirement for non-disruptive upgrade
+      def ha_presence_check
+        ret = {}
+        unless addon_installed? "ha"
+          ret[:errors] = [I18n.t("api.crowbar.ha_not_installed")]
+          return ret
+        end
+
+        founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
+
+        if founders.empty?
+          ret[:errors] = [I18n.t("api.crowbar.ha_not_configured")]
+        end
+        ret
+      end
+
       protected
 
       def lib_path
--- a/crowbar_framework/config/locales/crowbar/en.yml
+++ b/crowbar_framework/config/locales/crowbar/en.yml
@@ -773,6 +773,8 @@ en:
       upgrade_ongoing: 'Upgrade is already ongoing. Please wait.'
       zypper_locked: '%{zypper_locked_message}'
       upgrade_script_path: 'Could not find %{path}'
+      ha_not_installed: 'HA add-on is not installed.'
+      ha_not_configured: 'No Pacemaker cluster is configured.'
 
   # Global
   overview: 'Overview'
--- a/crowbar_framework/spec/models/api/upgrade_spec.rb
+++ b/crowbar_framework/spec/models/api/upgrade_spec.rb
@@ -22,6 +22,7 @@ describe Api::Upgrade do
       )
     )
   end
+  let!(:node) { NodeObject.find_node_by_name("testing") }
 
   before(:each) do
     allow(Api::Node).to(
@@ -53,22 +54,22 @@ describe Api::Upgrade do
   context "with a successful status" do
     it "checks the status" do
       allow(Crowbar::Sanity).to receive(:sane?).and_return(true)
-      allow_any_instance_of(Api::Crowbar).to(
+      allow(Api::Crowbar).to(
         receive(:features).and_return([])
+      )
+
+      allow(Api::Crowbar).to(
+        receive(:addons).and_return(
+          ["ha"]
+        )
+      )
+      allow(Api::Crowbar).to(
+        receive(:ha_presence_check).and_return({})
       )
 
       expect(subject.class).to respond_to(:status)
       expect(subject.class.status).to be_a(Hash)
       expect(subject.class.status.deep_stringify_keys).to eq(upgrade_status)
-    end
-  end
-
-  context "with a successful maintenance updates check" do
-    it "checks the maintenance updates on crowbar" do
-      allow(Crowbar::Sanity).to receive(:sane?).and_return(true)
-
-      expect(subject.class).to respond_to(:check)
-      expect(subject.class.check.deep_stringify_keys).to eq(upgrade_prechecks)
     end
   end
 
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -32,7 +32,8 @@ module Api
           maintenance_updates_installed: maintenance_updates_installed?,
           clusters_healthy: clusters_healthy?,
           compute_resources_available: compute_resources_available?,
-          ceph_healthy: ceph_healthy?
+          ceph_healthy: ceph_healthy?,
+          ha_deployed: ha_presence_check.empty?
         }
       end
 
@@ -132,6 +133,10 @@ module Api
 
       def compute_resources_available?
         Api::Crowbar.compute_resources_available?
+      end
+
+      def ha_presence_check
+        Api::Crowbar.ha_presence_check
       end
     end
   end
--- a/crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
+++ b/crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
@@ -21,9 +21,11 @@ describe Api::UpgradeController, type: :request do
     it "shows the upgrade status object" do
       allow(Crowbar::Sanity).to receive(:sane?).and_return(true)
       allow(Api::Crowbar).to(
-        receive(:addons).and_return([])
+        receive(:addons).and_return(["ha"])
       )
-
+      allow(Api::Upgrade).to(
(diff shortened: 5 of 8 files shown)


MaximilianMeister:
@jsuchome could you add a unit test for `ha_deployed?`

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -150,6 +150,22 @@ def compute_resources_available?
       true
     end
 
+    # Check for presence of HA setup, which is a requirement for non-disruptive upgrade
+    def ha_deployed?
```
dirkmueller: do we want to name that more specific, like e.g. control_plane_on_ha (or something like that)? I guess eventually we want to check that the relevant roles are deployed properly on a ha cluster like we expect them to be, and not only whether a proposal exists. 

Do we also want to return not only a boolean result but also a error message, like in a status code + error message combination? this could be displayed to the user then.

jsuchome: 1. It would be probably better to check more specifically yes (as I've asked in the PR message), but I'm not sure at this point what exactly to check.

2. Yes, error message should be shown, but we're currently redesigning the whole part, so I believe it will come as followup PR that would have unique way of returning errors for all the checks we're currently doing. I believe @MaximilianMeister is already working on something.

dirkmueller: ok, thanks for the update.

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -150,6 +150,24 @@ def compute_resources_available?
...
+      unless addon_installed? "ha"
+        errors.add(:base, I18n.t("api.crowbar.ha_not_installed"))
```
MaximilianMeister: do we plan to use this error somewhere? I think we should only write to the errors hash if we can't further process the instance of this object, not on every check.

jsuchome: I was expecting we're gonna read the errors somehow when constructing that pre-check map. Or do you think whole `ha_deployed` method should be changed to return the hash instead?

MaximilianMeister: as discussed on IRC, the errors hash is usually used outside the model where you call `@instance.errors` after some action. maybe we could use `self.errors` in the model though I don't know if that is a good practice, but I think here if we return a hash instead of a bool, it should be sufficient.

crowbar_framework/app/models/api/crowbar.rb
aspiers: Does this really belong in `crowbar-core`?

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -772,6 +772,8 @@ en:
     crowbar:
       upgrade_ongoing: 'Upgrade is already ongoing. Please wait.'
       zypper_locked: '%{zypper_locked_message}'
+      ha_not_installed: 'HA Addon is not installed.'
```
aspiers: "add-on"

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -150,6 +150,22 @@ def compute_resources_available?
       true
     end
 
+    # Check for presence of HA setup, which is a requirement for non-disruptive upgrade
+    def ha_deployed?
+      unless addon_installed? "ha"
+        Rails.logger.warn("HA Addon is not installed.")
```
aspiers: "add-on"

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -150,6 +150,22 @@ def compute_resources_available?
...
+        return false
+      end
+
+      founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
+
+      if founders.empty?
+        Rails.logger.warn("No pacemaker cluster is configured.")
```
aspiers: "Pacemaker"

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -772,6 +772,8 @@ en:
...
+      ha_not_configured: 'No pacemaker cluster is configured.'
```
aspiers: "Pacemaker"

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -151,21 +151,19 @@ def compute_resources_available?
     end
 
     # Check for presence of HA setup, which is a requirement for non-disruptive upgrade
-    def ha_deployed?
+    def ha_presence_check
```
aspiers: This commit should be squashed into the one which introduces `ha_deployed?`.

jsuchome:
> Does this really belong in crowbar-core?

Actually, yes. crowbar-ha might not be even installed - or does it has to be?

crowbar_framework/app/models/api/crowbar.rb
jsuchome: Adapted... but the comment is still shown here... :-(

aspiers:
> crowbar-ha might not be even installed - or does it has to be?

It will always be installed, but in theory we should aim to keep everything working even when it isn't.

aspiers:
Looks like you lost the original commit message when you did the squash - it has the message from the test suite update, which we don't want any more.

MaximilianMeister:
> It will always be installed, but in theory we should aim to keep everything working even when it isn't.

OFFTOPIC:
@aspiers if it is always installed, should https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/api/crowbar.rb#L85 then rather check if an addon is deployed? this will be used by the upgrade UI to determine if we have used ceph or pacemaker

jsuchome:
It's not off topic. If it might not be installed, than these checks definitely belong in crowbar-core, and only the more specific ones (like for cluster health) should go to crowbar-ha

jsuchome:
@aspiers so do you want me to move it to crowbar-ha, or is there any other blocker here?

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -156,6 +156,22 @@ def compute_resources_available?
         true
       end
 
+      # Check for presence of HA setup, which is a requirement for non-disruptive upgrade
+      def ha_presence_check
```
rsalevsky: As I mentioned here https://github.com/crowbar/crowbar-core/pull/689#discussion_r81117627, the name of this function is confusing and doesn't describe what she is doing. The current name somehow let's assumes the a boolen value is returned.

jsuchome: @rsalevsky I think the relation of these kind of methods is addressed in https://github.com/crowbar/crowbar-core/pull/672. It would be probably best if we could merge that one first and adapt the rest later

aspiers: Splitting functions into smaller parts is almost always a good idea! And sounds like it makes sense to me In this case.

crowbar_framework/spec/models/api/upgrade_spec.rb
```diff
@@ -53,25 +54,25 @@
...
+        receive(:ha_presence_check).and_return({})
+      )
+
       expect(subject.class).to respond_to(:status)
       expect(subject.class.status).to be_a(Hash)
       expect(subject.class.status.deep_stringify_keys).to eq(upgrade_status)
     end
   end
 
-  context "with a successful maintenance updates check" do
```
aspiers: Why are you removing this? It has nothing to do with the HA checks AFAICS.

jsuchome: No it does not. But it is wrong, the check has nothing to do with maintenance updates

aspiers: OK, then please fix it in a separate commit which has a commit message explaining this!

jsuchome: In the backport PR (https://github.com/crowbar/crowbar-core/pull/689) it is in the separate commit. I thought you wanted this PR to be gone (thus I created https://github.com/crowbar/crowbar-ha/pull/147 for master.

aspiers: Yes, I just remembered that and I'm looking at 
https://github.com/crowbar/crowbar-ha/pull/147 right now. But in general I feel that a lot of time is being wasted currently within the review process due to lack of commit hygiene. I strongly believe we should adhere to the same [standards in upstream OpenStack](https://wiki.openstack.org/wiki/GitCommitMessages). I'll raise it in the next standup.

aspiers:
I definitely prefer crowbar/crowbar-ha#147 to putting the check in this repo, so I guess this can be closed?  Although I'm a bit confused why this PR looks almost nothing like the other one.

jsuchome:
OK, let's close this one. The difference is in the testsuite, crowbar-ha apparently does not have anything.

PR #671, opened 2016-09-19
Provide a s390x ohai subtree similar to the dmi subtree

dirkmueller:
Since DMI does not exist on mainframe, we need to create
a custom ohai plugin to capture the DMI information that
we need for s390x

### DIFF
--- a/chef/cookbooks/ohai/files/default/plugins/linux/s390x.rb
+++ b/chef/cookbooks/ohai/files/default/plugins/linux/s390x.rb
@@ -1,0 +1,41 @@
+provides "s390x"
+
+if File.exist? "/proc/sysinfo"
+  retval = Mash.new.tap do |result|
+    result[:system] = {}
+
+    result[:system][:manufacturer] = "PR/SM"
+
+    File.open("/proc/sysinfo").each do |line|
+      key, val = line.split(":", 2)
+      next unless val
+
+      key.gsub!(/:$/, "")
+      val.strip!
+
+      if key.include? "Control Program"
+        if val.include? "KVM"
+          result[:system][:manufacturer] = "KVM"
+        end
+
+        if val.include? "z/VM"
+          result[:system][:manufacturer] = "z/VM"
+        end
+      end
+
+      if key.include? "UUID"
+        result[:system][:uuid] = val
+      end
+
+      if key.include? "Extended Name"
+        result[:system][:product_name] = val
+      end
+
+      if key.include? "Sequence Code"
+        result[:system][:serial_number] = val
+      end
+    end
+  end
+
+  s390x retval
+end


chef/cookbooks/ohai/files/default/plugins/linux/s390x.rb
```diff
@@ -0,0 +1,40 @@
+provides "s390x"
+
+if File.exist? "/proc/sysinfo"
```
rhafer: I guess this file something s390 specific? Or do we need some other way to ensure this plugin is only run on s390 machines?

dirkmueller: I have not been able to find that file on s390x, armv7, aarch64 or ppc64le. so I guess it is unique enough. the content of the file only makes sense on s390x at least.

dirkmueller:
one open thought here that I have. maybe we don't want to put it under "s390x" but something like "crowbar_hacks" and switch all usages to that instead ? the patch in crowbar-ha is certainly ugly (https://github.com/crowbar/crowbar-ha/pull/143)

vuntz:
It's fine to put it under s390x, but I guess we should have some more magic later on that takes the attributes we care about and copy them in a common namespace so that we don't need to add more and more `if` branches in places like crowbar-ha.

rhafer:
> It's fine to put it under s390x, but I guess we should have some more magic later on that takes the attributes we care about and copy them in a common namespace so that we don't need to add more and more if branches in places like crowbar-ha.

@vuntz Do you think adding yet another ohai plugin that depends on "dmi" and "s390x" and just adds those values on e.g. ```crowbar_ohai.system``` would work? Or do you have something else in mind?

chef/cookbooks/ohai/files/default/plugins/linux/s390x.rb
```diff
@@ -0,0 +1,40 @@
...
+  retval = Mash.new.tap do |result|
+    result[:system] = {}
+
+    result[:system][:manufacturer] = "PR/SM"
+
+    File.open("/proc/sysinfo") do |p|
```
aspiers: Nitpick: even though this temporary variable is only used once, I'd still prefer to avoid cryptic variable names, especially ones which are the same as a commonly used method in `Kernel`.  Maybe just `file` ?

chef/cookbooks/ohai/files/default/plugins/linux/s390x.rb
aspiers: Wouldn't it be safer to check `/proc/cpuinfo` for this?  Just an idea.

rhafer:
+1 (hello lgtm?)

chef/cookbooks/ohai/files/default/plugins/linux/s390x.rb
dirkmueller: @aspiers we plan to support any architecture other than armv7, aarch64, ppc64le, ppc64, x86_64, i586 ? none of those have that file.

chef/cookbooks/ohai/files/default/plugins/linux/s390x.rb
dirkmueller: we can just remove the need for that variable alltogether.

vuntz:
> @vuntz Do you think adding yet another ohai plugin that depends on "dmi" and "s390x" and just adds those values on e.g. crowbar_ohai.system would work? Or do you have something else in mind?

That'd work for me, at least :-) I'm fine if it's done in another PR.

rhafer:
> That'd work for me, at least :-) I'm fine if it's done in another PR.

Ok, that would be https://github.com/crowbar/crowbar-core/pull/677 then.

PR #672, opened 2016-09-21
adapt the prechecks API

MaximilianMeister:
we are including specific errors now in the response

example:

```json
{
  "checks": {
    "network_checks": {
      "required": true,
      "passed": true,
      "errors": {}
    },
    "maintenance_updates_installed": {
      "required": true,
      "passed": false,
      "errors": {
        "maintenance_updates_installed": {
          "data": [
            "ZYPPER_EXIT_INF_UPDATE_NEEDED: patches available for installation."
          ],
          "help": "make sure maintenance updates are installed"
        }
      }
    },
    "clusters_healthy": {
      "required": true,
      "passed": true,
      "errors": {}
    },
    "compute_resources_available": {
      "required": false,
      "passed": true,
      "errors": {}
    },
    "ceph_healthy": {
      "required": true,
      "passed": true,
      "errors": {}
    }
  }
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/crowbar/crowbar-core/672)

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -14,8 +14,6 @@
 # limitations under the License.
 #
 
-require "open3"
-
 module Api
   class Crowbar < Tableless
     class << self
@@ -111,25 +109,6 @@ module Api
           ret[:cloud][:repos][admin_architecture.to_sym] = {
             missing: ["SUSE OpenStack Cloud 8"]
           } unless cloud_available
-        end
-      end
-
-      def maintenance_updates_status
-        {}.tap do |ret|
-          Open3.popen3("zypper patch-check") do |_stdin, _stdout, _stderr, wait_thr|
-            case wait_thr.value.exitstatus
-            when 100
-              ret[:errors] ||= []
-              ret[:errors].push(
-                I18n.t("api.crowbar.maintenance_updates_status.patches_missing")
-              )
-            when 101
-              ret[:errors] ||= []
-              ret[:errors].push(
-                I18n.t("api.crowbar.maintenance_updates_status.security_patches_missing")
-              )
-            end
-          end
         end
       end
 
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -76,25 +76,6 @@ module Api
         end
       end
 
-      def maintenance_updates_installed?
-        Open3.popen3("zypper patch-check") do |_stdin, _stdout, _stderr, wait_thr|
-          case wait_thr.value.exitstatus
-          when 100
-            Rails.logger.warn(
-              "ZYPPER_EXIT_INF_UPDATE_NEEDED: patches available for installation."
-            )
-            false
-          when 101
-            Rails.logger.warn(
-              "ZYPPER_EXIT_INF_SEC_UPDATE_NEEDED: security patches available for installation."
-            )
-            false
-          else
-            true
-          end
-        end
-      end
-
       def repocheck
         # FIXME: once we start working on 7 to 8 upgrade we have to adapt the sles version
         zypper_stream = Hash.from_xml(
@@ -130,6 +111,25 @@ module Api
           ret[:cloud][:repos][admin_architecture.to_sym] = {
             missing: ["SUSE OpenStack Cloud 8"]
           } unless cloud_available
+        end
+      end
+
+      def maintenance_updates_status
+        {}.tap do |ret|
+          Open3.popen3("zypper patch-check") do |_stdin, _stdout, _stderr, wait_thr|
+            case wait_thr.value.exitstatus
+            when 100
+              ret[:errors] ||= []
+              ret[:errors].push(
+                I18n.t("api.crowbar.maintenance_updates_status.patches_missing")
+              )
+            when 101
+              ret[:errors] ||= []
+              ret[:errors].push(
+                I18n.t("api.crowbar.maintenance_updates_status.security_patches_missing")
+              )
+            end
+          end
         end
       end
 
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -22,18 +22,38 @@ module Api
       def status
         {
           crowbar: crowbar_upgrade_status,
-          checks: check
+          checks: checks
         }
       end
 
-      def check
-        {
-          sanity_checks: sanity_checks,
-          maintenance_updates_installed: maintenance_updates_installed?,
-          clusters_healthy: clusters_healthy?,
-          compute_resources_available: compute_resources_available?,
-          ceph_healthy: ceph_healthy?
-        }
+      def checks
+        {}.tap do |ret|
+          ret[:network_checks] = {
+            required: true,
+            passed: network_checks.empty?,
+            errors: sanity_check_errors
+          }
+          ret[:maintenance_updates_installed] = {
+            required: true,
+            passed: maintenance_updates_status.empty?,
+            errors: maintenance_updates_check_errors
+          }
+          ret[:compute_resources_available] = {
+            required: false,
+            passed: compute_resources_available?,
+            errors: compute_resources_check_errors
+          }
+          ret[:ceph_healthy] = {
+            required: true,
+            passed: ceph_healthy?,
+            errors: ceph_health_check_errors
+          } if Api::Crowbar.addons.include?("ceph")
+          ret[:clusters_healthy] = {
+            required: true,
+            passed: clusters_healthy?,
+            errors: clusters_health_check_errors
+          } if Api::Crowbar.addons.include?("ha")
+        end
       end
 
       def repocheck
@@ -171,12 +191,12 @@ module Api
         Api::Crowbar.upgrade
       end
 
-      def sanity_checks
-        ::Crowbar::Sanity.sane? || ::Crowbar::Sanity.check
+      def maintenance_updates_status
+        @maintenance_updates_status ||= Api::Crowbar.maintenance_updates_status
       end
 
-      def maintenance_updates_installed?
-        Api::Crowbar.maintenance_updates_installed?
+      def network_checks
+        @network_checks ||= ::Crowbar::Sanity.check
       end
 
       def ceph_healthy?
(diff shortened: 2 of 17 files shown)


MaximilianMeister:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;3TALQwDA&#x2F;121-5-upgrade-synchronize-api-responses-with-the-angular-app"> (5) #upgrade Synchronize API responses with the Angular app</a>

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -63,21 +63,24 @@ def upgrade!
...
-          true
+    def maintenance_updates_status
+      {}.tap do |ret|
+        ret[:passed] = false
+        ret[:errors] = []
+
+        Open3.popen3("zypper patch-check") do |_stdin, _stdout, _stderr, wait_thr|
+          case wait_thr.value.exitstatus
+          when 100
+            ret[:errors].push("ZYPPER_EXIT_INF_UPDATE_NEEDED: patches available for installation.")
```
jsuchome: Do we want to translate that message? (and others)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -20,18 +20,39 @@ module Api
...
-        ceph_healthy: ceph_healthy?
+        checks: {
+          sanity_checks: {
+            required: true,
+            passed: sanity_checks.empty?,
+            errors: sanity_check_errors
+          },
+          maintenance_updates_installed: {
+            required: true,
+            passed: maintenance_updates_status[:passed],
```
jsuchome: What about `maintenance_updates_check_errors.empty?` here, and not including :passed in maintenance_updates_status?

MaximilianMeister: good point, that is redundant information

crowbar_framework/app/models/api/crowbar.rb
MaximilianMeister: probably yes. will adapt it

jsuchome:
Maybe you could include empty check for `ha_deployed`

MaximilianMeister:
> Maybe you could include empty check for ha_deployed

@jsuchome what about getting https://github.com/crowbar/crowbar-core/pull/670 merged, and I will rebase on top?

jsuchome:
> @jsuchome what about getting #670 merged, and I will rebase on top?

Or that, sure

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -20,18 +20,39 @@ module Api
...
+            required: true,
+            passed: clusters_healthy?,
+            errors: clusters_health_check_errors
+          },
+          compute_resources_available: {
+            required: false,
+            passed: compute_resources_available?,
+            errors: compute_resources_check_errors
+          },
+          ceph_healthy: {
```
jsuchome: Shouldn't we pass the result of this check only when Ceph is actually present?

MaximilianMeister: fixed

aspiers:
`nova boot` failed in `testsetup` phase of gate.

aspiers:
@MaximilianMeister The example you gave includes a `sanity_checks` key, but it seems the PR is (wisely) removing that. Perhaps you can update the example to reflect a typical response?

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -63,21 +63,21 @@ def upgrade!
...
-        when 101
-          Rails.logger.warn(
-            "ZYPPER_EXIT_INF_SEC_UPDATE_NEEDED: security patches available for installation."
-          )
-          false
-        else
-          true
+    def maintenance_updates_status
+      {}.tap do |ret|
+        Open3.popen3("zypper patch-check") do |_stdin, _stdout, _stderr, wait_thr|
```
aspiers: Don't we also need to check that the update repo metadata is up to date?  Or does `zypper patch-check` always automatically refresh the update repos?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -20,19 +20,39 @@ module Api
...
+        ret[:checks][:compute_resources_available] = {
+          required: false,
+          passed: compute_resources_available?,
+          errors: compute_resources_check_errors
+        }
+        ret[:checks][:ceph_healthy] = {
+          required: true,
+          passed: ceph_healthy?,
+          errors: ceph_health_check_errors
+        } if Api::Crowbar.new.addons.include?("ceph")
```
aspiers: Why do we need to instantiate an object here? Is there really any useful state inside that object?  If not, shouldn't it be a class method?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -91,12 +111,12 @@ def crowbar_upgrade_status
       Api::Crowbar.new.upgrade
     end
 
-    def sanity_checks
-      ::Crowbar::Sanity.sane? || ::Crowbar::Sanity.check
+    def maintenance_updates_status
+      @maintenance_updates_status ||= Api::Crowbar.new.maintenance_updates_status
```
aspiers: Same confusion here about instances vs. class methods.

crowbar_framework/app/models/api/upgrade.rb
aspiers: Unless we also wanted to be able to return warnings from these checks, in which case the whole structure needs a redesign, because a single value for `required` would not be applicable to all warnings / errors.  For example:

```
{
  "checks": {
    ....
    "maintenance_updates_installed": {
      "warnings": [
        {
          "code": "maintenance_updates_metadata_old",
          "data": [
            "Updates repo metadata is 3 days old."
          ],
          "help": "Make sure maintenance update repo metadata is refreshed via zypper refresh, or turn on auto-refresh via zypper mr -r."
        }
      ],
      "errors": [
        {
          "code": "maintenance_updates_installed",
          "data": [
            "ZYPPER_EXIT_INF_UPDATE_NEEDED: patches available for installation."
          ],
          "help": "make sure maintenance updates are installed"
        }
      ]
    },
    ....
  }
}
```

MaximilianMeister:
> For network checks you are calling out to lib/crowbar/sanity.rb which is fine, but for e.g. maintenance update checks, they are implemented in app/models/api/crowbar.rb

@aspiers now that you mention it, we might move the maintenance updates check to the sanity library, what do you think?

crowbar_framework/app/models/api/upgrade.rb
aspiers: Agreed.

crowbar_framework/app/models/api/crowbar.rb
aspiers: @MaximilianMeister I don't think you answered this question anywhere?

crowbar_framework/app/models/api/upgrade.rb
aspiers: Also now addressed by #674.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -133,6 +153,76 @@ def clusters_healthy?
...
+        ]
+      end
+
+      def ceph_health_check_errors
+        return [] if ceph_healthy?
+
+        [
+          {
+            code: :ceph_health,
+            data: [], # TODO: implement ceph health check errors
```
jsuchome: So, could we adapt `ceph_healthy?` in api/crowbar.rb to return such kind of errors in this PR?

aspiers:
> @aspiers now that you mention it, we might move the maintenance updates check to the sanity library, what do you think?

Yes, that sounds good to me, since in the future they could be reused by code outside the upgrade process.

aspiers:
Sorry, this PR is a single commit which seems to be a big mix of lots of different changes, and no explanation of any them.  This makes it *really* hard to review.  Please can you split it into atomic commits, with a good explanation in each commit?  If you need help doing this, or this request doesn't make sense, please let me know and I'll do my best to show what I'm asking for.

crowbar_framework/app/models/api/crowbar.rb
MaximilianMeister: > Or does zypper patch-check always automatically refresh the update repos?

yes it does. 

> I don't think you answered this question anywhere?

sorry for the late response, I haven't been working on this PR since then. and I thought you might have tried it out yourself already

crowbar_framework/app/models/api/upgrade.rb
MaximilianMeister: we have had a short meeting with the ui team, and they don't require a separate warnings structure. also the example you chose here is maybe not the best, because the maintenance updates check already refreshes the metadata.

MaximilianMeister:
>> @aspiers now that you mention it, we might move the maintenance updates check to the sanity library, what do you think?

> Yes, that sounds good to me, since in the future they could be reused by code outside the upgrade process.

@aspiers done

> Please can you split it into atomic commits, with a good explanation in each commit?

I tried to :) hope it's a bit better now...

MaximilianMeister:
rebased

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Or we might do it in later PR...

jsuchome:
I'm basically fine with how this looks like, but I was wondering if @rsalevsky has something to say, as he commented other PR with somewhat similar issues...

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -191,6 +211,76 @@ def clusters_healthy?
...
+      #     help: String # "this is how you might fix the error"
+      #   },
+      #   ...
+      # ]
+      def sanity_check_errors
+        return [] if network_checks.empty?
+
+        [
+          {
+            code: :network_checks,
```
jsuchome: Sorry for coming too late with this, but when we have a code for each error, wouldn't it make sense to use it as a key for the error submap? Or is there any reason `errors` must be an array?
```
sanity_check_errors
  return {
    :network_checks: {
      :data: network_checks,
      :help: "help text what to do"
    }
  }
end

**Comment by jsuchome** [comment]
@aspiers , are you still requesting changes here?

File: crowbar_framework/app/models/api/upgrade.rb (hunk 8 above)
Comment (MaximilianMeister) [inline crowbar_framework/app/models/api/upgrade.rb]: good point, as it is now the array is somewhat useless

Comment (MaximilianMeister) [inline crowbar_framework/app/models/api/upgrade.rb]: @jsuchome updated

Comment (jsuchome) [inline crowbar_framework/app/models/api/upgrade.rb]: Thanks, could you adapt also the example in the PR description?

Comment (MaximilianMeister) [inline crowbar_framework/app/models/api/upgrade.rb]: done. thanks for the hint. that made me remember, I still had to update the api documentation example...

File: crowbar_framework/app/models/api/crowbar.rb [hunk 9]
```diff
@@ -133,6 +114,25 @@ def repocheck
...
+        {}.tap do |ret|
+          Open3.popen3("zypper patch-check") do |_stdin, _stdout, _stderr, wait_thr|
+            case wait_thr.value.exitstatus
+            when 100
+              ret[:errors] ||= []
+              ret[:errors].push(
+                I18n.t("api.crowbar.maintenance_updates_status.patches_missing")
+              )
+            when 101
+              ret[:errors] ||= []
```
MaximilianMeister: added as a commit on top

crowbar_framework/spec/fixtures/crowbar_maintenance.json
```diff
@@ -1,3 +1,4 @@
 {
-  "maintenance_updates_installed":true
+  "passed":true,
```
MaximilianMeister: agree that this is duplicated data, but more convenient for the API consumer. I'll check with @santiph (IIRC the requirement came from the UI)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -21,19 +21,39 @@ class Upgrade < Tableless
     class << self
       def status
         {
-          crowbar: crowbar_upgrade_status,
-          checks: check
-        }
+          crowbar: crowbar_upgrade_status
+        }.merge!(check)
```
MaximilianMeister: fixed

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -21,19 +21,39 @@ class Upgrade < Tableless
...
       end
 
       def check
```
MaximilianMeister: done

aspiers:
Reviewed 11 of 18 files at r1, 1 of 1 files at r2, 4 of 5 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/672)*

aspiers:
I'm experimenting with Reviewable.io as you can see :-) It's pretty cool!

PR #673, opened 2016-09-21
be more robust with deleted nodes

bmwiedemann:


bmwiedemann:
@rhafer yes, we (partially) deleted a node and got a [] access on nil in this line, that went away after applying this patch.

rhafer:
> yes, we (partially) deleted a node and got a [] access on nil in this line, that went away after applying this patch.

@bmwiedemann Thanks. "(partially) deleted" means what? Just the node-role, just the node-object, just some attributes or something else?

bmwiedemann:
We clicked forget in crowbar, but the node never fully went away, so I also did a `knife node delete $NODE` and a `knife role delete $NODEROLE` to get rid of it, but there were probably some parts left, because the next discovery failed with "cannot create $NODE - already exists"

aspiers:
+1

aspiers:
Wake up, LGTM! +1 again

PR #674, opened 2016-09-22
prefer class methods over instance methods and fix the addon availability check

MaximilianMeister:
needs (partial) backport

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -18,11 +18,11 @@ require "open3"
 
 module Api
   class Crowbar < Tableless
-    attr_reader :version
+    attr_reader :version, :addons
 
     def initialize
       @version = ENV["CROWBAR_VERSION"]
-      @addons = addons
+      @addons = features
     end
 
     def status
@@ -78,14 +78,6 @@ module Api
           false
         else
           true
-        end
-      end
-    end
-
-    def addons
-      [].tap do |list|
-        ["ceph", "ha"].each do |addon|
-          list.push(addon) if addon_installed?(addon)
         end
       end
     end
@@ -170,6 +162,14 @@ module Api
 
     def upgrade_script_path
       Rails.root.join("..", "bin", "upgrade_admin_server.sh")
+    end
+
+    def features
+      [].tap do |list|
+        ["ceph", "ha"].each do |addon|
+          list.push(addon) if addon_installed?(addon)
+        end
+      end
     end
 
     def addon_installed?(addon)
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -167,7 +167,7 @@ module Api
     def features
       [].tap do |list|
         ["ceph", "ha"].each do |addon|
-          list.push(addon) if addon_installed?(addon)
+          list.push(addon) if addon_installed?(addon) && addon_enabled?(addon)
         end
       end
     end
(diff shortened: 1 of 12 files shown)


crowbar_framework/app/models/api/node.rb
```diff
@@ -28,7 +28,7 @@ def repocheck(options = {})
             "available" => true,
             "repos" => {}
           }
-          platform = Api::Upgrade.new.target_platform(platform_exception: addon)
+          platform = Api::Upgrade.target_platform(platform_exception: addon)
```
jsuchome: This would ideally belong to previous commit

crowbar_framework/spec/models/api/upgrade_spec.rb
```diff
@@ -89,7 +85,14 @@
         receive(:find).with("state:crowbar_upgrade AND NOT run_list_map:pacemaker-cluster-member").
         and_return([])
       )
-      expect(subject.services).to eq([:ok, ""])
+      allow_any_instance_of(NodeObject).to receive(:ssh_cmd).with(
+        "/usr/sbin/crowbar-shutdown-services-before-upgrade.sh"
+      ).and_return(true)
+      allow_any_instance_of(NodeObject).to receive(:ssh_cmd).with(
+        "/usr/sbin/crowbar-shutdown-services-before-upgrade.sh"
+      ).and_return(true)
```
jsuchome: It would probably be better to differentiate between these 2 calls: both is executed on different set of nodes.

MaximilianMeister:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;8X7elkka&#x2F;82-cleanup-upgrade-related-mess"> Cleanup upgrade related mess</a>

crowbar_framework/spec/models/api/upgrade_spec.rb
MaximilianMeister: done in the latest commit

jsuchome: thanks!

aspiers: This is the kind of thing which ideally would belong in a separate PR - IIUC it's independent of the switch to class methods.

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -18,11 +18,11 @@
 
 module Api
   class Crowbar < Tableless
-    attr_reader :version
+    attr_reader :version, :addons
 
     def initialize
       @version = ENV["CROWBAR_VERSION"]
-      @addons = addons
+      @addons = features
```
aspiers: Why do this in `#initialize`?  If you're really worried about the calculation being repeated multiple times in a single request, then simply memoize the result the first time it's calculated, so that it's only calculated when it's really needed.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -48,7 +48,7 @@ def repocheck
       def target_platform(options = {})
         platform_exception = options.fetch(:platform_exception, nil)
 
-        case crowbar.version
+        case ENV["CROWBAR_VERSION"]
```
aspiers: Presumably this also eliminates the need to cache the value in `#initialize`?

aspiers:
BTW by "this removes DRY" I presume you mean "this obeys DRY and removes duplication"?  We don't want to remove the DRY rule, ever! ;-)

MaximilianMeister:
> So I would kindly ask, please keep commits as small as possible, and PRs as small as possible.

will do, I agree. If I would have had to review it, I would have run into the same issues you mentioned I guess

PR #675, opened 2016-09-22
api cleanup: move business logic into the model

jsuchome:
(cherry picked from commit f63124f35ab561de2f986f0de0fcd95f4d1e0e15)

backport of relavant parts of https://github.com/crowbar/crowbar-core/pull/666


abel-navarro:
I'm not familiar with this, but wouldn't we need the spec as well? https://github.com/crowbar/crowbar-core/pull/666/files#diff-e93b2c7e308af5036f6b27c7c4c7196eR148

jsuchome:
@abel-navarro the spec is there, I believe. I did not backport full PR though, second commit is only relevant to Cloud7

PR #676, opened 2016-09-23
Add ceph_client network (fate#319122)

tserong:
This adds a separate ceph client network, instead of having ceph
client traffic go over the admin network.

Note:

- The VLAN ID and IP address range were arbitrary choices.
- We might want to look at a different conduit, or at least this
  may be the right time to revisit a proposal for setting up bonds
  for ceph traffic (https://github.com/crowbar/crowbar-core/pull/64)
- I'm not sure what will happen during upgrades.  I guess I need
  to add a migration to make sure the new network gets added...?

### DIFF
--- a/chef/data_bags/crowbar/migrate/network/101_add_ceph_client_network.rb
+++ b/chef/data_bags/crowbar/migrate/network/101_add_ceph_client_network.rb
@@ -1,0 +1,11 @@
+def upgrade(ta, td, a, d)
+  a["networks"]["ceph_client"] = ta["networks"]["ceph_client"] unless a["networks"].key? "ceph_client"
+
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["networks"].delete "ceph_client" unless ta["networks"].key? "ceph_client"
+
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-network.json
+++ b/chef/data_bags/crowbar/template-network.json
@@ -161,6 +161,19 @@
             "host": { "start": "192.168.125.10", "end": "192.168.125.239" }
           }
         },
+        "ceph_client": {
+          "conduit": "intf1",
+          "vlan": 250,
+          "use_vlan": true,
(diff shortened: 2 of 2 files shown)


tserong:
This is used by https://github.com/crowbar/crowbar-ceph/pull/73

tserong:
Ping @mjura, @dirkmueller

tserong:
OK, I've added a migration, but... It doesn't do anything. If I take a crowbar install prior to this change, then install the new package, no new network is created. Nothing gets added to the network-config-default role, and I don't know where else to look.  Any and all help/direction/RTFMs appreciated.

(Sure, I can create the network manually by editing the default network proposal in raw mode, but, nobody's going to actually want to do that :-))

rhafer:
I guess what @vuntz wrote in https://github.com/crowbar/crowbar-core/pull/577#issuecomment-245866120 is valid here as well.

@tserong Regarding the failed migration are you sure, that he code actually ran? It should be possible to see that in ```/var/log/crowbar/component_install.log``` (you might need to run barclamp_install with ```--debug```.

tserong:
I'm not at all sure the migration code ran :-/ Shall investigate further tomorrow.

chef/data_bags/crowbar/migrate/network/101_add_ceph_client_network.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(ta, td, a, d)
+  a["networks"]["ceph_client"] = ta["networks"]["ceph_client"] unless a["networks"].key? "ceph_client"
```
mjura: Can you split this line? It is only small nitpick, but we would like to have Hound checks green.

mjura:
Besides Hound violation, it looks ok +1

dirkmueller:
as discussed, lets make this an attribute on the ceph barclamp directly, to avoid migration headaches (the separate network doesn't seem to be adding any additional security or performance as far as I can see).

PR #677, opened 2016-09-23
ohai: common ohai attribute for libvirt guest uuid

rhafer:
This adds a new attribute node[:crowbar_ohai][:libvirt][:guest_uuid]
which contains the UUID of the libvirt instance that the node is running
in. The purpose is to have a common place for this attribute independent
of the node's architecture (currently x86 ist storing this somewhere in
the "dmi" subtree and s390x in "s390x").

### DIFF
--- a/chef/cookbooks/ohai/files/default/plugins/linux/crowbar_libvirt.rb
+++ b/chef/cookbooks/ohai/files/default/plugins/linux/crowbar_libvirt.rb
@@ -1,0 +1,21 @@
+provides "crowbar_ohai"
+
+require_plugin "kernel"
+require_plugin "dmi"
+require_plugin "linux::s390x"
+
+libvirt_uuid = nil
+
+if kernel[:machine] == "s390x"
+  if s390x[:system][:manufacturer] == "KVM"
+    libvirt_uuid = s390x[:system][:uuid]
+  end
+else
+  manufacturer = dmi[:system] ? dmi[:system][:manufacturer] : "unknown"
+  if ["Bochs", "QEMU"].include? manufacturer
+    libvirt_uuid = dmi[:system][:uuid]
+  end
+end
+
+crowbar_ohai[:libvirt] = {}
+crowbar_ohai[:libvirt][:guest_uuid] = libvirt_uuid


rhafer:
This would go on top of:  https://github.com/crowbar/crowbar-core/pull/671

Alternatively, on s390x we could just write the attributes directly instead of copying them from the ```s390x``` subtree. This would remove the need for the s390x ohai plugin for now. But somehow I have the feeling we'll need that anyway for other things (e.g. the network barclamp is also using attributes from ```dmi``` in some cases).

chef/cookbooks/ohai/files/default/plugins/linux/crowbar_libvirt.rb
```diff
@@ -0,0 +1,23 @@
...
+require_plugin "linux::s390x"
+
+libvirt_uuid = nil
+
+if kernel[:machine] == "s390x"
+  if s390x[:system][:manufacturer] == "KVM"
+    libvirt_uuid = s390x[:system][:uuid]
+  end
+else
+  manufacturer = dmi[:system][:manufacturer] rescue "unknown"
```
rhafer: While I understand the reason behind Hound not liking this construct in general. I don't really see how it is bad in this special case.

Any hint how it could be done better is really appreciated. Otherwise I'd just keep it like it is currently.

aspiers: manufacturer = dmi[:system] ? dmi[:system][:manufacturer] : "unknown"

In this case I think I agree with Hound for the reasons stated in the link it gives.  Having said that, I just realised that I think we had decided to keep Hound as an optional status check, not a required one.

rhafer:
> ``` manufacturer = dmi[:system] ? dmi[:system][:manufacturer] : "unknown"```

@aspiers thanks, updated.

chef/cookbooks/ohai/files/default/plugins/linux/crowbar_libvirt.rb
```diff
@@ -0,0 +1,23 @@
...
+    libvirt_uuid = s390x[:system][:uuid]
+  end
+else
+  manufacturer = dmi[:system] ? dmi[:system][:manufacturer] : "unknown"
+  if ["Bochs", "QEMU"].include? manufacturer
+    libvirt_uuid = dmi[:system][:uuid]
+  end
+end
+
+unless libvirt_uuid.nil?
```
vuntz: Isn't it simpler to remove the condition? This will result in the guest_uuid attribute being nil when we're not virtualized, but that'd be reasonable, and would avoid checks to see if crowbar_ohai[:libvirt] is nil or not in other locations.

(no need to change if you're happy the current way)

rhafer: Yeah, makes sense. Updated the PRs.

rhafer:
@aspiers I addressed the HoundCI thing. Are you ok with the change now?

aspiers:
Looks like gating has failed once again due to [swift populate-dispersion timeout (Trello)](https://trello.com/c/ZepcSmJ4/223-swift-populate-dispersion-timeout).

PR #680, opened 2016-09-26
set bind directory to be readable by everybody (bsc#997530)

MaximilianMeister:
it has been readable before (see c955804f) and the current permissions
are breaking backup creation, as crowbar isn't allowed to read from
the directory

https://bugzilla.suse.com/show_bug.cgi?id=997530

### DIFF
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -43,7 +43,7 @@ end
 directory "/etc/bind" do
   owner "root"
   group bindgroup
-  mode 0770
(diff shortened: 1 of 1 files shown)


chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -43,7 +43,7 @@
 directory "/etc/bind" do
   owner "root"
   group bindgroup
-  mode 0770
+  mode 0774
```
bmwiedemann: are you sure that 774 is enough permission? I guess, 771 might work better.

PR #682, opened 2016-09-27
add api response examples to the api documentation

MaximilianMeister:
screenshot from 2016-09-27 08-43-34

### DIFF
--- a/crowbar_framework/app/controllers/api/backups_controller.rb
+++ b/crowbar_framework/app/controllers/api/backups_controller.rb
@@ -20,12 +20,36 @@ class Api::BackupsController < ApiController
 
   api :GET, "/api/crowbar/backups", "Returns a list of available backups"
   api_version "2.0"
+  example '
+  [
+    {
+      "id": 1,
+      "name": "testbackup",
+      "version": "4.0",
+      "size": 76815,
+      "created_at": "2016-09-27T06:05:10.208Z",
+      "updated_at": "2016-09-27T06:05:10.208Z",
+      "migration_level": 20160819142156
+    }
+  ]
+  '
   def index
     render json: Api::Backup.all
   end
 
   api :GET, "/api/crowbar/backups/:id", "Returns a specific backup"
   api_version "2.0"
+  example '
+  {
+    "id": 1,
+    "name": "testbackup",
+    "version": "4.0",
+    "size": 76815,
+    "created_at": "2016-09-27T06:05:10.208Z",
+    "updated_at": "2016-09-27T06:05:10.208Z",
+    "migration_level": 20160819142156
+  }
+  '
   def show
     render json: @backup
   end
@@ -35,6 +59,17 @@ class Api::BackupsController < ApiController
   param :backup, Hash, desc: "Backup info", required: true do
     param :name, String, desc: "Name of the backup", required: true
   end
+  example '
+  {
+    "id": 1,
+    "name": "testbackup",
+    "version": "4.0",
+    "size": 76815,
+    "created_at": "2016-09-27T06:05:10.208Z",
+    "updated_at": "2016-09-27T06:05:10.208Z",
+    "migration_level": 20160819142156
+  }
+  '
   def create
     @backup = Api::Backup.new(backup_params)
 
--- a/crowbar_framework/app/controllers/api/crowbar_controller.rb
+++ b/crowbar_framework/app/controllers/api/crowbar_controller.rb
@@ -17,6 +17,15 @@
 class Api::CrowbarController < ApiController
   api :GET, "/api/crowbar", "Show the crowbar object"
   api_version "2.0"
+  example '
+  {
+    "version": "4.0",
+    "addons": [
+      "ceph",
+      "ha"
+    ]
+  }
+  '
   def show
     render json: Api::Crowbar.status
   end
@@ -28,6 +37,20 @@ class Api::CrowbarController < ApiController
   end
 
   api :GET, "/api/crowbar/upgrade", "Status of Crowbar Upgrade"
+  example '
+  {
+    "version": "4.0",
+    "addons": [
+      "ceph",
+      "ha"
+    ],
+    "upgrade": {
+      "upgrading": false, # the crowbar admin server is currently upgrading
+      "success": false, # the crowbar admin server has been successfully upgraded
+      "failed": false # the crowbar admin server failed to upgrade
+    }
+  }
+  '
   api :POST, "/api/crowbar/upgrade", "Upgrade Crowbar"
   api_version "2.0"
   def upgrade
@@ -54,6 +77,24 @@ class Api::CrowbarController < ApiController
 
   api :GET, "/api/crowbar/repocheck", "Sanity check for Crowbar server repositories"
   api_version "2.0"
+  example '
+  {
+    "os": {
+      "available": true,
+      "repos": {}
+    },
+    "cloud": {
+      "available": false,
+      "repos": {
+        "x86_64": {
+          "missing": [
+            "SUSE OpenStack Cloud 7"
+          ]
+        }
+      }
+    }
+  }
+  '
   def repocheck
     check = Api::Crowbar.repocheck
 
--- a/crowbar_framework/app/controllers/api/errors_controller.rb
+++ b/crowbar_framework/app/controllers/api/errors_controller.rb
@@ -20,6 +20,20 @@ class Api::ErrorsController < ApiController
 
   api :GET, "/api/errors", "Show a list of errors"
   api_version "2.0"
+  example '
+  [
+    {
+      "id": 1,
+      "error": "StandardError",
+      "message": "This is an error",
+      "code": 500,
+      "caller": "apply_role",
+      "backtrace": "",
+      "created_at": "2016-09-13T08:04:24.128Z",
+      "updated_at": "2016-09-13T08:04:24.128Z"
+    }
+  ]
+  '
   def index
     render json: Api::Error.all
   end
@@ -27,6 +41,18 @@ class Api::ErrorsController < ApiController
   api :GET, "/api/errors/:id", "Show a specific error"
   param :id, Integer, desc: "Error ID", required: true
   api_version "2.0"
+  example '
+  {
+    "id": 1,
+    "error": "StandardError",
+    "message": "This is an error",
+    "code": null,
+    "caller": null,
+    "backtrace": "",
+    "created_at": "2016-09-13T08:04:24.128Z",
+    "updated_at": "2016-09-13T08:04:24.128Z"
+  }
+  '
   def show
     render json: @error
   end
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -17,6 +17,29 @@
 class Api::UpgradeController < ApiController
   api :GET, "/api/upgrade", "Show the Upgrade status object"
   api_version "2.0"
+  example '
+  {
+    "crowbar": {
+      "version": "4.0",
+      "addons": [
+        "ceph",
+        "ha"
+      ],
+      "upgrade": {
+        "upgrading": false,
+        "success": false,
+        "failed": false
+      }
+    },
+    "checks": {
+      "sanity_checks": true,
+      "maintenance_updates_installed": true,
+      "clusters_healthy": true,
+      "compute_resources_available": true,
+      "ceph_healthy": true
+    }
+  }
+  '
   def show
     render json: Api::Upgrade.status
   end
@@ -64,6 +87,15 @@ class Api::UpgradeController < ApiController
 
   api :GET, "/api/upgrade/prechecks", "Shows a sanity check in preparation for the upgrade"
   api_version "2.0"
+  example '
+  {
+    "sanity_checks": true,
+    "maintenance_updates_installed": true,
+    "clusters_healthy": true,
+    "compute_resources_available": true,
+    "ceph_healthy": true
+  }
+  '
   def prechecks
     render json: Api::Upgrade.check
   end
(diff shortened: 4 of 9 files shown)


MaximilianMeister:
[(5) #upgrade Synchronize API responses with the Angular app](https://trello.com/c/3TALQwDA/121-5-upgrade-synchronize-api-responses-with-the-angular-app)

crowbar_framework/app/controllers/repositories_controller.rb
```diff
@@ -135,11 +158,7 @@ def activate_all
     end
   end
 
-  #
-  # Deactivate all Repositories
-  #
-  # Provides the restful api call for
-  # Deactivate all Repositories   /utils/repositories/deactivate_all   POST   Destroys Repository DataBagItem
+  api :POST, "/utils/repositories/deactivate_all", "Dectivate all repositories. Destroys DataBagItems"
```
MaximilianMeister: fixed

crowbar_framework/app/controllers/repositories_controller.rb
```diff
@@ -93,12 +122,10 @@ def activate
     end
   end
 
-  #
-  # Deactivate a single Repository
-  #
-  # Provides the restful api call for
-  # Deactivate a Repository   /utils/repositories/deactivate   POST   Destroys Repository DataBagItem
-  # required parameters: platform, arch, repo
+  api :POST, "/utils/repositories/deactivate", "Deactivate a single repository. Destroys a DataBagItem"
```
MaximilianMeister: fixed

crowbar_framework/app/controllers/api/crowbar_controller.rb
```diff
@@ -54,6 +77,24 @@ def maintenance
 
   api :GET, "/api/crowbar/repocheck", "Sanity check for Crowbar server repositories"
   api_version "2.0"
+  example '
```
jsuchome: Have we already agreed on the format of the result here?

MaximilianMeister: no, but do we need to show something else besides the json response?

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -255,6 +437,63 @@ def proposals
   api :GET, "/crowbar/:barclamp/1.0/proposals/template",
     "Returns the content of a proposal template"
   param :barclamp, String, desc: "Name of the barclamp", required: true
+  example '
+  {
+    "id": "template-dns",
```
santiph: A proposal template for each type of barclamp supported must be documented

santiph:
Apipie Headers example: http://stackoverflow.com/questions/27028393/how-to-generate-header-for-apipie-rails-gem
https://github.com/Apipie/apipie-rails/issues/340
https://github.com/Apipie/apipie-rails/pull/341

crowbar_framework/app/controllers/api/backups_controller.rb
```diff
@@ -20,12 +20,36 @@ class Api::BackupsController < ApiController
...
+  ]
+  '
   def index
     render json: Api::Backup.all
   end
 
   api :GET, "/api/crowbar/backups/:id", "Returns a specific backup"
   api_version "2.0"
+  example '
+  {
```
MaximilianMeister: with the download endpoint

crowbar_framework/app/controllers/api/backups_controller.rb
```diff
@@ -35,6 +59,17 @@ def show
...
+    "id": 1,
+    "name": "testbackup",
+    "version": "4.0",
+    "size": 76815,
+    "created_at": "2016-09-27T06:05:10.208Z",
+    "updated_at": "2016-09-27T06:05:10.208Z",
+    "migration_level": 20160819142156
+  }
```
MaximilianMeister: why that? you mean it should be downloaded right after creation? I think that would be wrong

crowbar_framework/app/controllers/api/crowbar_controller.rb
```diff
@@ -30,6 +39,20 @@ def update
...
   api :GET, "/api/crowbar/upgrade", "Status of Crowbar Upgrade"
+  example '
+  {
+    "version": "4.0",
+    "addons": [
+      "ceph",
+      "ha"
+    ],
+    "upgrade": {
+      "upgrading": false,
```
MaximilianMeister: ok, I can add a comment next to it

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -19,6 +19,29 @@ class Api::UpgradeController < ApiController
 
   api :GET, "/api/upgrade", "Show the Upgrade status object"
```
MaximilianMeister: just look at the example :) `/api/crowbar/upgrade` doesn't include the checks

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -40,6 +40,39 @@ def controller_to_barclamp
 
   add_help(:barclamp_index)
   api :GET, "/crowbar", "Returns a list of string names and descriptions for all barclamps"
+  example '
```
MaximilianMeister: no. why do you think that?

crowbar_framework/app/controllers/barclamp_controller.rb
MaximilianMeister: that would be too much IMO, as this is just an example. does it really need to be that detailed? It would significantly bloat the LOC of the barclamp_controller

crowbar_framework/app/controllers/api/errors_controller.rb
```diff
@@ -20,13 +20,39 @@ class Api::ErrorsController < ApiController
 
   api :GET, "/api/errors", "Show a list of errors"
   api_version "2.0"
+  example '
+  [
+    {
+      "id": 1,
+      "error": "StandardError",
+      "message": "This is an error",
+      "code": null,
```
MaximilianMeister: > Apipie Headers example: http://stackoverflow.com/questions/27028393/how-to-generate-header-for-apipie-rails-gem
> Apipie/apipie-rails#340
> Apipie/apipie-rails#341

that would be done in another PR

MaximilianMeister:
> Apipie Headers example: http://stackoverflow.com/questions/27028393/how-to-generate-header-for-apipie-rails-gem
Apipie/apipie-rails#340
Apipie/apipie-rails#341

that would be done in another PR

crowbar_framework/app/controllers/api/errors_controller.rb
MaximilianMeister: added an example where you can see the datatype

crowbar_framework/app/controllers/api/crowbar_controller.rb
```diff
@@ -19,6 +19,15 @@ class Api::CrowbarController < ApiController
 
   api :GET, "/api/crowbar", "Show the crowbar object"
   api_version "2.0"
+  example '
+  {
+    "version": "4.0",
```
MaximilianMeister: cloud 6 is crowbar 3.0, cloud 7 is crowbar 4.0

santiph: Please Add that as part of the documentation

crowbar_framework/app/controllers/barclamp_controller.rb
santiph: I'd assume barclamp is an attribute of the crowbar entity, different than Version, for instance

crowbar_framework/app/controllers/barclamp_controller.rb
santiph: You cannot provide information through the API that is not being documented.
In other words, you cannot expect attributes or fields are being consumed by a client if they're not documented somewhere.

MaximilianMeister: in an ideal world I agree, that'd be super useful. but this would be an enormous amount of documentation, just imagine a proposal alone, and then repository metadata etc. we don't have the time to do that now

crowbar_framework/app/controllers/api/crowbar_controller.rb
MaximilianMeister: where? as a comment on that line it would be probably too long

crowbar_framework/app/controllers/barclamp_controller.rb
MaximilianMeister: this is the v1 api, I think you are referring to v2 `/api/crowbar`

crowbar_framework/app/controllers/barclamp_controller.rb
santiph: You can create tech debts to track that work separately and even in a different sprint, as long as PO (Vincent) is ok with it. I'm only raising issues as I see them.

crowbar_framework/app/controllers/api/upgrade_controller.rb
santiph: So `/api/upgrade` also contains `/api/crowbar/upgrade` results? Could you add that as a comment in the example?

crowbar_framework/app/controllers/barclamp_controller.rb
santiph: Good point. Thanks

crowbar_framework/app/controllers/api/crowbar_controller.rb
santiph: I'd prefer a long line in the example than a comment on a github review ;)
You can also add those lines outside the json example and make it multiline... 

I'm reading Apipie issues and it seems there is no better approach to represent the responses other than examples. A lot of people reference to swagger.io or raml.org, claiming they both support it...

crowbar_framework/app/controllers/api/backups_controller.rb
santiph: Well, I guess we can use the id to download the backup file. Although I would expect an href > download > link attribute to get it from this response

crowbar_framework/app/controllers/barclamp_controller.rb
santiph: well... This is not affecting our work right now, so I can still approve it if a tech debt to track this work is created and linked

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: just ignore it for now, `/api/upgrade` will change this sprint anyways, as soon as we start working on https://trello.com/c/OWehKX6K/126-5-implement-a-state-machine-for-the-upgrade

crowbar_framework/app/controllers/barclamp_controller.rb
MaximilianMeister: as I said there will be follow ups, I really want to avoid those huge PR's. but as this is a significant amount of work which is not necessary for the upgrade process, we have to postpone it.

the card to track this is https://trello.com/c/gs8pbS7M/1170-enhance-api-documentation

crowbar_framework/app/controllers/repositories_controller.rb
```diff
@@ -64,12 +95,10 @@ def sync
...
-  #
-  # Activate a single Repository
-  #
-  # Provides the restful api call for
-  # Activate a Repository   /utils/repositories/activate   POST  Creates Repository DataBagItem
-  # required parameters: platform, arch, repo
+  api :POST, "/utils/repositories/activate", "Activate a single repository. Creates a DataBagItem"
+  param :platform, String, desc: "Platform of the repository", required: true
+  param :arch, String, desc: "Architecture of the repository", required: true
+  param :repo, String, desc: "Name of the repository", required: true
```
MaximilianMeister: there is no response here in case of success, only 200 OK (like i think in all POST requests in crowbar, except backup create)

MaximilianMeister:
@santiph some comments are really implementation specific, like the backup create response, or would require changes there. this PR is really just to define the status quo in terms of documentation.

for the inline comments bit, I am not sure if that doesn't clutter the response examples too much. we may add some extra info below the json response, but to say it again, lets keep things simple and incrementally open smaller PR's to make the reviewers easier

crowbar_framework/app/controllers/api/crowbar_controller.rb
MaximilianMeister: @jsuchome @santiph as mentioned below, we may have to define some view to show extra information about response parameters that are unclear. but please in a separate small PR

vuntz:
@santiph let's not use this pull request as a way to change the APIs, as otherwise, this will never get merged (and tracking this here is a very good way to make sure that we will forget about some requested changes).

This pull request is just about documenting better the current behavior of the APIs, and once we have that, we can have some good discussion about what should be changed and how.

crowbar_framework/app/controllers/api/upgrade_controller.rb
santiph: Fair enough

crowbar_framework/app/controllers/repositories_controller.rb
santiph: HTTP Codes returned needs also to be documented. Later on, you can also use http codes for specific backend errors, when described

crowbar_framework/app/controllers/api/crowbar_controller.rb
aspiers: I agree with @santiph. Github PR comments should never be considered as documentation ...

crowbar_framework/app/controllers/repositories_controller.rb
MaximilianMeister: done now in https://github.com/crowbar/crowbar-core/pull/712

crowbar_framework/app/controllers/api/crowbar_controller.rb
MaximilianMeister: @aspiers me too :) not sure who would not ... :/

PR #683, opened 2016-09-27
[stable/3.0] partially backport #674

MaximilianMeister:
https://github.com/crowbar/crowbar-core/pull/674

### DIFF
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -15,10 +15,8 @@
 #
 
 class Api::UpgradeController < ApiController
-  before_action :set_upgrade
-
   def show
-    render json: @upgrade.status
+    render json: Api::Upgrade.status
   end
 
   def update
@@ -46,29 +44,17 @@ class Api::UpgradeController < ApiController
     end
   end
 
-  def services
-    if request.post?
-      head :not_implemented
-    else
-      render json: [], status: :not_implemented
-    end
-  end
-
   def prechecks
-    render json: @upgrade.check
+    render json: Api::Upgrade.check
   end
 
   def cancel
-    if @upgrade.cancel
+    cancel_upgrade = Api::Upgrade.cancel
+
+    if cancel_upgrade[:status] == :ok
       head :ok
     else
-      render json: { error: @upgrade.errors.full_messages.first }, status: :unprocessable_entity
+      render json: { error: cancel_upgrade[:message] }, status: cancel_upgrade[:status]
     end
-  end
-
-  protected
-
-  def set_upgrade
-    @upgrade = Api::Upgrade.new
   end
 end
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -18,15 +18,17 @@ require "open3"
 
 module Api
   class Crowbar < Tableless
-    attr_reader :version
+    attr_reader :version, :addons
 
     def initialize
       @version = ENV["CROWBAR_VERSION"]
+      @addons = features
     end
 
     def status
       {
-        version: @version
+        version: @version,
+        addons: @addons  
       }
     end
 
@@ -214,6 +216,28 @@ module Api
 
     def upgrade_script_path
       Rails.root.join("..", "bin", "upgrade_admin_server.sh")
+    end
+
+    def features
+      [].tap do |list|
+        ["ceph", "ha"].each do |addon|
+          list.push(addon) if addon_installed?(addon)
+        end
+      end
+    end
+
+    def addon_installed?(addon)
+      case addon
+      when "ceph"
+        CephService
+      when "ha"
+        PacemakerService
+      else
+        return false
+      end
+      true
+    rescue NameError
+      false
     end
 
     def repo_version_available?(products, product, version)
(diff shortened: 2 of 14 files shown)


crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -46,29 +44,17 @@ def prepare
     end
   end
 
-  def services
```
aspiers: I'm confused why this (and its corresponding tests) was removed in this commit?  Is it really related to the switch to class methods?

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -46,29 +44,17 @@ def prepare
...
 
   def cancel
-    if @upgrade.cancel
+    cancel_upgrade = Api::Upgrade.cancel
+
+    if cancel_upgrade[:status] == :ok
       head :ok
     else
-      render json: { error: @upgrade.errors.full_messages.first }, status: :unprocessable_entity
+      render json: { error: cancel_upgrade[:message] }, status: cancel_upgrade[:status]
```
aspiers: Is the change from `:unprocessable_entity` really related to the switch to class methods?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -18,59 +18,66 @@
...
-      service_object.revert_nodes_from_crowbar_upgrade
+      def cancel
+        service_object = CrowbarService.new(Rails.logger)
+        service_object.revert_nodes_from_crowbar_upgrade
 
-      true
-    rescue => e
-      errors.add(:base, e.message)
-      Rails.logger.error(e.message)
+        {
```
aspiers: It's not immediately obvious to me why you are changing the return value here, and the commit message doesn't explain.  It feels like maybe you are combining two logical commits into a single commit without any explanation, which makes it quite a bit harder to review.

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -18,15 +18,17 @@
 
 module Api
   class Crowbar < Tableless
-    attr_reader :version
+    attr_reader :version, :addons
```
aspiers: [I already queried this](https://github.com/crowbar/crowbar-core/pull/674/commits/72a52c584d5a3f12fd93a322793b4a97ed6639e6#r80055610) in the PR from which this is backporting stuff.  What's the point of having both `#addons` and `#features`, given that they return the same value?  The downside is a more confusing interface, and the only benefit I can see is that `@addons` memoizes the result of `#features`, so that it only executes it once, but then why not memoize on demand, instead of in the constructor, so that `#features` is only called if it's really needed?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -55,28 +55,32 @@ def cancel
 
       protected
 
+      def crowbar
+        Api::Crowbar.new
```
aspiers: But this gets removed again in the next commit which turns `Api::Crowbar` methods into class methods, so why is it being added?  I guess the answer is "because it's a backport of #674 which already did it that way"?  If so, OK, but please try to avoid that kind of thing in future, since it's pretty confusing to review a PR which makes some changes and then fully or partially reverts them in a later commit in the same PR.

Personally I usually end up doing quite a lot of fixups/squashing on my private branches before submitting PRs, to try to avoid this kind of thing.  Yes, it's more work, but a lot of it can be automated with the right git tools, and keeping the history really clean helps reviews and backports, and makes it easier for anyone reading the history later on to understand the changes.

Hope that makes sense :-)

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -18,236 +18,234 @@
...
-      founders.each do |n|
-        name = n.name
-        ssh_retval = n.run_ssh_cmd("crm status 2>&1")
-        if ssh_retval[:exit_code] != 0
-          crm_failures[name] = ssh_retval[:stdout]
-          Rails.logger.warn(
-            "crm status at node #{name} reports error:\n#{ssh_retval[:stdout]}"
-          )
-          next
+      def ceph_healthy?
```
aspiers: Looks like you reordered some methods, e.g. `ceph_healthy?` got moved before `compute_resources_available?`.  That's absolutely fine (I'm a fan of putting methods in a logical order), but it's a refactoring which should happen in a separate commit saying "reorder methods with no other changes", so that reviewers know (and can easily check if they want) that it is simply a method reordering, with no functional changes.  When it's mixed with other changes, it makes the review much harder, because each change can't be viewed atomically, and then it's easier to miss something in the review.

I know this is just a cherry-pick - sorry for not mentioning this earlier in the original review in #674. But I hope it's still worth mentioning here as something to bear in mind for the future.

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: stopping services is not needed on cloud 6 (which was not that obvious from the commit message I admit)

crowbar_framework/app/models/api/crowbar.rb
MaximilianMeister: this is no longer relevant, as it got dropped in the switch to class methods (should have been squashed somehow)

PR #663, opened 2016-09-14
Optional include the crowbar-ui config

rsalevsky:

PR #684, opened 2016-09-27
Cleanup apache conf

rsalevsky:
This change enables crowbar-ui while crowbar-init is running.

### DIFF
--- a/configs/crowbar.conf
+++ b/configs/crowbar.conf
@@ -1,8 +1,8 @@
 <VirtualHost *:80>
     DocumentRoot /opt/dell/crowbar_framework/public
 
-    ErrorLog /var/log/apache2/crowbar-error_log
-    CustomLog /var/log/apache2/crowbar-access_log combined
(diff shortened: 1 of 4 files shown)


configs/crowbar.conf
```diff
@@ -9,5 +9,10 @@
     ErrorDocument 422 /422.html
     ErrorDocument 500 /500.html
 
+    IncludeOptional /etc/apache2/conf.d/crowbar-ui.conf.partia[l]
```
nicolasbock: Why the pattern matching expression `partia[l]`?

rsalevsky: Otherwise the path is not allowed to be empty,  IncludeOptional is unfortunately not enough in this case. See https://github.com/crowbar/crowbar-core/pull/663

nicolasbock: Thanks! It's a bit strange that we have to add it, but oh well....

PR #687, opened 2016-09-29
document node show api

MaximilianMeister:
that has been overlooked somehow

### DIFF
--- a/crowbar_framework/app/controllers/nodes_controller.rb
+++ b/crowbar_framework/app/controllers/nodes_controller.rb
@@ -393,8 +393,47 @@ class NodesController < ApplicationController
     end
   end
 
-  # GET /nodes/1
-  # GET /nodes/1.xml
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/controllers/nodes_controller.rb
```diff
@@ -393,8 +393,47 @@ def hit
     end
   end
 
-  # GET /nodes/1
-  # GET /nodes/1.xml
+  api :GET, "/nodes/:id", "Show details of a node"
+  header "Accept", "application/json", required: true
+  param :id, String, desc: "Node name", required: true
+  example '
+  {
```
rsalevsky: Did you check if we can move this to a separate file?

MaximilianMeister: https://github.com/crowbar/crowbar-core/issues/690

PR #688, opened 2016-09-29
provisioner: Fix adding missing repos on nodes

vuntz:
"zypper --non-interactive --xmlout repos $non-existing-repo" of course
doesn't contain the attributes that we were expecting.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/base.rb
+++ b/chef/cookbooks/provisioner/recipes/base.rb
@@ -308,8 +308,10 @@ if node[:platform_family] == "suse" && !node.roles.include?("provisioner-server"
         `LANG=C zypper --non-interactive --xmlout repos #{name} 2> /dev/null`
       ).root.elements
 
-      current_priority = out["repo-list/repo"].attributes["priority"].to_i
-      current_url = out["repo-list/repo/url"].text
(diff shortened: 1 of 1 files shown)


chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -308,8 +308,13 @@
         `LANG=C zypper --non-interactive --xmlout repos #{name} 2> /dev/null`
       ).root.elements
 
-      current_priority = out["repo-list/repo"].attributes["priority"].to_i
-      current_url = out["repo-list/repo/url"].text
+      if out["repo-list/repo"].nil?
+        current_priority = nil
+        current_url = nil
```
MaximilianMeister: they are already set to nil some lines above

vuntz: fixed

vuntz:
> not 100% sure, but in which case are the repos missing? wouldn't that be a bug in the provisioner library (Provisioner::Repositories.get_repos)?

That happens if a user enables a repo in the webui after installing the node.

MaximilianMeister:
> That happens if a user enables a repo in the webui after installing the node.

hmm. not having a deeper understanding why, this change is good anyways. thanks

PR #689, opened 2016-09-29
[stable/3.0] Simple check for presence of HA-based configuration

jsuchome:
Adapted testsuite to addition of ha_deployed? check

(cherry picked from commit 0069bff189e2647ea6eaac281a56946c16330c52)

We should probably merge this one only after the **backport** for https://github.com/crowbar/crowbar-core/pull/672 is merged

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -201,6 +201,22 @@ module Api
         true
       end
 
+      # Check for presence of HA setup, which is a requirement for non-disruptive upgrade
+      def ha_presence_check
+        ret = {}
+        unless addon_installed? "ha"
+          ret[:errors] = [I18n.t("api.crowbar.ha_not_installed")]
+          return ret
+        end
+
+        founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
+
+        if founders.empty?
+          ret[:errors] = [I18n.t("api.crowbar.ha_not_configured")]
+        end
+        ret
+      end
+
       protected
 
       def lib_path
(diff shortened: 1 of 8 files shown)


crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -201,6 +201,22 @@ def compute_resources_available?
         true
       end
 
+      # Check for presence of HA setup, which is a requirement for non-disruptive upgrade
+      def ha_presence_check
```
rsalevsky: The name is a little bit confusing. I would name it `ha_present?` and return a true or false value.

jsuchome: It's backport from master. And we want to return specific errors somehow, not just true/false

rsalevsky: This makes it much more confusing then... Why don't you use the errors hash and return false in this function? At least this is the standard way in rails to do such stuff.

MaximilianMeister: we don't have any crowbar instance anymore, just class methods. so the errors hash is not an option here

aspiers:
Is this a backport of #670?  If so let's close it.

jsuchome:
It is a backport of https://github.com/crowbar/crowbar-core/pull/670, but I think we want it here. AFAIK `api/pacemaker` is not ported to 3.0 branch

crowbar_framework/app/models/api/crowbar.rb
jsuchome: @rsalevsky if you want to discuss how the methods should look like and what kind of return values should they return, please comment rather https://github.com/crowbar/crowbar-core/pull/672 . I'll adapt this one according to changes in 672 once it is merged

jsuchome:
This got too messy meanwhile, replacing by new PR: https://github.com/crowbar/crowbar-core/pull/751

PR #691, opened 2016-09-29
bsc#999637: Remove firewall sanity check

nicolasbock:
Commit
https://github.com/crowbar/crowbar-core/commit/73a0ce5ff0cd335c51e0bae2175c2d8c675f97fa
exposed an inconsistency between the firewall rules needed for the bmc-nat
barclamp and crowbar's firewall sanity check. The sanity check is too broad as
it does not tolerate _any_ iptables rule. A more sensible sanity check would
need to parse the firewall rules more intelligently to ensure that the admin
node is accessible and yet allow for rules added by individual barclamps.

In the interest of addressing https://bugzilla.suse.com/show_bug.cgi?id=999637
I am removing the sanity check for now.

### DIFF
--- a/crowbar_framework/config/locales/crowbar/en.yml
+++ b/crowbar_framework/config/locales/crowbar/en.yml
@@ -761,7 +761,6 @@ en:
       ip_resolved: 'Could not resolve %{fqdn} to an IPv4 or IPv6 address.'
       loopback_unresolved: '%{fqdn} resolves to a loopback address.'
       ip_configured: 'No local interface is configured with a correct IP address.'
-      firewall_disabled: 'Firewall is not completely disabled.'
       ping_succeeds: 'Failed to ping %{fqdn}; please check your network configuration.'
     check:
       cache_error: 'Failed to cache sanity checks'
--- a/crowbar_framework/lib/crowbar/checks/network.rb
+++ b/crowbar_framework/lib/crowbar/checks/network.rb
@@ -67,10 +67,6 @@ module Crowbar
         true
       end
 
-      def firewall_disabled?
-        !system("sudo LANG=C iptables -n -L | grep -qvE '^$|^Chain [^ ]|^target     prot'")
-      end
-
       def ping_succeeds?
         system("ping -c 1 #{fqdn} > /dev/null 2>&1")
       end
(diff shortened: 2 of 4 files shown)


MaximilianMeister:
can't we just adapt the check or ignore the new bmc-nat rule instead of deleting the whole check?

nicolasbock:
@MaximilianMeister We could but that check would be a bit more complicated. We could have to test for the exact same rules the bmc-nat barclamp is adding. And if we every add more rules in other barclamps we need to keep the sanity check up to date, duplicating the rule set every time we do that.

rsalevsky:
There was a reason why we have this check, but I forgot the exact details. Maybe @vuntz can explain why we have this check and if we still need it.

nicolasbock:
@rsalevsky I mumbled with @vuntz and the reason for this check was to ensure that the admin node is accessible and not blocked by a local firewall rule. But to make sure the it is, we need to take into account the firewall rules added by barclamps.

MaximilianMeister:
I dug out https://github.com/crowbar/crowbar/commit/15bae010ec2db25b40249ea5bb174841d2fafc5f

MaximilianMeister:
maybe we could just check for SuSEfirewall2 instead?

nicolasbock:
@MaximilianMeister Not a bad idea. Let me look into it.

vuntz:
> maybe we could just check for SuSEfirewall2 instead?

Sorry for the stupid question, but how does that help? I guess I'm missing something :-)

nicolasbock:
> maybe we could just check for SuSEfirewall2 instead?

I looked at the output of SuSEfirewall2. As far as I can tell it doesn't support a command or simple output that would allow us to test whether the firewall allows inbound access to the admin node. We would still have to parse the iptables rules.

MaximilianMeister:
> Sorry for the stupid question, but how does that help? I guess I'm missing something :-)

if you have the barclamps rules setup, and you execute `SuSEfirewall2 stop`, and it doesn't kill those wanted barclamp rules, then it would help to kill all other unwanted rules that are set up on the system. but i am not sure if that is the case, or if `SuSEfirewall2 stop` would just kill all the rules.

nicolasbock:
We could also write a helper function to manipulate the firewall. The helper function could register each rule and the crowbar sanity check would then call the helper function to verify that the registered rules weren't modified.

aspiers:
Sorry, I meant installation process not upgrade process.

nicolasbock:
> Is there something we can document about the [installation] process, in lieu of this check?

Aren't we doing that already? In the SOC 6 deployment manual for instance, we emphasize that the firewall has to be turned off (Sec. 3.5.2 and again in Sec. 6).

PR #692, opened 2016-09-29
Library for tracking the upgrade state

jsuchome:
We need to track the upgrade process so we know what is the current step/substep and return such kind of information to the called.

The progress is saved in the file so it is usable also by crowbar-init.

Example usage:
```
u = Upgrade.new
if u.current_step == "nodes_services"
  u.start_step
  services = go and execute the services step, return success/failure and the map with errors
  u.end_step(services[:ret], services[:errors])
end

```

- Backport to 3.0 is needed
- Requires https://github.com/crowbar/crowbar-core/pull/719

---
This change is [Reviewable](https://reviewable.io/reviews/crowbar/crowbar-core/692)


abel-navarro:
I think the idea is good but I see different pieces not fitting together. These are some things that I thing would need to be addressed:

#### Store state info per node or global?

Right now the state is saved in crowbar for all nodes, so it's global. It's more "in which state is crowbar" rather than "in which state is this node".

When you defined the `current_node` field I think you are designing on a upgrade done one node by one. This may be true for controllers, but we may want to do parallel upgrades later on, so this field won't apply.

#### Linear states

We are not jumping back and forth the states. Instead this is a well defined staged process where first goes A then B and then C. In such case you may not need to save the `next_step` on the state itself but can be global too.

#### Substeps

Are substeps needed? for me it would be simpler if we just have `step` and maybe identifying the substeps in the `step` string. I.e. "admin_repo_check_1" and "admin_repo_check_2".

#### Finished should be a step

#### Starting state

There's missing initialization of the starting state.

jsuchome:
Thanks @abel-navarro for review! Some comments:

- Store state info per node or global?

My idea is really the global state of the upgrade process, not of a node. As for 6-7 upgrade, we're definitely doing node-by-node upgrade so "current_node" or something similar should be sufficient I think.

- Not saving next step

You're right, if we provide some global list of steps (that is fixed) we should not need to include "next_step" in the status output.

- Substeps

This comes from the current implementation proposal. As steps, we have those shown in https://github.com/crowbar/crowbar-ui/wiki/Index-Upgrade67 Most of them do not have substeps, but the most important has it. The step is Upgrading Nodes (initiated by one API call) which has several substeps like upgrading controllers, migrating instances and upgrading compute nodes. So I think substeps make sense

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -109,6 +136,10 @@ def cancel
 
       protected
 
+      def progress_file_path
+        "/var/lib/crowbar/upgrade/progress.json"
```
MaximilianMeister:
```ruby
Pathname.new("/var/lib/crowbar/upgrade/progress.json")
```

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -26,6 +26,33 @@ def status
...
+        }
+        progress_file = progress_file_path
+        if File.exist? progress_file
+          upgrade_progress = JSON.load(File.read(progress_file))
+        end
+        upgrade_progress
+      end
+
+      def save_upgrade_progress(progress)
+        File.open(progress_file_path, "w") do |f|
```
MaximilianMeister:
```ruby
progress_file_path.open("w") do |f|
  f.write(JSON.pretty_generate(progress))
end
```

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -26,6 +26,33 @@ def status
...
+      def progress
+        upgrade_progress = {
+          finished: false,
+          current_step: "prechecks",
+          next_step: "admin_repo_check",
+          substep: "",
+          current_node: ""
+        }
+        progress_file = progress_file_path
+        if File.exist? progress_file
```
MaximilianMeister:
```ruby
if progress_file_path.exist?
```

jsuchome:
@abel-navarro I've incorporated some of your points.

So, the progress json scructure currently reflects only current state of upgrade process. @MaximilianMeister suggested that it should also cover state of previous steps, like in https://gist.github.com/MaximilianMeister/63de56a192b7eaaca4e386b918409291 .

I think it does not make sense to divide the steps into crowbar/nodes subgroup, but somehow we should some more information (as discussed elsewhere, we need to know e.g. which nodes have been upgraded already)

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -55,6 +55,59 @@ def supported?
...
 
+    class << self
+      def progress
+        upgrade_progress = {
+          current_step: "prechecks",
+          substep: "",
+          current_node: ""
+        }
+        if progress_file_path.exist?
+          upgrade_progress = JSON.load(File.read(progress_file_path))
```
MaximilianMeister: `progress_file_path.read`

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -55,6 +55,59 @@ def supported?
...
+
+      def save_current_step(step)
+        upgrade_progress = progress
+        upgrade_progress["current_step"] = step
+        save_upgrade_progress(upgrade_progress)
+      end
+
+      def finished?
+        upgrade_progress = progress
+        upgrade_progress["current_step"] == "finished"
```
MaximilianMeister: `progress["current_step"] == "finished"` should be enough here

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -55,6 +55,59 @@ def supported?
...
+      end
+
+      protected
+
+      # global list of the steps of the upgrade process
+      def upgrade_steps
```
MaximilianMeister: maybe we prefix the methods like the other ones? like `upgrade_steps_6_7`

MaximilianMeister:
@jsuchome sorry for not finding out earlier, but we may move this away from `lib/crowbar/upgrade.rb` to `lib/upgrade.rb` as it is not really only a crowbar context (it is the whole cloud upgrade). that would also free us also from squeezing it into the existing lib... what do you think?

crowbar_framework/lib/upgrade.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    def save_current_step(step)
+      upgrade_progress = progress
+      upgrade_progress["current_step"] = step
+      save_upgrade_progress(upgrade_progress)
+    end
+
+    def finished?
+      progress["current_step"] == "finished"
```
MaximilianMeister: `progress["current_step"] == upgrade_steps_6_7.last`

crowbar_framework/lib/upgrade.rb
```diff
@@ -0,0 +1,58 @@
+class Upgrade
+  class << self
+    # Return the current state of upgrade process.
+    # We're keeping the information in the file so is accessible by other applications
+    # and different crowbar versions.
+    def progress
+      upgrade_progress = {
+        current_step: "prechecks",
```
MaximilianMeister: I guess it should be `upgrade_prechecks`, or at least the same as in the list of the steps, so `current_step: upgrade_steps_6_7.first` should be better here i guess

rsalevsky:
I think it makes more sense to realize this via instance methods. This has the advantage the we can load  the file via lazy loading and we can save it only if we need to safe it. Also we can drop the `save_` prefix form the methods due to it get's obvious that they are setter methods.

crowbar_framework/lib/upgrade.rb
jsuchome: Oh, sure, it should be the same

crowbar_framework/lib/upgrade.rb
```diff
@@ -0,0 +1,62 @@
+class Upgrade
```
MaximilianMeister: can you add an `attr_accessor :upgrade_progress` here? then you also don't need the `set_progress` setter anymore.

and otherwise I think you can only access the `upgrade_progress` hash through `Upgrade.new.as_json` which might look a bit inconsistent in the controller

crowbar_framework/lib/upgrade.rb
```diff
@@ -0,0 +1,62 @@
...
+  # Return the current state of upgrade process.
+  # We're keeping the information in the file so is accessible by other applications
+  # and different crowbar versions.
+  def initialize
+    @upgrade_progress = {
+      current_step: upgrade_steps_6_7.first,
+      # substep is needed for more complex steps like upgrading the nodes
+      substep: "",
```
MaximilianMeister: shall we add a setter for the substep and current_node as well?

crowbar_framework/lib/upgrade.rb
jsuchome: Yes, I was thinking more about accessing specific parts, not the whole map. But it might be good, yes

crowbar_framework/lib/upgrade.rb
jsuchome: OK, good idea. But we still should track it in the file

jsuchome:
[(5) Implement a state machine for the upgrade](https://trello.com/c/OWehKX6K/126-5-implement-a-state-machine-for-the-upgrade)

jsuchome:
@MaximilianMeister @rsalevsky adapted

crowbar_framework/lib/upgrade.rb
```diff
@@ -0,0 +1,64 @@
...
+    }
+    if progress_file_path.exist?
+      @upgrade_progress = JSON.load(progress_file_path.read)
+    else
+      # in 'steps', we save the information about each step that was executed
+      @upgrade_progress["steps"] = upgrade_steps_6_7.map do |step|
+        [step, { status: "pending", errors: {} }]
+      end.to_h
+    end
+    @current_step = @upgrade_progress["current_step"]
```
rsalevsky: I would move this and the following lines to separate methods and just return the value there.

crowbar_framework/lib/upgrade.rb
```diff
@@ -0,0 +1,64 @@
...
+  def initialize
+    @upgrade_progress = {
+      current_step: upgrade_steps_6_7.first,
+      # substep is needed for more complex steps like upgrading the nodes
+      substep: "",
+      # current node is relevant only for the nodes_upgrade step
+      current_node: ""
+    }
+    if progress_file_path.exist?
+      @upgrade_progress = JSON.load(progress_file_path.read)
```
rsalevsky: If you split this into a extra methods then you can use lazy loading.

crowbar_framework/lib/upgrade.rb
jsuchome: @MaximilianMeister proposed `attr_accessor` to avoid this. However, I'm fine with any of these approaches.

crowbar_framework/lib/upgrade.rb
rsalevsky: Depends how you use it. I prefer in general loading it if we really access the data, then you can create as many objects as you want and it will not really hurt. Also you can create the object at the beginning of the call and use it in between and save it at the end of the call.

jsuchome: There's no point in creating many objects of this. We need to know what's the real status and that is the one saved in the file - e.g. for the cases of some unexpected failure between saving 2 states.

crowbar_framework/lib/upgrade.rb
```diff
@@ -0,0 +1,64 @@
...
+  attr_reader :steps
+
+  # Return the current state of upgrade process.
+  # We're keeping the information in the file so is accessible by other applications
+  # and different crowbar versions.
+  def initialize
+    @upgrade_progress = {
+      current_step: upgrade_steps_6_7.first,
+      # substep is needed for more complex steps like upgrading the nodes
+      substep: "",
```
aspiers: Wouldn't `nil` be a more reasonable default?

crowbar_framework/lib/upgrade.rb
```diff
@@ -0,0 +1,64 @@
...
+      # current node is relevant only for the nodes_upgrade step
+      current_node: ""
```
aspiers: `nil` here too.

crowbar_framework/lib/upgrade.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def finished?
+    @upgrade_progress["current_step"] == upgrade_steps_6_7.last
+  end
+
+  protected
+
+  # global list of the steps of the upgrade process
+  def upgrade_steps_6_7
```
aspiers: This doesn't quite feel right to me - it's hardcoding an assumption that the top-level steps are linear.  OTOH it's possible that for this one-off upgrade case that's a safe assumption, so I suppose we could live with that and then adapt later if needed.

crowbar_framework/lib/upgrade.rb
aspiers: I don't get these 4 lines.  Why are you copying data from one instance variable to another?  Seems like a DRY violation - am I missing something?

crowbar_framework/lib/upgrade.rb
jsuchome: Writing from two different objects would overwrite the status, yes. But how do you want to prevent it, reading the file each time we're accessing the attribute? That means more file accesses. Or going back to class method only?

crowbar_framework/lib/upgrade.rb
jsuchome: So I can access the attributes directly with `attr_accessor`

jsuchome: But I could use separate methods to manipulate with steps, substeps etc. just like @rsalevsky proposed. I do not mind which solution we're gonna use, just please agree on one.

aspiers: I probably don't mind as long as it doesn't violate DRY.  But why do you have to use `attr_accessor`?  If you want to store all the data in `@upgrade_process` then why not just do this?

    def current_step
        @upgrade_progress["current_step"]
    end

etc.

Or alternatively have a `current_step` accessor and don't store in `@upgrade_progress`.  But don't do both!

jsuchome:
@aspiers @rsalevsky still requesting changes?

crowbar_framework/lib/upgrade.rb
```diff
@@ -0,0 +1,71 @@
...
+  def current_substep
+    upgrade_progress[:current_substep]
+  end
+
+  def current_step
+    upgrade_progress[:current_step]
+  end
+
+  def next_step!
+    return true if finished?
```
MaximilianMeister: maybe return false when finished? or how shall the caller determin the end when this function is used a few times?

jsuchome: OK, I could return false here.
But for the case you mention I'd expect caller to check `finished?` himself.

MaximilianMeister:
not sure if we need some file locking here for the json file

jsuchome:
@MaximilianMeister I think I could just reuse `Crowbar::Lock::LocalBlocking.with_lock` for writing the file... ?

MaximilianMeister:
> I think I could just reuse Crowbar::Lock::LocalBlocking.with_lock for writing the file... ?

I have never used it, @aspiers should know better, but it seems that this could be used

aspiers:
@MaximilianMeister is right - you need some locking semantics.  This is non-trivial, but I guess the locking has to be handled by this class (or at least in `lib/`) because it will need to be consumed by both `crowbar-core` *and* `crowbar-init`, right?

It should be pretty clear that it's not good enough to simply obtain a lock when reading/writing the `.json` file - this would leave a gigantic race window as long as the time any single upgrade step takes.  Thinking about this more makes me question whether the current design of this library's interface really makes sense.  For example, let's say we're currently on step "11", whatever that is (it doesn't matter for this example), and we have two cloud operators simultaneously working on the same upgrade in different browsers.  Through a miscommunication, they both think they are responsible for driving the upgrade process.

1. API client A sends a request which is handled by server-side pid 11111 which calls `#next_step!` (which temporarily locks the `.json` file for writing, and updates it to step 12).
2. Server-side pid 11111 queries the current step (which temporarily locks the `.json` file for reading). It returns step "12" to client A (either as response to the above request, or to a separate request, depending on the front-end API design).
3. API client A sends a request which is handled by server-side pid 11111 which starts running upgrade step 12.  This takes a long time!
4. At any point before that upgrade step finishes, API client B sends a request which is handled by server-side pid 22222 handling which calls `#next_step!` (which temporarily locks the `.json` file for writing, and updates it to step 13).
5. Server-side pid 22222 queries the current step (which temporarily locks the `.json` file for reading). It returns step "13" to client B (either as response to the above request, or to a separate request, depending on the front-end API design).
6. API client B sends a request which is handled by server-side pid 11111 which starts running upgrade step 13.  This takes a long time!
7. Server-side pid 22222 starts running the next step (13) even before step 12 has completed.

To get the locking right, I think you probably need a library interface which allows for marking the start **and end** of each step.  Marking the start should obtain a lock (BTW should it persist even if the server is restarted, or is a per-process flock good enough?), and **it should not be released until the step is marked as completed**.

Also it seems strange that `#next_step!` doesn't return the next step.  This means that the consumer has to call `#current_step` immediately after, which feels like it's introducing another race window in between advancing the step and finding out what the new step is in order to be able to start executing it.

BTW TDD / BDD can help with getting code design right.  The idea is to "write the code you wish you had" as tests, and then implement.  So in this case you'd be writing tests whilst pretending that you are the frontend API consumer.  (OK I know that might be difficult since @santiph is the one who will do that.)

Let's at very least discuss this on Monday, and ideally do some pair programming.  Design by PR is not a good approach.  Even if we want a redesign, I think it makes sense to merge this and use it as a starting point, rather than making this an endlessly long and complex PR.  I'm already fed up of github's new review system, and I guess I'm not the only one ;-)  (Although if you want to use it as a test case for reviewable.io then that would be cool with me ...)

jsuchome:
(Without reading whole comment:) we have some library interface for locking. If I put writing to json file into the block of `Crowbar::Lock::LocalBlocking.with_lock`, second call of the same method will wait until the first block is finished.

jsuchome:
> Also it seems strange that #next_step! doesn't return the next step. This means that the consumer 
> has to call #current_step immediately after, which feels like it's introducing another race window in 
> between advancing the step and finding out what the new step is in order to be able to start
> executing it.

My understanding is that the application that does the transition between the steps already knows which step it is currently executing, so I do not see a need for that. But of course `next_step` could be changed in this way.

jsuchome:
What is needed:
- method saving the state of current step
- do not allow to transfer to next step when the latest one was not successfull

aspiers:
@MaximilianMeister @abel-navarro @rsalevsky and anyone else, please note that I agreed with @jsuchome that we will continue reviewing this particular PR via Reviewable.io as an experiment to see how well it works.  So if you want to join the review process, please click on the Reviewable button in the PR description (and remember to sign in to Reviewable!)  Thanks :-)

MaximilianMeister:
Reviewed 3 of 3 files at r3.
Review status: 2 of 3 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[crowbar_framework/spec/lib/upgrade.spec, line 20 at r4](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTruGnBRtUxfBPtZRX5:-KTruGnBRtUxfBPtZRX6:b-mfbqhu) ([raw file](https://github.com/crowbar/crowbar-core/blob/6dc3751d7cc7f03f913915240e40cf57f56194ad/crowbar_framework/spec/lib/upgrade.spec#L20)):*
> ```Ruby
> 
> describe Upgrade do
>   context "status file does not exist" do
> ```

please can you start the contexts with "with" like `context "with a status file that does not exist" do`? it just reads better and is more conventional

---

*[crowbar_framework/spec/lib/upgrade.spec, line 23 at r4](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTruo4Gn88k1Cl3Hixd:-KTruo4H396U-kvE8l85:bstesgl) ([raw file](https://github.com/crowbar/crowbar-core/blob/6dc3751d7cc7f03f913915240e40cf57f56194ad/crowbar_framework/spec/lib/upgrade.spec#L23)):*
> ```Ruby
>     it "should return first step as current step" do
>       allow(File).to receive(:exist?).and_return(false)
>       expect(subject.current_step).to eql "upgrade_prechecks"
> ```

it clearly doesn't hurt to test the 2 things below, but for that testcase I think testing until here is enough

---

*[crowbar_framework/spec/lib/upgrade.spec, line 28 at r4](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTrv1UjSIq0c0ubZ2zN:-KTrv1Uk9u2_Jt3r6Tf_:b-ym0l7z) ([raw file](https://github.com/crowbar/crowbar-core/blob/6dc3751d7cc7f03f913915240e40cf57f56194ad/crowbar_framework/spec/lib/upgrade.spec#L28)):*
> ```Ruby
>     end
> 
>     it "should move to next step when requested" do
> ```

i'd just say `it "moves to the next step" do`

---

*[crowbar_framework/spec/lib/upgrade.spec, line 39 at r4](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTrvFpU95nr9_Az4mRW:-KTrvFpU95nr9_Az4mRX:b-816vaw) ([raw file](https://github.com/crowbar/crowbar-core/blob/6dc3751d7cc7f03f913915240e40cf57f56194ad/crowbar_framework/spec/lib/upgrade.spec#L39)):*
> ```Ruby
>     end
> 
>     it "should not move to next step when current one failed" do
> ```

here the same, I think we can avoid the usage of "should" and just tell what it does .. `it "does not move to the next step when.." do`

---

*[crowbar_framework/spec/lib/upgrade.spec, line 42 at r4](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTrtkKyCU8eeUbj_y_X:-KTrtkKyCU8eeUbj_y_Y:b-5t6336) ([raw file](https://github.com/crowbar/crowbar-core/blob/6dc3751d7cc7f03f913915240e40cf57f56194ad/crowbar_framework/spec/lib/upgrade.spec#L42)):*
> ```Ruby
>       allow(File).to receive(:exist?).and_return(false)
>       expect(subject.current_step).to eql "upgrade_prechecks"
>       expect(subject.end_step(false, { :failure => "error message" })).to be false
> ```

(in theory) don't we need to start the step before ending it?

---

*[crowbar_framework/spec/lib/upgrade.spec, line 48 at r4](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTrvhOjAoWUlU8_hiT9:-KTrvhOjAoWUlU8_hiTA:b-2guby3) ([raw file](https://github.com/crowbar/crowbar-core/blob/6dc3751d7cc7f03f913915240e40cf57f56194ad/crowbar_framework/spec/lib/upgrade.spec#L48)):*
> ```Ruby
>     end
> 
>     it "should return finish when finished" do
> ```

same here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

jsuchome:
Review status: 2 of 3 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[crowbar_framework/spec/lib/upgrade.spec, line 23 at r4](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTruo4Gn88k1Cl3Hixd:-KTrxl48TuKXNJEJViQ2:bufhe5w) ([raw file](https://github.com/crowbar/crowbar-core/blob/6dc3751d7cc7f03f913915240e40cf57f56194ad/crowbar_framework/spec/lib/upgrade.spec#L23)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> it clearly doesn't hurt to test the 2 things below, but for that testcase I think testing until here is enough
</details>
I know, but does it really make sense to create separate test for those nil checks?

---

*[crowbar_framework/spec/lib/upgrade.spec, line 42 at r4](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTrtkKyCU8eeUbj_y_X:-KTry4FbRn07JPAaQEb8:b-8b11j9) ([raw file](https://github.com/crowbar/crowbar-core/blob/6dc3751d7cc7f03f913915240e40cf57f56194ad/crowbar_framework/spec/lib/upgrade.spec#L42)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> (in theory) don't we need to start the step before ending it?
</details>
Added

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

jsuchome:
Review status: 1 of 3 files reviewed at latest revision, 6 unresolved discussions.

---

*[crowbar_framework/spec/lib/upgrade.spec, line 20 at r4](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTruGnBRtUxfBPtZRX5:-KTrypGHqUEjwz220JDF:b-896fix) ([raw file](https://github.com/crowbar/crowbar-core/blob/6dc3751d7cc7f03f913915240e40cf57f56194ad/crowbar_framework/spec/lib/upgrade.spec#L20)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> please can you start the contexts with "with" like `context "with a status file that does not exist" do`? it just reads better and is more conventional
</details>
Done.

---

*[crowbar_framework/spec/lib/upgrade.spec, line 28 at r4](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTrv1UjSIq0c0ubZ2zN:-KTryrEfQImsx8yGbTqY:b-896fix) ([raw file](https://github.com/crowbar/crowbar-core/blob/6dc3751d7cc7f03f913915240e40cf57f56194ad/crowbar_framework/spec/lib/upgrade.spec#L28)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> i'd just say `it "moves to the next step" do`
</details>
Done.

---

*[crowbar_framework/spec/lib/upgrade.spec, line 39 at r4](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTrvFpU95nr9_Az4mRW:-KTryrr_lOEA7oniAx_c:b-896fix) ([raw file](https://github.com/crowbar/crowbar-core/blob/6dc3751d7cc7f03f913915240e40cf57f56194ad/crowbar_framework/spec/lib/upgrade.spec#L39)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> here the same, I think we can avoid the usage of "should" and just tell what it does .. `it "does not move to the next step when.." do`
</details>
Done.

---

*[crowbar_framework/spec/lib/upgrade.spec, line 42 at r4](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTrtkKyCU8eeUbj_y_X:-KTrysLel6VcnmknV4oQ:b-896fix) ([raw file](https://github.com/crowbar/crowbar-core/blob/6dc3751d7cc7f03f913915240e40cf57f56194ad/crowbar_framework/spec/lib/upgrade.spec#L42)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> Added
</details>
Done.

---

*[crowbar_framework/spec/lib/upgrade.spec, line 48 at r4](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTrvhOjAoWUlU8_hiT9:-KTryssi6Bay2NH9gHpJ:b-896fix) ([raw file](https://github.com/crowbar/crowbar-core/blob/6dc3751d7cc7f03f913915240e40cf57f56194ad/crowbar_framework/spec/lib/upgrade.spec#L48)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> same here
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

MaximilianMeister:
Reviewed 2 of 2 files at r5, 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[crowbar_framework/spec/lib/upgrade.spec, line 23 at r4](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTruo4Gn88k1Cl3Hixd:-KTrzmzepRK7EKCuQrqh:b-yu71rt) ([raw file](https://github.com/crowbar/crowbar-core/blob/6dc3751d7cc7f03f913915240e40cf57f56194ad/crowbar_framework/spec/lib/upgrade.spec#L23)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> I know, but does it really make sense to create separate test for those nil checks?
</details>
yeah why not? :) like this it feels a bit like squeezed into some other test

---

*[crowbar_framework/spec/lib/upgrade.spec, line 20 at r5](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTs-JXDlbwQ5CYoU22_:-KTs-JXDlbwQ5CYoU22a:b-jw5yc7) ([raw file](https://github.com/crowbar/crowbar-core/blob/2cae60425013e406d893944a9a1ec34997e672f9/crowbar_framework/spec/lib/upgrade.spec#L20)):*
> ```Ruby
> 
> describe Upgrade do
>   context "With a status file that does not exist" do
> ```

`s/With/with/`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

jsuchome:
Review status: 1 of 3 files reviewed at latest revision, 7 unresolved discussions.

---

*[crowbar_framework/spec/lib/upgrade.spec, line 23 at r4](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTruo4Gn88k1Cl3Hixd:-KTs0V68bUK8zEEHKEuV:b-896fix) ([raw file](https://github.com/crowbar/crowbar-core/blob/6dc3751d7cc7f03f913915240e40cf57f56194ad/crowbar_framework/spec/lib/upgrade.spec#L23)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> yeah why not? :) like this it feels a bit like squeezed into some other test
</details>
Done.

---

*[crowbar_framework/spec/lib/upgrade.spec, line 20 at r5](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTs-JXDlbwQ5CYoU22_:-KTs0TSbVdcPVTk4xd5A:b-896fix) ([raw file](https://github.com/crowbar/crowbar-core/blob/2cae60425013e406d893944a9a1ec34997e672f9/crowbar_framework/spec/lib/upgrade.spec#L20)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> `s/With/with/`
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

MaximilianMeister:
:lgtm:

---

Review status: 1 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-:-KTrsaGDnmu7znkK_-R_:bnfp4nl)*

MaximilianMeister:
Reviewed 2 of 2 files at r7, 1 of 1 files at r8.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

MaximilianMeister:
Reviewed 1 of 1 files at r9, 3 of 3 files at r10, 1 of 1 files at r11.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

aspiers:
Reviewed 2 of 2 files at r1, 1 of 3 files at r3.
Review status: all files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 23 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsOwWu0De1LOpxGwg8:-KTsOwWvhtIdR-jS4H6U:b-9hy2yr) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L23)):*
> ```Ruby
>   example '
>   {
>     "current_step": "upgrade_prechecks",
> ```

`current_step` sounds ambiguous to me - how does it distinguish between "the step which is currently executing" and "the step which we have to execute next (since right now we're not executing any step)"?

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 28 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsPP0_4JDLzcH_P3kB:-KTsPP0auAJ71dHhnWSS:b-diz4q5) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L28)):*
> ```Ruby
>     "steps": {
>       "upgrade_prechecks": {
>         "status": "pending",
> ```

Maybe this field is the answer to my question above? But I'd like to see the corresponding example code for the API caller.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 29 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsPnGuOcEgyaOmur0u:-KTsPnGuOcEgyaOmur0v:b-n8k66u) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L29)):*
> ```Ruby
>       "upgrade_prechecks": {
>         "status": "pending",
>         "errors": {}
> ```

If it's pending, why do we need to return `errors`?

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 31 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsQ6-4ZW3n-hQzjSTl:-KTsQ6-4ZW3n-hQzjSTm:bj2gndt) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L31)):*
> ```Ruby
>         "errors": {}
>       },
>       "admin_backup": {
> ```

What is the reason for including all the future steps in the response?  If the API callers need this info, maybe they also need the order of the future steps?

---

*[crowbar_framework/lib/upgrade.rb, line 1 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsQaazcR1DiU78Kw8f:-KTsQaazcR1DiU78Kw8g:bbui0te) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/lib/upgrade.rb#L1)):*
> ```Ruby
> class Upgrade
> ```

Shouldn't this be inside a `module Crowbar`, instead of polluting the global namespace?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

aspiers:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTz-l0kkK0sG7YOtVTF:-KTz-l0kkK0sG7YOtVTG:b-6h55sy) (no related file):*
Sorry, I wrote some comments the other day and forgot to publish, but now some of them may be out of date.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

aspiers:
Reviewed 1 of 1 files at r11.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[crowbar_framework/lib/upgrade.rb, line 16 at r11](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTz-zHPkE34V7eZnCaC:-KTz-zHPkE34V7eZnCaD:bxdijzz) ([raw file](https://github.com/crowbar/crowbar-core/blob/8deda6ff854d06ff99831a634d67af40783640e8/crowbar_framework/lib/upgrade.rb#L16)):*
> ```Ruby
>     }
>     if progress_file_path.exist?
>       Crowbar::Lock::LocalBlocking.with_lock(shared: true) do
> ```

This bit looks good now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

jsuchome:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 23 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsOwWu0De1LOpxGwg8:-KTzBGYf4-IRovkmIY1g:bv93ehs) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L23)):*
<details><summary><i>Previously, aspiers (Adam Spiers) wrote…</i></summary>
> `current_step` sounds ambiguous to me - how does it distinguish between "the step which is currently executing" and "the step which we have to execute next (since right now we're not executing any step)"?
</details>
This is really the step we're current executing.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 28 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsPP0_4JDLzcH_P3kB:-KTzCGstIgYVfqs3C4SY:b-4swlu2) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L28)):*
<details><summary><i>Previously, aspiers (Adam Spiers) wrote…</i></summary>
> Maybe this field is the answer to my question above? But I'd like to see the corresponding example code for the API caller.
</details>
I think the usage is visible from the `start_step` and `end_step` methods. `pending` status might be usable for the UI.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 29 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsPnGuOcEgyaOmur0u:-KTzBXu89VxD4CXiv17e:bplf2x6) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L29)):*
<details><summary><i>Previously, aspiers (Adam Spiers) wrote…</i></summary>
> If it's pending, why do we need to return `errors`?
</details>
We do not need it, but I think it's more convenient to always assume the "errors" key is always present.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 31 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsQ6-4ZW3n-hQzjSTl:-KTzCfcRn_LSwqfUg-fJ:b9y82au) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L31)):*
<details><summary><i>Previously, aspiers (Adam Spiers) wrote…</i></summary>
> What is the reason for including all the future steps in the response?  If the API callers need this info, maybe they also need the order of the future steps?
</details>
Good point: I think we need to make it an array, not hash.

---

*[crowbar_framework/lib/upgrade.rb, line 1 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsQaazcR1DiU78Kw8f:-KTzCn66UMuJ3cIBbaCx:b-cs0mea) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/lib/upgrade.rb#L1)):*
<details><summary><i>Previously, aspiers (Adam Spiers) wrote…</i></summary>
> Shouldn't this be inside a `module Crowbar`, instead of polluting the global namespace?
</details>
May be. I think maybe Max had some good reasoning for using just Upgrade, but I do not remember,,,

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

MaximilianMeister:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 31 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsQ6-4ZW3n-hQzjSTl:-KU11hkDhy0S1R_D6aCq:b9c4s0a) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L31)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> Good point: I think we need to make it an array, not hash.
</details>
why? isn't a hash always the same order too? I thought that is a python problem with dicts :)

---

*[crowbar_framework/lib/upgrade.rb, line 1 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsQaazcR1DiU78Kw8f:-KU121OqutqfhzxAv7bg:b-zh0dsd) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/lib/upgrade.rb#L1)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> May be. I think maybe Max had some good reasoning for using just Upgrade, but I do not remember,,,
</details>
I know it smells a bit, this would conflict with https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/upgrade.rb - but there may be a better naming that would allow us to put it in Crowbar's namespace?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

jsuchome:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 31 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsQ6-4ZW3n-hQzjSTl:-KU198aGgAejlPAvXq4q:bm2crjx) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L31)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> why? isn't a hash always the same order too? I thought that is a python problem with dicts :)
</details>
Ah, true, ruby hashes keep the order!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

Itxaka:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[crowbar_framework/lib/upgrade.rb, line 44 at r9](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KU1BJWIx4kPnTuyppU7:-KU1BJWIx4kPnTuyppU8:bknpjbp) ([raw file](https://github.com/crowbar/crowbar-core/blob/e35fe5a392dcbb47fa8c68e0e10a6a073a3a9446/crowbar_framework/lib/upgrade.rb#L44)):*
> ```Ruby
> 
>   def start_step
>     upgrade_progress[:steps][current_step][:status] = "running"
> ```

Shouldn't we `save` here so the current status gets saved everytime we start a different step?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

MaximilianMeister:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[crowbar_framework/lib/upgrade.rb, line 44 at r9](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KU1BJWIx4kPnTuyppU7:-KU1CHIXEnlv36jbJBLp:b-amxwxs) ([raw file](https://github.com/crowbar/crowbar-core/blob/e35fe5a392dcbb47fa8c68e0e10a6a073a3a9446/crowbar_framework/lib/upgrade.rb#L44)):*
<details><summary><i>Previously, Itxaka (Itxaka) wrote…</i></summary>
> Shouldn't we `save` here so the current status gets saved everytime we start a different step?
</details>
we'll do that from the outside in a more controlled/explicit way I assume

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

jsuchome:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[crowbar_framework/lib/upgrade.rb, line 44 at r9](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KU1BJWIx4kPnTuyppU7:-KU1DJ9S3naeSZWwlbi6:b-d43uc4) ([raw file](https://github.com/crowbar/crowbar-core/blob/e35fe5a392dcbb47fa8c68e0e10a6a073a3a9446/crowbar_framework/lib/upgrade.rb#L44)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> we'll do that from the outside in a more controlled/explicit way I assume
</details>
Hm, it might have sense to save hereby default, and not just assume caller cares about saving after calling `end_step`. That said, we could have saving in `start_step` and `end_step`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

jsuchome:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[crowbar_framework/lib/upgrade.rb, line 1 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsQaazcR1DiU78Kw8f:-KU1E2OzZHOqwsaLLCH6:bhdjcgy) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/lib/upgrade.rb#L1)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> I know it smells a bit, this would conflict with https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/upgrade.rb - but there may be a better naming that would allow us to put it in Crowbar's namespace?
</details>
Crowbar::UpgradeStatus , UpgradeState, UpgradeProgress ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

MaximilianMeister:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[crowbar_framework/lib/upgrade.rb, line 1 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsQaazcR1DiU78Kw8f:-KU1d9m3ZH1HDpHLdaXY:b-kziv0f) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/lib/upgrade.rb#L1)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> Crowbar::UpgradeStatus , UpgradeState, UpgradeProgress ?
</details>
`Crowbar::UpgradeStatus` sounds good to me

---

*[crowbar_framework/lib/upgrade.rb, line 44 at r9](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KU1BJWIx4kPnTuyppU7:-KU1dKXJJKiKwwGqzDPQ:b5lwzcx) ([raw file](https://github.com/crowbar/crowbar-core/blob/e35fe5a392dcbb47fa8c68e0e10a6a073a3a9446/crowbar_framework/lib/upgrade.rb#L44)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> Hm, it might have sense to save hereby default, and not just assume caller cares about saving after calling `end_step`. That said, we could have saving in `start_step` and `end_step`
</details>
not 100% sure here, @aspiers any opinion?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

Itxaka:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[crowbar_framework/lib/upgrade.rb, line 44 at r9](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KU1BJWIx4kPnTuyppU7:-KU1jfT9weQS6kmw6UT1:b-16fh9h) ([raw file](https://github.com/crowbar/crowbar-core/blob/e35fe5a392dcbb47fa8c68e0e10a6a073a3a9446/crowbar_framework/lib/upgrade.rb#L44)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> not 100% sure here, @aspiers any opinion?
</details>
Im guessing that if we want to be robust, we shouldn't assume that it will be done, but instead, saving it at each step, or with any status change, so we can be sure that the status for the steps its always updated, no matter what happens.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

jsuchome:
Review status: 1 of 5 files reviewed at latest revision, 7 unresolved discussions.

---

*[crowbar_framework/lib/crowbar/upgrade_status.rb, line 1 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsQaazcR1DiU78Kw8f:-KU218Q4oedxne7gzSYl:bi69w05) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/lib/crowbar/upgrade_status.rb#L1)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> `Crowbar::UpgradeStatus` sounds good to me
</details>
Changed to Crowbar::UpgradeStatus in latest commit. I will sqaush it into the first one once we agree that this is desired 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

jsuchome:
Review status: 1 of 5 files reviewed at latest revision, 7 unresolved discussions.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 29 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsPnGuOcEgyaOmur0u:-KU2BEfD5y4Q1fcyJ5Ne:bnwso4d) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L29)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> We do not need it, but I think it's more convenient to always assume the "errors" key is always present.
</details>
BTW, I removed those errors from steps with pending status

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

MaximilianMeister:
Reviewed 2 of 6 files at r12, 5 of 7 files at r13.
Review status: 0 of 5 files reviewed at latest revision, 10 unresolved discussions.

---

*[crowbar_framework/lib/crowbar/upgrade_status.rb, line 33 at r17](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KULPwCaoouh_JW2XOpj:-KULPwCaoouh_JW2XOpk:b3vtw0r) ([raw file](https://github.com/crowbar/crowbar-core/blob/632857461681ce261ba1515ce4999c3da83a915e/crowbar_framework/lib/crowbar/upgrade_status.rb#L33)):*
> ```Ruby
>           f.write(JSON.pretty_generate(progress))
>         end
>       end
> ```

Ideally `save` returns a boolean on success or failure. At least I would expect that, as it is the key function here for the persistency of the status

---

*[crowbar_framework/lib/crowbar/upgrade_status.rb, line 51 at r17](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KULQ2NmLLU3cK6A6xqM:-KULQ2NmLLU3cK6A6xqN:b-6wuaw6) ([raw file](https://github.com/crowbar/crowbar-core/blob/632857461681ce261ba1515ce4999c3da83a915e/crowbar_framework/lib/crowbar/upgrade_status.rb#L51)):*
> ```Ruby
>       progress[:steps][current_step][:status] = "running"
>       save
>       true
> ```

I'd let `save` determine the return value here

---

*[crowbar_framework/lib/crowbar/upgrade_status.rb, line 75 at r17](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KULQpZeQyOe83GqprSR:-KULQpZeQyOe83GqprSS:b9xlkf6) ([raw file](https://github.com/crowbar/crowbar-core/blob/632857461681ce261ba1515ce4999c3da83a915e/crowbar_framework/lib/crowbar/upgrade_status.rb#L75)):*
> ```Ruby
>       progress[:current_step] = upgrade_steps_6_7[i + 1]
>       save
>       true
> ```

same here. I think it makes much more sense to let `save` determine the return value here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

jsuchome:
Review status: 0 of 5 files reviewed at latest revision, 10 unresolved discussions.

---

*[crowbar_framework/lib/crowbar/upgrade_status.rb, line 44 at r9](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KU1BJWIx4kPnTuyppU7:-KULheAQZXp3G3RRA4ur:bwhxhk3) ([raw file](https://github.com/crowbar/crowbar-core/blob/e35fe5a392dcbb47fa8c68e0e10a6a073a3a9446/crowbar_framework/lib/crowbar/upgrade_status.rb#L44)):*
<details><summary><i>Previously, Itxaka (Itxaka) wrote…</i></summary>
> Im guessing that if we want to be robust, we shouldn't assume that it will be done, but instead, saving it at each step, or with any status change, so we can be sure that the status for the steps its always updated, no matter what happens.
</details>
So I've done saving in `start_step` and `end_step`

---

*[crowbar_framework/lib/crowbar/upgrade_status.rb, line 33 at r17](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KULPwCaoouh_JW2XOpj:-KULhXhc0X6RhOm6GKKq:bekzdpl) ([raw file](https://github.com/crowbar/crowbar-core/blob/632857461681ce261ba1515ce4999c3da83a915e/crowbar_framework/lib/crowbar/upgrade_status.rb#L33)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> Ideally `save` returns a boolean on success or failure. At least I would expect that, as it is the key function here for the persistency of the status
</details>
Good idea. Added in the latest comment

---

*[crowbar_framework/lib/crowbar/upgrade_status.rb, line 51 at r17](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KULQ2NmLLU3cK6A6xqM:-KULhkVxhak-CqGZ8ZzM:b-896fix) ([raw file](https://github.com/crowbar/crowbar-core/blob/632857461681ce261ba1515ce4999c3da83a915e/crowbar_framework/lib/crowbar/upgrade_status.rb#L51)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> I'd let `save` determine the return value here
</details>
Done.

---

*[crowbar_framework/lib/crowbar/upgrade_status.rb, line 75 at r17](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KULQpZeQyOe83GqprSR:-KULhlaraBqt-novqHDQ:b-896fix) ([raw file](https://github.com/crowbar/crowbar-core/blob/632857461681ce261ba1515ce4999c3da83a915e/crowbar_framework/lib/crowbar/upgrade_status.rb#L75)):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> same here. I think it makes much more sense to let `save` determine the return value here
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

MaximilianMeister:
:lgtm:

---

Review status: 0 of 5 files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-:-KULmWa-H9M45BvIUL9j:bnfp4nl)*

jsuchome:
Review status: 0 of 5 files reviewed at latest revision, 7 unresolved discussions.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 29 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsPnGuOcEgyaOmur0u:-KULn7UUrDeDzv1NFdKC:b-896fix) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L29)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> BTW, I removed those errors from steps with pending status
</details>
Done.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 31 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsQ6-4ZW3n-hQzjSTl:-KULn9En5fX-yJNGR8CO:b-896fix) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L31)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> Ah, true, ruby hashes keep the order!
</details>
Done.

---

*[crowbar_framework/lib/crowbar/upgrade_status.rb, line 1 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsQaazcR1DiU78Kw8f:-KULnBA6MDtWKnsFOeF1:b-896fix) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/lib/crowbar/upgrade_status.rb#L1)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> Changed to Crowbar::UpgradeStatus in latest commit. I will sqaush it into the first one once we agree that this is desired 
</details>
Done.

---

*[crowbar_framework/lib/crowbar/upgrade_status.rb, line 44 at r9](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KU1BJWIx4kPnTuyppU7:-KULnCD0hH5GGybsPOP6:b-896fix) ([raw file](https://github.com/crowbar/crowbar-core/blob/e35fe5a392dcbb47fa8c68e0e10a6a073a3a9446/crowbar_framework/lib/crowbar/upgrade_status.rb#L44)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> So I've done saving in `start_step` and `end_step`
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

MaximilianMeister:
Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTz-l0kkK0sG7YOtVTF:-KTwqabm5wE1mPiypM5l:b-vl4xer) (no related file):*
<details><summary><i>Previously, aspiers (Adam Spiers) wrote…</i></summary>
> Sorry, I wrote some comments the other day and forgot to publish, but now some of them may be out of date.
</details>
dismissed for testing

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

Itxaka:
Reviewed 2 of 6 files at r12, 5 of 7 files at r13.
Review status: 0 of 5 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

Itxaka:
Reviewed 2 of 6 files at r12, 1 of 3 files at r14, 4 of 6 files at r16, 1 of 2 files at r17, 1 of 1 files at r18.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

MaximilianMeister:
Reviewed 2 of 6 files at r12, 1 of 3 files at r14, 4 of 6 files at r16, 1 of 2 files at r17, 1 of 1 files at r18.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

aspiers:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 23 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsOwWu0De1LOpxGwg8:-KUMgJHmfKQvGxOY9P86:b-wq9fjb) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L23)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> This is really the step we're current executing.
</details>
So what is the value when no step is executing? If I understand correctly, it will be the step which just finished, or the first step if no step finished yet. If that's right, it's an inconsistent API design.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 28 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsPP0_4JDLzcH_P3kB:-KUMmwxLcaVWhOPQlKqT:btngdfh) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L28)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> I think the usage is visible from the `start_step` and `end_step` methods. `pending` status might be usable for the UI.
</details>
No, I meant code which would actually call `start_step` and `end_step`.  Writing that code should clarify whether the interface design is clean. Although you have examples of that in the test suite, which is good.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 31 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsQ6-4ZW3n-hQzjSTl:-KUMlx2nKxw-9ZBmC3W6:brjwlu2) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L31)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> Done.
</details>
Personally I don't like the idea of relying on insertion order, because it's implicit, hidden, brittle behaviour, which a future coder could easily break by accident. But if you're convinced that's a reliable approach in this particular case then I'm open to being persuaded otherwise.

---

*[crowbar_framework/lib/crowbar/upgrade_status.rb, line 44 at r9](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KU1BJWIx4kPnTuyppU7:-KU2C4r1fP6jvnCrUx-G:bobbhdk) ([raw file](https://github.com/crowbar/crowbar-core/blob/e35fe5a392dcbb47fa8c68e0e10a6a073a3a9446/crowbar_framework/lib/crowbar/upgrade_status.rb#L44)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> Done.
</details>
This still looks wrong to me, because it doesn't defend against two callers both starting a step at the same time.  Let's discuss on mumble.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

aspiers:
Reviewed 2 of 6 files at r12, 1 of 3 files at r14, 4 of 6 files at r16, 1 of 2 files at r17, 1 of 1 files at r18.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

jsuchome:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 23 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsOwWu0De1LOpxGwg8:-KUNPFeRjeJuFpCgFUSP:b9nijpc) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L23)):*
<details><summary><i>Previously, aspiers (Adam Spiers) wrote…</i></summary>
> So what is the value when no step is executing? If I understand correctly, it will be the step which just finished, or the first step if no step finished yet. If that's right, it's an inconsistent API design.
</details>
`end_step` advances the current step. So before calling `end_step`, current step is still being executed and it is still returned as current

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 28 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsPP0_4JDLzcH_P3kB:-KUNPXwtrIp4jhLYq4f3:b5qvtex) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L28)):*
<details><summary><i>Previously, aspiers (Adam Spiers) wrote…</i></summary>
> No, I meant code which would actually call `start_step` and `end_step`.  Writing that code should clarify whether the interface design is clean. Although you have examples of that in the test suite, which is good.
</details>
The example code is the the PR description

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

aspiers:
Reviewed 2 of 6 files at r12, 2 of 6 files at r16, 2 of 2 files at r19.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTz-l0kkK0sG7YOtVTF:-KUO4tXpUIb0FTA-_DtX:b-5xw7u0) (no related file):*
<details><summary><i>Previously, MaximilianMeister (Maximilian Meister) wrote…</i></summary>
> dismissed for testing
</details>
OK, so Jiri and I pair programmed and got this to a state where we proved that the interface *can* successfully handle concurrency, so I think it's OK to merge this.  There is still a bug where it doesn't handle concurrency between threads correctly yet because it relies on the in-memory cache of the state for reads, instead of always checking the json file, but it should be easy to fix this in a subsequent PR, and we'll have to do that since this is apparently blocking too much other work.  There are also a number of clean-ups which could be made to the interface, but again we ran out of time so will have to do them later.  As discussed I think this PR is another really good example of how pair-programming and pair-design can save us not just hours but days or even weeks of time vs. trying to do "design by PR".  

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 23 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsOwWu0De1LOpxGwg8:-KUO3tKRsEgS4-8EYBk_:b243ttx) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L23)):*
<details><summary><i>Previously, jsuchome (Jiří Suchomel) wrote…</i></summary>
> `end_step` advances the current step. So before calling `end_step`, current step is still being executed and it is still returned as current
</details>
So `current_step` actually returns "the current step if a step is running, or the next step if a step isn't running, unless it's the last step in which case it returns the last step". I think this is a pretty confusing interface design, but it seems we've run out of time to fix it in this PR, so we'll probably have to just live with it.

---

*[crowbar_framework/app/controllers/api/upgrade_controller.rb, line 31 at r8](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-KTsQ6-4ZW3n-hQzjSTl:-KUO4pMQ-GONRv-ekSo_:buh3acd) ([raw file](https://github.com/crowbar/crowbar-core/blob/ea9dbedb5c305d6652df17792f4046565bba7975/crowbar_framework/app/controllers/api/upgrade_controller.rb#L31)):*
<details><summary><i>Previously, aspiers (Adam Spiers) wrote…</i></summary>
> Personally I don't like the idea of relying on insertion order, because it's implicit, hidden, brittle behaviour, which a future coder could easily break by accident. But if you're convinced that's a reliable approach in this particular case then I'm open to being persuaded otherwise.
</details>
As discussed it seems we ran out of time to fix this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692)*

MaximilianMeister:
:lgtm:

---

Reviewed 2 of 6 files at r12, 2 of 6 files at r16, 2 of 2 files at r19.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-core/692#-:-KUR0dCuC8M4cSIlNl24:bnfp4nl)*

PR #694, opened 2016-09-30
bsc#999637: Remove firewall sanity check

nicolasbock:
Commit
https://github.com/crowbar/crowbar-core/commit/73a0ce5ff0cd335c51e0bae2175c2d8c675f97fa
exposed an inconsistency between the firewall rules needed for the bmc-nat
barclamp and crowbar's firewall sanity check. The sanity check is too broad as
it does not tolerate _any_ iptables rule. A more sensible sanity check would
need to parse the firewall rules more intelligently to ensure that the admin
node is accessible and yet allow for rules added by individual barclamps.

In the interest of addressing https://bugzilla.suse.com/show_bug.cgi?id=999637
I am removing the sanity check for now.

(cherry picked from commit 6f25f32374f31ca3938a6696092ce3cf182bde07)


rsalevsky:
Can you check if there is some reference in the UI to this check? I'm not sure anymore if there was one.

nicolasbock:
@rsalevsky I couldn't find any.

PR #702, opened 2016-10-06
Set the pre-upgrade pacemaker attribute for all nodes

jsuchome:
While executing "services" step, we're touching every node. This is a good time to set the pacemaker pre-upgrade attribute

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -26,6 +26,8 @@ return unless node[:platform_family] == "suse"
 
 upgrade_step = node["crowbar_wall"]["crowbar_upgrade_step"] || "none"
 
+Chef::Log.info("Current upgrade step: #{upgrade_step}")
+
 case upgrade_step
 
 when "revert_to_ready"
@@ -70,6 +72,11 @@ when "crowbar_upgrade"
 when "prepare-os-upgrade"
 
   include_recipe "crowbar::prepare-upgrade-scripts"
+
+  execute "set pre-upgrade node attribute" do
+    command "crm node attribute $(hostname) set pre-upgrade true"
+    only_if { node.roles.include? "pacemaker-cluster-member" }
+  end
 
 when "openstack_shutdown"
 


jsuchome:
[(13) Orchestrate upgrade of controller nodes in HA setup (upgrade step 12 part 2)](https://trello.com/c/2PMGauJx/118-13-orchestrate-upgrade-of-controller-nodes-in-ha-setup-upgrade-step-12-part-2)

jsuchome:
@aspiers @rsalevsky I've completely changed this. so please, another review... I mean, the code is the same, but executed from different place. The previous one was not really run on all nodes.

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -71,6 +71,13 @@
 
   include_recipe "crowbar::prepare-upgrade-scripts"
 
+  bash "set pre-upgrade node attribute"
```
aspiers: Since this is a one-liner, why not use the [`execute` resource](https://docs.chef.io/resource_execute.html) ?

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -71,6 +71,13 @@
...
+    code <<-EOF
+      crm node attribute $(hostname) set pre-upgrade true
+    EOF
+    only_if { node["run_list_map"].key? "pacemaker-cluster-member" }
```
aspiers: Is there a better way of doing this than checking `run_list_map`?  Ideally there should be a helper to do it. In the long term we might end up dynamically varying the run list (even though the new role recipes have gone in the opposite direction), so I'd expect `node["pacemaker"]["elements"]` to be a slightly more stable place to check.  But I'd be interested in @vuntz's opinion here.

vuntz: A couple of things:

- generally I agree that we don't want to look at `run_list_map`; I know we did that in 5->6, but that's really digging in the implementation details in a way that we shouldn't
- since we don't deactivate proposals (I think), `node.role? "pacemaker-cluster-member"` should work
- if we want to be safe and not look at the roles, I would actually check if we have `node[:pacemaker][:config][:environment]`, although this would match remote pacemaker nodes, which might not be desired

jsuchome: Adapted, a bit. I hope it is enough now.

PR #703, opened 2016-10-06
enhancing/speeding up node upgrade specs

MaximilianMeister:
the tests unneccessarily took way too long

### DIFF
--- a/crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
+++ b/crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
@@ -58,17 +58,27 @@ describe Api::UpgradeController, type: :request do
     end
 
     it "initiates the upgrade of nodes" do
-      allow(NodeObject).to(
-        receive(:find).and_return([NodeObject.find_node_by_name("testing")])
+      allow_any_instance_of(NodeObject).to receive(:run_ssh_cmd).and_return(
+        stdout: "",
+        tderr: "",
+        exit_code: 0
       )
+      allow(Api::Upgrade).to receive(:upgrade_controller_nodes).and_return(true)
+      allow_any_instance_of(Api::Node).to receive(:upgrade).and_return(true)
+
       post "/api/upgrade/nodes", {}, headers
       expect(response).to have_http_status(:ok)
     end
 
     it "initiates the upgrade of nodes and fails" do
-      allow(NodeObject).to(
-        receive(:find).and_return([])
+      allow_any_instance_of(NodeObject).to receive(:run_ssh_cmd).and_return(
+        stdout: "",
+        tderr: "",
+        exit_code: 1
       )
+      allow(Api::Upgrade).to receive(:upgrade_controller_nodes).and_return(false)
+      allow_any_instance_of(Api::Node).to receive(:upgrade).and_return(true)
+
       post "/api/upgrade/nodes", {}, headers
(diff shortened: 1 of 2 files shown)


crowbar_framework/spec/models/api/upgrade_spec.rb
```diff
@@ -178,12 +178,20 @@
         with("state:crowbar_upgrade AND (roles:database-server OR roles:rabbitmq-server)").
         and_return([NodeObject.find_node_by_name("testing.crowbar.com")])
       )
+      allow_any_instance_of(NodeObject).to receive(:run_ssh_cmd).and_return(
+        stdout: "",
+        stderr: "",
+        exit_code: 1
+      )
+
```
jsuchome: I'm not sure what consistency are we aiming for here, but to me it looks like you should have that empty line before the `expect` call and not between those 2 `allow`s

MaximilianMeister: oops, yes that was the intent. thanks :)

MaximilianMeister: fixed

crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
```diff
@@ -58,17 +58,15 @@
     end
 
     it "initiates the upgrade of nodes" do
-      allow(NodeObject).to(
-        receive(:find).and_return([NodeObject.find_node_by_name("testing")])
-      )
+      allow(Api::Upgrade).to receive(:nodes).and_return(true)
```
aspiers: Sorry, but I'm struggling to understand this commit message.  Which test is already covered by which model?  And how can a model cover a test anyway?  Surely it's the other way around - a test covering a model?

And how does the second sentence about slow testing relate to this commit?

This kind of disconnect in understanding seems to be quite a common problem at the moment, and it's really slowing reviews down, for me anyway.  I've suggested more frequent pair programming as a way of addressing it, but I'm open to other ideas too.  What would you suggest?  BTW did you read [these articles on the topic](http://mailman.suse.de/mlarch/SuSE/cloud-devel/2016/cloud-devel.2016.09/msg00089.html)?  I think they make a lot of good points.

crowbar_framework/spec/models/api/upgrade_spec.rb
```diff
@@ -178,6 +178,12 @@
         with("state:crowbar_upgrade AND (roles:database-server OR roles:rabbitmq-server)").
         and_return([NodeObject.find_node_by_name("testing.crowbar.com")])
       )
+      allow_any_instance_of(NodeObject).to receive(:run_ssh_cmd).and_return(
```
aspiers: Oh, so I guess this is what you were referring to in the other commit.  Maybe you split the changes into separate commits (if so, thank you for that!) but forgot to split the commit message too? This commit is missing an explanation.

crowbar_framework/spec/models/api/upgrade_spec.rb
```diff
@@ -184,12 +184,14 @@
         exit_code: 1
       )
       allow_any_instance_of(Api::Node).to receive(:upgrade).and_return(true)
+
       expect(subject.class.nodes).to be true
     end
     it "fails during the upgrade of controller nodes" do
```
aspiers: There should be a line break here too.

MaximilianMeister:
@aspiers updated

PR #704, opened 2016-10-06
[stable/3.0] enable collecting a repolist of missing/inactive repos

MaximilianMeister:
this becomes useful in the upgrade UI during the repochecks

(cherry picked from commit 4819c649a7f3acf99694f1518a110b9994670596)

PR #705, opened 2016-10-06
don't provoke repository enabling

MaximilianMeister:
when we check the status of the upgrade, we are trying an implicit
enabling of the repositories in the upgrade case.
at this stage this is just wrong, so we drop this check and
sacrifice the api response to just report status 200 instead of
the upgrade status in the response

(this needs a backport)

### DIFF
--- a/crowbar_framework/app/controllers/api/crowbar_controller.rb
+++ b/crowbar_framework/app/controllers/api/crowbar_controller.rb
@@ -61,7 +61,7 @@ class Api::CrowbarController < ApiController
       crowbar_upgrade = Api::Crowbar.upgrade!
 
       if crowbar_upgrade[:status] == :ok
-        render json: Api::Crowbar.upgrade
+        head :ok
       else
         render json: { error: crowbar_upgrade[:message] }, status: crowbar_upgrade[:status]
       end


MaximilianMeister:
this would likely obsolete https://github.com/crowbar/crowbar-core/pull/704

crowbar_framework/app/controllers/api/crowbar_controller.rb
```diff
@@ -61,7 +61,7 @@ def upgrade
       crowbar_upgrade = Api::Crowbar.upgrade!
 
       if crowbar_upgrade[:status] == :ok
-        render json: Api::Crowbar.upgrade
+        head :ok
```
jsuchome: Hm, we were expecting that 
```
            upgrading: upgrading?,
            success: success?,
            failed: failed?
``` 
result. Aren't we losing this info here?

jsuchome: Yes

jsuchome:
The change itself is correct, but I'm wondering - is there any reason to not have any specs for upgrade action?

abel-navarro:
Looks to be working when applied to cloud6 for the upgrade.

MaximilianMeister:
> The change itself is correct, but I'm wondering - is there any reason to not have any specs for upgrade action?

@jsuchome we have a spec for that: https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/spec/models/api/crowbar_spec.rb#L64

jsuchome:
@MaximilianMeister that's model only, but nothing for controller action. I was surprised that with the change of the possible return value in the controller action you do not need to change any spec as well.

MaximilianMeister:
> that's model only, but nothing for controller action. I was surprised that with the change of the possible return value in the controller action you do not need to change any spec as well.

@jsuchome ok, then we have https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/spec/controllers/api/crowbar_controller_spec.rb#L61
it just checks for success in terms of http code, so there was no change necessary

PR #708, opened 2016-10-07
Fix installation of UEFI nodes

dirkmueller:
For successful installation of a node that is booted via EFI
we need to create a /boot/efi partition.

(cherry picked from commit d3616e56431d0a392df1d5345ae5093a9c63a2c1)


mjura:
+1

mjura:
:+1:

PR #710, opened 2016-10-10
add api documentation for crowbar-init's upgrade related api

MaximilianMeister:
as crowbar-init doesn't support apipie there is now now way to nicely
document the endpoints that we have there.

and to not have the documentation in yet another place it might be
better to include it directly in the crowbar api docs.

the downside of that is that the methods don't live in the crowbar
codebase, and therefore need to be faked by adding an empty dummy
method

### DIFF
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -170,4 +170,69 @@ class Api::UpgradeController < ApiController
   def repocheck
     render json: Api::Upgrade.repocheck
   end
+
+  protected
+
+  api :POST, "/api/upgrade/new",
+    "Initialization of Crowbar during upgrade with creation of a new database.
+    NOTE: It is only possible to use this endpoint during the stage where crowbar-init is running."
+  header "Accept", "application/vnd.crowbar.v2.0+json", required: true
+  param :username, String, desc: "Username"
+  param :password, String, desc: "Password"
+  api_version "2.0"
+  example '
+  {
+    "database_setup": {
+      "success": true
+    },
+    "database_migration": {
+      "success": true
+    },
+    "schema_migration": {
+      "success": true
+    },
+    "crowbar_init": {
+      "success": false,
+      "body": {
+        "error": "crowbar_init: Failed to stop crowbar-init.service"
+      }
+    }
+  }
+  '
+  def dummy_crowbar_init_api_upgrade_new
+    # empty method to document crowbar-init's upgrade related API endpoints
+  end
+
+  api :POST, "/api/upgrade/connect",
+    "Initialization of Crowbar during upgrade with connection to an existing database.
+    NOTE: It is only possible to use this endpoint during the stage where crowbar-init is running."
+  header "Accept", "application/vnd.crowbar.v2.0+json", required: true
+  param :username, String, desc: "External database username"
+  param :password, String, desc: "External database password"
+  param :database, String, desc: "Database name"
+  param :host, String, desc: "External database host"
+  param :port, Integer, desc: "External database port"
+  api_version "2.0"
+  example '
+  {
+    "database_setup": {
+      "success": true
+    },
+    "database_migration": {
+      "success": true
+    },
+    "schema_migration": {
+      "success": true
+    },
+    "crowbar_init": {
+      "success": false,
+      "body": {
+        "error": "crowbar_init: Failed to stop crowbar-init.service"
+      }
+    }
+  }
+  '
+  def dummy_crowbar_init_api_upgrade_connect
+    # empty method to document crowbar-init's upgrade related API endpoints
+  end
 end
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -177,8 +177,20 @@ class Api::UpgradeController < ApiController
     "Initialization of Crowbar during upgrade with creation of a new database.
     NOTE: It is only possible to use this endpoint during the stage where crowbar-init is running."
   header "Accept", "application/vnd.crowbar.v2.0+json", required: true
-  param :username, String, desc: "Username"
-  param :password, String, desc: "Password"
+  param :username, /(?=^.{4,63}$)(?=^[a-zA-Z0-9_]*$)/,
+    desc: "Username
+      Min length: 4
+      Max length: 63
+      Only alphanumeric characters or underscores
+      Must begin with a letter [a-zA-Z] or underscore",
+    required: true
+  param :password, /(?=^.{4,63}$)(?=^[a-zA-Z0-9_]*$)(?=[a-zA-Z0-9_$&+,:;=?@#|'<>.^*()%!-]*$)/,
+    desc: "Password
+      Min length: 4
+      Max length: 63
+      Alphanumeric and special characters
+      Must begin with any alphanumeric character or underscore",
+    required: true
   api_version "2.0"
   example '
   {
@@ -199,6 +211,7 @@ class Api::UpgradeController < ApiController
     }
   }
   '
+  error 422, "Failed to initialize Crowbar, details are provided in the response hash"
   def dummy_crowbar_init_api_upgrade_new
     # empty method to document crowbar-init's upgrade related API endpoints
   end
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -170,4 +170,67 @@ def cancel
   def repocheck
     render json: Api::Upgrade.repocheck
   end
+
+  protected
+
+  api :POST, "/api/upgrade/new",
+    "Initialization of Crowbar during upgrade with creation of a new database"
```
rsalevsky: Can you add a hint that crowbar-init needs be running for this command?

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -170,4 +170,72 @@ def cancel
...
   end
+
+  protected
+
+  api :POST, "/api/upgrade/new",
+    "Initialization of Crowbar during upgrade with creation of a new database.
+    NOTE: It is only possible to use this endpoint during the stage where crowbar-init is running."
+  header "Accept", "application/vnd.crowbar.v2.0+json", required: true
+  param :username, String, desc: "Username"
+  param :password, String, desc: "Password"
```
santiph: For usarname and password:
Min length? Max length? Special characters required/allowed? Please update

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -170,4 +170,72 @@ def cancel
   def repocheck
     render json: Api::Upgrade.repocheck
   end
+
+  protected
+
+  api :POST, "/api/upgrade/new",
+    "Initialization of Crowbar during upgrade with creation of a new database.
```
santiph: It sounds like the database creation is part of this endpoint.
If that's not the case, please update the description

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -170,4 +170,72 @@ def cancel
...
+  end
+
+  api :POST, "/api/upgrade/connect",
+    "Initialization of Crowbar during upgrade with connection to an existing database.
+    NOTE: It is only possible to use this endpoint during the stage where crowbar-init is running."
+  header "Accept", "application/vnd.crowbar.v2.0+json", required: true
+  param :username, String, desc: "External database username"
+  param :password, String, desc: "External database password"
+  param :database, String, desc: "Database name"
+  param :host, String, desc: "External database host"
```
santiph: Ipv4? Ipv6? Hostname? Domain? What is supported here?

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -170,4 +170,72 @@ def cancel
...
+    # empty method to document crowbar-init's upgrade related API endpoints
+  end
+
+  api :POST, "/api/upgrade/connect",
+    "Initialization of Crowbar during upgrade with connection to an existing database.
+    NOTE: It is only possible to use this endpoint during the stage where crowbar-init is running."
+  header "Accept", "application/vnd.crowbar.v2.0+json", required: true
+  param :username, String, desc: "External database username"
+  param :password, String, desc: "External database password"
+  param :database, String, desc: "Database name"
```
santiph: For Username, Password and database name:
Min length? Max length? Special characters required/allowed? Please update

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -170,4 +170,72 @@ def cancel
...
+    },
+    "crowbar_init": {
+      "success": false,
+      "body": {
+        "error": "crowbar_init: Failed to stop crowbar-init.service"
+      }
+    }
+  }
+  '
+  error 406, "Connection to external database failed"
```
santiph: According to https://etherpad.nue.suse.com/p/cloud-upgrade-error-handling
there are several cases that could produce a failure in the connection:

- Host not found
- Incorrect port
- Wrong credentials: Return "wrong credentials", not "wrong user name" or "wrong password"
- Wrong database name

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -170,4 +170,72 @@ def cancel
...
+
+  api :POST, "/api/upgrade/new",
+    "Initialization of Crowbar during upgrade with creation of a new database.
+    NOTE: It is only possible to use this endpoint during the stage where crowbar-init is running."
+  header "Accept", "application/vnd.crowbar.v2.0+json", required: true
+  param :username, String, desc: "Username"
+  param :password, String, desc: "Password"
+  api_version "2.0"
+  example '
+  {
```
santiph: Didn't we agreed on split the endpoints between Database creation/connection and Crowbar initialisation?

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: it is the case (at least for now)

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: yes, we have to do this in a follow up, as this here is just to document what we have now.

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: I'll add that to the error description

MaximilianMeister:
@santiph I addressed your comments, and added a regex (from https://github.com/crowbar/crowbar-client/blob/master/lib/crowbar/client/mixin/database.rb) instead of a simple datatype to the database params

crowbar-init-doc-regex

santiph:
@MaximilianMeister Its good to have the regex validations, but we still need to get documentation for the parameters limitations and know the minimum and max length, if there is any special character required (number, symbols, upper or lowercase letters) so we can show a better message to the user than "Your input doesn't match ^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$", which is a terrible UX

MaximilianMeister:
> Its good to have the regex validations, but we still need to get documentation for the parameters limitations and know the minimum and max length, if there is any special character required (number, symbols, upper or lowercase letters) so we can show a better message to the user than "Your input doesn't match ^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-][a-zA-Z0-9]).)([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9-]*[A-Za-z0-9])$", which is a terrible UX

not sure how you want to do that, even if I add the text here in the apidoc, you will likely not read or parse it from the UI. don't you have to include that human readable form in the UI code?

santiph:
It's not to parse it, but to know it instead. We need to have those validation rules stored in the UI, but it's important for us to do so that we have documentation in place indicating such restrictions

MaximilianMeister:
@santiph right, it wouldn't hurt and looks more user friendly. thanks, fixed

param_desc

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -170,4 +170,115 @@ def cancel
...
+      Min length: 4
+      Max length: 63
+      Only alphanumeric characters or underscores
+      Must begin with a letter [a-zA-Z] or underscore",
+    required: true
+  param :password, /(?=^.{4,63}$)(?=^[a-zA-Z0-9_]*$)(?=[a-zA-Z0-9_$&+,:;=?@#|'<>.^*()%!-]*$)/,
+    desc: "Password
+      Min length: 4
+      Max length: 63
+      Alphanumeric and special characters
```
santiph: What are considered special characters here?
ä, ö, ü, ë, ñ?
Symbols?
Is there any restriction or everything is supported?

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -170,4 +170,115 @@ def cancel
...
+
+  api :POST, "/api/upgrade/connect",
+    "Initialization of Crowbar during upgrade with connection to an existing database.
+    NOTE: It is only possible to use this endpoint during the stage where crowbar-init is running."
+  header "Accept", "application/vnd.crowbar.v2.0+json", required: true
+  param :username, /(?=^.{4,63}$)(?=^[a-zA-Z0-9_]*$)/,
+    desc: "External database username
+      Min length: 4
+      Max length: 63
+      Only alphanumeric characters or underscores
```
santiph: and/or*

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -170,4 +170,115 @@ def cancel
...
+      Max length: 63
+      Alphanumeric and special characters
+      Must begin with any alphanumeric character or underscore",
+    required: true
+  param :host, /^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$/,
+    desc: "External database host, Ipv4 or Hostname
+      Min length: 4
+      Max length: 63
+      Numbers and period characters (only IPv4)
+      Hostnames:
```
santiph: Hostname vs Domains? Is there any difference we should care about?

santiph:
Thanks @MaximilianMeister 
API Documentation looks great now :)
I left a couple of comments more. I think it will be all before approve and merge it

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -170,4 +170,115 @@ def cancel
...
+       alphanumeric characters and hyphens
+       cannot start/end with digits or hyphen",
+    required: true
+  param :port, /(?=^.{1,5}$)(?=^[0-9]*$)/,
```
Itxaka: Dont we need to validate that this is a number lower than 65535?

MaximilianMeister: good point, though it would have been good to know that during https://github.com/crowbar/crowbar-client/pull/105 ;)

but here we are documenting the actual state, so first would be the fix in crowbar-client, and then another follow up for the api docs here.

santiph: +1

MaximilianMeister:

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: see https://github.com/crowbar/crowbar-client/pull/113

PR #711, opened 2016-10-10
[NotForMerge]UI to Allocate nodes from Rackscale systems

mmnelemane:
Provides a basic UI interface for RackScale systems:
 - Lists all the available Systems with their System IDs as a table of check-boxes
 - The UI allows the user to select one or more systems in the list for allocation
 - Clicking on the Allocate button creates a basic node object based on the data obtained through Redfish APIs from the RackScale simulator
 - The Nodes/Dashboard screen displays the list of allocated nodes and some basic info on the node.

NOTE: The UI need not be final and only done for easier test and a basic demo.

Here are a few sample screen-shot attachments.
screenshot_20161010_110908
screenshot_20161010_111006
screenshot_20161010_111030

This PR depends on https://github.com/crowbar/crowbar-core/pull/693


crowbar_framework/app/controllers/intelrsd_controller.rb
```diff
@@ -0,0 +1,245 @@
...
+  attr_reader :logger, :insecure
+
+#  def initialize()
+#    @redfish_client = RedfishHelper::RedfishClient.new('localhost', '8443')
+#    @node_object_list = []
+#  end
+
+  def show
+    @title = "Welcome to RackScale Design"
+    @redfish_client = RedfishHelper::RedfishClient.new('10.160.66.119', '8443')
```
Itxaka: Maybe we can initialise the client as part of the class so we dont have to create a client per method?

crowbar_framework/app/controllers/intelrsd_controller.rb
```diff
@@ -0,0 +1,245 @@
...
+    sys_list = get_all_systems()
+    p "SYSTEM LIST: #{sys_list}"
```
Itxaka: Rails.logger instead of p?

crowbar_framework/app/controllers/intelrsd_controller.rb
Itxaka: Also, missing getting the host+port from env variables.

PR #714, opened 2016-10-11
don't inherit API models from Tableless

aspiers:
These API models don't use any methods from `Tableless`, so don't inherit from it.

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -17,7 +17,7 @@
 require "open3"
 
 module Api
-  class Crowbar < Tableless
+  class Crowbar
     class << self
       def status
         {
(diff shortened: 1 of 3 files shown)


rsalevsky:
Nor sure, but I think we use the ActiveModel from this see https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/tableless.rb.

aspiers:
@rsalevsky Where is it used? I checked and couldn't see any evidence for this.

aspiers:
Still waiting for reviews on this.

rsalevsky:
Can we wait with merging this until we have the background processing PR merged? I want to test something first.

JanZerebecki:
This doesn't make sense for a discontinued product. Please reopen if you think otherwise.

PR #715, opened 2016-10-11
Fix test

nicolasbock:
The test operator `[` expects a value on the left hand side of `x -ge 8`.
However, if the `tput` command fails, for example because it doesn't recognize
the terminal type, then the expression fails. The `[[` operator is more
robust.

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/zzz-prompt.sh.erb
+++ b/chef/cookbooks/provisioner/templates/default/zzz-prompt.sh.erb
@@ -7,7 +7,7 @@ fi
 # Only set the crazy prompt on terminals that support colors
 # it looks horrible on monochromatic terminals
 
-if [ $(tput colors) -ge 8 ]; then
+if test $(tput colors 2> /dev/null || echo 0) -ge 8; then
     case $(readlink /proc/$$/exe 2>/dev/null) in
         */zsh)
             PS1=$'<%= @zsh_prompt_from_template.call %> '


dirkmueller:
I'm sort of ok with this but I'm not sure if [[]] works for non-bash shells. I would prefer to handle the case differently.

chef/cookbooks/provisioner/templates/default/zzz-prompt.sh.erb
```diff
@@ -7,7 +7,7 @@ fi
 # Only set the crazy prompt on terminals that support colors
 # it looks horrible on monochromatic terminals
 
-if [ $(tput colors) -ge 8 ]; then
+if [[ $(tput colors) -ge 8 ]]; then
```
dirkmueller: I think we need to also do 2>/dev/null here..

nicolasbock:
@dirkmueller I have updated the PR. I verified that the new logic works with `zsh`, `bash`, `ash`, `dash`, `ksh`, `pdkash`.

chef/cookbooks/provisioner/templates/default/zzz-prompt.sh.erb
```diff
@@ -7,7 +7,12 @@ fi
 # Only set the crazy prompt on terminals that support colors
 # it looks horrible on monochromatic terminals
 
-if [ $(tput colors) -ge 8 ]; then
+# Make sure that COLORS has a value even when tput does not recognize the
+# terminal type.
+COLORS=$(tput colors 2> /dev/null)
```
aspiers: Convention dictates that shell variables are only upper case iff they are either exported environment variables, or special variables set by the shell (e.g. `BASH_*`, `PWD` etc.) or used by the shell (e.g. `HOME`, `PATH`, `HISTFILE` etc.)

I don't think this falls into one of those categories, so I would suggest that it be made lowercase (i.e. `colors`).

nicolasbock: Good point. I updated the PR.

aspiers: Thanks!

chef/cookbooks/provisioner/templates/default/zzz-prompt.sh.erb
```diff
@@ -7,7 +7,12 @@ fi
 # Only set the crazy prompt on terminals that support colors
 # it looks horrible on monochromatic terminals
 
-if [ $(tput colors) -ge 8 ]; then
+# Make sure that colors has a value even when tput does not recognize the
+# terminal type.
+colors=$(tput colors 2> /dev/null)
```
dirkmueller: ugh, one ugly problem here is.. this is a global variable :-( can we do this without a named variable? imho there is no need to do that in multiple expressions

nicolasbock: Ok, ok, I updated the PR. It's much smoother now :)

PR #716, opened 2016-10-12
Script for deleting pacemaker resources

jsuchome:
We must delete old pacemaker resources from Cloud6 node
before the new ones get created from upgraded node.

See https://github.com/jsuchome/cloud-upgrade/blob/master/Step%2012.md , Step 4

### DIFF
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -57,7 +57,7 @@ new_alias = repo_alias.gsub("SP1", "SP2").gsub(node[:platform_version], target_p
 
 template "/usr/sbin/crowbar-upgrade-os.sh" do
   source "crowbar-upgrade-os.erb"
-  mode "0770"
+  mode "0755"
   owner "root"
   group "root"
   action :create
@@ -80,7 +80,7 @@ use_ha = node["run_list_map"].key? "pacemaker-cluster-member"
 
 template "/usr/sbin/crowbar-shutdown-services-before-upgrade.sh" do
   source "crowbar-shutdown-services-before-upgrade.sh.erb"
-  mode "0770"
+  mode "0755"
   owner "root"
   group "root"
   action :create
@@ -106,12 +106,23 @@ end
 
 template "/usr/sbin/crowbar-pre-upgrade.sh" do
   source "crowbar-pre-upgrade.sh.erb"
-  mode "0770"
+  mode "0755"
   owner "root"
   group "root"
   action :create
   variables(
     use_ha: use_ha,
     os_auth_url_v2: os_auth_url_v2
+  )
+end
+
+template "/usr/sbin/crowbar-delete-pacemaker-resources.sh" do
+  source "crowbar-delete-pacemaker-resources.sh.erb"
+  mode "0755"
+  owner "root"
+  group "root"
+  action :create
+  variables(
+    use_ha: use_ha
   )
 end
--- a/chef/cookbooks/crowbar/templates/default/crowbar-delete-pacemaker-resources.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-delete-pacemaker-resources.sh.erb
@@ -1,0 +1,70 @@
+#!/bin/bash
+#
+# Old pacemaker resources must be deleted in the old setup before they get created from upgraded node
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+UPGRADEDIR=/var/lib/crowbar/upgrade
+mkdir -p "`dirname "$LOGFILE"`"
+exec >>"$LOGFILE" 2>&1
+
+echo "Executing $BASH_SOURCE"
+
+set -x
+
+mkdir -p $UPGRADEDIR
+
+if [[ -f $UPGRADEDIR/crowbar-delete-pacemaker-resources-ok ]] ; then
+    echo "Pacemaker resources already successfully deleted"
+    exit 0
+fi
+
+rm -f $UPGRADEDIR/crowbar-delete-pacemaker-resources-failed
+
+<% if @use_ha %>
+
+echo "Stopping pacemaker resources..."
+
+# Services need to be stopped before deletion.
+# Otherwise we could have services starting up in unwanted locations while deleting constraints.
+
+for type in clone group ms primitive; do
+    for resource in $(crm configure show | awk "\$1 == \"$type\" && ! (\$2 ~ /drbd|stonith|vip-/) {print \$2}"); do
+        crm --wait resource stop $resource
+    done
+done
+
+ret=$?
+if [ $ret != 0 ] ; then
+    echo "Error occured during stopping resources."
+    echo $ret > $UPGRADEDIR/crowbar-delete-pacemaker-resources-failed
+    exit $ret
+fi
+
+echo "Deleting pacemaker resources..."
+
+crm_cmd=""
+for type in location clone group ms primitive; do
+    for resource in $(crm configure show | awk "\$1 == \"$type\" && ! (\$2 ~ /drbd|stonith|vip-/) {print \$2}"); do
+        printf -v crm_cmd "${crm_cmd}delete $resource\n"
+    done
+done
+
+# Delete the resources in single transaction - let pacemaker decide the ordering
+crm configure <<EOF
+$crm_cmd
+EOF
+
+ret=$?
+if [ $ret != 0 ] ; then
+    echo "Error occured during deleting resources"
+    echo $ret > $UPGRADEDIR/crowbar-delete-pacemaker-resources-failed
+    exit $ret
+fi
+
+<% else %>
+
+echo "No HA setup found..."
+
+<% end %>
+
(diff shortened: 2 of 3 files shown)


jsuchome:
[(13) S49P16 Orchestrate upgrade of controller nodes in HA setup (upgrade step 12 part 2)](https://trello.com/c/2PMGauJx/118-13-s49p16-orchestrate-upgrade-of-controller-nodes-in-ha-setup-upgrade-step-12-part-2)

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -115,3 +115,14 @@
     os_auth_url_v2: os_auth_url_v2
   )
 end
+
+template "/usr/sbin/crowbar-delete-pacemaker-resources.sh.erb" do
+  source "crowbar-delete-pacemaker-resources.sh.erb"
+  mode "0770"
```
vuntz: Why not 0755?

jsuchome: Or why not 0700? We only want to execute this by root anyway, so why 0755?

vuntz: it's just unusual for scripts to be unreadable if there's no reason for that.

vuntz: (so to be clear, no big deal, but unusual)

chef/cookbooks/crowbar/templates/default/crowbar-delete-pacemaker-resources.sh.erb
```diff
@@ -0,0 +1,36 @@
...
+
+for type in clone group ms primitive; do
+    for resource in $(crm configure show | awk "\$1 == \"$type\" && ! (\$2 ~ /drbd|stonith|vip-/) {print \$2}"); do
+        crm --wait resource stop $resource
+    done
+done
+
+echo "Deleting pacemaker resources..."
+
+for type in location clone group ms primitive; do
```
aspiers: What about `order` and `colocation`?

Also, there should be a comment explaining why it's crucial that services are stopped before deletion, in order to make it safe to remove constraints in separate singleton transactions before removing the corresponding services they refer to.

jsuchome: I haven't had any problems with `order` and `colocation`. Is it because it got deleted automatically with the resources they were referring too? Or because the order remains the same even with the new resources?

jsuchome: I've added the comment to the stop block. @aspiers, I've tried to simplify yours, I hope it is still correct

aspiers: > I haven't had any problems with order and colocation. Is it because it got deleted automatically with the resources they were referring too?

Possibly.  But the point is, we shouldn't do something which opens up the possibility of orphaned constraints, because we've already seen how that can give weird unwanted behaviour in the `g-haproxy` case.

jsuchome: Comment updated.
About collocations/order, I'll run the script this evening and check how the result looks like

jsuchome:
I've added one more commit that's not really relevant to the PR but does not belong anywhere else anyway...

jsuchome:
Now I've changed the script quite significantly. We should run it in the background, as it could take time and we should not execute long running scripts over ssh.

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -4,6 +4,7 @@
 
 LOGFILE=/var/log/crowbar/node-upgrade.log
 mkdir -p "`dirname "$LOGFILE"`"
+chmod 600 $LOGFILE
```
rhafer: Unless I am missing something ```$LOGFILE``` might not exist here yet, right?

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -57,7 +57,7 @@
 
 template "/usr/sbin/crowbar-upgrade-os.sh" do
   source "crowbar-upgrade-os.erb"
-  mode "0770"
+  mode "0775"
```
rhafer: Why not ```0755``` like everything else in ```/usr/sbin/```

Same for all the other scripts.

jsuchome: Sure, my fault

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
jsuchome: It probably does exist (from some previous script) but you are right I should move this to later part

jsuchome:
i've adapted the script to delete `order` constraints first (see the first commit)

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -13,7 +13,9 @@ set -x
 echo "Evacuating l3 agents out of this node..."
 
 if [[ -f /usr/bin/neutron-ha-tool ]] ; then
+    set -x
     source /root/.openrc
+    set +x
```
rhafer: I think you reversed the logic here. You need ```set +x``` before sourcing ```.openrc``` and ```set -x``` after it.

jsuchome: Ouch!

chef/cookbooks/crowbar/templates/default/crowbar-delete-pacemaker-resources.sh.erb
```diff
@@ -0,0 +1,71 @@
...
+<% if @use_ha %>
+
+echo "Stopping pacemaker resources..."
+
+# Delete order constraints before stopping the resources
+# so they do not cause starting of something that was already stopped.
+
+for order in $(crm configure show | grep ^order | cut -d " " -f2); do
+    crm configure delete $order
+done
```
jsuchome: Not  even this seems to help. After one execution of the script, keystone service is still running....

jsuchome:
So, with the help of Kristoffer Gronlund, I was able to adapt the script that now it really deletes what we want to, and does not start anything during deleting.

chef/cookbooks/crowbar/templates/default/crowbar-delete-pacemaker-resources.sh.erb
```diff
@@ -0,0 +1,72 @@
...
+    echo $ret > $UPGRADEDIR/pacemaker-resources-delete-failed
+    exit $ret
+fi
+
+echo "Deleting pacemaker resources..."
+
+crm_cmd=""
+for type in location clone group ms primitive; do
+    for resource in $(crm configure show | awk "\$1 == \"$type\" && ! (\$2 ~ /drbd|stonith|vip-/) {print \$2}"); do
+        crm configure delete $resource
```
aspiers: Shouldn't this line be dropped if you're doing it in a single transaction?

Also I think you could achieve the same thing without the inner `for` loop, e.g. (untested)

    crm_cmd="$crm_cmd"$( crm configure show | awk "\$1 == \"$type\" && ! (\$2 ~ /drbd|stonith|vip-/) {print \"delete \" \$2}" )

jsuchome: Oh, that's a bug indeed

jsuchome:
Double delete removed

jsuchome:
@aspiers I've deleted that double delete

chef/cookbooks/crowbar/templates/default/crowbar-delete-pacemaker-resources.sh.erb
```diff
@@ -0,0 +1,71 @@
...
+    exit $ret
+fi
+
+echo "Deleting pacemaker resources..."
+
+crm_cmd=""
+for type in location clone group ms primitive; do
+    for resource in $(crm configure show | awk "\$1 == \"$type\" && ! (\$2 ~ /drbd|stonith|vip-/) {print \$2}"); do
+        crm_cmd="$crm_cmd delete $resource
+"
```
rhafer: I think ```printf -v crm_cmd "${crm_cmd}delete $resource\n"``` would be better readable here (if it works). As it avoids the ugly line break.

jsuchome: I'd have to test it

jsuchome: I find Ralf's option more readable

PR #717, opened 2016-10-12
crowbar: Properly handle nodes rebooting at the end of chef

vuntz:
The reboot handler at the end of a chef run will cause /sbin/reboot,
which will result in the ssh session with chef-client to be suddenly
closed, resulting in what looks like a failure.

However, if we reach that reboot handler, then the chef run actually
completed since the handlers are run after the run itself.

This should fix the case of Xen compute nodes that cause the nova
proposal to fail to apply.

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1435,9 +1435,24 @@ class ServiceObject
       # check if the node is currently rebooting
       wait_for_reboot(node)
 
+      # don't use a cached node object here, as there might have been some chef
+      # run we were blocking on in the wait_for_chef_clients call before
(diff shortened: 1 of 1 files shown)


toabctl:
@vuntz hm. Why was that ever working before?

vuntz:
> @vuntz hm. Why was that ever working before?

As mentioned on IRC: we used to run chef-client a second time in case of errors. This got removed because it was evil (was hiding many issues); but this introduced this problem.

aspiers:
> As mentioned on IRC: we used to run chef-client a second time in case of errors. This got removed because it was evil (was hiding many issues)

Oh, nice!  I had either forgotten or never realised that we'd already made this change :-)

PR #718, opened 2016-10-12
[stable/3.0] fix cloud and sles versions in the admin repocheck

MaximilianMeister:
it seems that we haven't adapted the values correctly when doing the
backport


jsuchome:
gating failed in `testsetup`. This code is not really tested anywhere, could we merge? @rsalevsky ?

MaximilianMeister:
retriggered gating, lets see if it helps

PR #719, opened 2016-10-12
Added shared locks

jsuchome:
This will be needed for https://github.com/crowbar/crowbar-core/pull/692

### DIFF
--- a/crowbar_framework/lib/crowbar/lock.rb
+++ b/crowbar_framework/lib/crowbar/lock.rb
@@ -29,13 +29,14 @@ module Crowbar
     end
 
     def self.with_lock(options = {})
-      new(options).with_lock do
+      shared = options.delete(:shared) || false
+      new(options).with_lock(shared: shared) do
         yield
       end
     end
 
-    def with_lock
-      acquire
+    def with_lock(options = {})
+      acquire(options)
       yield if block_given?
     ensure
       release
--- a/crowbar_framework/lib/crowbar/lock/local_blocking.rb
+++ b/crowbar_framework/lib/crowbar/lock/local_blocking.rb
@@ -17,7 +17,7 @@
 
 module Crowbar
   class Lock::LocalBlocking < Lock
-    def acquire
+    def acquire(options = {})
       logger.debug("Acquire #{name} lock enter as uid #{Process.uid}")
       begin
         @file ||= File.new(path, File::RDWR | File::CREAT, 0644)
@@ -26,12 +26,15 @@ module Crowbar
         logger.error("cwd was #{Dir.getwd})")
         raise "Couldn't open #{path} for locking: #$!"
       end
-      logger.debug("Acquiring #{name} lock")
+      logger.debug("Acquiring #{name} lock with #{options}")
       count = 0
+      # specify different bits for shared vs. exclusive locks
+      bits = File::LOCK_NB
+      bits |= options[:shared] ? File::LOCK_SH : File::LOCK_EX
       loop do
         count += 1
         logger.debug("Lock #{path} attempt #{count}")
-        if @file.flock(File::LOCK_EX | File::LOCK_NB)
+        if @file.flock(bits)
           break
         end
         sleep 1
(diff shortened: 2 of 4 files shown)


jsuchome:
[(5) S49P4 Implement a state machine for the upgrade](https://trello.com/c/OWehKX6K/126-5-s49p4-implement-a-state-machine-for-the-upgrade)

crowbar_framework/spec/lib/crowbar/lock/local_blocking_spec.rb
```diff
@@ -26,6 +26,11 @@
       end
       expect(lock.locked?).to be false
     end
+  end
+
+  context "when a lock is acquired" do
+
```
aspiers: Extra blank line shouldn't be here.

jsuchome:
relevants commits squashed

crowbar_framework/lib/crowbar/lock/local_blocking.rb
```diff
@@ -17,21 +17,18 @@
...
-        logger.error("cwd was #{Dir.getwd})")
-        raise "Couldn't open #{path} for locking: #$!"
-      end
-      logger.debug("Acquiring #{name} lock")
+      ensure_lock_file_exists
+      logger.debug("Acquiring #{name} lock with #{options}")
       count = 0
+      # specify different bits for shared vs. exclusive locks
+      bits = File::LOCK_NB
+      bits |= options[:shared] ? File::LOCK_SH : File::LOCK_EX
```
MaximilianMeister: will that work when the option is missing? probably yes, but I think it is better to define a sane value above like
```ruby
def acquire(options = {})
  shared = options.fetch(:shared, false)
```

crowbar_framework/lib/crowbar/lock.rb
```diff
@@ -29,13 +29,14 @@ def initialize(options = {})
     end
 
     def self.with_lock(options = {})
-      new(options).with_lock do
+      shared = options.delete(:shared) || false
```
MaximilianMeister: why not `options.fetch(:shared, false)` ? we are using this in many places in crowbar btw

crowbar_framework/lib/crowbar/lock/local_blocking.rb
jsuchome: With `options[:shared]` being nil, the last part of ternary operator is used, so yes, it works.

crowbar_framework/spec/lib/crowbar/lock/local_blocking_spec.rb
aspiers: OK, this isn't a big enough reason for giving -1, but maybe worth mentioning this principle for future reference anyway: IMHO a PR should generally not add lines in one commit and remove them in a later commit, unless there is a good reason (e.g. a refactoring which affects both new and existing code).  Otherwise this can make the history (and consequently the review process) confusing.

jsuchome: I agree with the principle. But here it's not worth an extra work of changing existing commits.

crowbar_framework/lib/crowbar/lock/local_blocking.rb
```diff
@@ -54,5 +51,15 @@ def release
...
     end
+
+    private
+
+    def ensure_lock_file_exists
+      @file ||= File.new(path, File::RDWR | File::CREAT, 0o644)
+    rescue
+      logger.error("Couldn't open #{path} for locking: #$!")
+      logger.error("cwd was #{Dir.getwd})")
+      raise "Couldn't open #{path} for locking: #$!"
```
jsuchome: This was only refactoring this code out. I'm not sure what the possible user of this method expects on failure, so I think it's better to keep it the way it was

Itxaka:
LGTM +1

crowbar_framework/spec/lib/crowbar/lock/local_blocking_spec.rb
```diff
@@ -8,36 +8,77 @@
...
+      lock.acquire(options)
+      lock.release
+      expect(lock.locked?).to be false
+    end
+
+    it "can be acquired again after release" do
+      lock.acquire(options)
+      lock.release
+      lock.acquire(options)
+      expect(lock.locked?).to be true
```
MaximilianMeister: `lock.release`

jsuchome: It's realeased automatically in `after` block.

PR #721, opened 2016-10-13
Use the same key for Cloud product as with nodes repocheck call.

jsuchome:
Needs backport to 3.0

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -102,11 +102,11 @@ module Api
           } unless os_available
 
           cloud_available = repo_version_available?(products, "suse-openstack-cloud", "8")
-          ret[:cloud] = {
+          ret[:openstack] = {
             available: cloud_available,
             repos: {}
           }
-          ret[:cloud][:repos][admin_architecture.to_sym] = {
+          ret[:openstack][:repos][admin_architecture.to_sym] = {
             missing: ["SUSE OpenStack Cloud 8"]
           } unless cloud_available
         end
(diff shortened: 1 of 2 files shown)


Itxaka:
[(3) S49P14 bsc#999946 - [Upgrade] Check for repositories on admin server only checks for product](https://trello.com/c/sl8vHEaY/146-3-s49p14-bsc-999946-upgrade-check-for-repositories-on-admin-server-only-checks-for-product)

Itxaka:
+1

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -123,11 +123,11 @@ def repocheck
...
 
           cloud_available = repo_version_available?(products, "suse-openstack-cloud", "8")
-          ret[:cloud] = {
+          ret[:openstack] = {
             available: cloud_available,
             repos: {}
           }
-          ret[:cloud][:repos][admin_architecture.to_sym] = {
+          ret[:openstack][:repos][admin_architecture.to_sym] = {
             missing: ["SUSE OpenStack Cloud 8"]
```
Itxaka: names should be fixed like in the backport: https://github.com/crowbar/crowbar-core/pull/724
```
"SUSE-OpenStack-Cloud-8-Pool",
"SUSE-OpenStack-Cloud-8-Updates"
```

jsuchome: That backport is intentionally doing more, that's why it has 2 commits.

I do not think the other one is needed for master - this part is usable for 7-8 upgrade only and by that time we should not need to fake the responses, but have the backend repocheck method really return names of repositories that are missing

PR #723, opened 2016-10-13
Set proper versions to check for upgrade admin repos

Itxaka:
Sets the versions to check to suse 12.2 and openstack 7

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -83,7 +83,6 @@ module Api
       end
 
       def repocheck
-        # FIXME: once we start working on 7 to 8 upgrade we have to adapt the sles version
         zypper_stream = Hash.from_xml(
           `sudo /usr/bin/zypper-retry --xmlout products`
         )["stream"]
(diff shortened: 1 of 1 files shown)


Itxaka:
[(3) S49P14 bsc#999946 - [Upgrade] Check for repositories on admin server only checks for product](https://trello.com/c/sl8vHEaY/146-3-s49p14-bsc-999946-upgrade-check-for-repositories-on-admin-server-only-checks-for-product)

jsuchome:
No, these numbers are actually correct. The point is, admin server repo check is run from the non-upgraded Cloud, so for our case it's from Cloud6. This code is master, so Cloud7, and admin repo checks here are prepared for upgrading to Cloud8

Itxaka:
@jsuchome This is a PR against 3.0, cloud6. Im guessing that the original change was doing on master, then backported to 3.0 with no changes. This will never work on cloud6 to 7 upgrade!

jsuchome:
Ah, sorry, you are right.

But there's already an open PR for that... https://github.com/crowbar/crowbar-core/pull/718

Itxaka:
Ahh nice, thanks for pointing it out @jsuchome !

PR #725, opened 2016-10-13
Execute the post-upgrade actions for the node

jsuchome:
See https://github.com/jsuchome/cloud-upgrade/blob/master/Step%2012.md , step 6

Depends on
- ~~https://github.com/crowbar/crowbar-core/pull/716 for deleting resources~~ merged
- ~~https://github.com/crowbar/crowbar-ha/pull/152 for changing the cluster founder~~ merged

### DIFF
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -126,3 +126,15 @@ template "/usr/sbin/crowbar-delete-pacemaker-resources.sh" do
     use_ha: use_ha
   )
 end
+
+template "/usr/sbin/crowbar-post-upgrade.sh" do
+  source "crowbar-post-upgrade.sh.erb"
+  mode "0775"
+  owner "root"
+  group "root"
+  action :create
+  variables(
+    use_ha: use_ha,
+    has_drbd: use_ha && node.fetch("drbd", {})["rsc"]
+  )
+end
--- a/chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
@@ -1,0 +1,70 @@
+#!/bin/bash
+#
+# This does what is necessary directly after upgraded node is rebooted
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+UPGRADEDIR=/var/lib/crowbar/upgrade
+mkdir -p "`dirname "$LOGFILE"`"
+exec >>"$LOGFILE" 2>&1
+
+set -x
+
+mkdir -p $UPGRADEDIR
+
+if [[ -f $UPGRADEDIR/crowbar-post-upgrade-ok ]] ; then
+    echo "Post upgrade actions were already successfully executed"
+    exit 0
+fi
+
+<% if @use_ha %>
+
+<% if @has_drbd %>
+
+if [[ -f /sbin/drbdadm ]] ; then
+    echo "Recreating DRBD metadata after the upgrade to new DRBD version..."
+
+    # FIXME: workaround for bnc#1006105 : creating metatata works after 2nd run...
+    for i in 1 2; do
+        drbdadm -- --force create-md all
+
+        ret=$?
+        if [ $ret != 0 ] ; then
+            echo "Error occured during DRBD metadata upgrade"
+            echo $ret > $UPGRADEDIR/crowbar-post-upgrade-failed
+            exit $ret
+        fi
+
+        sleep 5
+    done
+fi
+
+# Starting pacemaker will cause DRBD service to start and synchronize with the other node
+echo "Starting pacemaker..."
+service pacemaker start
+
+ret=$?
+if [ $ret != 0 ] ; then
+    echo "Error occured during starting pacemaker service"
+    echo $ret > $UPGRADEDIR/crowbar-post-upgrade-failed
+    exit $ret
+fi
+
+if [[ -f /sbin/drbdadm ]] ; then
+    # Wait until DRBD is started
+    while crm resource status ms-drbd-{postgresql,rabbitmq} | grep "NOT running"; do
+        sleep 1
+    done
+    # Now wait after DRBD data are synced with the other node
+    while drbdadm status | grep replication; do
+        sleep 5
+    done
+fi
+
+<% end %>
+
+<% else %>
+
+echo "No HA setup found..."
+
+<% end %>
+touch $UPGRADEDIR/crowbar-post-upgrade-ok
--- a/chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
@@ -7,6 +7,8 @@ UPGRADEDIR=/var/lib/crowbar/upgrade
 mkdir -p "`dirname "$LOGFILE"`"
 exec >>"$LOGFILE" 2>&1
 
+echo "Executing $BASH_SOURCE"
+
 set -x
 
 mkdir -p $UPGRADEDIR
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -63,6 +63,35 @@ module Api
       false
     end
 
+    def wait_for_ssh_state(desired_state, action)
+      Timeout.timeout(300) do
+        loop do
+          ssh_status = @node.ssh_cmd("").first
+          break if desired_state == :up ? ssh_status == 200 : ssh_status != 200
+          sleep(5)
+        end
+      end
+      true
+    rescue Timeout::Error
+      save_error_state(
+        "Possible error at node #{@node.name}" \
+        "Node did not #{action} after 5 minutes of trying."
+      )
+      false
+    end
+
+    # Reboot the node and wait until it comes back online
+    def reboot_and_wait
+      ssh_status = @node.ssh_cmd("/sbin/reboot")
+      if ssh_status[0] != 200
+        save_error_state("Failed to reboot the machine. Could not ssh.")
+        return false
+      end
+
+      return false unless wait_for_ssh_state(:down, "reboot")
+      wait_for_ssh_state(:up, "come up")
+    end
+
     # Do the complete upgrade of one node
     def upgrade
       # FIXME: check the global status:
@@ -82,16 +111,7 @@ module Api
         return false
       end
 
-      # reboot the node after successful upgrade
-      ssh_status = @node.ssh_cmd("/sbin/reboot")
-      if ssh_status[0] != 200
-        save_error_state("Failed to reboot the machine. Could not ssh.")
-        return false
-      end
-
-      # FIXME: wait until node comes back
-
-      true
+      reboot_and_wait
     end
 
     # Disable "pre-upgrade" attribute for given node
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -20,12 +20,6 @@ module Api
       @node = NodeObject.find_node_by_name name
     end
 
-    def upgraded?
-      # FIXME: check this by looking at some upgraded-ok file on the node
-      current_platform = "#{@node[:platform]}-#{@node[:platform_version]}"
-      @upgraded ||= current_platform == @node[:target_platform]
-    end
-
     # execute script in background and wait for it to finish
     def execute_and_wait_for_finish(script, seconds)
       Rails.logger.info("Executing #{script} at #{@node.name}...")
@@ -99,7 +93,7 @@ module Api
 
       # this is just a fallback check, we should know by checking the global status that the action
       # should not be executed on already upgraded node
-      return true if upgraded?
+      return true if @node.file_exist? "/var/lib/crowbar/upgrade/node-upgraded-ok"
 
       unless pre_upgrade
         save_error_state("Error while executing pre upgrade script")
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -26,67 +26,41 @@ module Api
       @upgraded ||= current_platform == @node[:target_platform]
     end
 
+    # execute script in background and wait for it to finish
+    def execute_and_wait_for_finish(script, seconds)
+      Rails.logger.info("Executing #{script} at #{@node.name}...")
+      @node.wait_for_script_to_finish(script, seconds)
+      true
+    rescue StandardError => e
+      save_error_state(
+        e.message + "Check /var/log/crowbar/node-upgrade.log for details."
+      )
+      false
+    end
+
     def pre_upgrade
-      # TODO: save the global status info about this substep (we started prepare for upgrade)
-
-      # Migrate out the l3 agents and shut down pacemaker
-      script = "/usr/sbin/crowbar-pre-upgrade.sh"
-      ssh_status = @node.ssh_cmd(script)
-      if ssh_status[0] != 200
-        save_error_state("Executing of pre upgrade script has failed on node #{@node.name}.")
-        return false
+      if execute_and_wait_for_finish("/usr/sbin/crowbar-pre-upgrade.sh", 300)
+        save_node_state("Pre upgrade script run was successful.")
+        return true
       end
-
-      upgrade_ok = false
-      begin
-        Timeout.timeout(60) do
-          loop do
-            out = @node.run_ssh_cmd("test -e /var/lib/crowbar/upgrade/crowbar-pre-upgrade-ok")
-            if out[:exit_code].zero?
-              save_node_state("Pre upgrade script run was successful.")
-              upgrade_ok = true
-              break
-            end
-            out = @node.run_ssh_cmd("test -e /var/lib/crowbar/upgrade/crowbar-pre-upgrade-failed")
-            if out[:exit_code].zero?
-              save_error_state("Execution of pre upgrade script at node #{@node.name} has failed
-Check /var/log/crowbar/node-upgrade.log at #{@node.name} node for details.")
-              break
-            end
-            sleep(5)
-          end
-        end
-      rescue Timeout::Error
-        save_error_state("Possible error during pre-upgrade script execution.
-Action did not finish after 1 minute")
-        return false
-      end
-      upgrade_ok
+      false
     end
 
     def os_upgrade
-      # Upgrade one node
-      # TODO: save the global status info about this substep (we started upgrade of the node)
-      ssh_status = @node.ssh_cmd("/usr/sbin/crowbar-upgrade-os.sh")
-      if ssh_status[0] != 200
-        Rails.logger.error("Executing of os upgrade script has failed on node #{@node.name}.")
-        return false
+      if execute_and_wait_for_finish("/usr/sbin/crowbar-upgrade-os.sh", 600)
+        save_node_state("Package upgrade was successful.")
+        return true
       end
-      true
+      false
     end
 
+    # Execute post upgrade actions: prepare drbd and start pacemaker
     def post_upgrade
-      # TODO: save the global status info about this substep (we started post upgrade stuff)
-
-      # Join the cluster: start pacemaker and run selected recipes
-      script = "/usr/sbin/crowbar-post-upgrade.sh"
-      out = @node.run_ssh_cmd(script)
-      unless out[:exit_code].zero?
-        Rails.logger.error("Executing of post upgrade script has failed on node #{@node.name}.")
-        Rails.logger.error("Script location: #{script}")
-        Rails.logger.error("exit: #{out[:exit_code]}, out: #{out[:stdout]}, err: #{out[:stderr]}")
-        return false
+      if execute_and_wait_for_finish("/usr/sbin/crowbar-post-upgrade.sh", 300)
+        save_node_state("Post upgrade script run was successful.")
+        return true
       end
+      false
     end
 
     # Do the complete upgrade of one node
@@ -108,32 +82,6 @@ Action did not finish after 1 minute")
         return false
       end
 
-      # wait until the OS upgrade is finished
-      upgrade_failure = false
-      begin
-        Timeout.timeout(600) do
-          loop do
-            out = @node.run_ssh_cmd("test -e /var/lib/crowbar/upgrade/node-upgraded-ok")
-            if out[:exit_code].zero?
-              save_node_state("Package upgrade was successful.")
-              break
-            end
-            out = @node.run_ssh_cmd("test -e /var/lib/crowbar/upgrade/node-upgrade-failed")
-            if out[:exit_code].zero?
-              upgrade_failure = true
-              save_error_state("Installation of node #{@node.name} failed")
-              break
-            end
-            sleep(5)
-          end
-        end
-      rescue Timeout::Error
-        save_error_state("Error during upgrading node. Action did not finish after 10 minutes")
-        return false
-      end
-
-      return false if upgrade_failure
-
       # reboot the node after successful upgrade
       ssh_status = @node.ssh_cmd("/sbin/reboot")
       if ssh_status[0] != 200
@@ -143,10 +91,6 @@ Action did not finish after 1 minute")
 
       # FIXME: wait until node comes back
 
-      unless post_upgrade
-        save_error_state("Error while executing post upgrade script")
-        return false
-      end
       true
     end
 
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -76,8 +76,6 @@ Action did not finish after 1 minute")
     end
 
     def post_upgrade
-      # FIXME: so far, we have no post-upgrade script
-      return true
       # TODO: save the global status info about this substep (we started post upgrade stuff)
 
       # Join the cluster: start pacemaker and run selected recipes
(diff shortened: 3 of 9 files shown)


chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
```diff
@@ -0,0 +1,21 @@
...
+mkdir -p "`dirname "$LOGFILE"`"
+exec >>"$LOGFILE" 2>&1
+
+set -x
+
+<% if @use_ha %>
+
+echo "Unsetting the pre-upgrade node attribute..."
+
+crm node attribute $(hostname) set pre-upgrade false
```
jsuchome: No, that is wrong, pacemaker is not yet running...

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -115,3 +115,15 @@
...
+
+template "/usr/sbin/crowbar-post-upgrade.sh" do
+  source "crowbar-post-upgrade.sh.erb"
+  mode "0770"
+  owner "root"
+  group "root"
+  action :create
+  variables(
+    use_ha: use_ha,
+    have_drbd: use_ha && node.fetch("drbd", {})["rsc"]
```
abel-navarro: i'd name this `has_drbd` instead of `have_drbd`.

chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
```diff
@@ -0,0 +1,30 @@
...
+
+<% if @have_drbd %>
+
+if [[ -f /sbin/drbdadm ]] ; then
+    echo "Recreating DRBD metadata after the upgrade to new DRBD version..."
+    drbdadm -- --force create-md all
```
abel-navarro: this will be called whether it's drbd master or slave. Is this the correct behavior?

jsuchome:
[(13) S49P16 Orchestrate upgrade of controller nodes in HA setup (upgrade step 12 part 2)](https://trello.com/c/2PMGauJx/118-13-s49p16-orchestrate-upgrade-of-controller-nodes-in-ha-setup-upgrade-step-12-part-2)

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
aspiers: I agree

chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
jsuchome: So currently I'm executing it from `models/api/upgrade` ... still may be not the right place and time...

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -158,6 +161,18 @@ def upgrade_controller_nodes
...
         node_api.upgrade
 
+        # Explicitly mark node1 as the cluster founder
+        # this is one time action only (for each cluster)
+        Api::Pacemaker set_node_as_founder drbd_slave
+
+        # Remove "pre-upgrade" attribute from node1
+        # We must do it from the node where pacemaker is running
+        master_node_api = Api::Node.new drbd_master
+        master_node_api.remove_pre_upgrade_role_for drbd_slave
```
jsuchome: Maybe it would still be better to execute this action on the node where the role actually is present. But we need running pacemaker for that. So we might do it before we stop pacemaker, before the node is upgraded (in `node_api.upgrade`). However, we'd have to make sure services won't migrate to this node once we remove `pre-upgrade` attribute... so maybe we should put the node in the standby mode before stopping pacemaker and upgrading it?

chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
```diff
@@ -19,6 +19,15 @@ fi
...
 
+<% if @has_drbd %>
+
+if [[ -f /sbin/drbdadm ]] ; then
+    echo "Recreating DRBD metadata after the upgrade to new DRBD version..."
+    drbdadm -- --force create-md all
+fi
+
+<% end %>
+
```
rhafer: This is superfluous you already add exactly the same block in the previous commit.

crowbar_framework/app/models/api/node.rb
```diff
@@ -127,6 +127,31 @@ def upgrade
       true
     end
 
+    # Remove "pre-upgrade" attribute from given node
+    # We must do it from the node where pacemaker is running
+    def remove_pre_upgrade_role_for(name)
+      hostname = name.split(".").first
+      out = @node.run_ssh_cmd("crm node attribute #{hostname} set pre-upgrade false")
+      unless out[:exit_code].zero?
+        save_error_state("Changning the pre-upgrade role for #{hostname} from #{@node.name} failed")
```
rhafer: typo: "Changning"

crowbar_framework/app/models/api/upgrade.rb
rhafer: I am not sure why that would be better. At least on a first quick thought it sound more error prone.

chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
jsuchome: must be some rebase that went wrong

jsuchome:
We need to make post-upgrade action running in background as well...

chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
jsuchome: Fixed

crowbar_framework/app/models/api/node.rb
```diff
@@ -127,6 +127,31 @@ def upgrade
...
+      hostname = name.split(".").first
+      out = @node.run_ssh_cmd("crm node attribute #{hostname} set pre-upgrade false")
+      unless out[:exit_code].zero?
+        save_error_state("Channging the pre-upgrade role for #{hostname} from #{@node.name} failed")
+        return false
+      end
+      true
+    end
+
+    # Delete existing pacemaker resources, from other node in the cluster
```
aspiers: This code is not related to a particular node, so it doesn't make sense for it to be an instance method in the model.  Also, why is it in this model rather than in `Api::Upgrade`?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -213,6 +216,18 @@ def upgrade_controller_nodes
         # FIXME: save the global status information that this node is being upgraded
         node_api.upgrade
```
aspiers: Don't we need to check the return value of this?

crowbar_framework/app/models/api/upgrade.rb
aspiers: Yeah, I'm not sure why you are thinking of doing that either. What are the advantages?

Also, shouldn't you check the return value of this call?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -213,6 +216,18 @@ def upgrade_controller_nodes
...
+        # this is one time action only (for each cluster)
+        Api::Pacemaker.set_node_as_founder drbd_slave
+
+        # Remove "pre-upgrade" attribute from node1
+        # We must do it from the node where pacemaker is running
+        master_node_api = Api::Node.new drbd_master
+        master_node_api.remove_pre_upgrade_role_for drbd_slave
+
+        # FIXME: this should be one time action only (for each cluster)
+        master_node_api.delete_pacemaker_resources
```
aspiers: Again, need to check the return value.  (Or raise an exception on failure which gets caught.)

crowbar_framework/app/models/api/node.rb
```diff
@@ -142,11 +142,44 @@ def remove_pre_upgrade_role_for(name)
...
+      ssh_status = @node.ssh_cmd(script)
+      if ssh_status[0] != 200
+        save_error_state(
+          "Executing of 'delete pacemaker resources' has failed at node #{@node.name}.\
+Script location: #{script}."
+        )
+        return false
+      end
+
+      # wait until the the script is finished
```
aspiers: A comment stating the obvious is often an indication that the code can be split into a separate method.  In this case you could replace the comment with

    def await_pacemaker_resource_deletion

and then we'd have two easy-to-understand methods which both do one simple task well, rather than a longer and harder to understand method.

crowbar_framework/app/models/api/node.rb
```diff
@@ -142,11 +142,44 @@ def remove_pre_upgrade_role_for(name)
...
+          loop do
+            out = @node.run_ssh_cmd(
+              "test -e /var/lib/crowbar/upgrade/pacemaker-resources-delete-ok"
+            )
+            if out[:exit_code].zero?
+              save_node_state("Deleting pacemaker resources was successful.")
+              break
+            end
+            out = @node.run_ssh_cmd(
+              "test -e /var/lib/crowbar/upgrade/pacemaker-resources-delete-failed"
```
aspiers: Code duplication.  Would be cleaner to extend the model and then write `@node.file_exists("/var/lib/crowbar/...")`.

crowbar_framework/app/models/api/node.rb
```diff
@@ -142,11 +142,44 @@ def remove_pre_upgrade_role_for(name)
...
+            )
+            if out[:exit_code].zero?
+              upgrade_failure = true
+              save_error_state(
+                "Deleting of pacemaker resources at node #{@node.name} failed.\
```
aspiers: You can avoid breaking indentation here by replacing the `\` line continuation with simple string concatenation:

    save_error_state(
        "This is line one " +
        "and this is line two."
    )

crowbar_framework/app/models/api/node.rb
```diff
@@ -142,11 +142,44 @@ def remove_pre_upgrade_role_for(name)
...
+Check /var/log/crowbar/node-upgrade.log for details."
+              )
+              break
+            end
+            sleep(5)
+          end
+        end
+      rescue Timeout::Error
+        save_error_state("Error during deleting pacemaker resources.\
```
aspiers: Same here.

crowbar_framework/app/models/api/node.rb
```diff
@@ -127,6 +127,31 @@ def upgrade
       true
     end
 
+    # Remove "pre-upgrade" attribute from given node
+    # We must do it from the node where pacemaker is running
+    def remove_pre_upgrade_role_for(name)
+      hostname = name.split(".").first
+      out = @node.run_ssh_cmd("crm node attribute #{hostname} set pre-upgrade false")
+      unless out[:exit_code].zero?
+        save_error_state("Channging the pre-upgrade role for #{hostname} from #{@node.name} failed")
```
jsuchome: Already fixed

crowbar_framework/app/models/api/node.rb
jsuchome: Moving to Api:Upgrade...

jsuchome:
@aspiers I think I addressed all of your points

crowbar_framework/app/models/api/node.rb
```diff
@@ -110,43 +164,22 @@ def upgrade
...
+      true
+    end
 
-      # FIXME: wait until node comes back
-
-      unless post_upgrade
-        save_error_state("Error while executing post upgrade script")
+    # Remove "pre-upgrade" attribute from given node
+    # We must do it from the node where pacemaker is running
+    def remove_pre_upgrade_role_for(name)
```
jsuchome: This might be moved to Api::Pacemaker model...

crowbar_framework/app/models/api/node.rb
```diff
@@ -110,43 +164,22 @@ def upgrade
...
-      end
-
-      return false if upgrade_failure
+      return false unless reboot_and_wait
 
-      # reboot the node after successful upgrade
-      ssh_status = @node.ssh_cmd("/sbin/reboot")
-      if ssh_status[0] != 200
-        save_error_state("Failed to reboot the machine. Could not ssh.")
+      unless post_upgrade
```
jsuchome: Now I realized this is the wrong time to execute post upgrade actions (aka drbd sync and pacemaker start). Old resources need to be deleted (from the other node) before we start pacemaker...

crowbar_framework/app/models/api/node.rb
```diff
@@ -150,6 +150,18 @@ def upgrade
       true
     end
 
+    # Remove "pre-upgrade" attribute from given node
+    # We must do it from the node where pacemaker is running
+    def remove_pre_upgrade_role_for(name)
```
aspiers: It's not removing the attribute, it's setting it to `false`. That's the correct approach, but we should name the method correctly.  Also, it's an attribute not a role.  So I suggest `disable_pre_upgrade_attribute`.  The comment needs changing too.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -210,31 +210,102 @@ def nodes
...
+        # Explicitly mark node1 as the cluster founder
+        # and adapt DRBD config to the new founder situation.
+        # This shoudl be one time action only (for each cluster).
+        unless Api::Pacemaker.set_node_as_founder drbd_slave
+          save_error_state("Changing the cluster founder to #{drbd_slave} has failed")
+          return false
+        end
+
+        # Remove "pre-upgrade" attribute from node1
+        # We must do it from the node where pacemaker is running
```
aspiers: s/the/a/ - there could be more than one node currently running Pacemaker.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -210,31 +210,102 @@ def nodes
...
-        # FIXME: save the global status information that this node is being upgraded
-        node_api.upgrade
+        save_upgrade_state("Starting the upgrade of node #{drbd_slave}")
+        return false unless node_api.upgrade
+
+        # Explicitly mark node1 as the cluster founder
+        # and adapt DRBD config to the new founder situation.
+        # This shoudl be one time action only (for each cluster).
+        unless Api::Pacemaker.set_node_as_founder drbd_slave
+          save_error_state("Changing the cluster founder to #{drbd_slave} has failed")
```
aspiers: All these `return false` and `true` are suggesting to me that it would have been better to simply `raise` a custom `UpgradeFailure` exception when something goes wrong, and `rescue` it further up the stack.  Also that would allow for saving an error and raising the exception all within a single method call, rather than having to do them separately.  Something worth considering for the future (not in this PR).

crowbar_framework/app/models/node_object.rb
```diff
@@ -1064,6 +1064,30 @@ def ssh_cmd(cmd)
...
+  # Check for the presence of given file on the node
+  def file_exists?(file)
+    out = run_ssh_cmd("test -e #{file}")
+    out[:exit_code].zero?
+  end
+
+  # Waits until the background script finishes.
+  # We expect that the script generates two kinds of files to indicate success or failure
+  # Raises timeout exception if the waiting time exceedes 'seconds'
+  def wait_for_script_to_finish(seconds, ok_file, failed_file)
```
aspiers: Since it seems that `ok_file` and `failed_file` always have the identical path and prefix, why not just pass `script_name` instead?  Then you can calculate them inside this:

    base = "/var/lib/crowbar/upgrade/#{script_name}"
    ok_file = base + "-ok"
    failed_file = base + "-failed"

which eliminates duplication in the callers.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -210,31 +210,102 @@ def nodes
...
+            "Script location: #{script}."
+          )
+          return false
+        end
+
+        begin
+          finished_ok = node.wait_for_script_to_finish(
+            300,
+            "/var/lib/crowbar/upgrade/pacemaker-resources-delete-ok",
+            "/var/lib/crowbar/upgrade/pacemaker-resources-delete-failed"
```
aspiers: See my suggestion in the commit introducing `#wait_for_script_to_finish` for how to eliminate this duplication.

crowbar_framework/app/models/api/node.rb
```diff
@@ -72,7 +69,26 @@ def os_upgrade
         Rails.logger.error("Executing of os upgrade script has failed on node #{@node.name}.")
         return false
       end
-      true
+
+      begin
```
aspiers: There's a lot of duplicated code here.  Why not have a helper which calls `@node.ssh_cmd` and then `@node.wait_for_script_to_finish`?  You would just need to pass parameters for the script name, description, and failure / success messages.

crowbar_framework/app/models/api/node.rb
```diff
@@ -91,16 +91,36 @@ def os_upgrade
...
       # TODO: save the global status info about this substep (we started post upgrade stuff)
 
-      # Join the cluster: start pacemaker and run selected recipes
-      script = "/usr/sbin/crowbar-post-upgrade.sh"
-      out = @node.run_ssh_cmd(script)
-      unless out[:exit_code].zero?
-        Rails.logger.error("Executing of post upgrade script has failed on node #{@node.name}.")
-        Rails.logger.error("Script location: #{script}")
-        Rails.logger.error("exit: #{out[:exit_code]}, out: #{out[:stdout]}, err: #{out[:stderr]}")
+      ssh_status = @node.ssh_cmd("/usr/sbin/crowbar-post-upgrade.sh")
```
aspiers: We've got the same duplication here.

crowbar_framework/app/models/api/node.rb
```diff
@@ -125,6 +125,32 @@ def post_upgrade
...
+      if ssh_status[0] != 200
+        save_error_state("Failed to reboot the machine. Could not ssh.")
+        return false
+      end
+
+      begin
+        Timeout.timeout(300) do
+          loop do
+            ssh_status = @node.ssh_cmd("")
+            break if ssh_status[0] == 200
```
aspiers: I think you have to first wait for the `reboot` to kill `sshd`, and only then wait for `sshd` to come back post-reboot.  Otherwise I can imagine this passing straight through even before the node has rebooted.

crowbar_framework/app/models/api/node.rb
```diff
@@ -158,7 +152,7 @@ def upgrade
 
       # this is just a fallback check, we should know by checking the global status that the action
       # should not be executed on already upgraded node
-      return true if upgraded?
+      return true if @node.file_exists? "/var/lib/crowbar/upgrade/node-upgraded-ok"
```
aspiers: Along the same lines of my previous suggestion to calculate the path to these status files based on the script name, I think it would be nicer to do:

    return true if @node.step_completed? :admin_upgrade

crowbar_framework/app/models/api/node.rb
jsuchome: The problem is with those failure / success messages. I actually want to not only log a messages, but (in future) use a status library to update the current status of the upgrade. And I did not want to put such upgrade library call to the old `node_object.rb`

crowbar_framework/app/models/node_object.rb
jsuchome: Yes, it looks nicer, but relies on the fact that _ok_file and failed_file always have the identical path and prefix_ . It is true so far, but does it have to remain this way?

jsuchome: @aspiers What about this? https://gist.github.com/jsuchome/5a76ab1851e8c7934cd0c1a681f1a01e (diff)
https://gist.github.com/jsuchome/1e20c4a306bce77477014f6fe6a4a59b (new code)

aspiers: @jsuchome Looks good except for the FIXMEs.

aspiers: > It is true so far, but does it have to remain this way?

If we applied that argument everywhere, we would have duplicated code everywhere.  Better to do what's best for now, rather than worrying about a future which may never come.

crowbar_framework/app/models/api/node.rb
aspiers: Not sure I understand. A method called `#wait_for_script_to_finish` would just do that, not any extra stuff like updating upgrade status.  Any upgrade-specific stuff would be done in a separate method in one of the upgrade-specific classes. And it might make sense to build another upgrade-specific method which wraps both methods, e.g. `#upgrade_step_via_script` which calls `#wait_for_script_to_finish` and then `#end_step`.

jsuchome: I've already added a helper. However I do not understand why you keep mentioning `#end_step` here.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1064,6 +1064,43 @@ def ssh_cmd(cmd)
...
+  # Executes a script in background and Waits until it finishes.
+  # We expect that the script generates two kinds of files to indicate success or failure.
+  # Raise a timeout exception if the waiting time exceedes 'seconds'
+  def wait_for_script_to_finish(script, seconds)
+    ssh_status = ssh_cmd(script)
+    if ssh_status[0] != 200
+      raise "Executing of script #{script} has failed on node #{@node.name}."
+    end
+
+    base = script.split(".sh").first
```
aspiers: IIUC `script` is a full path, so this will mean `base` begins with something like `/usr/sbin/...` which is wrong.  That's why I suggested using `File.basename`.

crowbar_framework/app/models/api/node.rb
jsuchome: Does not seem relevant

crowbar_framework/app/models/api/node.rb
jsuchome: I've tried to adapt the code to address this issue. Not tested yet, though.

jsuchome: Update: yes, it seems to work. I iterated in the first loop twice before going to second one

crowbar_framework/app/models/node_object.rb
```diff
@@ -1064,6 +1064,43 @@ def ssh_cmd(cmd)
...
+  # We expect that the script generates two kinds of files to indicate success or failure.
+  # Raise a timeout exception if the waiting time exceedes 'seconds'
+  def wait_for_script_to_finish(script, seconds)
+    ssh_status = ssh_cmd(script)
+    if ssh_status[0] != 200
+      raise "Executing of script #{script} has failed on node #{@node.name}."
+    end
+
+    base = File.basename(script).split(".sh").first
+    ok_file = "/var/lib/crowbar/upgrade/" + base + "-ok"
```
jsuchome: ok

crowbar_framework/app/models/api/node.rb
```diff
@@ -26,67 +26,41 @@ def upgraded?
...
-            sleep(5)
-          end
-        end
-      rescue Timeout::Error
-        save_error_state("Possible error during pre-upgrade script execution.
-Action did not finish after 1 minute")
-        return false
+    def pre_upgrade
+      if execute_and_wait_for_finish("/usr/sbin/crowbar-pre-upgrade.sh", 300)
+        save_node_state("Pre upgrade script run was successful.")
```
jsuchome: Not really. All of these actions are part of `nodes` step, but they are different sub-steps. We haven't defined API for sub-steps yet, so there's only this preliminary `save_node_state` method which does not do anything so far

crowbar_framework/app/models/api/node.rb
jsuchome: This is not any `admin_upgrade` step, but all this is happening inside `nodes` step, as I wrote elsewhere. Indeed we'll need some API like this to check if particular sub-steps are finished, that's why the comment above this line says "this is just a fallback check"

crowbar_framework/app/models/api/node.rb
jsuchome: ok

crowbar_framework/app/models/node_object.rb
```diff
@@ -1064,6 +1064,43 @@ def ssh_cmd(cmd)
     [200, {}]
   end
 
+  # Check for the presence of given file on the node
+  def file_exists?(file)
```
MaximilianMeister: `exists` in Ruby is deprecated, so I'd rather call it `file_exist?`

crowbar_framework/app/models/node_object.rb
```diff
@@ -1064,6 +1064,43 @@ def ssh_cmd(cmd)
...
+
+    begin
+      Timeout.timeout(seconds) do
+        loop do
+          if file_exists? ok_file
+            break
+          end
+          if file_exists? failed_file
+            raise "Execution of script #{script} at node #{@node.name} has failed"
+          end
```
MaximilianMeister: maybe worth adding a log entry here that we're waiting for the script?

crowbar_framework/app/models/node_object.rb
jsuchome: But this is not invoking any Ruby method

crowbar_framework/app/models/node_object.rb
jsuchome: ok

crowbar_framework/app/models/node_object.rb
MaximilianMeister: that doesn't matter, the important bit here is that it is a ruby method

jsuchome: ok, changed

MaximilianMeister: maybe you forgot to push the change?

jsuchome: Ah, true :-( I've done it now

PR #728, opened 2016-10-17
Fix handling of devices without speed information

dirkmueller:
The previous code never triggered, as an empty list is true.

### DIFF
--- a/crowbar_framework/lib/crowbar/conduit_resolver.rb
+++ b/crowbar_framework/lib/crowbar/conduit_resolver.rb
@@ -106,7 +106,7 @@ module Crowbar
 
         sorted_ifs.each do |intf|
           speeds = cr_ohai_network[intf]["speeds"]
-          speeds = ["1g"] unless speeds # legacy object support
+          speeds = ["1g"] if speeds.nil? || speeds.empty? # legacy object support
           speeds.each do |speed|
             count = count_speed_map[speed] || 1
             result["#{speed}#{count}"] = intf


dirkmueller:
another variant of https://github.com/crowbar/crowbar-core/commit/3d7680c5a677a737f3e1028a86cd6f6d8b63e6d4

crowbar_framework/lib/crowbar/conduit_resolver.rb
```diff
@@ -106,7 +106,7 @@ def if_speed_remap
 
         sorted_ifs.each do |intf|
           speeds = cr_ohai_network[intf]["speeds"]
-          speeds = ["1g"] unless speeds # legacy object support
+          speeds = ["1g"] if speeds.nil? || speeds.empty? # legacy object support
```
rsalevsky: You can use `speeds.blank?`  here. 

http://api.rubyonrails.org/v4.2/classes/Object.html#method-i-blank-3F

dirkmueller: doesn't work from irb:

irb(main):003:0> [].blank?
NoMethodError: undefined method `blank?' for []:Array
        from (irb):3
        from /usr/bin/irb:11:in `<main>'

is that really supposed to work in crowbar_framework?

rhafer: ```blank?``` will not work here anyways. It's rails specific. This code can run outside rails (as it is also included in the barclamp chef recipe) via a symlink.

rsalevsky: We habe rails anyway installed so why not using it? 

Just require and it will work: `require "active_support/core_ext/object/blank"`

PR #729, opened 2016-10-17
Adapt the checks API to what pre-checks is expecting

jsuchome:
This is based on responses introduced in https://github.com/crowbar/crowbar-core/pull/672

Requires https://github.com/crowbar/crowbar-ha/pull/147 for HA presence check

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -74,27 +74,30 @@ module Api
         end
       end
 
-      def ceph_healthy?
-        ceph_node = NodeObject.find("roles:ceph-mon AND ceph_config_environment:*").first
-        return true if ceph_node.nil?
-        ssh_retval = ceph_node.run_ssh_cmd("LANG=C ceph health 2>&1")
-        unless ssh_retval[:stdout].include? "HEALTH_OK"
-          Rails.logger.warn("ceph cluster health check failed with #{ssh_retval[:stdout]}")
-          return false
+      def ceph_status
+        {}.tap do |ret|
+          ceph_node = NodeObject.find("roles:ceph-mon AND ceph_config_environment:*").first
+          return ret if ceph_node.nil?
+          ssh_retval = ceph_node.run_ssh_cmd("LANG=C ceph health 2>&1")
+          unless ssh_retval[:stdout].include? "HEALTH_OK"
+            ret["errors"] = [
+              "ceph cluster health check failed with #{ssh_retval[:stdout]}"
+            ]
+          end
         end
-        true
       end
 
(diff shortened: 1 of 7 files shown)


Itxaka:
to confirm, if there is no ceph nodes, then the response would be `ceph_healthy: {}` right? Sorry I cannot see it mentioned anywhere on #672 and from this PR it seems that would be the actual response.

jsuchome:
@Itxaka not sure if I understand. But if there's no ceph node, that the health check returns true (`ceph_healthy?` finds empty hash as a result of `ceph_status`)

Itxaka:
sorry @jsuchome I checked the other PR and mixed this one with the other on the response json and so on.

Ignore me :+1:

MaximilianMeister:

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -112,27 +112,30 @@ def repocheck
...
         ssh_retval = ceph_node.run_ssh_cmd("LANG=C ceph health 2>&1")
         unless ssh_retval[:stdout].include? "HEALTH_OK"
-          Rails.logger.warn("ceph cluster health check failed with #{ssh_retval[:stdout]}")
-          return false
+          ret["errors"] = [
+            "ceph cluster health check failed with #{ssh_retval[:stdout]}"
+          ]
         end
-        true
+        ret
```
MaximilianMeister: MINOR: again, for this cases where you're forced to explicitly return something, using `tap` is always a good option, as it always returns with it's block variable, whcih could be `ret`

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -199,17 +204,33 @@ def network_checks
...
+        @ceph_status ||= Api::Crowbar.ceph_status
+      end
+
       def ceph_healthy?
-        Api::Crowbar.ceph_healthy?
+        ceph_status.empty?
+      end
+
+      def ha_presence_status
+        @ha_presence_status ||= Api::Pacemaker.ha_presence_check
```
MaximilianMeister: I don't remember here, so pacemaker is always installed? meaning `Api::Pacemaker` is always available?

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -789,6 +789,9 @@ en:
         compute_resources_check:
           help:
             default: 'Make sure there is enough compute power'
+        ha_configured:
+          help:
+            defaults: 'Without HA setup, non-disruptive upgrade is not possible.'
```
MaximilianMeister: `s/defaults/default/`

PR #730, opened 2016-10-17
adapt api error responses for consistency

MaximilianMeister:
use the same format as we have in the prechecks, as this was already discussed as the best format

```ruby
code: {
  data: ... whatever data type ...,
  help: String # "this is how you might fix the error"
}
```


MaximilianMeister:
[(3) S49P7 Error reporting on crowbar level](https://trello.com/c/chzg85j4/142-3-s49p7-error-reporting-on-crowbar-level)

jsuchome:
Maybe @skazi0 would like to look at this from the UI side

PR #733, opened 2016-10-18
barclamp: Fix conduit_resolver from chef for s390x

rhafer:
For some reason s390x seems to need the require_relative, while it works
without that on the other archs.


rhafer:
The ci failure looks completely unrelated to me.

PR #735, opened 2016-10-19
Execute the pre-upgrade action in the background

jsuchome:


jsuchome:
5 minutes is probably too much. But I ran into problems that with current 15 seconds timeout, the script was killed before it executed.

I should probably pick something closer to 15 seconds than to 5 minutes

crowbar_framework/app/models/api/node.rb
```diff
@@ -31,14 +31,37 @@ def pre_upgrade
...
+            out = @node.run_ssh_cmd("test -e /var/lib/crowbar/upgrade/crowbar-pre-upgrade-ok")
+            if out[:exit_code].zero?
+              save_node_state("Pre upgrade script run was successful.")
+              upgrade_ok = true
+              break
+            end
+            out = @node.run_ssh_cmd("test -e /var/lib/crowbar/upgrade/crowbar-pre-upgrade-failed")
+            if out[:exit_code].zero?
+              save_error_state("Execution of pre upgrade script at node #{@node.name} has failed
+Check /var/log/crowbar/node-upgrade.log for details.")
```
rhafer: You should probably clarify that this log file is located on the node itself.

PR #736, opened 2016-10-19
add a status when running the prechecks

MaximilianMeister:

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -24,6 +24,10 @@ module Api
       end
 
       def checks
+        upgrade_status = ::Crowbar::UpgradeStatus.new
+        # the check for current_step means to allow running the step at any point in time
+        upgrade_status.start_step if upgrade_status.current_step == :upgrade_prechecks
+
         {}.tap do |ret|
           ret[:network_checks] = {
             required: true,
@@ -55,6 +59,15 @@ module Api
             passed: clusters_healthy?,
             errors: clusters_health_report_errors
           } if Api::Crowbar.addons.include?("ha")
+
+          return ret unless upgrade_status.current_step == :upgrade_prechecks
+
+          errors = ret.select { |_k, v| v[:required] && v[:errors].any? }.map { |_k, v| v[:errors] }
+          if errors.any?
+            upgrade_status.end_step(false, prechecks: errors)
+          else
+            upgrade_status.end_step
+          end
         end
       end
 


MaximilianMeister:
rebased to trigger a new gating (got stuck in testbuild)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -24,6 +24,9 @@ def status
       end
 
       def checks
+        upgrade_status = ::Crowbar::UpgradeStatus.new
+        upgrade_status.start_step
```
aspiers: Are you assuming that the step you are starting is `upgrade_prechecks`?  If so, can you really be sure of that, given that `::Crowbar::UpgradeStatus#initialize` checks for a progress file, and if it's already there, loads the state in from it?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -55,6 +58,13 @@ def checks
             passed: clusters_healthy?,
             errors: clusters_health_check_errors
           } if Api::Crowbar.addons.include?("ha")
+
+          errors = ret.select { |_k, v| v[:required] && v[:errors].any? }.map { |_k, v| v[:errors] }
+          if errors.any?
```
aspiers: Couldn't you just replace this whole `if` block with this?

    `upgrade_status.end_step(errors.any?, errors)`

And if so, to me that highlights the redundancy in the way the interface and state structures have been designed: they track both a list of errors, and whether there were any errors.  This needlessly complicates the code and also opens up the possibility of internal inconsistency, where these two values don't match (e.g. `success=false` and `errors={}` or vice-versa).

MaximilianMeister: thanks

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -24,6 +24,9 @@ def status
       end
 
       def checks
+        upgrade_status = ::Crowbar::UpgradeStatus.new
+        upgrade_status.start_step if upgrade_status.current_step == "upgrade_prechecks"
```
jsuchome: I think comment should be here explaining, that the check for current_step is there so we can allow executing this (and only this) step in any time

MaximilianMeister: added

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -24,6 +24,10 @@ def status
       end
 
       def checks
+        upgrade_status = ::Crowbar::UpgradeStatus.new
+        # the check for current_step means to allow running the step at any point in time
+        upgrade_status.start_step if upgrade_status.current_step == "upgrade_prechecks"
```
aspiers: This causes a race window to open in between the equality check and the call to `#start_step`.  This is almost exactly the same race condition I was predicting in [my earlier comment](https://github.com/crowbar/crowbar-core/pull/692#issuecomment-252298193):

> Also it seems strange that `#next_step!` doesn't return the next step. This means that the consumer has to call `#current_step` immediately after, which feels like it's introducing another race window in between advancing the step and finding out what the new step is in order to be able to start executing it.

The only difference is that here you are querying the current step *before* marking it as started, instead of after.

(Actually, there is another worse problem, which is that `current_step` returns the in-memory state which is not protected by a lock, and is consequently not thread-safe.  But we can fix that pretty easily by changing `#current_step` to always read from the JSON file and not cache in-memory.)

To do this correctly, the locking library needs to provide [an atomic test-and-set](https://en.wikipedia.org/wiki/Test-and-set) operation.  For example, the interface could be changed so that the caller (here) would do:

    upgrade_status.start_step("upgrade_prechecks")

and then the test and set would both happen inside the critical region protected by a single lock.  If the current step isn't `upgrade_prechecks` then this call would either raise an exception or return `nil` or `false` or something like that.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -55,6 +59,11 @@ def checks
             passed: clusters_healthy?,
             errors: clusters_health_check_errors
           } if Api::Crowbar.addons.include?("ha")
+
+          return ret unless upgrade_status.current_step == "upgrade_prechecks"
+
+          errors = ret.select { |_k, v| v[:required] && v[:errors].any? }.map { |_k, v| v[:errors] }
+          upgrade_status.end_step(errors.any?, errors)
```
aspiers: Just occurred to me that we should make the state library defend against one thread starting a step and other ending it.  This should never happen unless the caller is really buggy, but it's best to be safe.  However that is not related to this PR at all.

crowbar_framework/app/models/api/upgrade.rb
jsuchome: > `upgrade_status.start_step("upgrade_prechecks")`

@aspiers you had some PR for that already, but I cannot find it... 

Other than this (and https://github.com/aspiers/crowbar-core/pull/1) , do you think **this** PR (736) needs to be changed or could we proceed here and adapt the library later?

jsuchome:
[(5) S49P4 Implement a state machine for the upgrade](https://trello.com/c/OWehKX6K/126-5-s49p4-implement-a-state-machine-for-the-upgrade)

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -43,10 +43,14 @@ def upgrade!
...
 
         if upgrade_script_path.exist?
+          upgrade_status = ::Crowbar::UpgradeStatus.new
+          upgrade_status.start_step
           pid = spawn("sudo #{upgrade_script_path}")
           Process.detach(pid)
           Rails.logger.info("#{upgrade_script_path} executed with pid: #{pid}")
 
+          # we can't really call upgrade_status.end_step here yet as the upgrade is running
+          # in the background
```
skazi0: Where would it be called? `crowbar-init`? May be worth including this info in the comment.

MaximilianMeister: this is not yet sure or decided, so i havent hard coded this assumption into the code, but just in the commit message

rsalevsky: I think we have to do it at the end of the script.

jsuchome: I don't think so, the step is really finished after the system is up again, after the reboot

skazi0: Anyway, we need to decide it sooner or later and when we do, it would be good to note it here as this is an exception compared to other cases.

crowbar_framework/app/models/api/upgrade.rb
aspiers: IMHO the design is broken, but we could fix it quite quickly here. Alternatively we could keep adding new race conditions and new cards to fix them later, but is it really a good idea to keep accumulating technical debt which will need to be fixed before the release? I'd prefer to fix them now:

1. Make the code changes which make the tests in https://github.com/aspiers/crowbar-core/pull/1 green
2. Change the interface to implement [the atomic test-and-set operation described above](https://github.com/crowbar/crowbar-core/pull/736/files#r84703091)
3. Change this calling code to use that interface.

That should all be doable in 2-3 hours IMHO.

crowbar_framework/app/models/api/crowbar.rb
aspiers: It should be called by the code which finishes executing the step.  Maybe that will require the script calling the API to mark the step as done, in which case I guess we'd need a new endpoint.  Or the script could call a standalone Ruby script which uses the state persistence library to mark the step as done.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -90,6 +90,8 @@ def noderepocheck
       end
 
       def adminrepocheck
+        upgrade_status = ::Crowbar::UpgradeStatus.new
+        upgrade_status.start_step
```
aspiers: Same problem here with checking which step we're on in a thread-safe way.

crowbar_framework/app/controllers/api/backups_controller.rb
```diff
@@ -76,11 +76,19 @@ def show
   '
   error 422, "Failed to save backup, error details are provided in the response"
   def create
+    # TODO: move the upgrade related code to a new API endpoint in the upgrade_controller
+    upgrade_status = ::Crowbar::UpgradeStatus.new
+    upgrade_status.start_step if upgrade_status.current_step == :admin_backup
```
aspiers: Another race condition here.

crowbar_framework/app/controllers/api/backups_controller.rb
```diff
@@ -76,11 +76,19 @@ def show
...
     @backup = Api::Backup.new(backup_params)
 
     if @backup.save
+      upgrade_status.end_step if upgrade_status.current_step == :admin_backup
```
aspiers: And here.

crowbar_framework/app/controllers/api/backups_controller.rb
```diff
@@ -76,11 +76,19 @@ def show
...
       render json: @backup, status: :ok
     else
+      upgrade_status.end_step(
+        false,
+        admin_backup: @backup.errors.full_messages.first
+      ) if upgrade_status.current_step == :admin_backup
```
aspiers: And here.

MaximilianMeister:
I probably forgot to mention here that this PR was mostly to unblock UI testing. i just added the other commits besides the prechecks status for the ease of testability (to have some status for each step on cloud 6)

crowbar_framework/app/models/api/upgrade.rb
jsuchome: We need to call `upgrade_status.end_step(errors.empty?, errors)`, as empty == success

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -55,6 +59,11 @@ def checks
             passed: clusters_healthy?,
             errors: clusters_health_check_errors
           } if Api::Crowbar.addons.include?("ha")
+
+          return ret unless upgrade_status.current_step == "upgrade_prechecks"
```
jsuchome: Here you are checking for strings, and in other places for symbols. We need to check for symbols everywhere (esp. after https://github.com/crowbar/crowbar-core/pull/769)

jsuchome:
I know that @aspiers concerns are not addresed here, but I think it is probably better to merge this as it is and adapt the library API calls later, once they are ready.

crowbar_framework/app/models/api/upgrade.rb
jsuchome: @MaximilianMeister there's one more check for "upgrade_prechecks" (string, not symbol)

PR #737, opened 2016-10-19
symbolize all keys

MaximilianMeister:
otherwise we run into problems when trying to access hash keys

### needs a backport


rsalevsky:
Travis failed

aspiers:
This broke stuff which I've tried to fix in #752.  But thinking about it more, I really don't understand this PR at all. You wrote "otherwise we run into problems when trying to access hash keys" but I have no clue what kind of problems would be caused by accessing hash keys using strings, or why that would be fixed by switching to symbols. Again this highlights the need for verbose commit messages which [do not assume the reviewer knows as much as the submitter](https://slack.engineering/on-empathy-pull-requests-979e4257d158#.jbfj45i5w).

PR #738, opened 2016-10-19
[stable/3.0] adapt error messages of responses to be consistent

MaximilianMeister:
the error message format needs to be adapted to be consistent
over the other api responses that could throw an error

we are using the same format here as in the prechecks, as this
was discussed and designed properly

(cherry picked from commit b2bf06c56bb5fe7e69ed1f53023df2145c857370)

partially backports https://github.com/crowbar/crowbar-core/pull/730


MaximilianMeister:
[(3) S49P7 Error reporting on crowbar level](https://trello.com/c/chzg85j4/142-3-s49p7-error-reporting-on-crowbar-level)

PR #740, opened 2016-10-19
fix broken i18n

MaximilianMeister:
after adapting 8e9dee2 the i18n was falsely re-added, probably
due to copy and paste

see https://github.com/crowbar/crowbar-core/pull/672


Itxaka:
jenkins build went missing, retriggered manually

PR #741, opened 2016-10-19
Only load crowbar-init in production

rsalevsky:
We can't load crowbar-init in the development env via a Gemfile.

### DIFF
--- a/crowbar_framework/config/boot.rb
+++ b/crowbar_framework/config/boot.rb
@@ -108,10 +108,12 @@ else
   require "chef"
 end
 
-includes_path = Pathname.new("/var/lib/crowbar/includes/")
-if includes_path.directory?
-  includes_path.each_child(false) do |file|
-    next unless file.extname == ".rb"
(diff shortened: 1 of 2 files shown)


abel-navarro:
why only load crowbar-init in production?
who set `Rails.env.production`?

rsalevsky:
@abel-navarro We can't load crowbar-init in development because of the Gemfile restrictions. If we use a Gemfile bundle expect also deps for crowbar-init in the crowbar Gemfile but this is not the case so the loading of crowbar-init fails.

It's rails build in: http://api.rubyonrails.org/classes/Rails.html#method-c-env

MaximilianMeister:
> We can't load crowbar-init in development because of the Gemfile restrictions. If we use a Gemfile bundle expect also deps for crowbar-init in the crowbar Gemfile but this is not the case so the loading of crowbar-init fails.

would be good to have that in the commit message

rsalevsky:
@MaximilianMeister Updated the commit message.

PR #742, opened 2016-10-19
upgrade: workaround potential network issues during the upgrade

rhafer:
After evacuating routers and stopping the remaining services reset the
fail-mode of the ovs bridges to keep network connectivty during the
package upgrade.
https://bugzilla.suse.com/show_bug.cgi?id=1002734

Additionally adds a workaround for https://bugzilla.suse.com/show_bug.cgi?id=1011889 by forcing a wicked ifreload call in the preupgrade script.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
@@ -45,4 +45,10 @@ fi
 echo "No HA setup found..."
 
 <% end %>
+
+<% @bridges_to_reset.each do |bridge| %>
+echo "Resetting fail-mode on ovs-bridge <%= bridge %>"
+ovs-vsctl set-fail-mode <%= bridge %> standalone
+<% end %>
+
 touch $UPGRADEDIR/crowbar-pre-upgrade-ok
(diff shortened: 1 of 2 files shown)


chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -26,3 +26,8 @@ service pacemaker stop
 echo "No HA setup found..."
 
 <% end %>
+
+for bridge in $(ovs-vsctl --real list-br); do
```
rhafer: It should be enough (an probably safer) if we just do this for bridges which have ports connected to external devices (usually br-public and br-fixed). And only if the ml2 ovs driver is used. Probably easiest to figure out the bridge names in crowbar and set them template variables.

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -26,3 +26,8 @@ service pacemaker stop
...
+    echo "Resetting fail-mode on ovs-bridge $bridge"
+    ovs-vsctl del-fail-mode $bridge
```
abel-navarro: maybe we want to explicitly set the fail mode to *standalone* (`ovs-vsctl set-fail-mode bridge standalone`). This way openvswitchd can get over the missing controller (neutron agent).

rhafer:
> I think we still need to investigate whether upgrading OVS disconnects the admin network so we loose connectivity to the node. Otherwise these changes may not be necessary at all.

I checked that already. I does disconnect the admin network, when using the vlan driver and nova_fixed is same interface as admin.

jsuchome:
[networking after upgrade (bsc#1002734)](https://trello.com/c/07RTRTXu/178-networking-after-upgrade-bsc-1002734)

PR #743, opened 2016-10-20
Add background processing

rsalevsky:


Itxaka:
Everytime a scheduled/delayed job is stored on a database, a kitten dies :crying_cat_face: 

Are we really gonna store the jobs in a database? It doesnt scale, like at all.

Whats the purpose of the background processing? Whats gonna be used for? Small things? Big things? Moving every long task to the queue?

MaximilianMeister:
> Are we really gonna store the jobs in a database? It doesnt scale, like at all.

can you elaborate a bit more? not sure I fully understand the scaling issue.

> Whats the purpose of the background processing? Whats gonna be used for? Small things? Big things? Moving every long task to the queue?

I think it is planned to use this functionality for the nodes upgrade and maybe the services shutdown. so for tasks that can take a bit longer (on cloud 7+)

the idea behind this is we want to make use of something that is a well adopted (rails) standard from other developers that had the same problem and solved it with a lot of engineering.

rsalevsky:
Why it should not scale? It's just meta data which gets written if we schedule something and read via the scheduler. Basically it's the same what redis is doing but much simpler and for our purposes perfect. For now we only want to use it for some Upgrade API calls and later maybe for deploying proposals. That means very less workload and a redis based solution would be just using memory and cpu for idling the  most of the time.

Itxaka:
> Why it should not scale? It's just meta data which gets written if we schedule something and read via the scheduler.

For one, you have the polling vs pushing to the workers. On a database backend you have to keep polling the database to get new jobs, and this is time consuming, performance hitting and grows exponentially when scaling jobs *and* workers.

After that you have locking in a database (Although I believe postgres 9.5 has a SKIP LOCKED which can help for this) in which jobs need to keep locking rows, something that is nonexistent in other backends and again, hits performance.

And the last one is the size of the database, the more rows, the more slow the queries are, which again in other software more focused on queuing, its nonexistent.

> That means very less workload and a redis based solution would be just using memory and cpu for idling the most of the time.

was thinking more of a backend like zeromq or rabbitMQ which have a really small footprint when doing nothing and not a huge one when doing heavy work (unless you have a huge backlog, which is bad!)

> And it's just text so storing the jobs in the DB doesn't hurt at all, maybe it's sometimes even better because the job is not lost if the server crashes and the memory is gone.

IIRC, all focused backends for queues already save to disk from time to time so its nearly impossible to lose any jobs.

This is all thinking in scale, if this is gonna be for just a few tasks then it will be ok. If we believe in reusing the queue system to launch everything in the future (node status, node chef-runs) and you mix this with a hundred nodes, its gonna be slow, and Im sure there is other software that can plugin with rails that uses decent queues backends. 
This is all "My opinion:copyright:" of course, but is based on my experience with working with queues in a medium-to-low volume system (several hundred jobs per minute) :smiley:

rsalevsky:
You are definitely right for bigger use cases. But in our case we don't plan to do much via the queue so it should be fine. Also a quick google search didn't show any limitations and it is used it much bigger deployments and use cases. Also the database will be not an issue due to if the jobs succeeds  it will get deleted from the database. AFAIK there is no ActiveJob backend which is using rabbitmq or zeromq.

Itxaka:
Then for our use case sounds good :+1: 

> AFAIK there is no ActiveJob backend which is using rabbitmq or zeromq.

Good opportunity to port celery to ruby on rails? :stuck_out_tongue:

MaximilianMeister:
can you still add some explanation in the commit message as we talked about? usually I wouldn't mind too much, but as we introduce some new moving piece it might be good to write some history or reasoning about it

rsalevsky:
@MaximilianMeister Do we need this now? Otherwise I will do it later when I have time.

MaximilianMeister:
> Do we need this now? Otherwise I will do it later when I have time.

it's not really urgent, as we still have lots of things to do on cloud6

aspiers:
LGTM but I agree with @MaximilianMeister about the explanation in the commit message.

crowbar_framework/config/initializers/delayed_job_config.rb
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+Delayed::Worker.destroy_failed_jobs = false
+Delayed::Worker.delay_jobs = !Rails.env.test?
+Delayed::Worker.raise_signal_exceptions = :term
+if Rails.env.production?
+  Delayed::Worker.logger = Logger.new(File.join(ENV["CROWBAR_LOG_DIR"], "background_jobs.log"))
```
rsalevsky: https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/config/crowbar.env

PR #744, opened 2016-10-20
determine the best upgrade mode

MaximilianMeister:
none: when a required check fails
non-disruptive: when all required checks succeed
disruptive: when only some non-required checks fail

### related
* https://github.com/SUSE-Cloud/automation/pull/1378


MaximilianMeister:
[further adapt prechecks](https://trello.com/c/zxE4qqZF/154-further-adapt-prechecks)

PR #747, opened 2016-10-23
new recipe skipping mechanism to fix nova-ha-compute sync drift (bsc#1004758)

aspiers:
Add a new recipe skipping mechanism, and use it to fix sync marks drift during application of the `nova-ha-compute` role (bsc#1004758).

https://bugzilla.suse.com/show_bug.cgi?id=1004758

Requires https://github.com/crowbar/crowbar-ha/pull/163

### DIFF
--- a/chef/cookbooks/utils/libraries/role_recipe.rb
+++ b/chef/cookbooks/utils/libraries/role_recipe.rb
@@ -14,9 +14,42 @@
 #
 
 module CrowbarRoleRecipe
+  # Blacklist mechanism which uses role recipes to allow skipping of
+  # recipes in certain situations.
+  def self.skip_role(node, barclamp, role)
+    # If we're applying the nova-ha-compute batch, skip other roles to avoid
+    # chef-client drift between nodes (since sync marks are only reset before
+    # the first batch) - https://bugzilla.suse.com/show_bug.cgi?id=1004758
+    if node["crowbar"]["applying_for"].key?("nova") &&
+        node["crowbar"]["applying_for"]["nova"].include?("nova-ha-compute") &&
+        role != "nova-ha-compute"
+      return "not required in nova-ha-compute batch"
+    end
+
+    nil
+  end
+
   # Helper to decide whether a role should be run for a node, depending on the
   # state of the node.
   def self.node_state_valid_for_role?(node, barclamp, role)
+    # We always want deployer-client, both for heartbeat but also because it
+    # sets up the ability to use the barclamp library
+    return true if role == "deployer-client"
+
+    if node[:state] == "applying"
+      skip_reason = skip_role(node, barclamp, role)
+      if skip_reason
+        roles = node["crowbar"]["applying_for"].map { |k, v| v }.flatten.sort.uniq
+        if roles.include? role
+          raise "BUG: skip_role tried to skip #{role} whilst applying " \
+                "batch #{roles} for #{barclamp}"
+        end
+        Chef::Log.info("Skipping role #{role} whilst applying " \
+                       "batch #{roles} for #{barclamp}: #{skip_reason}")
+        return false
+      end
+    end
+
     if node.key? barclamp
       states_for_role = if node[barclamp].key? "element_states"
         # if nil, then this means all states are valid
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1290,21 +1290,16 @@ class ServiceObject
     batches.each do |roles, node_names|
       @logger.debug "Applying batch: #{node_names.inspect} for #{roles.inspect}"
 
-      threads = {}
-      node_names.each do |node_name|
-        next if node_attr_cache[node_name]["windows"]
-        ran_admin = true if node_attr_cache[node_name]["admin"]
-
-        pre_cached_nodes[node_name] ||= Node.find_by_name(node_name)
-        node = pre_cached_nodes[node_name]
-        node["crowbar"]["applying_for"] = {}
-        node["crowbar"]["applying_for"][@bc_name] = roles
-        node.save
-
-        filename = "#{ENV["CROWBAR_LOG_DIR"]}/chef-client/#{node_name}.log"
-        thread = run_remote_chef_client(node_name, "chef-client", filename)
-        threads[thread] = node_name
+      ran_admin = true if node_names.detect do |node_name|
+        node_attr_cache[node_name]["admin"]
       end
+
+      nodes_to_run = node_names.reject do |node_name|
+        node_attr_cache[node_name]["windows"]
+      end
+
+      threads = remote_chef_client_threads(nodes_to_run, pre_cached_nodes,
+                                           roles)
 
       # Invalidate cache as chef might have saved the nodes
       pre_cached_nodes = {}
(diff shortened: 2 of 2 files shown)


aspiers:
[P2: cloud-mkcloud7-job-ha-compute-x86_64](https://trello.com/c/ZKRtrPzC/227-p2-cloud-mkcloud7-job-ha-compute-x86-64)

MaximilianMeister:
@aspiers can you retrigger the gating somehow?

aspiers:
@MaximilianMeister I was wating for the first gate to trigger, so I could get to the `Rebuild` page and tweak the config, but it hasn't happened yet for some reason ...

aspiers:
But I tested locally and it seems to work - at least it skipped the right roles, and then failed later due to lack of support for libvirt STONITH in compute HA.

MaximilianMeister:
> I was wating for the first gate to trigger, so I could get to the Rebuild page and tweak the config, but it hasn't happened yet for some reason ...

@aspiers I see. I was wondering how to retrigger that manually in this case without force pushing (which is probably the easiest)

aspiers:
The [PR trigger job](https://ci.suse.de/job/cloud-crowbar-testbuild-pr/7191/) failed so I've triggered a rebuild of that.  Once that starts the mkcloud job then I can rebuild that using the compute HA scenario.

MaximilianMeister:
> The PR trigger job failed so I've triggered a rebuild of that. Once that starts the mkcloud job then I can rebuild that using the compute HA scenario.

@aspiers thanks, I triggered it in the meantime. thanks

aspiers:
@MaximilianMeister Cool thanks, looks good!

aspiers:
The ha_compute build failed because the merge of [crowbar/crowbar-openstack#568 (don't break whole cluster when not upgrading)](https://github.com/crowbar/crowbar-openstack/pull/568) had not yet taken effect. Rebuilding - should get further this time.

chef/cookbooks/utils/libraries/role_recipe.rb
```diff
@@ -14,9 +14,44 @@
...
+        if roles.include? role
+          raise "BUG: skip_role tried to skip #{role} whilst applying " \
+                "batch #{roles} for #{barclamp}"
+        end
+        Chef::Log.info("Skipping role #{role} whilst applying " \
+                       "batch #{roles} for #{barclamp}: #{skip_reason}")
+        return false
+      end
+
+      return true
```
vuntz: This is not fully correct: we should not return as some roles might no be run in the "applying" state. That's the case for ipmi-discover, for instance. So instead of returning, you should just keep running the rest of the code.

aspiers: @vuntz I don't follow you. Nodes have state like "applying". Roles don't.  And anyway, this `return true` is inside the `if node[:state] == "applying"` block, just like in the original version you wrote.  So if a node isn't in the `applying` state then it will already keep running.

vuntz: Well, my initial code was wrong ;-)

Let me try again: if you apply the IPMI barclamp, the ipmi-discover role recipe should be skipped. Assuming that we generalize the code here, then it will not be the case.

The point here is that the check that we add here is an additional check, but it should not cause the other checks to be skipped.

aspiers: I'm still struggling to understand you.

> if you apply the IPMI barclamp, the ipmi-discover role recipe should be skipped

Right, and currently `role_ipmi.rb` has:

    if CrowbarRoleRecipe.node_state_valid_for_role?(node, "ipmi", "ipmi")
      if ["discovering", "hardware-installing", "readying"].include? node[:state]
        include_recipe "ipmi::ipmi-discover"
      end

> Assuming that we generalize the code here,

What do you mean by "generalize the code"?  That's very vague.

> then it will not be the case.

It would, unless you are talking about modifying `role_ipmi.rb`?  Without any modifications to `#node_state_valid_for_role`, this role recipe will *never* run `ipmi-discover` on nodes in the `applying` state.

> The point is that the check that we add here is an additional check, and it should not cause the other checks to be skipped.

Why would it ever cause that?

aspiers:
@vuntz Also, I still don't understand what you were saying in Barcelona about libvirt STONITH for remote nodes.  The `ha_compute` gate is currently failing due to lack of a `libvirt` case in this code:

https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/nova/recipes/compute_ha.rb#L277

ISTR you pointing me to some other code which set remote STONITH modes to `manual`, but now I can't find it anywhere.

vuntz:
> @vuntz Also, I still don't understand what you were saying in Barcelona about libvirt STONITH for remote nodes. The ha_compute gate is currently failing due to lack of a libvirt case in this code:
> 
> https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/nova/recipes/compute_ha.rb#L277
> 
> ISTR you pointing me to some other code which set remote STONITH modes to manual, but now I can't find it anywhere.

It's in crowbar-ha, in the crowbar-pacemaker cookbook (there's probably a stonith.rb recipe), and I think in Barcelona, I realized that this code is only running for corosync nodes, while it should possibly also run for remote nodes.

chef/cookbooks/utils/libraries/role_recipe.rb
aspiers: We mumbled and I finally understand this - "passing it should not mean that all other checks can be ignored" refers to other checks which don't exist yet, but which could live further down in this method (i.e. not even including the ones which may be added to `#skip_role` in the future).

aspiers:
@vuntz Rebased, and dropped that `return true`.

aspiers:
@vuntz I'll rebase, but do you still need changes?

vuntz:
> @vuntz I'll rebase, but do you still need changes?

I don't think so, from a quick glance. I'll look deeper tomorrow. But we need https://github.com/crowbar/crowbar-ha/pull/163 to go in to get this to work.

aspiers:
@vuntz wrote:
> But we need crowbar/crowbar-ha#163 to go in to get this to work.

Why?  You also (IIUC) mentioned this [here](https://github.com/crowbar/crowbar-ha/pull/163#issuecomment-270604880) without saying why.  Perhaps you already told me the reason but if so I can't remember, and there is no obvious overlap between STONITH attributes affected by that PR and the attributes used by this one.

vuntz:
@aspiers we had discussed it: with this code, the translation for the stonith mode in the cookbook is not executed anymore, and the compute HA recipe depends on this translation being done (as it doesn't expect "libvirt", but "per_node"). That's why I moved it in the other PR to the rails app.

aspiers:
Got it, thanks - I vaguely remember now.

aspiers:
@vuntz Please take another look - now rebased with conflicts fixed, plus a couple of extra refactorings.

aspiers:
Hmm, apparently I messed something up. Will take a look.

aspiers:
@dirkmueller This should be good now - if the gates pass and we merge, it should fix a lot of the ha-compute job failures.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1430,6 +1438,23 @@ def add_role_to_instance_and_node(barclamp, instance, name, prop, role, newrole)
     true
   end
 
+  def remote_chef_client_threads(node_names, pre_cached_nodes, roles)
+    threads = {}
+    node_names.each do |node_name|
+      pre_cached_nodes[node_name] ||= NodeObject.find_by_name(node_name)
```
rsalevsky: NodeObject is obsoleted by Node

aspiers: @rsalevsky Fixed

PR #748, opened 2016-10-24
Always return the result of a check if ha is present

MaximilianMeister:
# needs backport


jsuchome:
I think we need to change `checks` to always return `:ha_configured` part

MaximilianMeister:
> Change looks fine but please get in the habit of writing proper commit messages. This would not be accepted in upstream OpenStack, and I think we should aim for the same standards in Crowbar.

you're right. I updated the message, as first gating got aborted too

aspiers:
@MaximilianMeister Great, thanks a lot :)

PR #749, opened 2016-10-24
Do not add the pre-upgrade attribute from the recipe

jsuchome:
When done from recipe, it could be possibly executed later when
upgraded node is doing its first chef-client run.

So let's rather do it explicitely at the point when we are preparing
all nodes for the upgrade.

(This partially reverts https://github.com/crowbar/crowbar-core/pull/702)

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -146,16 +146,10 @@ module Api
           }
         end
 
-        # Initiate the services shutdown by calling scripts on all nodes.
-        # For each cluster, it is enough to initiate the shutdown from one node (e.g. founder)
-        NodeObject.find("state:crowbar_upgrade AND pacemaker_founder:true").each do |node|
-          node.ssh_cmd("/usr/sbin/crowbar-shutdown-services-before-upgrade.sh")
-        end
-        # Shutdown the services for non clustered nodes
-        NodeObject.find("state:crowbar_upgrade AND NOT run_list_map:pacemaker-cluster-member").
-          each do |node|
-          node.ssh_cmd("/usr/sbin/crowbar-shutdown-services-before-upgrade.sh")
-        end
+        # Initiate the services shutdown for all nodes
+        NodeObject.find("state:crowbar_upgrade").each(
+          &:shutdown_services_before_upgrade
+        )
 
         {
           status: :ok,
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -73,11 +73,6 @@ when "prepare-os-upgrade"
 
   include_recipe "crowbar::prepare-upgrade-scripts"
 
-  execute "set pre-upgrade node attribute" do
-    command "crm node attribute $(hostname) set pre-upgrade true"
-    only_if { node.roles.include? "pacemaker-cluster-member" }
-  end
-
 when "openstack_shutdown"
 
   include_recipe "crowbar::stop-services-before-upgrade"
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -1103,6 +1103,20 @@ class NodeObject < ChefObject
     end
   end
 
+  def shutdown_services_before_upgrade
+    if @node.roles.include?("pacemaker-cluster-member")
+      # For all nodes in cluster, set the pre-upgrade attribute
+      ssh_cmd("crm node attribute $(hostname) set pre-upgrade true")
+      # Shutdown of services could be done from the founder node
+      if @node["pacemaker"] && @node["pacemaker"]["founder"]
+        ssh_cmd("/usr/sbin/crowbar-shutdown-services-before-upgrade.sh")
+      end
+    else
+      # For non clustered nodes, we need to shutdown the services at each node
+      ssh_cmd("/usr/sbin/crowbar-shutdown-services-before-upgrade.sh")
+    end
+  end
+
   def net_rpc_cmd(cmd)
     case cmd
     when :power_cycle
--- a/crowbar_framework/spec/models/api/upgrade_spec.rb
+++ b/crowbar_framework/spec/models/api/upgrade_spec.rb
@@ -114,38 +114,13 @@ describe Api::Upgrade do
         :prepare_nodes_for_os_upgrade
       ).and_return(true)
       allow(NodeObject).to(
-        receive(:find).with("state:crowbar_upgrade AND pacemaker_founder:true").
+        receive(:find).with("state:crowbar_upgrade").
         and_return([NodeObject.find_node_by_name("testing.crowbar.com")])
       )
-      allow(NodeObject).to(
-        receive(:find).with("state:crowbar_upgrade AND NOT run_list_map:pacemaker-cluster-member").
-        and_return([])
+      allow_any_instance_of(NodeObject).to(
+        receive(:shutdown_services_before_upgrade).
+        and_return(true)
       )
-      allow_any_instance_of(NodeObject).to receive(:ssh_cmd).with(
-        "/usr/sbin/crowbar-shutdown-services-before-upgrade.sh"
-      ).and_return(true)
-
-      expect(subject.class.services).to eq(
-        status: :ok,
-        message: ""
-      )
-    end
-
-    it "prepares and shuts down services on non clustered nodes" do
-      allow_any_instance_of(CrowbarService).to receive(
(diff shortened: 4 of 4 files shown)


aspiers:
I can't understand from the commit message why this change is necessary.

> When done from recipe, it could be possibly executed later when upgraded node is doing its first chef-client run.

What would the problem be in that case which we need to fix?

jsuchome:
>> When done from recipe, it could be possibly executed later when upgraded node is doing its first 
>> chef-client run.

> What would the problem be in that case which we need to fix?

That means that the recipe would add pre-upgrade attribute to the already upgraded node. 
Yes, we do not need to remove it from the recipe, we might ensure the recipe is only executed ones (by properly changing the "upgrade_step" for the node). But I think moving it out from recipe is more in line with other actions, because now we do most of the stuff during the upgrade by executing remote scripts anyway.

aspiers:
OK, now I understand, thanks.  In future please can you avoid over-estimating my intelligence and ability to understand PRs ;-)  Specifically, try to make the commit messages focus on the *why*, not the *how*, because that will make reviews a lot easier and quicker.  E.g. in this case the title could have been "don't set pre-upgrade attribute again after upgrade" or similar.

aspiers:
BTW I have not looked thought about this in detail yet, but it seems to me that there is a lot of business logic being placed in the models, when it should belong in the controllers.  I've mentioned this several times already, but please can you clarify the exact scope of each of these new models? i.e. what objects/concepts do they represent, and what is the expected state which they are tracking?

jsuchome:
>  Specifically, try to make the commit messages focus on the why, not the how, 

I've tried to change commit message one more time.

jsuchome:
> it seems to me that there is a lot of business logic being placed in the models, when it should
> belong in the controllers.

I'm not sure I'm the right one to answer that. There have been already accepted PR for new models, so now I'm trying to add new parts along with older ones.
Maybe that whole idea of putting new models and controllers under API was wrong.

MaximilianMeister:
> it seems to me that there is a lot of business logic being placed in the models, when it should belong in the controllers.

why should the business logic be in the controllers?

aspiers:
> > it seems to me that there is a lot of business logic being placed in the models, when it should belong in the controllers.
>
> why should the business logic be in the controllers?

I probably worded that wrong.  There is nothing wrong with having some business logic in the models. But when logic is more strongly associated with a particular API request ("do this upgrade step") than a particular model ("update the state for this node") then it probably makes sense to put it in the controller not the model.  In reality the code will be split between controller and model, and it's a question of finding the right balance.  But at the moment it feels like not too much thought has gone into that.

Also, the current `app/models/api/*.rb` models don't make much sense to me because it's not at all clear what they represent and what their scope / internal state is supposed to be.  [I have mentioned this](https://github.com/crowbar/crowbar-core/pull/725#pullrequestreview-5321539) a few times but still haven't heard an explanation.  If I understood that better then maybe I would also understand why certain code is in those models.

> > Maybe that whole idea of putting new models and controllers under API was wrong.
>
> why that? and this should ideally have been brought up during #555

Putting controllers under `api/` was not wrong and we did discuss that at great length in #555.

I struggle to understand why the models are under `api/`, for reasons [I already mentioned](https://github.com/crowbar/crowbar-core/pull/725#pullrequestreview-5321539).  Yes, I should have spotted that when reviewing #555 - sorry for missing that, I think at the time I was distracted by the API versioning debate.  Maybe I am missing some information and it's actually OK the way it is currently, but again I would really appreciate someone clarifying my above questions about what the models represent and what their scope / internal state is supposed to be.  If noone can do that then it's probably a sign that the design is not right yet.

aspiers:
BTW I'm still pretty uncomfortable with the amount of HA-specific code which is going into crowbar-core here :-/  Is there really no way to avoid this?

jsuchome:
> BTW I'm still pretty uncomfortable with the amount of HA-specific code which is going into 
> crowbar-core here :-/ Is there really no way to avoid this?

I'm not sure. From some higher level we need to distinguish the actions for HA and none HA setups.

abel-navarro:
Looks correct to me but do agree with @aspiers that we are adding maybe too much HA code into crowbar-core. We do not have upgrade models in crowbar-ha so I don't know how difficult would be to consider it doing it there, otherwise we can get lots of HA pollution.

jsuchome:
I understand the concern, however I do not see a nice way to do it via crowbar-ha. 
What do you think should be moved there, the part for 'pre-upgrade' attribute? Or even the call of pre upgrade script? But that one has also non-ha parts...

abel-navarro:
@jsuchome yes, the part that calls `crm node attribute` doesn't semantically belong here. But I also don't see a clean way to do it.

aspiers:
Maybe there isn't an easy clean way to move *all* the HA-specific stuff to `crowbar-ha`.  Probably at least *some* of it could be moved simply by extracting HA-specific stuff into to methods which then get called.

But it might be easier to figure out the right design if we had a clear foundation to base it on.  Currently I'm still waiting for an answer to my questions:

> please can you clarify the exact scope of each of these new models? i.e. what objects/concepts do they represent, and what is the expected state which they are tracking?

jsuchome:
> Currently I'm still waiting for an answer to my questions:

I think it was already discussed here, but anyway I think such discussion does not belong to this PR. Could you clarify it with @MaximilianMeister elsewhere?

aspiers:
> I think it was already discussed here

Where?  I didn't see it.

> but anyway I think such discussion does not belong to this PR. Could you clarify it with @MaximilianMeister elsewhere?

I'm fine discussing it elsewhere.  But we should involve everyone working on this code, not just me and @MaximilianMeister.  I suggest a 20 minute mumble, the outcome of which would most likely be filing a new story or extending an existing one.

jsuchome:
> Where? I didn't see it.

https://github.com/crowbar/crowbar-core/pull/749#issuecomment-255981805 and following comments

aspiers:
For the record, @MaximilianMeister @jsuchome and I had a productive mumble discussion and came up with the following:

https://etherpad.nue.suse.com/p/cloud-upgrade-models

We agreed that renaming the models and moving code around was a bit disruptive, so we should not do it immediately, but wait for a good time when there are not so many PRs open, and then do the renames/moves in the smallest number of PRs (just one if possible).

We also discussed what to do about cleanly separating HA code from non-HA code, and agreed that we can at least keep HA code in separate methods within the `crowbar-core` models, since it's easy to move methods into different files or even repos if it is clear what the method does.

Additionally if it is not too painful in terms of multiple PRs with dependencies between `crowbar-ha` and `crowbar-core`, we could try to move some of these HA methods into `crowbar-ha`, and just call them from methods in `crowbar-core`.

The ideal situation would be for Crowbar to have a proper plugin framework, with the ability to add hooks at any point within the Rails app.  But of course it doesn't and we don't have time for that now ...

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -147,14 +147,18 @@ def services
...
         # Initiate the services shutdown by calling scripts on all nodes.
-        # For each cluster, it is enough to initiate the shutdown from one node (e.g. founder)
-        NodeObject.find("state:crowbar_upgrade AND pacemaker_founder:true").each do |node|
-          node.ssh_cmd("/usr/sbin/crowbar-shutdown-services-before-upgrade.sh")
-        end
-        # Shutdown the services for non clustered nodes
-        NodeObject.find("state:crowbar_upgrade AND NOT run_list_map:pacemaker-cluster-member").
-          each do |node|
-          node.ssh_cmd("/usr/sbin/crowbar-shutdown-services-before-upgrade.sh")
+        NodeObject.find("state:crowbar_upgrade").each do |node|
```
aspiers: In light of the mumble discussion just now on smaller methods, it would be good to put this `#each` block into a method called `shutdown_services`. (Actually that would help even when forgetting about the HA / non-HA discussion.)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -147,14 +147,18 @@ def services
...
+          if node.roles.include?("pacemaker-cluster-member")
```
aspiers: Can we write the body of the `#each` block like this?

    node.shutdown_services

Then we can move the code for shutting down services on a single node into the `shutdown_services` instance method. Then later, HA nodes could implement service shutdown differently to non-HA nodes, either by subclasses (ugly), or by concerns (nicer).

crowbar_framework/app/models/api/upgrade.rb
jsuchome: So I placed it to the `node_object`, which I hope is good for now, but should be cleaned up once we have model for nodes that are being upgraded ...

PR #751, opened 2016-10-25
[stable/3.0] Backport various pre-check commits to 3.0

jsuchome:
Mostly it's backport of https://github.com/crowbar/crowbar-core/pull/729 + parts of https://github.com/crowbar/crowbar-core/pull/689/commits/905dbae00e267e21fdeda314eae76b11536e4e5d

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -153,6 +153,17 @@ module Api
         end
       end
 
+      # Check for presence of HA setup, which is a requirement for non-disruptive upgrade
+      def ha_presence_check
+        unless addon_installed? "ha"
+          return { errors: [I18n.t("api.upgrade.prechecks.ha_configured.not_installed")] }
+        end
+        founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
+        founders.empty? ?
+          { errors: [I18n.t("api.upgrade.prechecks.ha_configured.not_configured")] }
+          : {}
+      end
+
       protected
 
       def lib_path
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -127,27 +127,30 @@ module Api
         cluster_health.empty?
       end
 
-      def ceph_healthy?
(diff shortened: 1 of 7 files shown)


jsuchome:

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -202,11 +234,15 @@ def compute_resources_check_errors
...
         {
           compute_resources: {
-            data: [], # TODO: implement cluster health check errors
+            data: compute_resources_status[:errors],
             help: I18n.t("api.upgrade.prechecks.compute_resources_check.help.default")
           }
         }
       end
+
+      def ha_presence_check
```
MaximilianMeister: I think you have to either drop this method, as it is not used, or reference it from `ha_presence_status`

crowbar_framework/spec/models/api/crowbar_spec.rb
```diff
@@ -195,7 +195,48 @@
...
         and_return(exit_code: 0, stdout: "HEALTH_WARN", stderr: "")
       )
-      expect(subject.class.ceph_healthy?).to be false
+      expect(subject.class.ceph_status).to_not be_empty
+    end
+  end
+
+  context "with HA deployed" do
+    it "succeeds to confirm that HA is deployed" do
+
```
MaximilianMeister: can you drop this empty line? same below

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -127,27 +127,46 @@ def clusters_healthy?
...
+              "Found only one compute node of #{virt} type; non-disruptive upgrade is not possible"
+            )
+          end
         end
-        true
+      end
+
+      # Check for presence of HA setup, which is a requirement for non-disruptive upgrade
+      def ha_presence_check
+        ret = {}
```
MaximilianMeister: would you mind using `tap` here, as in `ceph_status` and `compute_resource_status`?

jsuchome: Actually... I'll make it the same as in master branch (https://github.com/crowbar/crowbar-ha/blob/master/crowbar_framework/app/models/api/pacemaker.rb#L21) which is probably different commit that I used here, so I wont call it cherry pick any more

skazi0:
lgtm :)

PR #754, opened 2016-10-26
Move missing methods onto the upgrade api model

Itxaka:
During the movement of the api repochecks/adminrepochecks
from crowbar to the upgrade namespace a couple of needed
methods were left that are necessary for the calls to work

Reference commit: https://github.com/crowbar/crowbar-core/commit/26371aac9a21cf384c3047f82f799b18744086bf

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -140,15 +140,6 @@ module Api
         Api::Node.repocheck(addon: addon)[addon]["available"]
       end
 
-      def repo_version_available?(products, product, version)
-        products.any? do |p|
-          p["version"] == version && p["name"] == product
-        end
-      end
-
-      def admin_architecture
-        NodeObject.admin_node.architecture
-      end
     end
   end
 end
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -352,6 +352,16 @@ module Api
           }
         }
       end
+
+      def repo_version_available?(products, product, version)
+        products.any? do |p|
+          p["version"] == version && p["name"] == product
+        end
+      end
+
+      def admin_architecture
+        NodeObject.admin_node.architecture
+      end
     end
   end
 end


Itxaka:
[(3) S49P19 #32 Admin Server Repo Checks: Service Integration](https://trello.com/c/oGniN5Lf/149-3-s49p19-32-admin-server-repo-checks-service-integration)

Itxaka:
@rsalevsky  That may be true, but this commit is about fixing both apis that were broken on a commit and unblocking the UI upgrade team and QA to continue working on the UI stuff :)

We can open a card to move this into an utils module afterwards, but I dont think is in the scope of the commit to refactor this right now.

rsalevsky:
@Itxaka Agree, so it was maybe also wrong to move the noderepocheck? This code is also very generic... So what do you think about moving this code also back to the crowbar model? I guess this will also make @aspiers happy.

Itxaka:
@rsalevsky I think thats a good idea, but then it should be redo to be more....generic, like just return a list of the repos available. They way that its now, its looking for some hardcoded repos so its only useful for the 6->7 upgrade. A more generic method would mean that the API response and the check for proper repos could be separated, one a generic way of getting repos and one more focused depending on the upgrade path.

But still, I think it should fall in a different card in order to track a good way of doing it while we unblock the UI and QA to keep testing the upgrade properly.

rsalevsky:
@Itxaka Just removing my review and merging is the wrong way to solve this problem. This makes the complete review process somehow obsolete... Specially if you can fix the problem with just changing one line of code.

Itxaka:
@rsalevsky you were asking for a refactor of the API methods, which is not on scope of this commit. This was fixing a commit done several days ago that broke a couple of api calls and was blocking 2 teams.

I totally agree on you that maybe we should have a look at those functions and move them, but I dont think it had anything to do with this bugfix.

aspiers:
For reference:

- I agree that code which is not upgrade-specific should (at some point) be moved outside upgrade-specific classes.
- I doubt there's a good reason to have *any* model under the `api` namespace (although it's fine for controllers).
- I strongly disagree with calling any model in the Crowbar codebase `...::Crowbar`, because that is so vague it's useless. Every model in the Crowbar codebase is obviously related to Crowbar, so it's pointless to reflect that in the final component of the class name.

IOW, IMHO `app/models/api/*.rb` should almost certainly be killed at some point, and especially `app/models/api/crowbar.rb`.  If I've missed some good reason why they should be kept then please let me know!

PR #756, opened 2016-10-27
Show the response from cluster health check in the upgrade prechecks

jsuchome:
I made it more special than the other checks, based on current return values of `Api::Pacemaker.health_report`. But maybe it is wrong and I should rather adapt `Api::Pacemaker.health_report` to return `:errors` just like the other checks ... ideas?

This is the kind of response that current `Api::Pacemaker.health_report` could return:

```
 # Simple check if HA clusters report some problems
 # If there are no problems, empty hash is returned.
 # If this fails, information about failed actions for each cluster founder is
 # returned in a hash that looks like this:
       {
           "crm_failures" => {
                   "node1" => "reason for crm status failure"
           },
           "failed_actions" => {
                   "node2" => "Failed actions at this node",
                   "node3" => "Failed actions at this node"

           }
       }
 # User has to manually clean pacemaker resources before proceeding with the upgrade.
```

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -53,7 +53,7 @@ module Api
           ret[:clusters_healthy] = {
             required: true,
             passed: clusters_healthy?,
-            errors: clusters_health_check_errors
+            errors: clusters_health_report_errors
           } if Api::Crowbar.addons.include?("ha")
         end
       end
@@ -268,9 +268,12 @@ module Api
         ha_presence_status.empty?
       end
 
+      def clusters_health_report
+        @clusters_health_report ||= Api::Pacemaker.health_report
+      end
+
       def clusters_healthy?
-        # FIXME: to be implemented
-        true
+        clusters_health_report.empty?
       end
 
       def compute_resources_status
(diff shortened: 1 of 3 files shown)


crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -786,7 +786,9 @@ en:
         ceph_health_check:
           help:
             default: 'Make sure Ceph is healthy'
-        clusters_health_check:
+        clusters_health:
+          crm_failures: '"crm status" failed at nodes %nodes.'
+          failed_actions: '"crm staus" is showing some failed actions at nodes %nodes.'
```
MaximilianMeister: status

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -331,15 +334,27 @@ def ha_presence_errors
...
+            nodes: crm_failures.join(",")
+          )
+        } if crm_failures
+        ret[:clusters_health_failed_actions] = {
+          data: failed_actions.values,
+          help: I18n.t(
+            "api.upgrade.prechecks.clusters_health.failed_actions",
+            nodes: failed_actions.keys.join(",")
+          )
         }
```
MaximilianMeister: `} if failed_actions` ?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -331,15 +334,27 @@ def ha_presence_errors
...
-            help: I18n.t("api.upgrade.prechecks.clusters_health_check.help.default")
-          }
+      def clusters_health_report_errors
+        ret = {}
+        return ret if clusters_healthy?
+
+        crm_failures = clusters_health_report["crm_failures"]
+        failed_actions = clusters_health_report["failed_actions"]
+        ret[:clusters_health_crm_failures] = {
+          data: crm_failures.values,
```
MaximilianMeister: somehow `data` could also be a hash that holds `clusters_health_crm_failures` and `clusters_health_failed_actions`, but then the help text wouldn't be that individual anymore...

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Oh yes

crowbar_framework/app/models/api/upgrade.rb
jsuchome: And than the `data` would nest more errors inside, unlike with other check results

PR #763, opened 2016-11-03
nfs-client: Install nfs-client package

vuntz:
Without this, we don't have mount.nfs, which is going to be an issue :-)

(cherry picked from commit eb16eddf78379b45d780306e154079cffb41111a)

Backport of https://github.com/crowbar/crowbar-core/pull/762


vuntz:
@s-t-e-v-e-n-k @aplanas since you reviewed the original PR, any +1 here? :-)

PR #764, opened 2016-11-04
fix ipmi vlan setup, convert vlan id to string

jdsn:
The VLAN id is expected as a string, but it comes in as an int.
So it needs to be converted to a string.


jdsn:
@kbaikov  and @jdsn reran the test on a fresh cloud first without the patch, then with patch.
Without the IPMI config fails and the node starts to reboot. With the patch the ipmi config can be done without any errors on the console and it can be allocated and installed.

PR #765, opened 2016-11-04
Do not remove the OS values from crowbar

jsuchome:
The OS values are updated directly by the upgrade script, see https://github.com/crowbar/crowbar/pull/2282

This is a revert of https://github.com/crowbar/crowbar-core/commit/5626f2fc324ddd3f701de2c4a3cd0873f1b9694c

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -43,13 +43,6 @@ module Api
         end
 
         if upgrade_script_path.exist?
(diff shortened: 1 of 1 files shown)


jsuchome:
[Bug 1006770 - during upgrade, default_os and target_platform set incorrectly](https://trello.com/c/TNUan7h4/167-bug-1006770-during-upgrade-default-os-and-target-platform-set-incorrectly)

jsuchome:
@MaximilianMeister @abel-navarro could I get a review?

PR #766, opened 2016-11-04
Thread-safe locking

aspiers:
Make the library thread-safe, so that multiple clients (web UI or `crowbarctl`) can simultaneously read/write upgrade state in a safe manner.

Still TODO:

- [ ] fix pending test
- [ ] change `#start_step` and `#end_step` to take a step name as mandatory parameter, and add (and test) checks within the critical code regions protected by locks which validate that it's permitted to start or end the given step

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade_status.rb
+++ b/crowbar_framework/lib/crowbar/upgrade_status.rb
@@ -6,8 +6,8 @@ module Crowbar
     # Return the current state of upgrade process.
     # We're keeping the information in the file so is accessible by
     # external applications and different crowbar versions.
-    def initialize(json_file = "/var/lib/crowbar/upgrade/progress.json")
-      @progress_file_path = Pathname.new(json_file)
+    def initialize(yaml_file = "/var/lib/crowbar/upgrade/progress.yml")
+      @progress_file_path = Pathname.new(yaml_file)
       load
     end
 
@@ -29,14 +29,14 @@ module Crowbar
       }
       # in 'steps', we save the information about each step that was executed
       @progress[:steps] = upgrade_steps_6_7.map do |step|
-        [step, { status: "pending" }]
+        [step, { status: :pending }]
       end.to_h
       save
     end
 
     def load!
       Crowbar::Lock::LocalBlocking.with_lock(shared: true) do
-        @progress = JSON.load(progress_file_path.read)
+        @progress = YAML.load(progress_file_path.read)
       end
     end
 
@@ -58,7 +58,7 @@ module Crowbar
           Rails.logger.warn("The step has already been started")
           return false
         end
-        progress[:steps][current_step][:status] = "running"
+        progress[:steps][current_step][:status] = :running
         save
       end
     end
@@ -80,13 +80,13 @@ module Crowbar
     def running?(step_name = nil)
       step = progress[:steps][step_name || current_step]
       return false unless step
-      step[:status] == "running"
+      step[:status] == :running
     end
 
     def pending?(step_name = nil)
       step = progress[:steps][step_name || current_step]
       return false unless step
-      step[:status] == "pending"
+      step[:status] == :pending
     end
 
     def finished?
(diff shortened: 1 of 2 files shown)


crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -97,7 +97,7 @@ def finished?
 
     def save
       progress_file_path.open("w") do |f|
-        f.write(JSON.pretty_generate(progress))
+        f.write(YAML.dump(progress))
```
MaximilianMeister: still not convinced why to use yaml for the state, at least not from the commit message. isn't readability a better reason to use yaml, not some internal restrictions with symbols?
and not that it matters for much for crowbar, but when we have to read the file from our automation (bash) wouldn't json be better?

jsuchome: Shoudln't we rather go back to symbolizing the string values, like we've had before in https://github.com/crowbar/crowbar-core/commit/b7909220d9f5572b4f695d1b81f093143e72cb8f (and reverted by https://github.com/crowbar/crowbar-core/commit/52589eee5b121fa198570967cb2bcfe830feb0e8)

aspiers: @MaximilianMeister Yes, sorry for the bad commit message - I was very tired when working on this and forgot to put justification. Will fix that!

@jsuchome No I don't think we should, otherwise I wouldn't have proposed this change to YAML.

aspiers:
Commit messages improved.  Thanks for keeping me honest on this, and please point out any time in future where I forget to write clear commit messages ;-)

crowbar_framework/lib/crowbar/upgrade_status.rb
jsuchome: Shouldn't it be in separate PR then? It's not related to thread locking and current library does not work with its mix of strings and symbols.

aspiers: No, it's related because it was necessary to fix this in order to get the test suite working.

jsuchome: Test suite might be working until now, but library is not, because at some point it checks for strings instead of symbols or vice versa. It is fixed by either deep_symbolize_keys call or by moving to YAML, but it is not specific to thread safe topic

MaximilianMeister:
@aspiers thanks for clarification, but TBH i am still not convinced with using yaml, and it feels to me we are only using it because of the serialization feature.

the question is also: why does the status file even need to be human readable?

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -1,27 +1,42 @@
 module Crowbar
   class UpgradeStatus
+    attr_reader :progress_file_path
     attr_accessor :progress
 
     # Return the current state of upgrade process.
     # We're keeping the information in the file so is accessible by
     # external applications and different crowbar versions.
-    def initialize
+    def initialize(json_file = "/var/lib/crowbar/upgrade/progress.json")
```
jsuchome: `yaml_file = "/var/lib/crowbar/upgrade/progress.yml`

jsuchome:

crowbar_framework/lib/crowbar/upgrade_status.rb
aspiers: Thanks! Will fix.

aspiers:
> @aspiers thanks for clarification, but TBH i am still not convinced with using yaml

Why not? And why do you think JSON would be better? Let's discuss facts not feelings ;-)

> and it feels to me we are only using it because of the serialization feature.

Not only, but avoiding the need to carefully convert certain parts of the data structure between string and symbol on read/write is certainly one advantage.

> the question is also: why does the status file even need to be human readable?

Because it makes debugging way easier!  One simple example:

    watch -d cat /var/lib/crowbar/upgrade/progress.yml

crowbar_framework/lib/crowbar/upgrade_status.rb
jsuchome: I extracted the commit here https://github.com/crowbar/crowbar-core/pull/769. I believe we need this before the thread safe locking is done to finally make use of the progress library and unblock UI and QA.

aspiers:
Fixed and rebased.

PR #767, opened 2016-11-04
Add guard support

aspiers:
It's indispensable for TDD coding.

### DIFF
--- a/crowbar_framework/Guardfile
+++ b/crowbar_framework/Guardfile
@@ -1,0 +1,50 @@
+#!/usr/bin/ruby
+#
+# More info at https://github.com/guard/guard#readme
+
+guard_opts = {
+  cmd: "bundle exec rspec --fail-fast",
+  all_on_start:   true,
+  all_after_pass: true,
+  failed_mode: :focus
+}
+
+DEBUG = false
+
+def reload(target)
+  puts "-> #{target}" if DEBUG
+  target
+end
+
+def all_specs;      reload "all_specs";      "spec"           end
+def library_specs;  reload "library_specs";  "spec/libraries" end
+def provider_specs; reload "provider_specs"; "spec/providers" end
+
+group :rspec do
+  guard :rspec, guard_opts do
+    watch('app/controllers/application_controller.rb') {
+      "spec/controllers"
+    }
+    watch(%r{^app/(.+)\.rb$})              { |m| "spec/#{m[1]}_spec.rb" }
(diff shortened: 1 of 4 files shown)


rsalevsky:
I would prefer if we integrate it into https://github.com/crowbar/crowbar/blob/master/Guardfile. Then we have all the guard code at one place.

MaximilianMeister:
> I would prefer if we integrate it into https://github.com/crowbar/crowbar/blob/master/Guardfile. Then we have all the guard code at one place.

but that will not work anymore if you just clone this repo... only works if you set up a development environment with the barclamps folder that has this repo here, which not everybody does.

crowbar_framework/Guardfile
```diff
@@ -0,0 +1,50 @@
...
+
+def all_specs;      reload "all_specs";      "spec"           end
+def library_specs;  reload "library_specs";  "spec/libraries" end
+def provider_specs; reload "provider_specs"; "spec/providers" end
+
+group :rspec do
+  guard :rspec, guard_opts do
+    watch('app/controllers/application_controller.rb') {
+      "spec/controllers"
+    }
```
MaximilianMeister: why do we mention the `application_controller` here explicitely?

aspiers: Because if the `application_controller` changes, all controllers need to be retested.

aspiers:
@rsalevsky I would prefer the opposite, because they have no goals or code in common.

MaximilianMeister (approved):

rsalevsky:
There are two reasons why I think it is better if we do it in the crowbar repo:

1.  We save the duplicated code for the gem definition.
2. You just need to run one command to sync to the remote and see the results. @MaximilianMeister I think if you do crowbar development then you should use also the dev setup. Otherwise I'm not sure how the code should be tested.

If you get another +1 feel free to merge it I don't want to block merging this if it helps making faster progress.

aspiers:
> We save the duplicated code for the gem definition.

I agree that's a benefit, but only a very small one which is outweighed by the other considerations below:

> You just need to run one command to sync to the remote and see the results.

This is mixing two completely different things.  One is syncing to a remote server, which is useful for testing deployed code, and the other is running unit tests directly from within a development environment.  Sometimes you might want to do one, sometimes the other, and sometimes both at the same time.  So it doesn't make sense to couple them together.

Other reasons:

- It must be possible to update the code and the corresponding Guardfile in a single PR, for example if you add a new subdirectory or filetype which is not covered by the existing patterns.  This is pretty important.
- It should be possible to quickly check out just `crowbar-core` and hack on it without requiring other repos.  This is not so important but still worth having.

aspiers:
Rebased

PR #768, opened 2016-11-07
Revert "fix breakage from conversion of upgrade state to symbols"

MaximilianMeister:
This partially reverts commit 52589eee5b121fa198570967cb2bcfe830feb0e8.

The fix satisfied the tests but it was actually wrong because in the
end_step method we were mixing symbols and strings again

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade_status.rb
+++ b/crowbar_framework/lib/crowbar/upgrade_status.rb
@@ -15,7 +15,7 @@ module Crowbar
       }
       if progress_file_path.exist?
         Crowbar::Lock::LocalBlocking.with_lock(shared: true) do
-          @progress = JSON.load(progress_file_path.read)
+          @progress = JSON.load(progress_file_path.read).deep_symbolize_keys
         end
       else
         # in 'steps', we save the information about each step that was executed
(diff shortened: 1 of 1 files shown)


jsuchome:
I'm fine with this one. @aspiers do you have any problems with symbolizing the keys from json? Looks better than using YAML, IMHO.

aspiers:
> Looks better than using YAML, IMHO.

@jsuchome Why?

jsuchome:
Seems to me like simple and straightforward

aspiers:
YAML is also simple and straightforward, and I have expanded the commit message in #766 to explain why using JSON is not a clean solution. What is the *exact* problem you see with using YAML?

jsuchome:
I do not have any technical argument if you want to hear that. Just that it looks we could have working library sooner with this quick fix.

MaximilianMeister:
> What is the exact problem you see with using YAML?

i posted something in https://github.com/crowbar/crowbar-core/pull/766#issuecomment-259066016 where it gets introduced

as @jsuchome said we just have a broken status lib right now, and I prefer to have it working and not put too much focus on 2 extra `to_sym` calls. Then I'd rather go for `Marshal` as I don't see any advantage in the status file being human readable.

jsuchome:
We're going with YAML instead.

PR #769, opened 2016-11-08
don't forget to save the failed state

jsuchome:
save also failed state for the steps

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade_status.rb
+++ b/crowbar_framework/lib/crowbar/upgrade_status.rb
@@ -74,6 +74,8 @@ module Crowbar
           errors: errors
         }
         next_step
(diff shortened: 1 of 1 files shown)


jsuchome:

aspiers:
I don't really understand the need to split this out of #766 when we could just merge #766 (it will cause merge conflicts).  But if you want to do that then that's fine by me.

However why did you truncate my commit message which explains exactly why YAML was chosen over JSON or Marshal?  This is worth explaining, and presumably Max agrees since he was (quite rightly) asking for the justification.

jsuchome:
> I don't really understand the need to split this out of #766 when we could just merge #766 

Mostly because I believed it would be faster.

PR #770, opened 2016-11-08
adding a status to the rest of the upgrade steps on cloud 6 side

MaximilianMeister:
### IMPORTANT TO BACKPORTING

the backups commit needs to be backported to use the old `Backups` model in the old backups_controller

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -94,6 +94,8 @@ module Api
       end
 
       def adminrepocheck
+        upgrade_status = ::Crowbar::UpgradeStatus.new
+        upgrade_status.start_step
         # FIXME: once we start working on 7 to 8 upgrade we have to adapt the sles version
         zypper_stream = Hash.from_xml(
           `sudo /usr/bin/zypper-retry --xmlout products`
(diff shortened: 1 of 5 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -115,6 +117,12 @@ def adminrepocheck
           ret[:openstack][:repos][admin_architecture.to_sym] = {
             missing: ["SUSE OpenStack Cloud 8"]
           } unless cloud_available
+
+          if ret.any? { |_k, v| !v["available"] }
```
jsuchome: `!v[:available]`

jsuchome:
> not to be merged yed, mainly to have something for testing

I think we could make it real PR, right? What's missing, other than addressing @aspiers' points about possible race conditions?

MaximilianMeister:
> I think we could make it real PR, right? What's missing, other than addressing @aspiers' points about possible race conditions?

mixing upgrade related code into the backups controller is not really wanted still. we would probably need a better solution for that.

MaximilianMeister:
we could either create a new endpoint in the `upgrade_controller` or just pass another parameter to `POST /api/crowbar/backup` like `with_upgrade`.

creating a new endpoint would require changing the url in the clients.
providin a `with_upgrade` parameter would at least make sure not to always run the statuslib code when we just want to create a "normal" backup during runtime (not in an upgrade case)

MaximilianMeister:
@jsuchome updated the backups api for an additional upgrade parameter

MaximilianMeister:
as described in https://github.com/crowbar/crowbar-client/pull/120#issuecomment-259929989 it is probably better to create a new endpoint in the `upgrade_controller`.

jsuchome:

MaximilianMeister:
[create a dedicated upgrade backup endpoint for the admin server](https://trello.com/c/HrwLujIn/176-create-a-dedicated-upgrade-backup-endpoint-for-the-admin-server)

PR #771, opened 2016-11-08
Add handling for prompts returned from zypper.

skazi0:
Sometimes zypper has some pending prompt which will be shown on every
invocation (e.g. repo key confirmation). In such case, repocheck fails
silently with 500 and log does not contain much useful information.

This change adds proper handling for such cases with the prompt text
returned as a response from the call.


jsuchome:
Hm, can't we just pass some argument to zypper to always confirm any questions?

skazi0:
Hmm, `--gpg-auto-import-keys` would (as the name implies) import the keys without asking which is not what every user would want. Same for `--auto-agree-with-licenses`.

The only one which makes sense is `--non-interactive` but it will silently reject the signing key and silenty skip the repo. If this is what we want then it will be an easier fix.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -90,12 +90,21 @@ def adminrepocheck
           if zypper_stream["message"] =~ /^System management is locked/
             return {
               status: :service_unavailable,
-              message: I18n.t(
+              error: I18n.t(
```
jsuchome: I know it's not part of your change, we have another problem, that even the normal repo check performed bellow does not return "error"...

jsuchome: OK, maybe not. I meant the result of admin repo check when the repos are actually not there, but we do not have to treat it like an error, just a special kind of result.

skazi0: Yes, these are not errors. They are just 'negative' result of checks.

PR #779, opened 2016-11-10
Fixes for cluster controllers upgrade

jsuchome:
This also includes fixes for both  DRBD and non-DRBD cases

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
@@ -53,12 +53,14 @@ fi
 
 if [[ -f /sbin/drbdadm ]] ; then
     # Wait until DRBD is started
-    while crm resource status ms-drbd-{postgresql,rabbitmq} | grep "NOT running"; do
+    while [[ $(drbdadm status | wc -c) -eq 0 ]]; do
+        echo "Waiting for DRBD to bring the devices up..."
         sleep 1
     done
     # Now wait after DRBD data are synced with the other node
-    while drbdadm status | grep replication; do
-        sleep 5
+    while drbdadm status | grep -q Inconsistent; do
+        echo "Replication status: `drbdadm status`"
+        sleep 10
     done
 fi
 
--- a/chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
@@ -69,4 +69,8 @@ fi
 echo "No HA setup found..."
 
 <% end %>
+
+# Move node to ready state and execute chef-client for the first time
+crowbar_join --start
+
 touch $UPGRADEDIR/crowbar-post-upgrade-ok
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -349,7 +349,16 @@ module Api
         # and the existing cluster has been cleaned up by deleting most of resources
         return false unless node_api.post_upgrade
 
-        # FIXME: if upgrade went well, continue with next node(s)
+        # FIXME: Delete router namespaces on non-upgraded node
+
+        save_upgrade_state("Starting the upgrade of node #{drbd_master}")
+        return false unless master_node_api.upgrade
+        return false unless master_node_api.post_upgrade
+        # Remove pre-upgrade attribute after chef-client run because pacemaker is already running
+        # and we want the configuration to be updated first
+        return false unless node_api.disable_pre_upgrade_attribute_for drbd_master
+
+        save_upgrade_state("Nodes in DRBD-based cluster successfully upgraded")
         true
       end
 
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -254,6 +254,53 @@ module Api
       def upgrade_controller_nodes
         drbd_nodes = NodeObject.find("drbd_rsc:*")
         return upgrade_drbd_nodes unless drbd_nodes.empty?
+
+        founder = NodeObject.find(
+          "state:crowbar_upgrade AND pacemaker_founder:true"
+        ).first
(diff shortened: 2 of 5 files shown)


jsuchome:
[(8) S50P7 (S49P16) Orchestrate upgrade of controller nodes in HA setup (upgrade step 12 part 2 -- HA/no-DRBD/one 3-node cluster)](https://trello.com/c/8ulaY77q/171-8-s50p7-s49p16-orchestrate-upgrade-of-controller-nodes-in-ha-setup-upgrade-step-12-part-2-ha-no-drbd-one-3-node-cluster)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -238,9 +238,65 @@ def prepare(options = {})
...
+        cluster_env = founder[:pacemaker][:config][:environment]
+
+        non_founder = NodeObject.find(
+          "state:crowbar_upgrade AND pacemaker_founder:false AND " \
+          "pacemaker_config_environment:#{cluster_env}"
+        ).first
+
+        # 1. upgrade the founder
+        save_upgrade_state("Starting the upgrade of node #{founder.name}")
+        founder_api = Api::Node.new founder.name
```
MaximilianMeister: why not name it `founder_node`?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -238,9 +238,65 @@ def prepare(options = {})
...
+        remaining_nodes = NodeObject.find(
+          "state:crowbar_upgrade AND pacemaker_config_environment:#{cluster_env}"
+        )
+        remaining_nodes.each do |node|
+          cmd = "for router_ns in `ip netns | grep qrouter`; do ip netns delete \\$router_ns; done"
+          node.run_ssh_cmd(cmd)
+        end
+        # 6. upgrade the rest of nodes in the same cluster
+        remaining_nodes.each do |node|
+
```
MaximilianMeister: empty line

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -238,9 +238,65 @@ def prepare(options = {})
...
+          name = node.name
+          save_upgrade_state("Starting the upgrade of node #{name}")
+
+          node_api = Api::Node.new name
```
MaximilianMeister: ...continuing about the naming from the above comment: i think at some point when we rename the models, the `api` doesn't fit anymore

jsuchome: It's probably true we should move the methods from api/node to node_object, but I'd do it in some later PR

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -293,7 +348,17 @@ def upgrade_controller_nodes
         # and the existing cluster has been cleaned up by deleting most of resources
         return false unless node_api.post_upgrade
 
-        # FIXME: if upgrade went well, continue with next node(s)
+        # Delete router namespaces on non-upgraded node
+        master_node_api.delete_router_namespace
```
jsuchome: So are we going back to the proposal with some extra pacemaker constraints for neutron-agents?

abel-navarro: @rhafer is is true this way we don't have much control over the router migration timing. What you propose sounds good, but I don't know how to face that implementation.

jsuchome: So... do we need a different action for `pacemaker_transaction "neutron ha tool"` in case of `CrowbarPacemakerHelper.being_upgraded?`

https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/neutron/recipes/network_agents_ha.rb#L203

If we are able to set the target role as stopped, we could be able to start it once we want...

jsuchome: In the recipe, that would be easy, I agree. But how do we create it easily once we need it? We did not plan second chef-client run on the first node

vuntz: Would it make sense to implement this as a follow-up PR? This would help move forward for the rest of the code.

FWIW, on how to implement this: with HA, controller nodes can ssh into each other, so we could have a script running from the upgraded controller that loops over the namespaces in the second controller, and do what @rhafer suggests.

vuntz: > In the recipe, that would be easy, I agree. But how do we create it easily once we need it? We did not plan second chef-client run on the first node

Could we possibly have it created by the last upgraded controller? I know it's kind of an exception as usually we let the founder create the resources, but it kind of makes sense here I think (without too much thinking, though).

jsuchome: > Would it make sense to implement this as a follow-up PR? This would help move forward for 
> the rest of the code.

Probably good idea. I'll remove the relevant commit for now.

jsuchome:
UPDATE: as per @vuntz suggestion, I've removed deleting of router namespaces for now. This should be solved by followup PR.
(For reference, it was using the code from https://github.com/crowbar/crowbar-core/pull/783)

chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
```diff
@@ -53,12 +53,13 @@ fi
 
 if [[ -f /sbin/drbdadm ]] ; then
     # Wait until DRBD is started
-    while crm resource status ms-drbd-{postgresql,rabbitmq} | grep "NOT running"; do
+    while [[ $(drbdadm status | wc -c) -eq 0 ]]; do
         sleep 1
```
rhafer: IMO we should log something here. Otherwise it will be hard to debug if for some reason a drbd is never started correctly. It should also probably timeout after a while.

jsuchome: 1. It actually logs the command command (see `set -x` at start of the script) .. although extra echo might help.

2. timeout is there, see the execution if the script via post-upgrade methid in api/node

jsuchome: Added extra log line

PR #782, opened 2016-11-14
Fix breaking change of codeclimate-test-reporter

rsalevsky:
Since codeclimate-test-reporter v1.0 is released our travis runs are
broken for master.

See https://github.com/codeclimate/ruby-test-reporter/blob/master/README.md


rsalevsky:
Tested locally but gating will not help here.

PR #783, opened 2016-11-14
Delete router namespaces to avoid router dups

abel-navarro:
Even if agents are migrated, we want to explicitly destroy the router
namespaces because in some upgrade cases, the routers remain active
when there are no valid destinations for the migration.

Example case, cluster with two nodes. The first one is upgraded, so
routers are moved to the second one. Once first node finishes the
upgrade, routers exist in the first and second nodes at once (because of
chef-client). Killing those routers in the pre-upgrade step of the
second node avoids that situation.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
@@ -25,6 +25,13 @@ if [[ -f /usr/bin/neutron-ha-tool ]] ; then
     OS_AUTH_URL="<%= @os_auth_url_v2 %>" neutron-ha-tool --l3-agent-evacuate $(hostname)
 fi
 
+# Even if agents are migrated, we want to explicitly destroy the router namespaces
+# because in some upgrade cases, the routers remain active when there are no
+# valid destinations for the migration
+for router_ns in `ip netns | grep qrouter`; do
+    ip netns delete $router_ns
+done
+
 # Shutdown pacemaker so the remaining OpenStack services are stopped
 # This is needed so that the zypper dup won't trigger any db migrations on package update
 service pacemaker stop


abel-navarro:
As I was discussing with @jsuchome this approach may not meet the requirements for the **non-disruptive upgrade**. Manual tests in a single node show connectivity loss to the VMs of more than 20 seconds.

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -25,6 +25,13 @@ if [[ -f /usr/bin/neutron-ha-tool ]] ; then
     OS_AUTH_URL="<%= @os_auth_url_v2 %>" neutron-ha-tool --l3-agent-evacuate $(hostname)
 fi
 
+# Even if agents are migrated, we want to explicitly destroy the router namespaces
+# because in some upgrade cases, the routers remain active when there are no
+# valid destinations for the migration
```
rhafer: @abel-navarro I don't quiet understand that. When does it happen that there are no valid destinations for a router migration?

jsuchome: @rhafer btw, I've moved this code to https://github.com/crowbar/crowbar-core/pull/779

abel-navarro: @rhafer first note that this PR may be superseded by https://github.com/crowbar/crowbar-core/pull/779

These two PRs address the same problem. In a three controller scenario, when you want to upgrade the first one, the routers are migrated to some other controller. This works flawlessly. Let's say we just upgraded first controller (node1) and routers migrated to the third controller (node3). So, right after node1 is upgraded, chef-clients runs there and Neutron and the agents are created. It happens then that node1 has a router but so does node3. This is a conflicting scenario that often leads to VM connectivity loss.

So the solution is to mercilessly kill the routers in the other controllers once the first one is upgraded. This is what this PR aimed to to, but I think it's better addresses in #779 . There's going to be for sure an interval where the two routers will co-exist so we expect misbehavior for some (hopefully short) time.

rhafer:
@abel-navarro Can this be closed?

abel-navarro:
@rhafer right, it will be tackled in #779 or in a new approach to be discussed.

PR #785, opened 2016-11-14
[stable/3.0] fix broken i18n

MaximilianMeister:
after adapting 8e9dee2 the i18n was falsely re-added, probably
due to copy and paste

(cherry picked from commit dfd77601dfd4cda30d2b07475bf527c9eb7959d3)


jsuchome:
I do not understand, why can't we translate those messages?

MaximilianMeister:
> I do not understand, why can't we translate those messages?

unfortunately we don't have a clear policy on i18n's, but IMO a lib should be reusable, which means we shouldn't use i18n

PR #787, opened 2016-11-15
[stable/3.0] make naming consistent

MaximilianMeister:


jsuchome:
Can't you make it part of https://github.com/crowbar/crowbar-core/pull/786 ? You're already renaming there...

MaximilianMeister:
> Can't you make it part of #786 ? You're already renaming there...

theoretically yes, but I decided to open this one because we might need to rethink the issue with the different check_id's decribed in https://github.com/crowbar/crowbar-client/pull/122

regarding:
> on top of that it should be insured that every error response is
wrapped into the same check_id as the top level check_id to provide
additional security.

this PR fixes one issue where the check_id differs, but we still have `clusters_health_crm_failures` and `clusters_health_failed_actions` which are special case here, that caused the bug that gets fixed with https://github.com/crowbar/crowbar-client/pull/122

i wanted to leave some place for discussion about that here in this PR as it fits the topic.

so long story short: should we think about joining `clusters_health_crm_failures` and `clusters_health_failed_actions` under one single check_id to have always the same reliable API response structure which is:

```
"check_id": {
  "passed": false,
  ...
  "errors": {
=>  "check_id": {
      "data": ...,
      "help": "Message"
    }
  }
}
```

or is it ok to have all checks returning the above structure, **except** `cluisters_health` which could look like:

```
"check_id": {
  "passed": false,
  ...
  "errors": {
=>  "another_check_id": {
      "data": ...,
      "help": "Message"
    }
  }
}
```

MaximilianMeister:
rebased

vuntz:
(Not giving a +1 because I don't know the impact here, but if you think it's needed, that's good enough for me to merge it)

PR #792, opened 2016-11-16
Check only for one drbd resource when picking first node for upgrade

jsuchome:
When both resources (postgres, rabbitmq) have different allocations
of master/slave nodes, previous call would not be able to find
the slave node.

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -306,15 +306,18 @@ module Api
       def upgrade_drbd_nodes
         save_upgrade_state("Upgrading controller nodes with DRBD-based storage")
 
-        # TODO: find the controller node that needs to be upgraded now
-        # First node to upgrade is DRBD slave
+        # Find the controller node that needs to be upgraded now
+
+        # First node to upgrade is DRBD slave. There might be more resources using DRBD backend
+        # but the Master/Slave distribution might be different in the among them.
+        # Therefore, we decide only by looking at the first DRBD resource we find in the cluster.
         drbd_slave = ""
         drbd_master = ""
         NodeObject.find(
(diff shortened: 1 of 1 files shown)


jsuchome:
[(13) S50P6 (S49P16) Orchestrate upgrade of controller nodes in HA setup (upgrade step 12 part 2 -- HA/DRBD/1cluster)](https://trello.com/c/2PMGauJx/118-13-s50p6-s49p16-orchestrate-upgrade-of-controller-nodes-in-ha-setup-upgrade-step-12-part-2-ha-drbd-1cluster)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -257,7 +257,7 @@ def upgrade_controller_nodes
         NodeObject.find(
           "state:crowbar_upgrade AND (roles:database-server OR roles:rabbitmq-server)"
         ).each do |db_node|
-          cmd = "LANG=C crm resource status ms-drbd-{postgresql,rabbitmq}\
+          cmd = "LANG=C crm resource status ms-drbd-postgresql\
```
rhafer: @jsuchome postgresql and rabbitmq could run in different clusters. With this change we would not find a node on the cluster just running rabbitmq. 
It also possible I think to configure one of them to use drbd but the other one for a different type of shared storage.

That being said, it seem this function seems to have some other issues if there are multiple drbd clusters. So maybe we should handle that in a separate PR/Trello card?

jsuchome: @rhafer what about this - take only one resource that is actually running there:

`LANG=C crm resource status ms-drbd-{postgresql,rabbitmq} | sort | head -n 2 | grep \\$(hostname)`

rhafer: Yeah, makes sense

PR #796, opened 2016-11-16
Replace deep_clone with deep_dup

rsalevsky:
Active Support has already a native object dup method.


rsalevsky:
Backport is https://github.com/crowbar/crowbar-core/pull/797

PR #797, opened 2016-11-16
[3.0] Replace deep_clone with deep_dup

rsalevsky:
Active Support has already a native object dup method.

(cherry picked from commit 79670c10536071d6e627e726e7f890c9fffc5a66)


aspiers:
@rsalevsky When you are backporting, please reference the original PR in the backported PR's description, and also prefix the PR title with something like `[3.0]`.  This helps a lot with dealing with a large volume of PRs.  Thanks!

aspiers:
There are also labels which we agreed to use for tracking back and forward porting.

PR #799, opened 2016-11-17
[stable/3.0] let adminrepocheck be called multiple times

MaximilianMeister:
it has to be handeled the same way as the prechecks, because the user
might call it multiple times, especially when the repochecks failed

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -101,7 +101,7 @@ module Api
 
       def adminrepocheck
         upgrade_status = ::Crowbar::UpgradeStatus.new
-        upgrade_status.start_step
+        upgrade_status.start_step if upgrade_status.current_step == :admin_repo_checks
         # FIXME: once we start working on 7 to 8 upgrade we have to adapt the sles version
         zypper_stream = Hash.from_xml(
           `sudo /usr/bin/zypper-retry --xmlout products`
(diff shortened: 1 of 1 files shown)


jsuchome:
I think it might be correct to enable calling it multiple times, but I'm unsure with the reasoning "especially when the repochecks failed". If it has failed, the `end_step` would not move the progress to the next step, so it's totally fine to call it again, right?

MaximilianMeister:
> think it might be correct to enable calling it multiple times, but I'm unsure with the reasoning "especially when the repochecks failed". If it has failed, the end_step would not move the progress to the next step, so it's totally fine to call it again, right?

yes, i'll update the commit message, you're right. the bug was actually this:

```
<skazi> mmeister: bug: admin repo checks move to next step even if they should not. e.g. run repo checks, they pass, current step is admin_upgrade, run checks again and current step changes to database
```

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -151,8 +151,12 @@ def adminrepocheck
               missing_repo_arch = v[:repos].keys.first.to_sym
               v[:repos][missing_repo_arch][:missing]
             end.flatten.join(", ")
-            upgrade_status.end_step(false, adminrepocheck: "Missing repositories: #{missing_repos}")
+            upgrade_status.end_step(
+              false,
+              adminrepocheck: "Missing repositories: #{missing_repos}"
+            ) if upgrade_status.current_step == :admin_repo_checks
           else
+            next unless upgrade_status.current_step == :admin_repo_checks
```
skazi0: Why are you using `next` here?

MaximilianMeister: it returns with `ret` from the {}.tap block. for a similar example look at https://github.com/crowbar/crowbar-init/commit/c04698738fb754baa3cec375df94d96c0421ca94

skazi0: aha, yet another ruby magic trick :)

PR #800, opened 2016-11-17
Uses World Wide Name as most persistent name for disks

sayalilunkad:


vuntz:
There should be code in the rails app that is similar to this, can you also modify it there?

sayalilunkad:
@vuntz It looks a little different to me how it is done in the crowbar_framework https://github.com/sayalilunkad/crowbar-core/blob/master/crowbar_framework/app/models/node_object.rb#L1539
Do we want to change the way it is selected there for the wwn filter?

vuntz:
@sayalilunkad yes, I believe we also want to favor wwn there too.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1585,16 +1585,25 @@ def unique_device_for(device)
...
-
+      unless result.empty?
+        result = result[0]
+        match = result.find { |b| b =~ /^wwn-/ } ||
+          result.find { |b| b =~ /^scsi-[a-zA-Z]/ } ||
+          result.find { |b| b =~ /^scsi-[^1]/ } ||
+          result.find { |b| b =~ /^scsi-/ } ||
+          result.find { |b| b =~ /^ata-/ } ||
+          result.find { |b| b =~ /^cciss-/ } ||
+          result.first
```
vuntz: Because of `result.first`, I guess you don't need the `result = result[0]` line.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1585,16 +1585,25 @@ def unique_device_for(device)
...
+      end
       # virtio disk might have neither by-path nor by-id links, use the /dev/vdX
       # name in that case
-      if result.empty?
+      if match.empty?
```
vuntz: Isn't it possible that `match` is not defined (in case we don't enter in the `unless result.empty?` branch)?

crowbar_framework/app/models/node_object.rb
sayalilunkad: No, the problem is the meta["disk"][type] returns a nested list with all the elements in the [0] position.

vuntz: Ah, yes, sorry, I read the code too fast. What about simply doing `match = result[0].find...`?

crowbar_framework/app/models/node_object.rb
```diff
@@ -1585,19 +1585,27 @@ def unique_device_for(device)
...
+          "/dev/#{device}"
+        else
+          "/dev/disk/#{match}"
+        end
       else
-        "/dev/disk/#{result.compact.first}"
+        nil
       end
-    else
-      nil
```
vuntz: You're losing the `else` here, which doesn't seem intended.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1585,19 +1585,27 @@ def unique_device_for(device)
...
+          result[0].find { |b| b =~ /^scsi-/ } ||
+          result[0].find { |b| b =~ /^ata-/ } ||
+          result[0].find { |b| b =~ /^cciss-/ } ||
+          result[0].first
+        # virtio disk might have neither by-path nor by-id links, use the /dev/vdX
+        # name in that case
+        if match.empty?
+          "/dev/#{device}"
+        else
+          "/dev/disk/#{match}"
```
vuntz: I may be wrong, but I don't think it's fully correct. In the old code, we had (summarizing):
```ruby
return "/dev/disk/#{type}/#{meta["disks"][type].first}"
```
In the node, we miss the `/#{type}` part.

sayalilunkad: right, fixed it.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1585,17 +1585,24 @@ def unique_device_for(device)
...
+          result[0].find { |b| b =~ /^scsi-/ } ||
+          result[0].find { |b| b =~ /^ata-/ } ||
+          result[0].find { |b| b =~ /^cciss-/ } ||
+          result[0].first
+        # virtio disk might have neither by-path nor by-id links, use the /dev/vdX
+        # name in that case
+        if match.empty?
+          "/dev/#{device}"
+        else
+          "/dev/disk/#{type}/#{match}"
```
vuntz: But `type` is not defined here anymore ;-)

crowbar_framework/app/models/node_object.rb
```diff
@@ -1585,17 +1585,24 @@ def unique_device_for(device)
...
+        end
     else
```
vuntz: So you restorered that `else`, but there's no `else` for ` unless result.empty?` anymore? Is that intended?

sayalilunkad: yes that was intended..I did assume it would do that anyways.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1585,16 +1585,25 @@ def unique_device_for(device)
...
+            b =~ /^scsi-[^1]/ ||
+            b =~ /^scsi-[^1]/ ||
+            b =~ /^ata-/ ||
+            b =~ /^cciss-/
+          } || meta["disks"][type].first
+          unless match.nil?
+            "#{type}/#{match}"
+          end
         end
       end
```
vuntz: For clarity (and to fix the small bugs), I'd rewrite the whole code to:

```ruby
[...]
end

candidates = disk_lookups.map do |type|
  disks_for_type = meta["disks"][type]
  unless disks_for_type.nil? || disks_for_type.empty?
    disk_for_type = meta["disks"][type].find do |b|
      b =~ /^wwn-/ ||
      b =~ /^scsi-[a-zA-Z]/ ||
      b =~ /^scsi-[^1]/ ||
      b =~ /^scsi-/ ||
      b =~ /^ata-/ ||
      b =~ /^cciss-/
    end
    disk_for_type ||= disks_for_type.first
    unless disk_for_type.nil?
      "#{type}/#{disk_for_type}"
    end
  end
end
candidates.compact!

# virtio disk might have neither by-path nor by-id links, use the /dev/vdX
# name in that case
if candidates.empty?
  "/dev/#{device}"
else
  "/dev/disk/#{candidates.first}"
end

File: crowbar_framework/app/models/node_object.rb [hunk 8]
```diff
@@ -1585,16 +1585,25 @@ def unique_device_for(device)
...
       result = disk_lookups.map do |type|
         if meta["disks"][type] and not meta["disks"][type].empty?
-          "#{type}/#{meta["disks"][type].first}"
+          match = meta["disks"][type].find{ |b| b =~ /wwn-/ ||
```
vuntz: It's `^wwn-`, not `wwn-`.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1585,16 +1585,25 @@ def unique_device_for(device)
...
+            b =~ /^scsi-[a-zA-Z]/ ||
+            b =~ /^scsi-[^1]/ ||
+            b =~ /^scsi-[^1]/ ||
```
vuntz: Duplicated lines; the second one should be `/^scsi-/`, I think.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1585,16 +1585,25 @@ def unique_device_for(device)
...
+          unless match.nil?
+            "#{type}/#{match}"
+          end
         end
       end
-
+      result = result.compact.first
       # virtio disk might have neither by-path nor by-id links, use the /dev/vdX
       # name in that case
       if result.empty?
```
vuntz: This may fail, as `result` can be `nil` with your new line above.

crowbar_framework/app/models/node_object.rb
sayalilunkad: makes sense.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1482,18 +1482,30 @@ def unique_device_for(device)
       unless hardware =~ /VirtualBox/i
         disk_lookups.unshift "by-id"
       end
-      result = disk_lookups.map do |type|
-        if meta["disks"][type] and not meta["disks"][type].empty?
-          "#{type}/#{meta["disks"][type].first}"
+      candidates = disk_lookups.map do |type|
+        disks_for_type = meta["disks"][type]
+        next unless disks_for_type.nil? || disks_for_type.empty?
```
vuntz: `next if`

crowbar_framework/app/models/node_object.rb
```diff
@@ -1482,18 +1482,30 @@ def unique_device_for(device)
       unless hardware =~ /VirtualBox/i
         disk_lookups.unshift "by-id"
       end
-      result = disk_lookups.map do |type|
-        if meta["disks"][type] and not meta["disks"][type].empty?
-          "#{type}/#{meta["disks"][type].first}"
+      candidates = disk_lookups.map do |type|
+        disks_for_type = meta["disks"][type]
+        next if disks_for_type.nil? || disks_for_type.empty?
+        disk_for_type = meta["disks"][type].find do |b|
```
vuntz: s/`meta["disks"][type]`/` disks_for_type`/

PR #801, opened 2016-11-18
crowbar: Stop defining SLES 12 SP1 repos

vuntz:
We moved to SES 4, so no usage of SP1 is left.


vuntz:
> Is there any remanent in qa_crowbarsetup.sh to take care of?

Nothing I'm aware of.

PR #803, opened 2016-11-18
provisioner: Avoid crashing if node has incomplete admin network mapping

vuntz:
This could happen if some ohai data is missing, or if the network
proposal is not providing a conduit map that is valid for a node.

If we keep crashing, then this actually breaks chef-client on the admin
server, which can then break DHCP, DNS and more for new nodes, or on
state transitions. However, there's nothing the admin server can do
about a node with wrong data, and we should avoid breaking the
deployment infrastructure because of this.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -14,12 +14,32 @@
 #
 
 def find_node_boot_mac_addresses(node, admin_data_net)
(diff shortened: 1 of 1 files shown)


vuntz:
This is an alternative to the end of https://github.com/crowbar/crowbar-core/pull/575.

rsalevsky:
@vuntz Can you please add the bsc number to the commit message header?

vuntz:
> @vuntz Can you please add the bsc number to the commit message header?

Was a bit of a challenge to add this without making the line too long, but done.

PR #805, opened 2016-11-18
crowbar_register: Add entry to /etc/hosts for resolving our hostname

vuntz:
This helps make sure that, in case the DNS server is down, we can
still resolve our hostname. We already do that in the autoyast profile,
but we forgot to add this to crowbar_register.

See https://github.com/crowbar/barclamp-provisioner/pull/344


aplanas:
commit description: duplicate `sure`

vuntz:
> commit description: duplicate sure

fixed

PR #808, opened 2016-11-18
[ntp] add documentation

bmwiedemann:
about how NTP clients and servers synchonize time with external servers

This was lost during merge of barclamp-ntp into crowbar-core


rsalevsky:
Not sure but I think that's not the right place for such documentation.

vuntz:
@bmwiedemann would it make sense to mention that to the doc team?

vuntz:
@fsundermeyer @taroth21 what do you think? Does it makes sense to include that in our doc instead?

fsundermeyer:
Makes sense to add it to the documentation, see https://trello.com/c/NcP5wUge/

bmwiedemann:
I'd still like to have a home for the ntp.dot source of the .svg, which will probably not be in the manual.

PR #811, opened 2016-11-18
Enable puma_worker_killer (bsc#1010211)

rsalevsky:
This change will keep the memory usage under controll if puma runs for a
long time in production.


crowbar_framework/config/puma.rb
```diff
@@ -31,6 +31,17 @@
 
 bind "tcp://#{LISTEN}:#{PORT}"
 
+before_fork do
+  PumaWorkerKiller.config do |config|
+    config.ram           = `free -m`.split(" ")[7]
```
aplanas: you can remove the `" "`, but IMHO reading `/proc/meminfo` and using a regex to get the `MemTotal` line is a more robust solution

crowbar_framework/config/puma.rb
```diff
@@ -31,6 +31,17 @@
...
+    config.frequency     = 60
```
aplanas: can you document the units? are this seconds?

crowbar_framework/config/puma.rb
```diff
@@ -31,6 +31,17 @@
...
+    config.percent_usage = 0.98
+    config.rolling_restart_frequency = 12 * 3600
```
aplanas: so this check two times per day? can be more clear `60 * 60 * 12` or a comment

PR #814, opened 2016-11-21
upgrade: Iterate over all possible DRBD clusters

jsuchome:
It's possible to have more clusters with DRBD-enabled nodes.


jsuchome:
@rhafer I think this should address one of your earlier concerns

PR #816, opened 2016-11-22
upgrade: Allow repeating some steps

jsuchome:
User might want to repeat some of the steps even when they already
succeeded.

Besides the Cloud6 backport, API change in crowbar-init is also required.

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -26,7 +26,7 @@ module Api
       def checks
         upgrade_status = ::Crowbar::UpgradeStatus.new
         # the check for current_step means to allow running the step at any point in time
-        upgrade_status.start_step if upgrade_status.current_step == :upgrade_prechecks
+        upgrade_status.start_step(:upgrade_prechecks)
 
         {}.tap do |ret|
           ret[:network_checks] = {
@@ -95,7 +95,7 @@ module Api
 
       def adminrepocheck
         upgrade_status = ::Crowbar::UpgradeStatus.new
-        upgrade_status.start_step if upgrade_status.current_step == :admin_repo_checks
+        upgrade_status.start_step(:admin_repo_checks)
         # FIXME: once we start working on 7 to 8 upgrade we have to adapt the sles version
         zypper_stream = Hash.from_xml(
           `sudo /usr/bin/zypper-retry --xmlout products`
@@ -148,9 +148,8 @@ module Api
             upgrade_status.end_step(
               false,
               adminrepocheck: "Missing repositories: #{missing_repos}"
-            ) if upgrade_status.current_step == :admin_repo_checks
+            )
           else
-            next unless upgrade_status.current_step == :admin_repo_checks
             upgrade_status.end_step
           end
         end
@@ -242,7 +241,7 @@ module Api
       end
 
       def prepare(options = {})
-        ::Crowbar::UpgradeStatus.new.start_step
+        ::Crowbar::UpgradeStatus.new.start_step(:upgrade_prepare)
 
         background = options.fetch(:background, false)
 
@@ -574,7 +573,10 @@ module Api
         true
       rescue => e
         message = e.message
-        ::Crowbar::UpgradeStatus.new.end_step(false, { prepare_nodes_for_crowbar_upgrade: message })
+        ::Crowbar::UpgradeStatus.new.end_step(
+          false,
+          prepare_nodes_for_crowbar_upgrade: message
+        )
         Rails.logger.error message
 
         false
--- a/crowbar_framework/lib/crowbar/upgrade_status.rb
+++ b/crowbar_framework/lib/crowbar/upgrade_status.rb
@@ -75,13 +75,19 @@ module Crowbar
       progress[:steps][current_step] || {}
     end
 
-    def start_step
+    # 'step' is name of the step user wants to start.
+    def start_step(step_name)
       ::Crowbar::Lock::LocalBlocking.with_lock(shared: false, logger: @logger, path: lock_path) do
-        if running?
+        if running? step_name
           @logger.warn("The step has already been started")
           return false
         end
-        progress[:steps][current_step][:status] = :running
+        unless step_allowed? step_name
+          @logger.error("The start of step #{step_name} is requested in the wrong order")
+          return false
+        end
+        progress[:current_step] = step_name
+        progress[:steps][step_name][:status] = :running
         save
       end
     end
@@ -153,6 +159,24 @@ module Crowbar
         :nodes_upgrade,
         :finished
       ]
+    end
+
+    # Return true if user is allowed to execute given step
+    # In normal cases, that should be true only for next step in the sequence.
+    # But for some cases, we allow repeating of the step that has just passed.
+    def step_allowed?(step)
+      return true if step == current_step
+      if [
+        :upgrade_prechecks,
+        :admin_repo_checks,
+        :nodes_repo_checks
+      ].include? step
+        # Allow repeating one of these steps if it was the last one finished
+        # and no other one has been started yet.
+        i = upgrade_steps_6_7.index step
+        return upgrade_steps_6_7[i + 1] == current_step && pending?(current_step)
+      end
+      false
     end
 
     def lock_path
--- a/crowbar_framework/spec/lib/crowbar/upgrade_status_spec.rb
+++ b/crowbar_framework/spec/lib/crowbar/upgrade_status_spec.rb
@@ -49,7 +49,7 @@ describe Crowbar::UpgradeStatus do
     it "determines whether current step is pending" do
       expect(subject.current_step).to eql :upgrade_prechecks
       expect(subject.pending?).to be true
-      expect(subject.start_step).to be true
+      expect(subject.start_step(:upgrade_prechecks)).to be true
       expect(subject.pending?).to be false
     end
 
@@ -58,7 +58,7 @@ describe Crowbar::UpgradeStatus do
       expect(subject.pending?).to be true
       expect(subject.pending?(:upgrade_prechecks)).to be true
       expect(subject.pending?(:admin_backup)).to be true
-      expect(subject.start_step).to be true
+      expect(subject.start_step(:upgrade_prechecks)).to be true
       expect(subject.pending?).to be false
       expect(subject.pending?(:upgrade_prechecks)).to be false
       expect(subject.pending?(:admin_backup)).to be true
@@ -68,7 +68,7 @@ describe Crowbar::UpgradeStatus do
       expect(subject.current_step).to eql :upgrade_prechecks
       expect(subject.running?).to be false
       expect(subject.running?(:upgrade_prechecks)).to be false
-      expect(subject.start_step).to be true
+      expect(subject.start_step(:upgrade_prechecks)).to be true
       expect(subject.running?).to be true
       expect(subject.running?(:upgrade_prechecks)).to be true
     end
@@ -77,7 +77,7 @@ describe Crowbar::UpgradeStatus do
       expect(subject.current_step).to eql :upgrade_prechecks
       other_status = new_status
       expect(other_status.running?).to be false
-      expect(subject.start_step).to be true
+      expect(subject.start_step(:upgrade_prechecks)).to be true
       other_status.load
       expect(other_status.running?).to be true
     end
@@ -85,7 +85,7 @@ describe Crowbar::UpgradeStatus do
     it "determines whether a given step is running" do
       expect(subject.current_step).to eql :upgrade_prechecks
       expect(subject.running?(:upgrade_prepare)).to be false
-      expect(subject.start_step).to be true
+      expect(subject.start_step(:upgrade_prechecks)).to be true
       expect(subject.running?(:upgrade_prepare)).to be false
     end
 
@@ -93,7 +93,7 @@ describe Crowbar::UpgradeStatus do
       expect(subject.current_step).to eql :upgrade_prechecks
       expect(subject.current_step_state[:status]).to eql :pending
       expect(subject.running?).to be false
-      expect(subject.start_step).to be true
+      expect(subject.start_step(:upgrade_prechecks)).to be true
       expect(subject.running?).to be true
       expect(subject.running?(:upgrade_prepare)).to be false
     end
@@ -101,14 +101,14 @@ describe Crowbar::UpgradeStatus do
     it "moves to next step when requested" do
       expect(subject.current_step).to eql :upgrade_prechecks
       expect(subject.current_step_state[:status]).to eql :pending
-      expect(subject.start_step).to be true
+      expect(subject.start_step(:upgrade_prechecks)).to be true
       expect(subject.end_step).to be true
       expect(subject.current_step).to eql :upgrade_prepare
     end
 
     it "does not move to next step when current one failed" do
       expect(subject.current_step).to eql :upgrade_prechecks
-      expect(subject.start_step).to be true
+      expect(subject.start_step(:upgrade_prechecks)).to be true
       expect(subject.end_step(false, failure: "error message")).to be false
       expect(subject.current_step).to eql :upgrade_prechecks
       expect(subject.current_step_state[:status]).to eql :failed
@@ -117,7 +117,7 @@ describe Crowbar::UpgradeStatus do
 
     it "does not allow to end step when it is not running" do
       expect(subject.current_step).to eql :upgrade_prechecks
-      expect(subject.start_step).to be true
+      expect(subject.start_step(:upgrade_prechecks)).to be true
       expect(subject.end_step).to be true
       expect(subject.current_step).to eql :upgrade_prepare
       expect(subject.end_step).to be false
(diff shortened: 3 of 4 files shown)


crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -75,13 +75,19 @@ def current_step_state
...
       ::Crowbar::Lock::LocalBlocking.with_lock(shared: false, logger: @logger, path: lock_path) do
-        if running?
+        if running? step_name
           @logger.warn("The step has already been started")
           return false
         end
-        progress[:steps][current_step][:status] = :running
+        if step_name != current_step && !out_of_order
+          @logger.error("The step #{step_name} is requested in the wrong order")
+          return false
```
jsuchome: We probably need to raise an exception instead, so we could distinguish between different kind of errors...

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -75,13 +75,19 @@ def current_step_state
       progress[:steps][current_step] || {}
     end
 
-    def start_step
+    # 'step' is name of the step user wants to start.
+    # If `out_of_order` is true, the step could be executed out of order
+    def start_step(step_name, out_of_order = false)
```
skazi0: this would require changing all current uses of `start_step` to include the step, right? there's no backward compatibility with current "argumentless" call.

abel-navarro:
I don't think our upgrade process is idempotent. If we offer the customer the possibility to re-run single steps we will have to support that option, which can be a nightmare. From a developer point of view it may make sense though to have that option. If we can prevent customers for using this, I'm ok with the change.

crowbar_framework/lib/crowbar/upgrade_status.rb
jsuchome: Right, we'd need to adapt all the calls

jsuchome:
> If we offer the customer the possibility to re-run single steps we will have to support that option, which > can be a nightmare

We do not allow end user to pick a step. We need an API to do so for some of them (like the `prechecks`), but definitely not for all; that's why the `out_of_order` argument.

jsuchome:

crowbar_framework/lib/crowbar/upgrade_status.rb
skazi0: Shouldn't it be done in the same PR? otherwise we're left with a broken system.

jsuchome: You can see that the changes for library invocation are already present in this PR

skazi0: hmm, don't know how I missed these. thx.

jsuchome: I've added them later and fixed it up to the existing commits, sorry

MaximilianMeister:
loogs good! could you maybe add another PR to update crowbar-init? then we should have all dependencies covered

crowbar_framework/lib/crowbar/upgrade_status.rb
skazi0: aha :)

one more question:
wouldn't it be better to remove this `out_of_order` flag and embed the knowledge about steps which could be run out of order in the library itself?

jsuchome: I was thinking about it as well, I'm not sure...

skazi0: Well, it depends on the role we assign to this library. If it's supposed to be a simple wrapper around the status file then no, such a mapping doesn't belong there. On the other hand, maybe we should put some more functionality in the library to make it more convenient for the client code?

jsuchome:
> could you maybe add another PR to update crowbar-init?

Might not be needed if we decide to store the information about nodes that could be called any time inside the library. What do you think?

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -75,13 +75,19 @@ def current_step_state
...
       ::Crowbar::Lock::LocalBlocking.with_lock(shared: false, logger: @logger, path: lock_path) do
-        if running?
+        if running? step_name
```
aspiers: No, this allows running two different steps at the same time.  You don't really want that, right?

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -75,13 +75,19 @@ def current_step_state
...
           @logger.warn("The step has already been started")
```
aspiers: This needs changing to `"The #{current_step} step has already been started"`, because with this API change the already-running step can now be different to the one the caller is requesting to start.

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -75,13 +75,19 @@ def current_step_state
...
           return false
         end
-        progress[:steps][current_step][:status] = :running
+        if step_name != current_step && !out_of_order
```
aspiers: https://twitter.com/codinghorror/status/506010907021828096

In this case, "naming things" is the problem - [as I mentioned on 13th October](https://github.com/crowbar/crowbar-core/pull/692#issuecomment-253575454), this code will be pretty confusing for any newbies to read.  It is checking that the step the caller wants to start is the same as `current_step`, and then it is starting `current_step`.  So `current_step` doesn't mean "current step" at all!

It could be renamed to something like `next_or_current_step`. But even that's not quite right, because then the library is attempting to predict the future which is impossible because it has no idea what the next step will be: the user might want to proceed or repeat some steps.

A more correct name would be `current_or_next_pending_step`, but the length of that is a bad smell which highlights that the variable is trying to do two things at once:

1. track what is the next *pending* step (not necessarily the next one the user will take)
2. track what step is currently running, if any

Luckily we can fix this quite easily by splitting `current_step` into two:

1. `next_pending_step`
2. `current_step` - the current step, **or `nil` if there is no step running!!**

I am not suggesting to make those changes in this PR though - we can do a separate PR for that.

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -92,17 +92,22 @@ def start_step(step_name, out_of_order = false)
...
+    def end_step(step_name = nil, out_of_order = false, success = true, errors = {})
       ::Crowbar::Lock::LocalBlocking.with_lock(shared: false, logger: @logger, path: lock_path) do
-        unless running?
+        step_name ||= current_step
+        unless running? step_name
           @logger.warn("The step is not running, could not be finished")
           return false
         end
-        progress[:steps][current_step] = {
+        if step_name != current_step && !out_of_order
```
aspiers: How can this ever be true, unless we allow multiple steps running concurrently (which surely we don't want)?

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -92,17 +92,22 @@ def start_step(step_name, out_of_order = false)
...
+          @logger.error("The end of step #{step_name} is requested in the wrong order")
+          return false
+        end
+        progress[:steps][step_name] = {
           status: success ? :passed : :failed,
           errors: errors
         }
-        next_step
+        next_step unless out_of_order
```
aspiers: This doesn't feel right to me.  Shouldn't choosing whether to break the normal order should be decided when starting a step, not when ending it?

crowbar_framework/spec/lib/crowbar/upgrade_status_spec.rb
```diff
@@ -77,47 +77,51 @@ def new_status
...
     end
 
     it "does not move to next step when current one failed" do
       expect(subject.current_step).to eql :upgrade_prechecks
-      expect(subject.start_step).to be true
-      expect(subject.end_step(false, failure: "error message")).to be false
+      expect(subject.start_step(:upgrade_prechecks)).to be true
+      expect(
+        subject.end_step(
+          :upgrade_prechecks, false, false, failure: "error message"
```
aspiers: Code smell - this `false, false` is unintuitive which to me hints that the design is not quite right.  I gave more details in my review, but just wanted to add a mention of the useful maxim from BDD which I found really helpful to remember: tests are a good opportunity to ["write the code you wish you had"](http://waysysweb.com/unittestframework/testing_as_design.htm).

crowbar_framework/lib/crowbar/upgrade_status.rb
skazi0: This one is tricky. Running steps "out of order" should be called rather something like running steps in "detached" mode. This is meant for steps like prechecks or repochecks which should be callable any time even when some other step is running.
With this implementation, this would cause the status to show those "detached" steps as "running" for a short while which might cause some unwanted side effects.

crowbar_framework/lib/crowbar/upgrade_status.rb
skazi0: As mentioned in my comment below, those "out of order" steps are not always going to be "current" when running. They should be callable any time, even when some other step is "current".

crowbar_framework/lib/crowbar/upgrade_status.rb
skazi0: You do for some steps. See below.

crowbar_framework/lib/crowbar/upgrade_status.rb
jsuchome: `current_step` is meant as the step that **should** be current step in the normal order of things. The exception to this rule is the step that would be called "out of order", but than we still want to remember which one *should* be current

crowbar_framework/lib/crowbar/upgrade_status.rb
jsuchome: > How can this ever be true

Easily, when user requests to start 3rd step while 2nd hasn't been run yet. But if user requests this with `out_of_order = true`, than it would be allowed.

crowbar_framework/lib/crowbar/upgrade_status.rb
jsuchome: The point is, we do not want to move `current_step` to next one if we were executing out of order. And the called has the knowledge about out-of-order status, so I think it's better to provide this info both when starting and ending the step

crowbar_framework/spec/lib/crowbar/upgrade_status_spec.rb
jsuchome: Yes, that would be solved with @skazi0 's proposal, see https://github.com/crowbar/crowbar-core/pull/816#discussion_r89111780

crowbar_framework/lib/crowbar/upgrade_status.rb
jsuchome: This is mostly to prevent one step for being started twice in parallel. But you are right it does not exactly prevent us to having 2 steps running at the same time (if one of them is `out_of_order`). We might allow that (as @skazi0 suggests), or we might extend the `running?` method to check if *any* step is currently running.

crowbar_framework/lib/crowbar/upgrade_status.rb
aspiers: Oh right I see now, I temporarily forgot that `current_step` doesn't mean the current step.  So this code actually needs to read:

    if step_name != next_pending_step && !out_of_order

crowbar_framework/lib/crowbar/upgrade_status.rb
aspiers: > we do not want to move current_step to next one if we were executing out of order.

Why not?

> And the called has the knowledge about out-of-order status, so I think it's better to provide this info both when starting and ending the step

The caller has knowledge only about the step being ended.  It should not be expected to predict the future beyond that.  So I'm really not convinced about this.  Let's mumble.

crowbar_framework/spec/lib/crowbar/upgrade_status_spec.rb
aspiers: I agree with @skazi0's proposal to make the library enforce which order steps can be run in.  However even without that change I still think this parameter does not belong in `#end_step`.

crowbar_framework/lib/crowbar/upgrade_status.rb
skazi0: Ping me when you have a mumble session.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -574,7 +574,10 @@ def prepare_nodes_for_crowbar_upgrade
         true
       rescue => e
         message = e.message
-        ::Crowbar::UpgradeStatus.new.end_step(false, { prepare_nodes_for_crowbar_upgrade: message })
+        ::Crowbar::UpgradeStatus.new.end_step(
```
vuntz: Missing name of the step?

jsuchome: ah, true

aspiers:
@jsuchome IIUC this API change will require changes to crowbar-init, right? If so, should mention it in the PR description.

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -75,24 +75,35 @@ def current_step_state
...
+    def end_step(step_name = nil, success = true, errors = {})
       ::Crowbar::Lock::LocalBlocking.with_lock(shared: false, logger: @logger, path: lock_path) do
-        unless running?
+        step_name ||= current_step
+        unless running? step_name
           @logger.warn("The step is not running, could not be finished")
           return false
         end
-        progress[:steps][current_step] = {
+        unless step_allowed? step_name
```
aspiers: This doesn't make sense here - `step_allowed?` is testing whether a user is allowed to start a step, not end it.

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -155,6 +166,24 @@ def upgrade_steps_6_7
       ]
     end
 
+    # Return true if user is allowed to execute given step
+    # In normal cases, that should be true only for next step in the sequence.
+    # But for some cases, we allow repeating of the step that has just passed.
+    def step_allowed?(step)
```
aspiers: This should get some unit tests.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -146,11 +146,11 @@ def adminrepocheck
...
             end.flatten.join(", ")
             upgrade_status.end_step(
+              :admin_repo_checks,
               false,
               adminrepocheck: "Missing repositories: #{missing_repos}"
-            ) if upgrade_status.current_step == :admin_repo_checks
+            )
           else
-            next unless upgrade_status.current_step == :admin_repo_checks
             upgrade_status.end_step
```
jsuchome: No, we can take default values here

crowbar_framework/lib/crowbar/upgrade_status.rb
jsuchome: Hm, true, it's not needed any more, since at this point, there's only `current_step`. Speaking of which... passing step name to `end_step` is no longer needed with this state of API...

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -155,6 +166,24 @@ def upgrade_steps_6_7
...
+      return true if step == current_step
+      if [
+        :upgrade_prechecks,
+        :admin_repo_checks,
+        :nodes_repo_checks
+      ].include? step
+        # Allow repeating one of these steps if it was the last one finished
+        # and no other one has been started yet.
+        i = upgrade_steps_6_7.index step
+        return upgrade_steps_6_7[i + 1] == current_step && !running?(current_step)
```
skazi0: This doesn't match the description in comment above. We should be checking if it's not started instead of not running (hint: passed = !running but passed != !started).

jsuchome: Ah, true, current step might have failed

crowbar_framework/lib/crowbar/upgrade_status.rb
jsuchome: I believe it's now covered by those 3 new test cases added with the commit

jsuchome:
@aspiers BTW with the latest state of the PR, `current_step` really points only to the current step, i.e. the step that really is currently running or the first one pending (while nothing else is running). 

So I understand it's still covering 2 different states, but I believe it is not as bad as with original PR.

aspiers:
> @aspiers BTW with the latest state of the PR, current_step really points only to the current step, i.e. the step that really is currently running or the first one pending (while nothing else is running).

Nope, AFAICS it can also refer to the predicted next step when no step is running (and the prediction is unreliable because we have no way to know what the user will do next).

> So I understand it's still covering 2 different states, but I believe it is not as bad as with original PR.

I don't see any change at all in this respect.  But it doesn't matter because I'm working on a separate PR to address this.

My only remaining concern with this PR as mentioned on IRC is that it's not 100% to have `#end_step()` without a `step_name` parameter.  But that's a minor issue which is also covered by the `pending` unit test regarding ownership of each step, so we can fix that later too.

jsuchome:
> Nope, AFAICS it can also refer to the predicted next step when no step is running

That's what I meant by *or the first one pending (while nothing else is running)*

> (and the prediction 
> is unreliable because we have no way to know what the user will do next).

User can only start the first one pending, or the last one executed (when none is pending and when the step is repeatable)

aspiers:
> That's what I meant by or the first one pending (while nothing else is running)

Oh OK, I thought you were talking about the very first step only.

> User can only start the first one pending, or the last one executed (when none is pending and when the step is repeatable)

Sure, but this still does not change that a step is not current until it's actually running.  Again, I am not saying there is a bug here in any *behaviour*, only that there's a misleading variable name and the design will be improved by my upcoming PR which eliminates overloading of one variable with two meanings.  I don't think this PR changes that in any way.

PR #817, opened 2016-11-22
webserver: enable crowbar to change workers/threads itself

MaximilianMeister:


rsalevsky:
Tested and it works...

PR #819, opened 2016-11-23
[3.0] upgrade: bring repo names in consistent format

MaximilianMeister:
this format is used in the noderepochecks through the repos-cloud.yml
file

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -134,7 +134,7 @@ module Api
             repos: {}
           }
           ret[:os][:repos][admin_architecture.to_sym] = {
-            missing: ["SUSE Linux Enterprise Server 12 SP2"]
+            missing: ["SUSE-Linux-Enterprise-Server-12-SP2"]
           } unless os_available
 
           cloud_available = repo_version_available?(products, "suse-openstack-cloud", "7")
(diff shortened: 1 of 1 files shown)


jsuchome:
Note: in forwardport we need to adapt the help text as well

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -134,7 +134,7 @@ def adminrepocheck
             repos: {}
           }
           ret[:os][:repos][admin_architecture.to_sym] = {
-            missing: ["SUSE Linux Enterprise Server 12 SP2"]
+            missing: ["SUSE-Linux-Enterprise-Server-12-SP2"]
```
jsuchome: Um... should it actually be `SLES12-SP2-Pool` and `SLES12-SP2-Updates`? 
Both of them until we have a way to finding out which one is missing

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -143,7 +143,7 @@ def adminrepocheck
             repos: {}
           }
           ret[:openstack][:repos][admin_architecture.to_sym] = {
-            missing: ["SUSE OpenStack Cloud 7"]
+            missing: ["SUSE-OpenStack-Cloud-7"]
```
jsuchome: `["SUSE-OpenStack-Cloud-7-Pool", "SUSE-OpenStack-Cloud-7-Updates"]`

@skazi0 , is this what UI expects?

skazi0: UI doesn't currently check `missing` anywhere but int the main functionality it expects "Pool" and "Updates" exactly as in your comment.

PR #823, opened 2016-11-23
[provisioner] Add autoyast self update option

nicolasbock:


chef/data_bags/crowbar/template-provisioner.json
```diff
@@ -9,6 +9,8 @@
       "access_keys": "",
       "shell_prompt": "USER@HOST:CWD SUFFIX",
       "enable_pxe": true,
+      "self_update": false,
+      "self_update_url": "http://updates.suse.com/sle12/$arch",
```
vuntz: Shouldn't we default this to a repo that is mirrored on the admin server, like all the other repos we have there?

vuntz: Also, I'm not convinced we need a boolean to enable this or not. Either the repo is present locally (we enable), or the user provides an external URL (we enable), or the repo is not present locally and there's no url (we disable)

nicolasbock: @vuntz Thanks. I changed the PR.

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
```diff
@@ -26,6 +26,7 @@
       <import_gpg_key config:type="boolean">true</import_gpg_key>
     </signature-handling>
     <storage/>
+    <self_update_url><%= @self_update_url %></self_update_url>
```
vuntz: Didn't we want to add some `unless @self_update_url.nil? || @self_update_url.empty?`

chef/data_bags/crowbar/template-provisioner.json
```diff
@@ -9,6 +9,7 @@
       "access_keys": "",
       "shell_prompt": "USER@HOST:CWD SUFFIX",
       "enable_pxe": true,
+      "self_update_url": "http://crowbar/sle12/$arch",
```
vuntz: I would change the default to: `http://<ADMINWEB>/suse-12.2/$arch/repos/installer-update` (or something similar).

A couple of things:
- `<ADMINWEB>` should be substituted in the rails app (see glance & tempest barclamps)
- if we put `suse-12.2` in the path, then maybe that attribute should live in a versioned suse attribute path somewhere (we already have that in the provisioner barclamp, I think)
- we need a doc bug to document how to mirror that repo in the location

crowbar_framework/app/views/barclamp/provisioner/_edit_attributes.html.haml
```diff
@@ -7,6 +7,9 @@
     %span.help-block
       = t(".access_keys_hint")
 
+    = boolean_field :self_update
```
vuntz: This one is gone :-)

chef/data_bags/crowbar/template-provisioner.json
nicolasbock: - I've add the `<ADMINWEB>` alias.
- Opened doc bug: https://bugzilla.suse.com/show_bug.cgi?id=1012469
- I reverted the default to "" which defaults to `scc.suse.com`. Or would you prefer a versioned local URL?

crowbar_framework/app/views/barclamp/provisioner/_edit_attributes.html.haml
nicolasbock: Good catch! I removed it.

crowbar_framework/config/locales/provisioner/en.yml
```diff
@@ -21,6 +21,7 @@ en:
       edit_attributes:
         access_keys: 'Additional SSH keys'
         access_keys_hint: 'Each SSH key must be on its own line'
+        self_update_url: 'AutoYaST Self-update URL (The alias <ADMINWEB> can be used to specify the admin node)'
```
vuntz: s/admin node/admin server/

chef/data_bags/crowbar/template-provisioner.json
```diff
@@ -9,6 +9,7 @@
       "access_keys": "",
       "shell_prompt": "USER@HOST:CWD SUFFIX",
       "enable_pxe": true,
+      "self_update_url": "",
```
vuntz: So we don't provide a default value that just works? :-)

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -290,6 +290,7 @@ def find_node_boot_mac_addresses(node, admin_data_net)
           append << "ifcfg=dhcp4 netwait=60"
           append << "squash=0" # workaround bsc#962397
           append << "autoupgrade=1" if mnode[:state] == "os-upgrading"
+          append << "self_update"
```
vuntz: Shouldn't we pass this only if `self_update_url` is not empty?

vuntz: Also, if the URL contains `<ADMINWEB>`, I think we could do some quick check with `File.exists?` so that if we use a default with `<ADMINWEB>` but the repo is not mirrored, we just don't enable the update feature.

chef/data_bags/crowbar/template-provisioner.json
nicolasbock: :) Well, the default in this case is to use `scc.suse.com` which works. But we could also use a crowbar local default. Which would you prefer?

chef/cookbooks/provisioner/recipes/update_nodes.rb
nicolasbock: I added settings to `autoyast.xml` so that failures in self update don't mean failure anymore. AutoYaST will log those failure but continue with the installation process.

nicolasbock: I'll add a check whether the repository is actually there.

chef/data_bags/crowbar/template-provisioner.json
vuntz: I think the default should be the mirror on the admin server.

nicolasbock: ok, done.

nicolasbock:
@vuntz Here is where I am at with this PR:
- The self update feature is _always_ enabled now
- I added some settings to the autoyast file so that a missing or unavailable self update URL is not fatal. A warning is shown and after 10 seconds the installation process continues. The warning is logged and will show up in a supportconfig.
- The default URL points to a mirrored repository on the admin server
- We don't verify that that mirror exists and instead rely on the logged error message from AutoYaST. To verify we need to know the http server root directory and replace the proper parts of the URI with that directory. Since we don't know what arch we will deploy the role on, we have to be happy with _any_ `$arch` and testing for that seemed more than a line of code. Of course I might very well not seeing the simple solution here, so please correct me if I am making this too complicated :wink:.
- An empty URL means update from `scc.suse.com`
- I still don't fully understand how the defaults from a role end up in a node's attributes. This PR replaces the `<ADMINWEB>` shortcut in the chef recipe. I tried to follow the glance barclamp but couldn't get it to work here. i.e. if I replace `<ADMINWEB>` in `apply_role_pre_chef_call` in `models/provider_service.rb` the replace URI does not end up in the node.

rsalevsky:
@nicolasbock Can you please change the prefix from `[provisioner]` to `provisioner:`?

nicolasbock:
@rsalevsky oop, sorry, of course.

nicolasbock:
@aspiers I have added a description to the commit. Could you have another look?

chef/data_bags/crowbar/template-provisioner.schema
```diff
@@ -19,6 +19,7 @@
             "access_keys": { "type": "str", "required": true },
             "shell_prompt": { "type": "str", "required": true },
             "enable_pxe": { "type": "bool", "required": true },
+            "self_update_url": { "type": "str", "required": true },
```
vuntz: Sorry, I failed to notice this earlier on, but we have a `suse` attribute, and this should live in there. See line 56.

chef/cookbooks/provisioner/recipes/update_nodes.rb
vuntz: I guess the check you mention here is still not there? :-)

nicolasbock: I added it now.

chef/data_bags/crowbar/template-provisioner.schema
nicolasbock: Moved

nicolasbock:
@aspiers I have added a commit message and a PR comment.

nicolasbock:
@vuntz I think I have addressed the issues you raised. Could you have another look?

aspiers:
@nicolasbock Removed my -1.  I can't give +1 because I don't even understand why autoyast would need to self-update :-)  If there's newer autoyast code then why wouldn't it be simply provided via TFTP/HTTP the first time around?

nicolasbock:
@aspiers Thanks! Regarding your question: I haven't checked but am guessing that the autoyast comes from the image pulled from tftp on the admin node. I don't know though how and where from it is updated and I think your question is very valid :smile:.

I forgot to add [this Trello card](https://trello.com/c/Zj9q4AsW/181-p3-add-ability-to-use-installer-update-from-autoyast).

nicolasbock:
@vuntz Could you have another look, please?

PR #824, opened 2016-11-24
upgrade: Unset db-sync flag before upgrading the nodes, skip keystone bootstrap

jsuchome:
I think it makes sense to do both action from the point where we're accessing all nodes (to be upgraded), and before we actually do the upgrades.

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -385,6 +385,10 @@ class CrowbarService < ServiceObject
     upgrade_nodes.each do |node|
       node["target_platform"] = admin_node["provisioner"]["default_os"]
       node["crowbar_wall"]["crowbar_upgrade_step"] = "prepare-os-upgrade"
+      # skip initial keystone bootstrapping
+      if node["run_list_map"].key? "keystone-server"
+        node["keystone"]["bootstrap"] = true
+      end
       node.save
     end
 
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -392,6 +392,9 @@ class CrowbarService < ServiceObject
       node.save
     end
 
+    # unset `db_synced` flag for OpenStack components
+    ::Openstack::Upgrade.unset_db_synced
+
     commit_and_check_proposal
   end
 


jsuchome:
Hm, I should move that bootstrap call to extra method, so I do not have to repeat this call for all scenarios...

jsuchome:
[Solve Issues during controlles upgrade](https://trello.com/c/bkqIuMNB/190-solve-issues-during-controlles-upgrade)

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -385,9 +385,14 @@ def prepare_nodes_for_os_upgrade
     upgrade_nodes.each do |node|
       node["target_platform"] = admin_node["provisioner"]["default_os"]
       node["crowbar_wall"]["crowbar_upgrade_step"] = "prepare-os-upgrade"
+      # skip initial keystone bootstrapping
+      node["keystone"]["bootstrap"] = true if node["keystone"]
```
vuntz: I'd use `if node.roles.include?("keystone-server")` or something like this -- this feels slightly more robust.

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -385,9 +385,14 @@ def prepare_nodes_for_os_upgrade
...
       node.save
     end
 
+    # unset `db_synced` flag for OpenStack components
+    ::Openstack::Upgrade.unset_db_synced
```
vuntz: I'm wondering if it's really the best location for this, as this will have an impact if the user cancels the upgrade. Isn't there a place where we can do that a bit later?

jsuchome: I don't think there's an option to cancel the upgrade at this point (which is called "stopping the services" in the UI). This is already Cloud7 side.

crowbar_framework/app/models/crowbar_service.rb
jsuchome: Hm, somehow it looks that the change to `node.roles` broke it... I need to investigate more...

jsuchome:
@vuntz I changed the check for keystone-server again, so your current approval might not be valid....

vuntz:
> @vuntz I changed the check for keystone-server again, so your current approval might not be valid....

It's okay. I'm not a big fan of the current way, but we have bigger things to focus on :-)

PR #825, opened 2016-11-24
[forward-port] upgrade: bring repo names in consistent format

MaximilianMeister:
forward-port https://github.com/crowbar/crowbar-core/pull/819


nicolasbock:
@jsuchome good catch!

MaximilianMeister:
> https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/controllers/api/upgrade_controller.rb#L266

updated

PR #828, opened 2016-11-29
upgrade: Execute crowbar_join from extra script

jsuchome:
The DRBD-sync actions in post-upgrade script could take lot of time,
let's decouple them from crowbar_join (and chef-client) runs.


jsuchome:
[Solve Issues during controlles upgrade](https://trello.com/c/bkqIuMNB/190-solve-issues-during-controlles-upgrade)

jsuchome:
@vuntz I think it's better to have different long-running actions in extra script, so if one of them fails, we do not repeat all of them. Also, one script has one timeout, and both DRBD sync and chef-client run take quite some time; I think with separate timeouts we (and user) have better understanding of what actually fails/times out.

PR #830, opened 2016-11-29
[3.0] upgrade: restrict api access through the status library

MaximilianMeister:
there is also a backport of https://github.com/crowbar/crowbar-core/pull/772 in here

and it needs a forward port of the other commits about the api restriction

### DIFF
--- a/crowbar_framework/app/controllers/api/crowbar_controller.rb
+++ b/crowbar_framework/app/controllers/api/crowbar_controller.rb
@@ -42,6 +42,16 @@ class Api::CrowbarController < ApiController
     else
       render json: Api::Crowbar.upgrade
     end
+  rescue Crowbar::Error::StartStepRunningError,
+         Crowbar::Error::StartStepOrderError => e
+    render json: {
+      errors: {
+        admin_upgrade: {
+          data: e.message,
+          help: I18n.t("api.crowbar.upgrade.help.default")
+        }
+      }
+    }, status: :unprocessable_entity
   end
 
   def maintenance
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -36,6 +36,16 @@ class Api::UpgradeController < ApiController
         }
       }, status: :unprocessable_entity
     end
+  rescue Crowbar::Error::StartStepRunningError,
+         Crowbar::Error::StartStepOrderError => e
+    render json: {
+      errors: {
+        upgrade_prepare: {
+          data: e.message,
+          help: I18n.t("api.upgrade.prepare.help.default")
+        }
+      }
+    }, status: :unprocessable_entity
   end
 
   def prechecks
@@ -72,6 +82,17 @@ class Api::UpgradeController < ApiController
     else
       render json: check
     end
+  rescue Crowbar::Error::StartStepRunningError,
+         Crowbar::Error::StartStepOrderError,
+         Crowbar::Error::EndStepRunningError => e
+    render json: {
+      errors: {
+        admin_repo_checks: {
+          data: e.message,
+          help: I18n.t("api.upgrade.adminrepocheck.help.default")
+        }
+      }
+    }, status: :unprocessable_entity
   end
 
   def adminbackup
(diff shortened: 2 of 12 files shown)


crowbar_framework/spec/lib/crowbar/upgrade_status_spec.rb
```diff
@@ -196,28 +196,28 @@ def new_status
...
 
     it "prevents repeating steps that do not allow repetition" do
-      expect(subject.start_step(:upgrade_prepare)).to be false
-      expect(subject.start_step(:admin_backup)).to be false
-      expect(subject.start_step(:admin_upgrade)).to be false
-      expect(subject.start_step(:database)).to be false
-      expect(subject.start_step(:nodes_services)).to be false
-      expect(subject.start_step(:nodes_db_dump)).to be false
-      expect(subject.start_step(:nodes_upgrade)).to be false
+      expect{subject.start_step(:upgrade_prepare)}.to raise_error(Crowbar::UpgradeStatus::StartStepOrderError)
```
MaximilianMeister: if someone is wondering why i switched to a block syntax, it is necessary apparently when an error is raised (see http://stackoverflow.com/questions/17212754/expect-to-raise-error-and-test-fails-because-it-raises-this-error)

MaximilianMeister:

crowbar_framework/app/controllers/api/crowbar_controller.rb
```diff
@@ -42,6 +42,16 @@ def upgrade
     else
       render json: Api::Crowbar.upgrade
     end
+  rescue Crowbar::UpgradeStatus::StartStepRunningError,
+         Crowbar::UpgradeStatus::StartStepOrderError => e
```
jsuchome: Can't we check for StandardError here, or better, for some common predecessor of these exceptions?

crowbar_framework/app/controllers/api/crowbar_controller.rb
```diff
@@ -42,6 +42,16 @@ def upgrade
...
+    render json: {
+      errors: {
+        prepare: {
```
jsuchome: That's not `prepare` step, right?

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -76,19 +86,29 @@ def adminrepocheck
...
     else
       upgrade_status.end_step(
         false,
         admin_backup: @backup.errors.full_messages.first
-      ) if upgrade_status.current_step == :admin_backup
+      )
       render json: { error: @backup.errors.full_messages.first }, status: :unprocessable_entity
     end
+  rescue Crowbar::UpgradeStatus::StartStepRunningError,
+         Crowbar::UpgradeStatus::StartStepOrderError => e
```
jsuchome: Don't we need to check for end_step exceptions here as well?

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -76,19 +86,29 @@ def adminrepocheck
...
+    render json: {
+      errors: {
+        prepare: {
```
jsuchome: Wrong step name

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: i don't think so, or will it ever happen at that stage that the step is not running?

jsuchome: Well it could happen. Theoretically some other user could call the `end_step` already.

MaximilianMeister: afaik a user is not able to call the `end_step` at all directly

MaximilianMeister: i added it anyways as it doesn't hurt

MaximilianMeister:
rebased

jsuchome:
On note, not specific to this PR: we'll need to adapt crowbarctl to correctly handle error messages coming from the exception handlers.

MaximilianMeister:
> There's no check for library exceptions from adminrepocheck method.

oh right! thanks, fixed

aplanas:
@MaximilianMeister I can not properly review the change : ((

jsuchome:
Tested again (with curl) - I'm getting (expected) errors when trying to start the step twice or out of order.

PR #831, opened 2016-11-29
Do router migration back to founder node

abel-navarro:
As part of the founder upgrade, we moved all routers to a different node
in order to keep connectivity to the VMs. Once the founder node has been
upgraded it is safe to move back the routers.

**Requires:** https://github.com/crowbar/crowbar-openstack/pull/619

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
@@ -1,0 +1,45 @@
+#!/bin/bash
+#
+# This script migrates all neutron routers to the node where the script is
+# executed. As part of the upgrade, the founder node disbanded all routers
+# to other nodes. After the founder is upgraded, all routers have to come
+# back.
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+UPGRADEDIR=/var/lib/crowbar/upgrade
+mkdir -p "`dirname "$LOGFILE"`"
+exec >>"$LOGFILE" 2>&1
+
+echo "Executing $BASH_SOURCE"
+
+set -x
+
+mkdir -p $UPGRADEDIR
+
+if [[ -f $UPGRADEDIR/crowbar-router-migration-ok ]] ; then
+    echo "router migration actions were already successfully executed"
+    exit 0
+fi
+
+set +x
+source /root/.openrc
+set -x
+
+upgraded_agent=$(neutron agent-list --binary neutron-l3-agent --host $(hostname) -f value -c id)
+routers=$(neutron router-list -f value -c id)
+for router in $routers; do
+    agent=$(neutron l3-agent-list-hosting-router $router -f value -c id)
+    old_host=$(neutron agent-show $agent -f value -c host)
+
+    # delete old router assignment to agent
+    neutron l3-agent-router-remove $agent $router
+
+    # assign router to agent in upgraded node
+    neutron l3-agent-router-add $upgraded_agent $router
+
+    # delete old routers namespaces. Unassignment didn't delete the namespaces
+    ssh $old_host ip netns delete qrouter-$router
+
+done
+
+touch $UPGRADEDIR/crowbar-router-migration-ok
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -48,6 +48,14 @@ module Api
       false
     end
 
+    def router_migration
+      if execute_and_wait_for_finish("/usr/sbin/crowbar-router-migration.sh", 600)
+        save_node_state("Router migration was successful.")
+        return true
+      end
+      false
+    end
+
     # Execute post upgrade actions: prepare drbd and start pacemaker
     def post_upgrade
       if execute_and_wait_for_finish("/usr/sbin/crowbar-post-upgrade.sh", 600)
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -278,14 +278,16 @@ module Api
         return false unless non_founder_api.disable_pre_upgrade_attribute_for founder.name
 
         # 3. delete old pacemaker resources
-        # FIXME: couldn't we do it right after migrating l3 routers from the first node?
         delete_pacemaker_resources non_founder.name
 
         # 4. start crowbar-join at the first node
         return false unless founder_api.post_upgrade
         return false unless founder_api.join_and_chef
 
-        # 5. upgrade the rest of nodes in the same cluster
+        # 5. migrate routers from nodes being upgraded
+        return false unless founder_api.router_migration
+
+        # 6. upgrade the rest of nodes in the same cluster
         NodeObject.find(
           "state:crowbar_upgrade AND pacemaker_config_environment:#{cluster_env}"
         ).each do |node|
@@ -376,7 +378,8 @@ module Api
         # - initiate the first chef-client run
         return false unless node_api.join_and_chef
 
-        # FIXME: Delete router namespaces on non-upgraded node
+        # migrate routers from drbd_master to recently upgraded drbd_slave
+        return false unless node_api.router_migration
 
         save_upgrade_state("Starting the upgrade of node #{drbd_master}")
         return false unless master_node_api.upgrade
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -122,6 +122,14 @@ template "/usr/sbin/crowbar-delete-pacemaker-resources.sh" do
   )
 end
 
+template "/usr/sbin/crowbar-router-migration.sh" do
+  source "crowbar-router-migration.sh.erb"
+  mode "0755"
+  owner "root"
+  group "root"
+  action :create
+end
+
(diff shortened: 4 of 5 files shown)


chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -0,0 +1,29 @@
...
+
+source /root/.openrc
+routers=$(neutron router-list -f value -c id)
+for router in $routers; do
+    # delete old router assignment to agent
+    agent=$(neutron l3-agent-list-hosting-router $router -f value -c id)
+    neutron l3-agent-router-remove $agent $router
+
+    # assign router to agent in upgraded node
+    upgraded_agent=$(neutron agent-list --binary neutron-l3-agent --host $(hostname) -f value -c id)
```
vuntz: This can be done outside of the loop.

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -0,0 +1,29 @@
...
+    neutron l3-agent-router-add $upgraded_agent $router
+
+    # delete old routers namespaces. Unassignment didn't delete the namespaces
+    old_host=$(neutron agent-show $agent -f value -c host)   # check if we can use '$' inside '$()'
```
vuntz: This should probably be run before we do `neutron l3-agent-router-remove`.

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
abel-navarro: agreed, will do.

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
vuntz: I don't understand. Why can't you do:
```
agent=$(neutron l3-agent-list-hosting-router $router -f value -c id)
old_host=$(neutron agent-show $agent -f value -c host)
```
at the beginning?

The reason I'm suggesting this is that if we can move all the neutron calls (that are taking some time) early on, then the migration will be faster.

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -0,0 +1,29 @@
...
+for router in $routers; do
+    # delete old router assignment to agent
+    agent=$(neutron l3-agent-list-hosting-router $router -f value -c id)
+    neutron l3-agent-router-remove $agent $router
+
+    # assign router to agent in upgraded node
+    neutron l3-agent-router-add $upgraded_agent $router
+
+    # delete old routers namespaces. Unassignment didn't delete the namespaces
+    old_host=$(neutron agent-show $agent -f value -c host)
```
vuntz: Since my old "thread" about this is hidden by github by default: I think this line should be moved at the beginning of the block inside the loop, after the line defining `$agent`.

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+#
+# This script starts the upgrade of the node server to the latest version of
+# system and Cloud product.
```
jsuchome: Please fix the comment :-)

crowbar_framework/app/models/api/node.rb
```diff
@@ -48,6 +48,14 @@ def os_upgrade
       false
     end
 
+    def router_migration
+      if execute_and_wait_for_finish("/usr/sbin/crowbar-router-migration.sh", 600)
```
jsuchome: If we use this method (`execute_and_wait_for_finish`), the script needs to create those `-ok` and `-failed` files. Also, the timeout might be shorted, I believe.

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
abel-navarro: oops, will fix that.

crowbar_framework/app/models/api/node.rb
abel-navarro: Totally missed that, will handle those file creations.

abel-navarro:
self-note: we may want to call `neutron-ha-tool --l3-agent-rebalance` after all cluster nodes have been updated. Otherwise all the routers will remain in the founder node.

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -0,0 +1,42 @@
...
+set -x
+
+mkdir -p $UPGRADEDIR
+
+if [[ -f $UPGRADEDIR/router-migration-ok ]] ; then
+    echo "router migration actions were already successfully executed"
+    exit 0
+fi
+
+source /root/.openrc
```
jsuchome: We need to suppress logging around this, so we do not log the password. See https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb#L26

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -0,0 +1,42 @@
...
+mkdir -p "`dirname "$LOGFILE"`"
+exec >>"$LOGFILE" 2>&1
+
+echo "Executing $BASH_SOURCE"
+
+set -x
+
+mkdir -p $UPGRADEDIR
+
+if [[ -f $UPGRADEDIR/router-migration-ok ]] ; then
```
jsuchome: This needs to be full name of the file + "-ok" suffix, see https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/node_object.rb#L1089

jsuchome:
@abel-navarro Please call the `router_migration` method also from `upgrade_drbd_cluster` (there's even a FIXME for that)

jsuchome:
> we may want to call neutron-ha-tool --l3-agent-rebalance after all cluster nodes have been updated.

Good idea, you might even add such call to the end of controller upgrade methods. I think there's no need for a long running script in this case, so `node.run_ssh_cmd(cmd)` should be enought....

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -278,13 +278,15 @@ def upgrade_controller_nodes
         return false unless non_founder_api.disable_pre_upgrade_attribute_for founder.name
 
         # 3. delete old pacemaker resources
-        # FIXME: couldn't we do it right after migrating l3 routers from the first node?
         delete_pacemaker_resources non_founder.name
 
-        # 4. start crowbar-join at the first node
+        # 4. migrate routers from nodes being upgraded
+        return false unless founder_api.router_migration
```
rhafer: I think we need to check either here or in ```router_migration``` if the node really has the "neutron-network" role. And skip calling the script if it doesn't.

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
abel-navarro: done

crowbar_framework/app/models/api/upgrade.rb
abel-navarro: good point, addressing that.

abel-navarro: After looking at the code, I think we do not deal with a separated network cluster properly with our present code. We are basing the controller upgrade on the notion of [pacemaker] founder node and we assume it is also a network node (we move routers away from it).

The change you propose is right, but it involves more steps in order to make the solution fully compatible with a separated network cluster. I'm proposing at this point we still consider all clusters are in the same nodes and later on handle the network cluster separation. What do you think?

abel-navarro:
rebased

crowbar_framework/app/models/api/upgrade.rb
jsuchome: I think simple check like 
`@node["run_list_map"].key? "neutron-network"` inside `router_migration` should be enough, but yes, we can also do it in followup PR.

rhafer: I am fine doing this in followup PR as well. We just should not forget about it.

abel-navarro:

abel-navarro:
@jsuchome added the router migration on the drbd upgrade path. Notice that I put it before the post_upgrade step as I did on the non-drbd path.

jsuchome:
> Notice that I put it before the post_upgrade step as I did on the non-drbd path.

Ah, I did not see that. That's wrong, I think. neutron-server is running only after `join_and_chef` action

jsuchome:
So, I've tested it (at `vi7.cloud`, still available for anyone to look), and it has worked for me as expected. That's very close to approval, although I'd rather wait for further tests...

jsuchome:
@abel-navarro please, squash some commits (where it makes sense), and add an "upgrade: " prefix to the remaining ones

abel-navarro:
Squashed commits. Added `upgrade:` tag to commit message.

abel-navarro:
tested with DRBD. Working for me. I think this is ready to merge.

PR #835, opened 2016-11-30
cleanup: remove old landing page

MaximilianMeister:
for the upgrade from cloud 5 to 6 it was necessary to have a landing
page to choose to continue from cloud 5 upgrade or to perform a
fresh installation.
this is no longer needed for cloud 7 and we can immediately go to
the installer.


crowbar_framework/config/routes.rb
```diff
@@ -194,7 +194,7 @@
   end
 
   scope :installer do
-    root to: "installer#index",
+    root to: "installer/installers#show",
```
jsuchome: This is already implemented somewhere?

MaximilianMeister: yes, that's the installer page

rsalevsky:
Can we also drop the old upgrade related parts? See https://github.com/crowbar/crowbar-core/pull/667

vuntz:
> Can we also drop the old upgrade related parts? See #667

Yes, we should. But it doesn't need to be in this PR :-)

PR #836, opened 2016-11-30
Adds node alias to chef errors seen in crowbar UI

sayalilunkad:


crowbar_framework/app/models/service_object.rb
```diff
@@ -1297,9 +1297,11 @@ def apply_role(role, inst, in_queue, bootstrap = false)
 
       next if bad_nodes.empty?
 
-      message = "Failed to apply the proposal to: "
+      message = "Failed to apply the proposal to "
       bad_nodes.each do |node|
-        message += "#{node}\n"
+        node_obj = NodeObject.find_node_by_name(node)
+        node_alias = node_obj.alias
```
vuntz: You can simply replace these two lines with: `node_alias = pre_cached_nodes[node].alias`, I think.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1297,9 +1297,11 @@ def apply_role(role, inst, in_queue, bootstrap = false)
 
       next if bad_nodes.empty?
 
-      message = "Failed to apply the proposal to: "
+      message = "Failed to apply the proposal to "
```
vuntz: also, I wouldn't change that line as things can be a bit complicated in case many nodes have a failure, and the `:` make it clearer that it may be a list.

sayalilunkad:
@vuntz it does, tested it. Thanks!

rsalevsky:
Gating failed but seams unrelated.

sayalilunkad:
@rsalevsky yeah it is not related but I can re-run it to check is something changed.

aspiers:
Nice idea, thanks!

PR #838, opened 2016-12-01
fix typo and comparison in migration

MaximilianMeister:


jsuchome:
Oh, I think these commits could be squashed, right? There's no need to have 3 commits modifying one line...

MaximilianMeister:
> Oh, I think these commits could be squashed, right? There's no need to have 3 commits modifying one line...

sometimes it's good to see the succession of changes, even if the result seems really small. e.g. i was not sure about the regex usage (which doesn't have anything to do with the typo)

MaximilianMeister:
> tested: 'crowbarctl upgrade database new' now works for cloud6->cloud7

thanks for testing that

PR #839, opened 2016-12-01
upgrade: set status and restrict api for noderepochecks and service shutdown

MaximilianMeister:
## Depends on
~~https://github.com/crowbar/crowbar-core/pull/830~~ https://github.com/crowbar/crowbar-core/pull/847

@jsuchome I thought about doing the node upgrade status as well, but I'd like to wait for the implementation to be done before setting the status, because there are many different cases for setting the status as I saw in the code, and I am not sure if the model will still change, so I kept it untouched for now.

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -85,10 +85,23 @@ module Api
       end
 
       def noderepocheck
+        upgrade_status = ::Crowbar::UpgradeStatus.new
+        upgrade_status.start_step(:nodes_repo_checks)
+
         response = {}
         addons = Api::Crowbar.addons
         addons.push("os", "openstack").each do |addon|
           response.merge!(Api::Node.repocheck(addon: addon))
+        end
+
+        unavailable_repos = response.select { |_k, v| !v["available"] }
+        if unavailable_repos.any?
+          upgrade_status.end_step(
+            false,
+            nodes_repo_checks: "#{unavailable_repos.keys.join(", ")} repositories are missing"
+          )
+        else
+          upgrade_status.end_step
         end
         response
       end
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -183,6 +183,9 @@ module Api
 
       # Shutdown non-essential services on all nodes.
       def services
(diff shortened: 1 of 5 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -85,11 +85,20 @@ def best_method
...
+        upgrade_status.start_step(:nodes_repo_checks)
+
         response = {}
         addons = Api::Crowbar.addons
         addons.push("os", "openstack").each do |addon|
           response.merge!(Api::Node.repocheck(addon: addon))
         end
+
+        if response.any? { |_k, v| !v["available"] }
+          upgrade_status.end_step(false, "Some node repositories are missing")
```
jsuchome: Aren't we passing the error hash elsewhere?

MaximilianMeister: yes, thanks for the hint. fixed

jsuchome:
I think you can include node upgrade already there. The extra statuses we'll need to do are those substeps and current node, for which we're still missing the API. But start_step/end_step could be already added

MaximilianMeister:
> I think you can include node upgrade already there. The extra statuses we'll need to do are those substeps and current node, for which we're still missing the API. But start_step/end_step could be already added

i was just not sure about the `save_upgrade_state` and `save_error_state` methods, and what you had in mind with it, so this would be just the end_step invocation then?

jsuchome:
> i was just not sure about the save_upgrade_state and save_error_state methods

I'm not sure yet - save_upgrade_state is probably for the substep setting, but I think it's fine just to ignore both of them for now

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -170,23 +183,37 @@ def target_platform(options = {})
...
 
         # Initiate the services shutdown for all nodes
-        NodeObject.find("state:crowbar_upgrade").each(
-          &:shutdown_services_before_upgrade
-        )
+        errors = []
+        upgrade_nodes = NodeObject.find("state:crowbar_upgrade")
+        upgrade_nodes.each do |upgrade_node|
+          cmd = upgrade_node.shutdown_services_before_upgrade
+          next if cmd[0] == 200
```
jsuchome: `shutdown_services_before_upgrade` does not return anything, so we fail here

MaximilianMeister:

PR #840, opened 2016-12-01
Final actions needed after upgrade of cluster / compute node is done

jsuchome:
Needed together with https://github.com/crowbar/crowbar-ha/pull/168

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -30,7 +30,7 @@ Chef::Log.info("Current upgrade step: #{upgrade_step}")
 
 case upgrade_step
 
-when "revert_to_ready"
+when "revert_to_ready", "done_os_upgrade"
 
   service "crowbar_join" do
     action :enable
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -150,9 +150,13 @@ module Api
       )
       if state == "upgraded"
         progress = status.progress
-        remaining = progress[:remaining_nodes] - 1
-        upgraded = progress[:upgraded_nodes] + 1
-        ::Crowbar::UpgradeStatus.new.save_nodes(upgraded, remaining)
+        # This should not really happen, but in some corner cases,
+        # repeated upgrade of the node could have been invoked
+        if progress[:remaining_nodes] > 0
+          remaining = progress[:remaining_nodes] - 1
+          upgraded = progress[:upgraded_nodes] + 1
+          ::Crowbar::UpgradeStatus.new.save_nodes(upgraded, remaining)
+        end
       end
       @node.upgrade_state = state
     end
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -66,6 +66,13 @@ module Api
     end
 
     def join_and_chef
+      # Mark this upgrade step, so chef-client run can also start disabled services
+      unless @node["crowbar_wall"]["crowbar_upgrade_step"] == "done_os_upgrade"
+        # Make sure we save with the latest node data
+        @node = ::Node.find_by_name(@node.name)
+        @node["crowbar_wall"]["crowbar_upgrade_step"] = "done_os_upgrade"
+        @node.save
+      end
       begin
         execute_and_wait_for_finish("/usr/sbin/crowbar-chef-upgraded.sh", 600)
       rescue StandardError => e
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -165,7 +165,9 @@ module Api
           ::Crowbar::UpgradeStatus.new.save_nodes(upgraded, remaining)
         end
       end
-      @node.upgrade_state = state
+      @node = ::Node.find_by_name(@node.name)
+      @node["crowbar_wall"]["node_upgrade_state"] = state
+      @node.save
     end
 
     class << self
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -600,9 +600,24 @@ module Api
       # crowbar_upgrade_step will not be needed after node is upgraded
       def finalize_node_upgrade(node)
         return unless node["crowbar_wall"].key? "crowbar_upgrade_step"
+
         node["crowbar_wall"].delete "crowbar_upgrade_step"
         node["crowbar_wall"].delete "node_upgrade_state"
         node.save
+
+        scripts_to_delete = [
+          "prepare-repositories",
+          "upgrade-os",
+          "shutdown-services-before-upgrade",
+          "delete-cinder-services-before-upgrade",
+          "evacuate-host",
+          "pre-upgrade",
+          "delete-pacemaker-resources",
+          "router-migration",
+          "post-upgrade",
+          "chef-upgraded"
+        ].map { |f| "/usr/sbin/crowbar-#{f}.sh" }.join(" ")
+        node.run_ssh_cmd("rm -f #{scripts_to_delete}")
       end
 
       def finalize_nodes_upgrade
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -526,6 +526,8 @@ module Api
         if substep == "computes"
           upgrade_all_compute_nodes
         end
+
+        finalize_nodes_upgrade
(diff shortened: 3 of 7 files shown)


abel-navarro:
looks good. A placeholder for that neutron-ha-tool rebalance?

jsuchome:
Ah, yes. I believe it could be in the same method...

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -254,6 +254,19 @@ def prepare(options = {})
 
       protected
 
+      def finalize_cluster_upgrade(cluster)
+        save_upgrade_state("Finalizing the upgrade of nodes in cluster #{cluster}")
+
+        NodeObject.find("pacemaker_config_environment:#{cluster}").each do |node|
+          if node["pacemaker"]["founder"] && node.roles.include?("neutron-network")
+            node.run_ssh_cmd("source /root/.openrc; neutron-ha-tool --l3-agent-rebalance")
```
vuntz: Can we actually have multiple commands in `run_ssh_cmd` (as in: "does it take arbitrary bash code?")

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -254,6 +254,19 @@ def prepare(options = {})
...
+          end
+          node["crowbar_wall"]["crowbar_upgrade_step"] = "done"
+          node.save
+        end
+        # FIXME: we need to enable chef-client and crowbar-join services at some point
```
jsuchome: Hm.... that we could, true. Although, `join_and_chef` is still for one node only, executed when others are not upgraded.

crowbar_framework/app/models/api/upgrade.rb
jsuchome: I think it should work yes, but TBH I did not test it.

jsuchome: Yes, finally confirmed, this works.

crowbar_framework/app/models/api/upgrade.rb
rhafer: Hm, I am not sure I understand your proposal here. Wasn't the suggestion to just enable crowbar-join and chef-client as part of ```node_api.join_and_chef```? I.e. inside ```crowbar-chef-upgraded.sh```. Or is there a problem with enabling them at that stage already? (pacemaker sync-marks maybe?)

crowbar_framework/app/models/api/upgrade.rb
rhafer: There is an issue with relying on ```.openrc``` to be present  when calling any openstack cli during the upgrade. IIRC .openrc is only present on the node(s) that hold the keystone-server role. So in a multi-cluster setup the file might not be there. (We don't need to fix this now, but we should you track it somewhere)

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Enabling `crowbar-join` and `chef-client` directly from `node_api.join_and_chef` would work and not break sync marks, but it would be reset by the same chef-client run when executing `crowbar_upgrade` recipe (https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb#L67).

So I think we need another `crowbar_upgrade_step`, that will 1. tell crowbar_upgrade to start the service again and 2. not break skipping of pacemaker sync marks (as that is needed until we can declare whole cluster ready)

Maybe it's overcomplicated, but I do not see other way now...

crowbar_framework/app/models/api/upgrade.rb
jsuchome: What would the options be here? Copying/creating .openrc? Or passing credentials as arguments to ssh command (looks more ugly from the security POV)

jsuchome: @rhafer I've added a task to 
https://trello.com/c/MquyXrTn/2058-8-s49p16-orchestrate-upgrade-of-controller-nodes-in-ha-setup-upgrade-step-12-part-2-ha-multiple-clusters

jsuchome:
@rhafer @vuntz I've added commit that tries to deal with that chef-client/crowbar_join states to illustrate my point.
We would need to adapt `CrowbarPacemakerHelper.being_upgraded?` to add this new step into account

rhafer:
> I've added commit that tries to deal with that chef-client/crowbar_join states to illustrate my point.
> We would need to adapt CrowbarPacemakerHelper.being_upgraded? to add this new step into account

@jsuchome Ok, understood. And do you think there is something wrong with that? I actually thing the proposal is fine.

jsuchome:
> I actually thing the proposal is fine.

I also think this is correct, but I'm waiting for counter-proposals :-)

rhafer:
> but I'm waiting for counter-proposals

You won't get one from me. ;)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -285,6 +285,9 @@ def finalize_cluster_upgrade(cluster)
         save_upgrade_state("Finalizing the upgrade of nodes in cluster #{cluster}")
```
aspiers: Are you sure this commit message is right?  The routers should be distributed fairly equally across all upgraded nodes by the end of the upgrade, because when each node is upgraded, its routers are migrated to some random other node.

Having said that, a rebalance is still a good idea, but the commit message should be fixed.

jsuchome:
@aspiers Commit message adapted (it was also caused by the fact that this PR is older than latest work on routers migrations)

jsuchome:
Hm, actually maybe the bit about starting chef-client service is probably wrong. It could mean chef-client is started on some upgraded node while trying to upgrade another one in the cluster and it does not seem to end well.

jsuchome:
[Bring the nodes to correct state after they are upgraded](https://trello.com/c/oENUIUvd/2196-bring-the-nodes-to-correct-state-after-they-are-upgraded)

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -73,6 +73,25 @@
...
+    "pre-upgrade",
+    "delete-pacemaker-resources",
+    "router-migration",
+    "post-upgrade",
+    "chef-upgraded"
+  ].each do |f|
+    file "/usr/sbin/crowbar-#{f}.sh" do
+      action :delete
+    end
+  end
```
jsuchome: Or is it better to do it from rails app once, so chef-client won't by trying to delete those files on each run?

rhafer: Yeah, somehow I think we should do it from the rails app. BTW, do we have somehow remove the "crowbar-upgrade.rb " recipe from the run_list(s) once all nodes are upgraded? (I remember that was somehow tricky for the last release)

jsuchome:
About the role removal: yes, that is not covered here, I know. It is not critical, but I agree we should do it at some point.

jsuchome:
@rhafer adapted (without the role removal, I'd rather address it later)

crowbar_framework/app/models/api/node.rb
```diff
@@ -150,9 +150,13 @@ def save_node_state(role, state = "upgrading")
...
-        remaining = progress[:remaining_nodes] - 1
-        upgraded = progress[:upgraded_nodes] + 1
-        ::Crowbar::UpgradeStatus.new.save_nodes(upgraded, remaining)
+        # This should not really happen, but in some corner cases,
+        # repeated upgrade of the node could have been invoked
+        if progress[:remaining_nodes] > 0
+          remaining = progress[:remaining_nodes] - 1
+          upgraded = progress[:upgraded_nodes] + 1
+          ::Crowbar::UpgradeStatus.new.save_nodes(upgraded, remaining)
+        end
```
rhafer: Strictly speaking this doesn't seem to belong to this commit. In the future please create separate commits of unrelated changes.

jsuchome: Yes and no. In fact that "corner case" was introduced by the same commit - `finalize_node_upgrade` removes "node_upgrade_state" which makes it impossible to check if node is `upgraded?` Sure, it should not be needed after `finalize_node_upgrade`, but..

jsuchome:
@rhafer I've added one more commit after your review to make sure node changes are actually written.

rhafer:
> I've added one more commit after your review to make sure node changes are actually written.

@jsuchome Whould you mind explaining why that last change is needed? Under which circumstance might the node object that we already have be outdated?

jsuchome:
@rhafer  The node object is saved several times during the node upgrade, e.g. in `join_and_chef`. The last time it should be in `save_node_state`, but I can't say why it was not being saved for me without the last commit.

PR #841, opened 2016-12-02
[3.0] fix typo and comparison in migration

MaximilianMeister:
backport of https://github.com/crowbar/crowbar-core/pull/838


jsuchome:
Could you maybe squash the commits (original first)?
Also, I do not think this is really "upgrade:" related, it's just a fix for a bug that was introduced earlier and was not upgrade related...

PR #853, opened 2016-12-05
Make router migration synchronous

abel-navarro:
The fact that we call to neutron l3-agent-router-add doesn't mean the
actual router migration occurs immediately, although neutron call does.
When migrating several routers the phasing adds up which increate the
outtage times.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
@@ -25,7 +25,8 @@ set +x
 source /root/.openrc
 set -x
 
-upgraded_agent=$(neutron agent-list --binary neutron-l3-agent --host $(hostname) -f value -c id)
+hostname=$(hostname)
+upgraded_agent=$(neutron agent-list --binary neutron-l3-agent --host $hostname -f value -c id)
 routers=$(neutron router-list -f value -c id)
 for router in $routers; do
     agent=$(neutron l3-agent-list-hosting-router $router -f value -c id)
@@ -39,6 +40,18 @@ for router in $routers; do
 
     # delete old routers namespaces. Unassignment didn't delete the namespaces
     ssh $old_host ip netns delete qrouter-$router
+
+    # wait for router ports to migrate
+    ports=$(neutron router-port-list $router -f value -c id)
+    for port in $ports; do
+        port_host=$(neutron port-show $port -c binding:host_id -f value)
+        num_tries=60
+        while [ $port_host != $hostname ] || [ $num_tries -gt 0 ]; do
+            sleep 1
+            num_tries=$(( $num_tries - 1 ))
+            port_host=$(neutron port-show $router -c binding:host_id -f value)
+        done
+    done
 
 done
 


abel-navarro:
shamelessly based on @matelakat [work](https://github.com/matelakat/cookbook-openstack-network/blob/add-router-move/files/default/move-routers.sh).

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -40,6 +40,16 @@ for router in $routers; do
...
     ssh $old_host ip netns delete qrouter-$router
 
+    # wait for router ports to migrate
+    ports=$(neutron router-port-list $router -f value -c id)
+    for port in $ports; do
+        port_host=$(neutron port-show $router -c binding:host_id -f value)
+        while [ $port_host != $(hostname) ]; do
+            sleep 1
+            port_host=$(neutron port-show $router -c binding:host_id -f value)
+        done
```
vuntz: Should we have some timeout to give up at some point?

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -40,6 +40,16 @@ for router in $routers; do
     # delete old routers namespaces. Unassignment didn't delete the namespaces
     ssh $old_host ip netns delete qrouter-$router
 
+    # wait for router ports to migrate
+    ports=$(neutron router-port-list $router -f value -c id)
+    for port in $ports; do
+        port_host=$(neutron port-show $router -c binding:host_id -f value)
```
vuntz: s/$router/$port/

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -40,6 +40,16 @@ for router in $routers; do
...
+        while [ $port_host != $(hostname) ]; do
```
vuntz: Can we avoid evaluating `$(hostname)` all the time? We should just do it once outside of all loops :-)

abel-navarro: deal.

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
abel-navarro: done.

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -40,6 +40,16 @@ for router in $routers; do
     # delete old routers namespaces. Unassignment didn't delete the namespaces
     ssh $old_host ip netns delete qrouter-$router
 
+    # wait for router ports to migrate
+    ports=$(neutron router-port-list $router -f value -c id)
```
vuntz: It does matter a bit for error handling, but okay.

jsuchome:
Please also lower the timeout for this script execution ( https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/api/node.rb#L52 ) 
Or do you think it could take 10 minutes?

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -40,6 +41,18 @@ for router in $routers; do
...
+    for port in $ports; do
+        port_host=$(neutron port-show $port -c binding:host_id -f value)
+        num_tries=60
+        while [ $port_host != $hostname ] || [ $num_tries -gt 0 ]; do
+            sleep 1
+            num_tries=$(( $num_tries - 1 ))
+            port_host=$(neutron port-show $router -c binding:host_id -f value)
```
jsuchome: `$port` ?

jsuchome:
@rhafer so, do I understand it right that this PR would be obsolete by some work done for `neutron-ha-tool`? 
Do we still need this step (`crowbar-router-migration.sh.erb`), or do we need to adapt it to use `neutron-ha-tool` as well?

rhafer:
> so, do I understand it right that this PR would be obsolete by some work done for neutron-ha-tool?

Yeah, I think this is obsolete. At least partially.

> Do we still need this step (crowbar-router-migration.sh.erb), or do we need to adapt it to use neutron-ha-tool as well?

Not sure I fully understand the question. We'll need the step, yes. But if we still need the script (modified to call neutron-ha-tool) or if we trigger the execution from crowbar directly, I can't say yet. (My guess is we will still need some tiny wrapper around the neutron-ha-tool call. If just for the creation of the -ok and -failed files.

jsuchome:

sidthe:
@abel-navarro any chance this can be closed today?

abel-navarro:
Since the work is going to be overriden by https://github.com/SUSE-Cloud/cookbook-openstack-network/pull/7 I'm closing this one.

aspiers:
[P2: bsc#1008055: L3: network HA cluster maintenance issue](https://trello.com/c/ni7N64ev/1-p2-bsc-1008055-l3-network-ha-cluster-maintenance-issue)

PR #856, opened 2016-12-05
Add ssh connection Remove NodeObject dependency from Api::Node

felixsch:
**Refactoring NodeObject and Api::Node to a unified class**

Long term goal is to melt Api::Node and NodeObject to a unified class and move as most as possible to the database. Using the Api::Node class as base class because it's the not overly complicated one of both. Renaming the class after refactoring should be no problem.

**in this commit**
By duplicating the already existing ssh connection code from NodeObject to a own Class it is possibly to decouple most of the code from Api::Node from NodeObject and move SSH Connection implementation in its own namespace. This is neccessary to make Api::Node independent to NodeObject. Later NodeObject will inherit from Api::Node.

Follow up pull request will be removing now duplicated ssh code from NodeObject.

### DIFF
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -16,16 +16,20 @@
 
 module Api
   class Node < Tableless
+    attr_reader :name
+    attr_reader :ssh
+
     def initialize(name = nil)
-      @node = NodeObject.find_node_by_name name
+      @name = name
+      @ssh = ::Crowbar::Connection::SSH.new("root", name)
     end
 
     # execute script in background and wait for it to finish
     def execute_and_wait_for_finish(script, seconds)
-      Rails.logger.info("Executing #{script} at #{@node.name}...")
-      @node.wait_for_script_to_finish(script, seconds)
+      Rails.logger.info("Executing #{script} at #{name}...")
+      ssh.exec_script(script, seconds)
       true
-    rescue StandardError => e
+    rescue RuntimeError => e
       save_error_state(
         e.message + "Check /var/log/crowbar/node-upgrade.log for details."
       )
@@ -76,15 +80,15 @@ module Api
     def wait_for_ssh_state(desired_state, action)
       Timeout.timeout(300) do
         loop do
-          ssh_status = @node.ssh_cmd("").first
-          break if desired_state == :up ? ssh_status == 200 : ssh_status != 200
+          ssh_status = ssh.exec_in_background("")
+          break if ssh_status == (desired_state == :up)
           sleep(5)
         end
       end
       true
     rescue Timeout::Error
       save_error_state(
-        "Possible error at node #{@node.name}" \
+        "Possible error at node #{name}" \
         "Node did not #{action} after 5 minutes of trying."
       )
       false
@@ -92,8 +96,7 @@ module Api
 
     # Reboot the node and wait until it comes back online
     def reboot_and_wait
-      ssh_status = @node.ssh_cmd("/sbin/reboot")
-      if ssh_status[0] != 200
+      unless ssh.exec_in_background("/sbin/reboot")
         save_error_state("Failed to reboot the machine. Could not ssh.")
         return false
       end
@@ -109,7 +112,7 @@ module Api
 
       # this is just a fallback check, we should know by checking the global status that the action
       # should not be executed on already upgraded node
-      return true if @node.file_exist? "/var/lib/crowbar/upgrade/node-upgraded-ok"
+      return true if ssh.file_exist? "/var/lib/crowbar/upgrade/node-upgraded-ok"
 
       unless pre_upgrade
         save_error_state("Error while executing pre upgrade script")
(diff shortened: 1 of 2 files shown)


felixsch:
[(5) #crowbar Create node object that is backed by database](https://trello.com/c/EsVoyO2G/202-5-crowbar-create-node-object-that-is-backed-by-database)

crowbar_framework/app/models/api/node.rb
```diff
@@ -16,16 +16,20 @@
 
 module Api
   class Node < Tableless
+    attr_reader :node_name
```
vuntz: Why not simply `name`?

crowbar_framework/app/models/api/node.rb
```diff
@@ -16,16 +16,20 @@
...
+    attr_reader :ssh
+
     def initialize(name = nil)
-      @node = NodeObject.find_node_by_name name
+      @node_name = name
+      @ssh = Crowbar::Connection::SSH.new("root", name)
```
vuntz: Does it really make sense to initialize this here? It sounds to me like there should be some class method like: `Crowbar::Connection::SSH.exec(hostname, user, script)`

crowbar_framework/app/models/api/node.rb
felixsch: I did this before but then I realized there will be lot of methods which will take a name so I decided to name the accessor more explicitly.
But yes could be named also simply 'name'

felixsch: But you're right when moving to ActiveRecord the column name will be name I guess so it makes sense to change it to name!

jsuchome:
I'm not sure about this. Actually there was an idea that we should move all methods from Api::Node directly into NodeObject or some new child object of NodeObject...

crowbar_framework/app/models/api/node.rb
```diff
@@ -76,24 +80,23 @@ def join_and_chef
     def wait_for_ssh_state(desired_state, action)
       Timeout.timeout(300) do
         loop do
-          ssh_status = @node.ssh_cmd("").first
-          break if desired_state == :up ? ssh_status == 200 : ssh_status != 200
+          ssh_status = ssh.exec_in_background("")
+          break if ssh_status == (desired_state == :up ? true : false)
```
MaximilianMeister:
```ruby
break if ssh_status == (desired_state == :up)
```
should work too

crowbar_framework/lib/crowbar/connection/ssh.rb
```diff
@@ -0,0 +1,104 @@
+#
+# Copyright 2011-2013, Dell
+# Copyright 2013-2014, SUSE LINUX Products GmbH
```
MaximilianMeister: old header :)

rsalevsky:
@jsuchome The idea was to create a new NodeObject which is ActiveRecord based and only uses some functions from the old NodeObject. Felix currently works on the task to create a NodeObject which is ActiveRecord based to do so we need to get rid of the old NodeObject references first. I know it's a bit confusing but I guess we missed that this task was also on the list.

jsuchome:
Well I'm confused because we've had another discussion about reorganization of Api::Node and NodeObject with @MaximilianMeister and @aspiers , see https://etherpad.nue.suse.com/p/cloud-upgrade-models

MaximilianMeister:
> Well I'm confused because we've had another discussion about reorganization of Api::Node and NodeObject with @MaximilianMeister and @aspiers , see https://etherpad.nue.suse.com/p/cloud-upgrade-models

I think the discussion doesn't directly have to do with this change here, as we were only talking about the upgrade related node object methods, and reorganizing them.

jsuchome:
Yes, but currently everything in `Api::Node` is upgrade related.

rsalevsky:
We just decided recently that we want to put parts of the NodeObject into the database. Before it was not clear when and if we do this. I was not aware of the discussion about reorganizing the API::Node but AFIK the idea is based that we keep the old NodeObject. Due to this we should wait till we finished moving stuff in the database to avoid duplicating work. If there are still open questions or concerns then we should do a quick chat in mumble.

jsuchome:
CC @aspiers . I still do not think we should work on Api::Node at all, and we should rather move it's functionality into NodeObject

felixsch:
Do you mean the better way would be make NodeObject a ActiveRecord and let Api::Node inherit from NodeObject? Goal is to melt both classes into one.
From a perspective of reworking it makes no sense to take the 2000 lines class and extend it further. If create a new seperate class (Node?) and move refactored parts into it. Just my cents :)

jsuchome:
I think the idea of https://etherpad.nue.suse.com/p/cloud-upgrade-models was to create children of  NodeObject and move all upgrade related code there (and get rid of Api::Node totally). Currently some upgrade related parts are in Api:Node and some in NodeObject. 

But I'd let @MaximilianMeister and @aspiers comment here, as they were involved in that proposal.

felixsch:
@aspiers I updated the description to hopefully clarify things.
Since I'm out at school (the next two weeks) I will answer here:

- The dry violation is intentional since I wanted to start refactoring only in one place. Hopefully updated description makes it more clear.
- Yes goal is to melt both classes into one class. Using the smaller class as base class is more easy to deal with (from my point of view).
- The card I'm working on is to make the NodeObject/Api::Node database dependent. If there is a problem, this should be discussed. 
- I thought squashing all commits into one is the way to go.
- Your definitely right. I updated the description

aspiers:
> @aspiers I updated the description to hopefully clarify things.

Thanks!

> Since I'm out at school (the next two weeks) I will answer here:
> 
> - The dry violation is intentional since I wanted to start refactoring only in one place. Hopefully updated description makes it more clear.

Yes, it does, thanks.  But I really don't think that duplicating code is the right way to achieve a unified class, for all the reasons which make DRY such an important rule.  Instead, please consider *moving* code.

> - Yes goal is to melt both classes into one class. Using the smaller class as base class is more easy to deal with (from my point of view).

It might be easier, but it's just wrong because that class should not exist (and is only used in an upgrade context).  So it's making the problem worse not better.

> - The card I'm working on is to make the NodeObject/Api::Node database dependent. If there is a problem, this should be discussed.

This can be OK as long as it's not duplicating a part of the model between Chef and the DB.  There should be a single source of truth for each piece of data, otherwise it introduces all kinds of new problems.

> - I thought squashing all commits into one is the way to go.

Not in this case.  Please read https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes

PR #859, opened 2016-12-06
upgrade: adapt the schema migration logic to the new workflow

MaximilianMeister:


crowbar_framework/lib/schema_migration.rb
```diff
@@ -25,6 +25,12 @@ def self.run
   end
 
   def self.run_for_bc bc_name
+    return unless File.exist?("/var/lib/crowbar/install/crowbar-installed-ok")
+    return if File.exist?("/var/run/crowbar/admin-server-upgrading")
+    unless system("systemctl is-active postgresql")
+      raise "Cannot run schema migrations. Database server is not running"
```
rsalevsky: Can you mention the bc_name here too?

MaximilianMeister: done

PR #860, opened 2016-12-06
provisioner: Cache chef_client_runs (bsc#1013556)

rsalevsky:
To avoid heavy spamming of the log due to the regular database queries
let's cache the value till it changes.

### DIFF
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -528,13 +528,8 @@ class NodeObject < ChefObject
     return "unknown" if (@node.nil? or @role.nil?)
     if self.crowbar["state"] === "ready" and @node["ohai_time"]
       since_last = Time.now.to_i-@node["ohai_time"].to_i
-      crowbar_proposal = Proposal.where(barclamp: "provisioner").first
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/node_object.rb
```diff
@@ -528,18 +528,23 @@ def state
...
       return "noupdate" if since_last > max_last
     end
     return self.crowbar["state"] || "unknown"
   end
 
+  def max_last
+    frequent_time = Rails.cache.read("chef_client_runs")
+    return frequent_time + 300 unless frequent_time.blank?
+
+    proposal = Proposal.where(barclamp: "provisioner").first
```
aspiers: As I asked previously, why you are querying the proposal this way? Maybe you already answered by email and I didn't see, since I'm way behind on email currently.  But this is inconsistent with the way pretty much everywhere else we read attributes from proposals for a specific node.  The proposal's attributes are already applied to the node via the proposal role, so you can simply read the attribute directly off the Chef node object.

rsalevsky: Maybe I'm doing something wrong but there is no chef_client_runs in `@node`.

PR #861, opened 2016-12-06
upgrade: only allow requests to the upgrade api during the upgrade

MaximilianMeister:
UPDATE (jsuchome): requires https://github.com/crowbar/crowbar-core/pull/883

### DIFF
--- a/crowbar_framework/app/controllers/application_controller.rb
+++ b/crowbar_framework/app/controllers/application_controller.rb
@@ -33,6 +33,10 @@ class ApplicationController < ActionController::Base
     Crowbar::Installer.successful? || \
     Rails.env.test?
   }
+  before_filter :upgrade, if: proc {
+    File.exist?("/var/lib/crowbar/upgrade/6-to-7-progress.yml") && \
+      !File.exist?("/var/lib/crowbar/upgrade/6-to-7-upgraded-ok")
+  }
   before_filter :sanity_checks, unless: proc {
     Rails.env.test? || \
     Rails.cache.fetch(:sanity_check_errors).empty?
@@ -192,6 +196,18 @@ class ApplicationController < ActionController::Base
       end
       format.json do
         render json: { error: I18n.t("error.before_install") }, status: :unprocessable_entity
+      end
+    end
+  end
+
+  def upgrade
+    respond_to do |format|
+      format.html do
+        # redirect to the upgrade ui wizard
+        redirect_to "/upgrade"
+      end
+      format.json do
+        render json: { error: I18n.t("error.during_upgrade") }, status: :service_unavailable
       end
     end
   end
--- a/crowbar_framework/app/controllers/application_controller.rb
+++ b/crowbar_framework/app/controllers/application_controller.rb
@@ -202,12 +202,15 @@ class ApplicationController < ActionController::Base
 
   def upgrade
     respond_to do |format|
-      format.html do
-        # redirect to the upgrade ui wizard
-        redirect_to "/upgrade"
-      end
       format.json do
-        render json: { error: I18n.t("error.during_upgrade") }, status: :service_unavailable
+        if request.post?
+          render json: { error: I18n.t("error.during_upgrade") }, status: :service_unavailable
+        else
+          return
+        end
+      end
+      format.html do
+        render "errors/during_upgrade", status: :service_unavailable
       end
     end
   end
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -15,6 +15,8 @@
 #
 
 class Api::UpgradeController < ApiController
+  skip_before_filter :upgrade
+
   api :GET, "/api/upgrade", "Show the Upgrade progress"
   header "Accept", "application/vnd.crowbar.v2.0+json", required: true
   api_version "2.0"
--- a/crowbar_framework/app/views/errors/during_upgrade.html.haml
+++ b/crowbar_framework/app/views/errors/during_upgrade.html.haml
@@ -1,0 +1,28 @@
+.navbar.navbar-inverse.navbar-fixed-top
+  .container
+    .navbar-header
+      %button.navbar-toggle
+        %span.sr-only
+          Navigation
+        %span.icon-bar
+        %span.icon-bar
+        %span.icon-bar
+      %a{ href: "/" }
+        .navbar-brand
+          %span.title
+            Crowbar
+    .collapse.navbar-collapse.navbar-right
+      %ul.nav.navbar-nav
+        %li
+          %a{ href: "/upgrade" }
+            Go to upgrade wizard
+
+.panel.panel-default.panel-warning
+  .panel-heading
+    The requested service is temporarily unavailable
+
+  .panel-body
+    %p
+      You are currently upgrading your deployment and therefore access to the Crowbar UI is
+      restricted to the <a href="/upgrade">upgrade wizard</a>.
+      Please finish the upgrade first.
--- a/crowbar_framework/config/locales/crowbar/en.yml
+++ b/crowbar_framework/config/locales/crowbar/en.yml
@@ -900,6 +900,7 @@ en:
     chef_server_down: 'Chef server is not available'
     param_missing: 'A required parameter is missing'
     before_install: 'This API is not available before installation'
+    during_upgrade: 'This API is not available during upgrade'
 
   state:
     unknown: 'Unknown'
--- a/crowbar_framework/spec/controllers/machines_controller_spec.rb
+++ b/crowbar_framework/spec/controllers/machines_controller_spec.rb
@@ -25,6 +25,9 @@ describe MachinesController do
   describe "GET index" do
     before do
       allow(File).to receive(:exist?).and_return(true)
+      allow(File).to receive(:exist?).with(
+        "/var/lib/crowbar/upgrade/6-to-7-progress.yml"
+      ).and_return(false)
     end
 
     it "is successful" do


crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -37,6 +37,9 @@ class ApplicationController < ActionController::Base
     Rails.env.test? || \
     Rails.cache.fetch(:sanity_check_errors).empty?
   }
+  before_filter :upgrade, if: proc {
+    File.exist?("/var/run/crowbar/admin-server-upgrading")
```
jsuchome: Well... this is only present when admin server is being upgraded. Don't we want to limit all non-upgrade calls during the while upgrade process?

MaximilianMeister:

jsuchome:
Actually, I'm not sure this is correct. I'm not able to do the checks I usually do during the upgrade:

```
crowbar_node_state status
Failed to talk to service show: 503: {"error":"This API is not available during upgrade"}
```

MaximilianMeister:
@jsuchome now GET requests via `json` are allowed and all rehular UI access will render:
ui-restriction

crowbar_framework/public/503_upgrade.html
```diff
@@ -0,0 +1,67 @@
...
+      <div class="panel panel-default panel-warning">
+        <div class="panel-heading">
+          <h2>
+            The requested service is temporarily unavailable
+          </h2>
+        </div>
+        <div class="panel-body">
+          <p>
+            You are currently upgrading your cloud and some API is restricted.
+            Please finish the upgrade and try again.
```
jsuchome: Please finish the upgrade first.

crowbar_framework/public/503_upgrade.html
```diff
@@ -0,0 +1,67 @@
...
+    <div class="container" id="errorpage">
+      <div class="panel panel-default panel-warning">
+        <div class="panel-heading">
+          <h2>
+            The requested service is temporarily unavailable
+          </h2>
+        </div>
+        <div class="panel-body">
+          <p>
+            You are currently upgrading your cloud and some API is restricted.
```
jsuchome: Maybe "some API calls are forbidden"? Or even "most of API calls"?

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -195,4 +198,26 @@ def sanity_checks
       end
     end
   end
+
+  def upgrade
+    if request.post?
+      respond_to do |format|
+        format.html do
+          render file: "public/503_upgrade.html"
+        end
```
vuntz: Can't we redirect to the upgrade UI?

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -37,6 +37,9 @@ class ApplicationController < ActionController::Base
     Rails.env.test? || \
     Rails.cache.fetch(:sanity_check_errors).empty?
   }
+  before_filter :upgrade, if: proc {
+    File.exist?("/var/lib/crowbar/upgrade/progress.yml")
+  }
```
vuntz: Not sure it matters, but I would have put this before the sanity checks.

MaximilianMeister: you're right, it may matter in some cases

crowbar_framework/app/controllers/application_controller.rb
vuntz: But that's what we do for the installer :-)

MaximilianMeister: yes, it is, but anyone who hasn't installed crowbar yet will likely not want to go to the barclamp page. so there it makes sense IMO. and in `crowbarctl` we have a similar warning.

crowbar_framework/public/503_upgrade.html
```diff
@@ -0,0 +1,67 @@
...
+          </h2>
+        </div>
+        <div class="panel-body">
+          <p>
+            You are currently upgrading your cloud and therefore access to the Crowbar UI is
+            restricted to the <a href="/upgrade">upgrade wizard</a>.
+            Please finish the upgrade first.
+          </p>
+        </div>
+      </div>
```
MaximilianMeister: the footer is actually rendered by Rails, that's why I took it out from the HTML

crowbar_framework/app/controllers/application_controller.rb
vuntz: But any user who starts an upgrade will not be surprised to be redirected to the upgrade page :-) Anyway, I don't mind much.

crowbar_framework/public/503_upgrade.html
vuntz: Hrm, can we add a HTML comment about this? Because I'm sure someone will add it in the future to make the page consistent with the other static pages.

(Btw, can't we have rails render other bits too?)

MaximilianMeister: hmm, after all it might be better to render it from `views/errors` so we can leave the public folder for apache only

MaximilianMeister:
updated screenshot with the `haml` view:

ui-restriction

MaximilianMeister:
@vuntz could you have another look?

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -200,13 +200,23 @@ def sanity_checks
...
+        format.html do
+          render "errors/during_upgrade", status: :service_unavailable
+        end
+        format.json do
+          render json: { error: I18n.t("error.during_upgrade") }, status: :service_unavailable
+        end
+      end
+    else
+      respond_to do |format|
+        format.html do
```
aspiers: AFAICS this is needless code duplication - you could just keep the original structure and put the `request.post?` conditional inside the `format.json` block.  Am I missing something?

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -33,6 +33,9 @@ class ApplicationController < ActionController::Base
     Crowbar::Installer.successful? || \
     Rails.env.test?
   }
+  before_filter :upgrade, if: proc {
+    File.exist?("/var/lib/crowbar/upgrade/progress.yml")
```
skazi0: I thought the plan was that this file will remain in the filesystem after the upgrade is finished. Otherwise the status API will not work correctly.

MaximilianMeister: do we need the upgrade status api after the upgrade?

crowbar_framework/app/controllers/application_controller.rb
MaximilianMeister: thanks, fixed

crowbar_framework/app/controllers/application_controller.rb
skazi0: Don't know. IIRC you were the one to consider this as needed.
1. How would the API respond if the file is missing?
2. When will this file be deleted and by which component? (or maybe this is done already?)
3. When does "upgrade" begin? (i.e. when is this file created?)
4. What about "cancel"? Does it remove the file?

jsuchome: So, the library could have *another* file indicating that the upgrade is currently running, like `/var/lib/crowbar/upgrade/running`. That would be deleted by progress library when the library reaches the last step (and it succeeds). 
It's obviously not adding any new information to what we already have, its only reason would be so the check here in the `application_controller.rb` could be as simple (and fast) as possible.

jsuchome: Or, we can just grep for current step of the progress file, the last value should be `:finished`. Isn't it faster than loading and parsing the YML?

jsuchome: Also, to answer those questions:

1. if it were missing, it's considered that the upgrade is not running and any API can go through
2. Good question indeed.
3. When the `start_step` is called for the first timr
4. Yes

MaximilianMeister: @skazi0 it should be as cheap as possible, as we are running this check on every request. there might be even better ways, but a check for file existence seems quicker than a full parse of a status file. 
i thought about `Rails.cache` too which would be really quick I guess. It could be written at the time of the prechecks and after finishing the upgrade. The only problem is that it won't persist after a reboot.

skazi0: I think adding separate `running` file is the best of options so far. I don't like removing the status file after upgrade is `finished` as this would prevent us from even noticing this. As soon as the `current_step` would be set to `finished` the status API would stop responding.
Removing the `running` file should be pretty simple as it can be bound to step transition in status library. Removing the whole state of status file library is IMO more fragile.

aspiers: I'd prefer a solution which will work cleanly if we use a similar mechanism for 7->8 upgrades.  The current file gives no indication of which product version it refers to.  So it will lie around for a long time until 7->8 upgrade, and then what happens?  OTOH, we could have it called

    /var/lib/crowbar/upgrade/6-to-7-progress.yml

and then on completion it could be renamed to

    /var/lib/crowbar/upgrade/6-to-7-completed.yml

This would allow the same quick file existence tests but also make it future-proof.

MaximilianMeister: updated

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -33,6 +33,9 @@ class ApplicationController < ActionController::Base
     Crowbar::Installer.successful? || \
     Rails.env.test?
   }
+  before_filter :upgrade, if: proc {
+    File.exist?("/var/lib/crowbar/upgrade/6-to-7-progress.yml")
```
jsuchome: That's not enough. We are not deleting `6-to-7-progress.yml` after the upgrade is done, we're touching `/var/lib/crowbar/upgrade/6-to-7-upgraded-ok` once the upgrade is finished...

MaximilianMeister: oh true, i confused the files

PR #865, opened 2016-12-07
crowbar: Fix crash when displaying chef log on error

vuntz:
The node is not necessarily cached.

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1302,7 +1302,7 @@ class ServiceObject
       bad_nodes.each do |node|
         pre_cached_nodes[node] ||= NodeObject.find_node_by_name(node)
         node_alias = pre_cached_nodes[node].alias
-        message += "#{node_alias}: #{node}\n"
+        message += "#{node_alias} (#{node}):\n"
         message += get_log_lines(node)
       end
       update_proposal_status(inst, "failed", message)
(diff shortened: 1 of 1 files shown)


vuntz:
@sayalilunkad my suggestion in https://github.com/crowbar/crowbar-core/pull/836 caused some crash, sorry.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1300,8 +1300,9 @@ def apply_role(role, inst, in_queue, bootstrap = false)
 
       message = "Failed to apply the proposal to:\n"
       bad_nodes.each do |node|
+        pre_cached_nodes[node] ||= NodeObject.find_node_by_name(node)
         node_alias = pre_cached_nodes[node].alias
-        message += "#{node_alias}: #{node}\n"
+        message += "#{node_alias} (#{node})\n"
```
sayalilunkad: The colon is missing here. Is that wanted?

vuntz: Fixed

crowbar_framework/app/models/service_object.rb
```diff
@@ -1300,8 +1300,9 @@ def apply_role(role, inst, in_queue, bootstrap = false)
 
       message = "Failed to apply the proposal to:\n"
       bad_nodes.each do |node|
+        pre_cached_nodes[node] ||= NodeObject.find_node_by_name(node)
         node_alias = pre_cached_nodes[node].alias
-        message += "#{node_alias}: #{node}\n"
+        message += "#{node_alias} (#{node}):\n"
```
sayalilunkad: The colon should be in between like this "#{node_alias}: (#{node})\n" Having it at the end does not make much sense because there is no other message after that.

vuntz: No, it's at the end because after this line, we have the log for this host.

PR #867, opened 2016-12-07
rspec: fix unit tests

MaximilianMeister:
PR #839 wasn't fully rebased at the time of merging, so another
change in the upgrade status lib requires the unit tests to be adapted


jsuchome:
It's still upgrade related, right? (-> commit message prefix)

MaximilianMeister:
> It's still upgrade related, right? (-> commit message prefix)

should be OK, as it already got approved

PR #868, opened 2016-12-07
network: Run wicked ifreload only once per chef-client run (bsc#1012587)

rhafer:
This reworks the fix for https://bugzilla.suse.com/show_bug.cgi?id=1011889
to ensure that wicked ifreload is only execute at most once for a single
chef-client run. Previouly it was triggered once per updated ifcfg file
and that caused problems for certain configurations. E.g. when a single
chef-client run set up an "normal" physcial interface and a vlan
interface on top of that. (Happens e.g. in "dual" network setups when
allocating a public IP and when public and admin are on different
physical interfaces.)

https://bugzilla.suse.com/show_bug.cgi?id=1012587

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -490,7 +490,7 @@ when "suse"
         interfaces: ifs,
         nic: nic
       })
-      notifies :run, "bash[wicked-ifreload-#{nic.name}]", :immediately
+      notifies :create, "ruby_block[wicked-ifreload-required]", :immediately
     end
     if ifs[nic.name]["gateway"]
       template "/etc/sysconfig/network/ifroute-#{nic.name}" do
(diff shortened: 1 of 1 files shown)


chef/cookbooks/network/recipes/default.rb
```diff
@@ -499,23 +499,40 @@ def kill_nic(nic)
...
+  # way (instead of notifying the "wicked-ifreload-all" resource directly),
+  # is to make sure that the ifrelaod is only run once after all ifcfg
+  # file have been update and independent of how many of them were changed.
+  ruby_block "wicked-ifreload-required" do
+    block do
+      r = resources(bash: "wicked-ifreload-all")
+      a = Array.new(r.action)
+      a << :run unless a.include?(:run)
+      a.delete(:nothing) if a.include?(:run)
+      r.action(a)
```
vuntz: Wow, that's some pretty amazing creativity here :-) I'm tempted of giving a +1 because you're teaching me some new magic here.

Asking out loud about other options first, though:
- is it maybe possible to have a global variable that the `ruby_block` would change and that would be used in the `code` of the `bash` resource?
- what about using a file as a marker to do the action?
- and what about simply running the `bash` resource all the time? Since we start with a `ifcheck`, it should not hurt that much, should it?

rhafer: >  I'm tempted of giving a +1 because you're teaching me some new magic here

@vuntz I must admit that this wasn't my idea :)

> * is it maybe possible to have a global variable that the ruby_block would change and that would be used in the code of the bash resource?
> * what about using a file as a marker to do the action?

Hm, I considered the 2nd approach for a while, but I find that really ugly. And I think it's more prone to errors than the way it's currently implemented. For the 1st suggestion, I havn't tried that. I can give it a spin.

> * and what about simply running the bash resource all the time? Since we start with a ifcheck, it should not hurt that much, should it?

Yeah, I considered that as well. But I am not 100% sure if it really doesn't do any harm. If you e.g. run "wicked ifreload" on a cluster node with VIPs assigned it will cause the VIPs to removed for a while (until the pacemaker resource adds them back). That still should happen normally (because of the ifcheck). But I'd really like to restrict to the cases when the config file got updated.

rhafer:
> is it maybe possible to have a global variable that the ruby_block would change and that would be used in the code of the bash resource?

@vuntz I've now changed the implementation in that way. Seems to work as well. But is of course a lot less magic. **\*sniff\*** ;-)

PR #869, opened 2016-12-07
return err406 on alias rename failure

sidthe:
this change complements the PR https://github.com/crowbar/crowbar-client/pull/128 and introduces a custom failure for alias rename failure

### DIFF
--- a/crowbar_framework/app/controllers/machines_controller.rb
+++ b/crowbar_framework/app/controllers/machines_controller.rb
@@ -88,10 +88,20 @@ class MachinesController < BarclampController
 
   add_help(:rename, [:id])
   def rename
-    @machine.alias = params[:alias]
+    alias_error = nil
+    begin
+      @machine.alias = params[:alias]
+    rescue StandardError => e
+      alias_error = e.message
+    end
 
     respond_to do |format|
-      if @machine.save
+      if alias_error
+        format.json do
+          render json: { error: "#{I18n.t("cannot_save_node", scope: "error")} #{alias_error}" },
+                 status: :not_acceptable
+        end
+      elsif @machine.save
         format.json { head :ok }
       else
         format.json do


sidthe:
[P5: bsc#1011581 - Renaming of node to already taken name via crowbarctl fails with 500 Internal Server Error](https://trello.com/c/vk9zvmGN/204-p5-bsc-1011581-renaming-of-node-to-already-taken-name-via-crowbarctl-fails-with-500-internal-server-error)

crowbar_framework/app/controllers/machines_controller.rb
```diff
@@ -88,10 +88,19 @@ def role
 
   add_help(:rename, [:id])
   def rename
-    @machine.alias = params[:alias]
+    begin
+      @machine.alias = params[:alias]
+    rescue StandardError => e
```
MaximilianMeister: can we catch a specific exception here? or can it always be a different exceprtion?

crowbar_framework/app/controllers/machines_controller.rb
```diff
@@ -88,10 +88,19 @@ def role
...
+      alias_error = e.message
+    end
 
     respond_to do |format|
-      if @machine.save
+      if !alias_error.nil?
```
MaximilianMeister: i think that won't work in the case where the variable doesn't exist. you'd have to set it to nil explicitly somewhere above

crowbar_framework/app/controllers/machines_controller.rb
sidthe: well, that catches all exceptions specific to what happens in the functions that proceed rename in the stack. And those pretty much deal just with alias validation.

crowbar_framework/app/controllers/machines_controller.rb
sidthe: yeah, I though so as well, but when tested, works properly

sidthe: @MaximilianMeister I added it there just in case, but as I was saying, in testing that did not show issues

MaximilianMeister: just `if alias_error` should work the same way

crowbar_framework/app/controllers/machines_controller.rb
MaximilianMeister: just looked it up, we don't throw any specific exception there

crowbar_framework/app/controllers/machines_controller.rb
rsalevsky: `unless alias_error.blank?` would be better here

sidthe: @rsalevsky excuse my ignorance, but why would blank be better?

sidthe:
@rsalevsky I just did if alias_error - its cleaner

sidthe:
> unless alias_error.blank? would be better here

@rsalevsky just did an if alias_error, it's cleaner

PR #870, opened 2016-12-07
upgrade: Remove unsupported cinder services right before the upgrade

jsuchome:
See Known Issues at http://docs.openstack.org/releasenotes/cinder/newton.html#id4

Cinder services are now automatically downgrading RPC messages to be understood
by the oldest version of a service among all the deployment. Disabled and dead
services are also taken into account. It is important to keep service list up to date,
without old, unused records. This can be done using cinder-manage service remove command.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
@@ -41,3 +41,20 @@ do
 done
 
 <% end %>
+
+<% if @cinder_controller %>
+
+# all cinder services must be stopped
+while cinder-manage service list 2>/dev/null | grep -q ":-)"; do
+  echo "Some cinder services are still running"
+  sleep 5
+done
+
+for service in cinder-volume cinder-scheduler; do
+    hosts=$(cinder-manage service list | grep $service | tr -s ' ' | cut -d ' ' -f 2)
+    for host in $hosts; do
+        cinder-manage service remove $service $host 2>/dev/null
+    done
+done
+
+<% end %>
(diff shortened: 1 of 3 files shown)


chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
```diff
@@ -11,6 +11,29 @@ echo "Executing $BASH_SOURCE"
...
+# we need to stop scheduler so we can remove it from list of cinder services
+<% if @use_ha %>
+crm --wait resource stop cinder-scheduler
+<% else %>
+/etc/init.d/openstack-cinder-scheduler stop
+<% end %>
+
+# FIXME we need to do this only after all cinder-volume services are stopped!
+for service in cinder-volume cinder-scheduler; do
+    hosts=$(cinder service-list --binary $service| grep $service | sed 's/ //g' | cut -d "|" -f 3)
```
rhafer: Is there a special reason why you use ```cinder``` to list here and ```cinder-manage``` to remove? AFAICT you can do both steps with ```cinder-manage```. Which also has the advantage of not needing the cinder-api to run (it works by accessing the db models directly AFAIK). That would mean you could move the code to a later point, where the cluster (apart from the database) is already shutdown.

jsuchome: Good point. (In my tests, `cinder-manage` somehow did not work after some shutdowns, but that was probably some problem between chair and keyboard)

jsuchome:
> In the commit message you mention the services being removed after the upgrade

I think before upgrade is better, because after upgrade I would need to do it after the (partially failed) chef-client run. I'll adapt the commit message.

jsuchome:
I've added special handling of cinder-volume service. But if it is not relevant and we should expect no openstack-cinder-volume service will be running separately (in non-HA fassion), i could drop that commit.

jsuchome:
Thanks for the reviews. I still did not test it fully (executed from crowbar API), so I'm waiting with merging until I really do it.

jsuchome:
[cinder issues during the nodes upgrade](https://trello.com/c/IRStF4aq/196-cinder-issues-during-the-nodes-upgrade)

jsuchome:
Yep, it works.

PR #874, opened 2016-12-08
[3.0] upgrade: fix adminrepochecks on missing repositories

MaximilianMeister:
when just one product has missing repositories and another one
shows empty (successful) repochecks, we need to skip trying to
parse the missing repositories when there aren't any.

fixes: NoMethodError (undefined method `to_sym' for nil:NilClass)


MaximilianMeister:
https://github.com/crowbar/crowbar-core/pull/873 will fix travis

jsuchome:
Can we have some unit test for this scenario?

MaximilianMeister:
> Can we have some unit test for this scenario?

@jsuchome added

PR #875, opened 2016-12-08
upgrade: upgrading compute nodes, live-migrate VM's in the process

jsuchome:
This currently solves only one case, there are missing parts needed for non-ovs network setup. I believe that could be solved later

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
@@ -4,6 +4,8 @@
 
 LOGFILE=/var/log/crowbar/node-upgrade.log
 UPGRADEDIR=/var/lib/crowbar/upgrade
+RUNDIR=/var/run/crowbar
+RUNFILE=$RUNDIR/node-pre-upgrading
 mkdir -p "`dirname "$LOGFILE"`"
 exec >>"$LOGFILE" 2>&1
 
@@ -11,6 +13,7 @@ echo "Executing $BASH_SOURCE"
 
 set -x
 
+mkdir -p $RUNDIR
 mkdir -p $UPGRADEDIR
 rm -f $UPGRADEDIR/crowbar-pre-upgrade-failed
 
@@ -18,6 +21,26 @@ if [[ -f $UPGRADEDIR/crowbar-pre-upgrade-ok ]] ; then
     echo "Pre upgrade actions were already successfully executed"
     exit 0
 fi
+
+cleanup()
+{
+    echo "cleaning up after interrupt or exit"
+
+    # cleanup upgrading indication so that the action can be restarted
+    rm -f $RUNFILE
+
+    exit 2
+}
+
+if [[ -f $RUNFILE ]] ; then
+    echo "Exit: Pre-Upgrade is already running..."
+    exit 1
+fi
+
+# Signalize that the pre-upgrade script is running
+touch $RUNFILE
+
+trap cleanup INT EXIT
 
 <% if @use_ha %>
 
--- a/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
@@ -55,4 +55,52 @@ ovs-vsctl set-fail-mode <%= bridge %> standalone
 echo "Reloading wicked configuration for all interfaces"
 wicked ifreload all
 
+<% if @compute_node %>
+
+echo "Upgrading services needed to act as nova-compute node"
+
+zypper --non-interactive up openstack-nova-compute
+# This is ugly, but fixing neutron spec to upgrade to correct version of pecan is even uglier:
+zypper --non-interactive up python-pecan
+zypper --non-interactive install crudini
+
+echo "Upgrading configuration for Newton based packages"
+
+crudini --set /etc/nova/nova.conf.d/200-crowbar-upgrade.conf DEFAULT use_neutron true
+
+zypper --non-interactive up <%= @neutron_agent %>
+<% case @neutron_agent %>
+<% when "openstack-neutron-openvswitch-agent" %>
+crudini --set /etc/neutron/neutron-openvswitch-agent.conf.d/200-crowbar-upgrade.conf ovs ovsdb_interface vsctl
+crudini --set /etc/neutron/neutron-openvswitch-agent.conf.d/200-crowbar-upgrade.conf ovs of_interface ovs-ofctl
+<% when "openstack-neutron-linuxbridge-agent" %>
+<% # FIXME: configuration changes for linuxbridge %>
+<% end %>
+systemctl start <%= @neutron_agent %>
+
+<% if @l3_agent %>
+zypper --non-interactive up <%= @l3_agent %>
+<% # FIXME: configuration changes for l3 agent %>
+systemctl start <%= @l3_agent %>
+<% end %>
+
+<% if @metadata_agent %>
+zypper --non-interactive up <%= @metadata_agent %>
+<% # FIXME: configuration changes for metadata agent %>
+systemctl start <%= @metadata_agent %>
+<% end %>
+
+systemctl start openstack-nova-compute.service
+
+<% if @cinder_volume %>
+zypper --non-interactive up openstack-cinder-volume
+systemctl start openstack-cinder-volume.service
+<% end %>
+
+# Remove temporary config options
+rm -f /etc/nova/nova.conf.d/200-crowbar-upgrade.conf
+rm -f /etc/neutron/neutron-openvswitch-agent.conf.d/200-crowbar-upgrade.conf
+
+<% end %>
+
 touch $UPGRADEDIR/crowbar-pre-upgrade-ok
--- a/chef/cookbooks/crowbar/templates/default/crowbar-prepare-repositories.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-prepare-repositories.sh.erb
@@ -1,0 +1,34 @@
+#!/bin/bash
+#
+# This script prepares the repositories needed for following node upgrade.
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+mkdir -p "`dirname "$LOGFILE"`"
+exec >>"$LOGFILE" 2>&1
+
+echo "Executing $BASH_SOURCE"
+
+set -x
+
+UPGRADEDIR=/var/lib/crowbar/upgrade
+mkdir -p $UPGRADEDIR
+
+if [[ -f $UPGRADEDIR/crowbar-prepare-repositories-ok ]] ; then
+    echo "Repositories already correctly prepared"
+    exit 0
+fi
+
+
+echo "Removing old repositories..."
+<% @old_repos.each do |name, _| %>
+zypper --non-interactive removerepo <%= name %>
+<% end %>
+zypper --non-interactive removerepo <%= @old_base_repo %>
+
+echo "Adding new repositories..."
+<% @new_repos.each do |name, attrs| %>
+zypper --non-interactive addrepo <%= attrs[:url] %> <%= name %>
+<% end %>
+zypper --non-interactive addrepo <%= @new_base_repo %> <%= @new_alias %>
+
+touch $UPGRADEDIR/crowbar-prepare-repositories-ok
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -304,7 +304,7 @@ module Api
         end
 
         if substep == "computes"
-          return false unless upgrade_compute_nodes
+          return false unless upgrade_all_compute_nodes
         end
         # FIXME: mark the whole step as done
         true
@@ -470,8 +470,62 @@ module Api
         Rails.logger.error(message)
       end
 
-      def upgrade_compute_nodes
-        # TODO: not implemented
+      def upgrade_all_compute_nodes
+        ["kvm", "xen"].each do |virt|
+          return false unless upgrade_compute_nodes virt
+        end
+        true
+      end
+
+      def upgrade_compute_nodes(virt)
+        save_upgrade_state("Upgrading compute nodes of #{virt} type")
+        compute_nodes = NodeObject.find("roles:nova-compute-#{virt}")
+        return true if compute_nodes.empty?
+
+        controller = NodeObject.find("roles:nova-controller").first
+        if controller.nil?
+          save_error_state("No nova controller node was found!")
+          return false
+        end
+
+        # FIXME: action in this block could be executed in parallel for all compute nodes!
+        # should we use chef-client for that?
+        compute_nodes.each do |node|
+          node_api = Api::Node.new node.name
+          # prepare the repositories (we could actually do this in paralel for all nodes!)
+          return false unless node_api.prepare_repositories
+          # upgrade specific services (nova-compute, neutron agent, cinder)
+          # adapt the config files so the upgraded services can run
+          # start the services
+          return false unless node_api.pre_upgrade
+        end
+
+        compute_nodes.each do |n|
+          node_api = Api::Node.new n.name
+
+          # FIXME: this is the main part of the whole upgrade
+          # from controller node (ha!) execute "nova host-evacuate-live"
+          # ... and wait until live-migration finishes
+
+          # upgrade and reboot the node
+          return false unless node_api.os_upgrade
+          return false unless node_api.reboot_and_wait
+          return false unless node_api.post_upgrade
+          return false unless node_api.join_and_chef
+
+          out = controller.run_ssh_cmd(
+            "source /root/.openrc; nova service-enable #{n.name} nova-compute"
+          )
+          unless out[:exit_code].zero?
+            save_error_state(
+              "Enabling nova-compute service for #{n.name} has failed!" \
+              "Check nova log files at #{controller.name} and #{n.name}."
+            )
+            return false
+          end
+          save_upgrade_state("Node #{n.name} successfully upgraded.")
+        end
+        # FIXME: finalize compute nodes (move upgrade_step to done etc.)
         true
       end
 
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -494,6 +494,44 @@ module Api
         true
       end
 
+      # Take a list of nodes and execute given script at each node in the background
+      # Wait until all scripts at all nodes correctly finish or until some error is detected
+      def execute_scripts_and_wait_for_finish(nodes, script, seconds)
+        nodes.each do |node|
+          ssh_status = node.ssh_cmd(script).first
+          if ssh_status != 200
+            save_error_state("Failed to connect to node #{node.name}!")
+            raise "Execution of script #{script} has failed on node #{node.name}."
+          end
+        end
+
+        scripts_status = {}
+        begin
+          Timeout.timeout(seconds) do
+            nodes.each do |node|
+              # wait until sript on this node finishes, than move to check next one
+              loop do
+                status = node.script_status(script)
+                scripts_status[node.name] = status
+                break if status != "running"
+                sleep 1
+              end
+            end
+          end
+          failed = scripts_status.select { |_, v| v == "failed" }.keys
+          unless failed.empty?
+            raise "Execution of script #{script} has failed at node(s) " \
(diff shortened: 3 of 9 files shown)


jsuchome:

MaximilianMeister:
seems you need to add one more mock to `context "upgrading the nodes"`

```ruby
allow(NodeObject).to(
  receive(:find).
  with("roles:nova-compute-kvm").
  and_return([NodeObject.find_node_by_name("testing.crowbar.com")])
)
```

jsuchome:
@MaximilianMeister thanks, adapted

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -464,8 +483,60 @@ def save_error_state(message = "")
...
+          # upgrade and reboot the node
+          return false unless node_api.os_upgrade
+          return false unless node_api.reboot_and_wait
+          return false unless node_api.post_upgrade
+          return false unless node_api.join_and_chef
+
+          out = controller.run_ssh_cmd(
+            "source /root/.openrc; nova service-enable #{n.name} nova-compute"
+          )
+          unless out[:exit_code].zero?
```
jsuchome: Hm, not really, I want to exit from the method....

abel-navarro:
after running the upgrade on all computes, the last compute node won't have any VMs running there because all will have been migrated previously.

jsuchome:
> after running the upgrade on all computes, the last compute node won't have any VMs running there 
> because all will have been migrated previously.

But that's not a bug, right? Once nova-api is available, users will create/move/destroy instances and soon, the last node's gonna be used again.

abel-navarro:
yep, definitely not a bug although a nice improvement.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -464,8 +481,57 @@ def save_error_state(message = "")
...
+        return true if compute_nodes.empty?
+
+        controller = NodeObject.find("roles:nova-controller").first
+        if controller.nil?
+          save_error_state("No nova controller node was found!")
+          return false
+        end
+
+        # FIXME: action in this block could be executed in parallel for all compute nodes!
+        # should we use chef-client for that?
```
jsuchome: @rhafer do you have any idea for this?

rhafer: Not a quick one unfortunately. But it definitely makes sense to do this step in parallel and utilizing chef also doesn't sound wrong. OTOH you could also do something similar to what ```ServiceObject::apply_role``` does (starting here: https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/service_object.rb#L1275) and maybe utilize ```run_remote_chef_client``` (Note: appart from its name that function seems to allow calling other command that "chef-client" as well.)

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -54,4 +54,26 @@ ovs-vsctl set-fail-mode <%= bridge %> standalone
 echo "Reloading wicked configuration for all interfaces"
 wicked ifreload all
 
+<% if @compute_node %>
+
+echo "Upgrading services needed to act as nova-compute node"
+
+zypper --non-interactive up openstack-nova-compute
+zypper --non-interactive up openstack-neutron-openvswitch-agent
+zypper --non-interactive up python-pecan
```
rhafer: Wouldn't this better be fixed in the RPM dependencies of neutron?

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -54,4 +54,26 @@ ovs-vsctl set-fail-mode <%= bridge %> standalone
...
+zypper --non-interactive install crudini
+zypper --non-interactive up openstack-cinder-volume
+
+echo "Upgrading configuration for Newton based packages"
+
+crudini --set /etc/neutron/plugins/ml2/openvswitch_agent.ini ovs ovsdb_interface vsctl
+crudini --set /etc/neutron/plugins/ml2/openvswitch_agent.ini ovs of_interface ovs-ofctl
```
rhafer: This is only need when the openvswitch driver is enable in the neutron proposal

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -54,4 +54,26 @@ ovs-vsctl set-fail-mode <%= bridge %> standalone
...
+crudini --set /etc/nova/nova.conf DEFAULT use_neutron true
+
+systemctl start openstack-nova-compute.service
+systemctl start openstack-neutron-openvswitch-agent.service
```
rhafer: Hm, which agent(s) to start here really depends on the neutron proposal. E.g. it might also be linuxbridge. And in certain setups (DVR) even l3-agent (and maybe others are needed).

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -54,4 +54,26 @@ ovs-vsctl set-fail-mode <%= bridge %> standalone
...
+systemctl start openstack-cinder-volume.service
```
rhafer: This is only requires if the node has the cinder-volume role.

crowbar_framework/app/models/api/upgrade.rb
jsuchome: I could also use `node.ssh_cmd`, which executes the script in background. But I'm also not sure how to track progress and/or possible errors... and that is also ugly for chef execution

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
jsuchome: Last time I spoke with @jgrassler , it seemed like a bad idea for some reason. But indeed it looks strange...

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
jsuchome: Yes, the code needs to be adapted to more/different agents...

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
jsuchome: True

crowbar_framework/app/models/api/upgrade.rb
rhafer: ```run_remote_chef_client``` allows to keep track. Just see how it is implemented in ```apply_role```

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
jsuchome: Sigh. Might be easiest to search for installed packages than into neutron role attributes...

jsuchome:
@rhafer so, I've made the openvswitch changes depend on presence of package, somehow it feels cleaner in the bash script than the need to search for neutron-server role from each node's chef-client run... But it's wrong anyway, isn't it? The package might be installed and still different service configured.

rhafer:
> so, I've made the openvswitch changes depend on presence of package, somehow it feels cleaner in the bash script than the need to search for neutron-server role from each node's chef-client run... But it's wrong anyway, isn't it? The package might be installed and still different service configured.

Yeah, that doesn't sound very robust do me. I guess I would have gone for the search + template variables (it's bascially what the nova/neutron cookbooks to as well).

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -55,4 +55,48 @@ ovs-vsctl set-fail-mode <%= bridge %> standalone
 echo "Reloading wicked configuration for all interfaces"
 wicked ifreload all
 
+<% if @compute_node %>
+
+echo "Upgrading services needed to act as nova-compute node"
+
+zypper --non-interactive up openstack-nova-compute
+# This is ugly, but fixing neutron spec to upgrade to correct version of pecan is even uglier:
+zypper --non-interactive up python-pecan
```
dirkmueller: isn't that fixable by fixing dependencies?

jsuchome: Actually, with recent changes done by @toabctl it should be fixed. I've yet to test it...

jsuchome: So latest update is, "no we have to install it explicitely, as the dependencies can't be easily fixed"

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -55,4 +55,48 @@ ovs-vsctl set-fail-mode <%= bridge %> standalone
...
+zypper --non-interactive install crudini
+
+echo "Upgrading configuration for Newton based packages"
+
+crudini --set /etc/nova/nova.conf DEFAULT use_neutron true
```
jsuchome: So this is apparently already obsolete by latest changes in Newton nova config packages....

toabctl: After the config structure change, better would be ```/etc/nova/nova.conf.d/200-crowbar-pre-upgrade.conf``` and at the end you delete the file again.

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -55,4 +55,48 @@ ovs-vsctl set-fail-mode <%= bridge %> standalone
...
+
+zypper --non-interactive up <%= @neutron_agent %>
+<% case @neutron_agent %>
+<% when "openstack-neutron-openvswitch-agent" %>
+  crudini --set /etc/neutron/plugins/ml2/openvswitch_agent.ini ovs ovsdb_interface vsctl
```
toabctl: Imo ```/etc/neutron/neutron-openvswitch-agent.conf.d/200-crowbar-pre-upgrade.conf``` would be better

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -57,12 +57,15 @@ wicked ifreload all
 
 <% if @compute_node %>
 
+# FIXME: remove this once we can install Cloud6 from non-devel repositories
+zypper --non-interactive install crudini
+crudini --set /etc/zypp/zypp.conf main solver.allowVendorChange true
```
rhafer: I'd really prefer to not merge this. Would it be possible to add this to mkcloud/qa-crowbarsetup?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -481,6 +481,44 @@ def upgrade_all_compute_nodes
         true
       end
 
+      def execute_scripts_and_wait_for_finish(nodes, script, seconds)
+        nodes.each do |node|
+          ssh_status = node.ssh_cmd(script).first
+          if ssh_status != 200
+            save_error_state("Failed to connect to node #{node.name}!")
+            raise "Execution of script #{script} has failed on node #{node.name}."
```
jsuchome: I do not think it really matters. When the whole function is executed again, and scripts are run again on the nodes, they are not run again when there's already -ok file.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1072,6 +1072,15 @@ def file_exist?(file)
     out[:exit_code].zero?
   end
 
+  def script_finished(script)
+    base = "/var/lib/crowbar/upgrade/" + File.basename(script, ".sh")
+    ok_file = base + "-ok"
+    failed_file = base + "-failed"
+    return "ok" if file_exist? ok_file
+    return "failed" if file_exist? failed_file
```
jsuchome: Well... I don't know. I do not see it as a problem.
And I see it as more readable than the alternative threads option.

> And we call script_finished every two seconds on every compute node

Only for those where we do not have the result yet.

jsuchome:
@rhafer as for the main FIXME's, I've created https://trello.com/c/8XevJ4WD/233-solve-non-default-configuration-for-compute-nodes-upgade - that should probably transfer to regular (non-noise) card.

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
jsuchome: Hm... maybe. 
(Although it's just different code that should be removed later)

crowbar_framework/app/models/api/upgrade.rb
jsuchome: This actually is serious error, the node cannot be accessed. I think user should be alerted immediately.

chef/cookbooks/crowbar/templates/default/crowbar-prepare-repositories.sh.erb
```diff
@@ -0,0 +1,35 @@
...
+<% @old_repos.each do |name, _| %>
+zypper --non-interactive removerepo <%= name %>
+<% end %>
+zypper --non-interactive removerepo <%= @old_base_repo %>
+
+echo "Adding new repositories..."
+<% @new_repos.each do |name, attrs| %>
+zypper --non-interactive addrepo <%= attrs[:url] %> <%= name %>
+<% end %>
+zypper --non-interactive addrepo <%= @new_base_repo %> <%= @new_alias %>
```
MaximilianMeister: do we need to refresh the repos after adding them?

chef/cookbooks/crowbar/templates/default/crowbar-prepare-repositories.sh.erb
```diff
@@ -0,0 +1,35 @@
...
+
+touch $UPGRADEDIR/crowbar-prepare-repositories-ok
+
```
MaximilianMeister: empty line

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -457,8 +474,105 @@ def save_error_state(message = "")
...
+            "/usr/sbin/crowbar-pre-upgrade.sh",
+            300
+          )
+          save_upgrade_state("Services at compute nodes upgraded and prepared.")
+        rescue StandardError => e
+          save_error_state(e.message)
+          return false
+        end
+
+        # Next part must be done sequentionaly, only one compute node can be upgraded at a time
```
MaximilianMeister: `s/sequentionaly/sequentially`

crowbar_framework/app/models/node_object.rb
```diff
@@ -1072,11 +1072,22 @@ def file_exist?(file)
     out[:exit_code].zero?
   end
 
+  def script_finished(script)
+    base = "/var/lib/crowbar/upgrade/" + File.basename(script, ".sh")
+    ok_file = base + "-ok"
+    failed_file = base + "-failed"
+    return "ok" if file_exist? ok_file
```
MaximilianMeister: can we use symbols here for the `return` value as in the upgrade status? that would minimally enhance the memory footprint

chef/cookbooks/crowbar/templates/default/crowbar-prepare-repositories.sh.erb
jsuchome: Hm, none of mine tests needed it, so I think it was probably done automatically when the package install was requested.

crowbar_framework/app/models/node_object.rb
jsuchome: Ok

crowbar_framework/app/models/node_object.rb
rhafer: > Well... I don't know. I do not see it as a problem.

Given that this means 100 ssh-connections/second for a 100 compute node setup I am not sure I agree with it :).
I would also be pretty easy get down to half the number, by checking for both files in one command. It also might be an option to just poll a single node and check the next one only after the first finished. (IIRC mkcloud does it this way when waiting for nodes to be ready).

> And I see it as more readable than the alternative threads option.

Note, I didn't say that the Threads approach is the better way here. As it would involves a lot of longer-lived ssh connections it might actually be worse (but we don't actually know).

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
jsuchome: Hopefully addressed by https://github.com/SUSE-Cloud/automation/pull/1544

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Hm, good point. We might adapt the scritps to handle it (like we have it for upgrade os one: https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.erb#L48

Or we migt rather adapt the calling code here, like you're suggesting.

chef/cookbooks/crowbar/templates/default/crowbar-prepare-repositories.sh.erb
MaximilianMeister: probably yes. but lets say it fails to refresh (for some dependency issue) should it fail here or before the installation?

jsuchome: Explicit refresh actually happens in next script

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -55,4 +55,52 @@ ovs-vsctl set-fail-mode <%= bridge %> standalone
...
+
+crudini --set /etc/nova/nova.conf.d/200-crowbar-upgrade.conf DEFAULT use_neutron true
+
+zypper --non-interactive up <%= @neutron_agent %>
+<% case @neutron_agent %>
+<% when "openstack-neutron-openvswitch-agent" %>
+crudini --set /etc/neutron/neutron-openvswitch-agent.conf.d/200-crowbar-upgrade.conf ovs ovsdb_interface vsctl
+crudini --set /etc/neutron/neutron-openvswitch-agent.conf.d/200-crowbar-upgrade.conf ovs of_interface ovs-ofctl
+<% when "openstack-neutron-linuxbridge-agent" %>
+# FIXME: configuration changes?
```
rhafer: I believe in the processed file on the destination node this looks really weird. I guess it will be something like:

```
crudini --set /etc/neutron/neutron-openvswitch-agent.conf.d/200-crowbar-upgrade.conf ovs ovsdb_interface vsctl
crudini --set /etc/neutron/neutron-openvswitch-agent.conf.d/200-crowbar-upgrade.conf ovs of_interface ovs-ofctl
# FIXME: configuration changes?
systemctl start neutron-openvswitch-agent
```

As the comment is actually meant for the template code (mainly for the ```when``` statement), I guess it would make sense to not have it appear in the template's output. I'd suggest to used ```<%# comment %> here. Same of the other FIXMEs as well.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -481,6 +481,44 @@ def upgrade_all_compute_nodes
...
+          end
+        end
+
+        scripts_status = {}
+        begin
+          Timeout.timeout(seconds) do
+            nodes.each do |node|
+              # wait until sript on this node finishes, than move to check next one
+              loop do
+                finished = node.script_finished(script)
```
rhafer: Nitpick: I'd suggest to rename this (both, the variable and the method) to ```script_status```. By just looking at the name, I expected ```script_finished``` to just return a boolean (which it doesn't)

jsuchome:
@rhafer rebased and addressed your comments

jsuchome:
> can't we already call end_step with the error on all cases where we call the save_error_state 
> placeholder?

I would like to do once this is merged, so we have both compute and controllers code merged. I already created a checkbox in https://trello.com/c/StuV7nHt/203-8-use-sub-steps-during-the-last-part-of-the-upgrade-part-i-controller-nodes

jsuchome:
So, @rhafer what do you think? Could we merge this? 
(Than you could rebase your https://github.com/crowbar/crowbar-core/pull/935)

PR #877, opened 2016-12-08
upgrade: some cleanup of controller node upgrade code

jsuchome:


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -281,6 +281,41 @@ def prepare(options = {})
 
       protected
 
+      def upgrade_first_cluster_node(node_name, next_node_name)
+        save_upgrade_state("Starting the upgrade of node #{node_name}")
+        node_api = Api::Node.new node_name
+        next_node_api = Api::Node.new next_node_name
+        # upgrade the first node
+        return false unless node_api.upgrade
```
Itxaka: Dont we want to `save_error_state` in here as well?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -281,6 +281,41 @@ def prepare(options = {})
...
+
+        # Explicitly mark the first node as cluster founder
+        # and in case of DRBD setup, adapt DRBD config accordingly.
+        unless Api::Pacemaker.set_node_as_founder node_name
+          save_error_state("Changing the cluster founder to #{node_name} has failed")
+          return false
+        end
+        # remove pre-upgrade attribute, so the services can start
+        return false unless next_node_api.disable_pre_upgrade_attribute_for node_name
```
Itxaka: Same here and below, shouldnt we be saving the error state in each of this actions?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -295,44 +330,13 @@ def upgrade_controller_nodes
...
-
-        # 4. start crowbar-join at the first node
-        return false unless founder_api.post_upgrade
-        return false unless founder_api.join_and_chef
-
-        # 5. migrate routers from nodes being upgraded
-        return false unless founder_api.router_migration
+        return false unless upgrade_first_cluster_node founder.name, non_founder.name
 
         # 6. upgrade the rest of nodes in the same cluster
```
Itxaka: Comment number does not make sense anymore

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -281,6 +281,41 @@ def prepare(options = {})
...
+        # start crowbar-join at the first node
+        return false unless node_api.post_upgrade
+        return false unless node_api.join_and_chef
+        # migrate routers from nodes being upgraded
+        node_api.router_migration
+      end
+
+      def upgrade_next_cluster_node(node_name)
+        save_upgrade_state("Starting the upgrade of node #{node_name}")
+        node_api = Api::Node.new name
```
Itxaka: `name` -> `node_name` ?

crowbar_framework/app/models/api/upgrade.rb
jsuchome: It's actually called by node_api.upgrade

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Again `disable_pre_upgrade_attribute_for` does it. (But we'll need to revisit those save_error_state parts later, currently it only logs)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -281,6 +281,41 @@ def prepare(options = {})
...
+        return false unless node_api.upgrade
+        return false unless node_api.post_upgrade
+        return false unless node_api.join_and_chef
+        # Remove pre-upgrade attribute after chef-client run because pacemaker is already running
+        # and we want the configuration to be updated first
+        node_api.disable_pre_upgrade_attribute_for
```
jsuchome: Actually here the argument is not needed as it is doing on the node itself. I've adapted `disable_pre_upgrade_attribute_for` for that.

jsuchome: Or maybe... I could pass the argument and not change `disable_pre_upgrade_attribute_for` at all, that might be better

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -377,43 +382,9 @@ def upgrade_drbd_cluster(cluster)
...
-
-        # Execute post-upgrade actions after the node has been upgraded, rebooted
-        # and the existing cluster has been cleaned up by deleting most of resources:
-        # - start pacemaker and sync DRBD devices
-        return false unless node_api.post_upgrade
-        # - initiate the first chef-client run
-        return false unless node_api.join_and_chef
-
-        # migrate routers from drbd_master to recently upgraded drbd_slave
-        return false unless node_api.router_migration
```
abel-navarro: i guess we'll have to remove old router migration code now that we are using a new approach.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -281,6 +281,42 @@ def prepare(options = {})
 
       protected
 
+      def upgrade_first_cluster_node(node_name, next_node_name)
```
rhafer: I believe ```next_node_name``` is a bit of a misnomer here. Technically it doesn't really matter whether the node passed there is upgraded next after the first node (I guess the only thing you need here is a cluster member that still running pacemaker?). Additionally the implementation also doesn't really make sure it's upgraded next (AFAICS, ```upgrade_cluster_nodes``` simply reruns the search for non-upgrade nodes, and I believe chef doesn't make any assumptions in which order they will be returned).

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -281,6 +281,42 @@ def prepare(options = {})
...
+        # delete old pacemaker resources (from the node where old pacemaker is running)
+        return false unless delete_pacemaker_resources next_node_name
+        # start crowbar-join at the first node
+        return false unless node_api.post_upgrade
+        return false unless node_api.join_and_chef
+        # migrate routers from nodes being upgraded
+        node_api.router_migration
+      end
+
+      def upgrade_next_cluster_node(node_name)
```
rhafer: Same here about the naming. Would it make sense to rename the functions to ```upgrade_cluster_founder_node```/```upgrade_cluster_non_founder_node```, or is that also leading in the wrong direction?

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Do you think the whole `node_api.router_migration` method should be removed? My impression is that it is gonna call some different script, but not that it is not needed...

rhafer: I am pretty sure we'll still need the method. But I think we should worry about that in a different PR.

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Yes, it does not really mean "next", just "another". Should be renamed

crowbar_framework/app/models/api/upgrade.rb
jsuchome: It's a bit misdirecting in a different way. True, once the first node is upgraded, it has to change itself to be the founder (even if it originally wasn't), but that's exactly it, it might not have been founder at the time we call this method.

Actually here "next" sounds still correct to me - we are really calling this for every other node that is not the first one.

jsuchome:
> And I wonder if it would be feasible to reduce duplication any further by just having one 
> upgrade_cluster_node with a flag passed in to indicated whether the current node is the first to be 
> upgraded or a subsequent node. Or would that make the code unreadable?

I agree we could still remove some duplication and yes, I thought current situation with two methods is better for readability (and for explictely expressing the fact that we're treating the first node differently).

I'll try to think if it could be made better...

crowbar_framework/app/models/api/upgrade.rb
rhafer: Hm, right I misread the code a bit. I still find ```upgrade_next_cluster_node``` a bit misleading, but lack a better idea now. So I better shutup...

PR #886, opened 2016-12-13
[3.0] show sanity check errors in the json response

MaximilianMeister:
if you don't have an UI which lists you the sanity errors, it is very
difficult from the back end to determine what went wrong


MaximilianMeister:
> I guess the old i18n is now unused, if yes we should delete it here too.

it's still used in the `enforce_installer` filter (one method above)

PR #888, opened 2016-12-14
[3.0] only allow requests to the upgrade api during the upgrade

MaximilianMeister:
backport of https://github.com/crowbar/crowbar-core/pull/861

NOTE: the difference to the original commit is that i dropped the check for `/var/lib/crowbar/upgrade/6-to-7-upgraded-ok`


jsuchome:
Let's wait until we solve the issue with node's transitioning.

MaximilianMeister:
> Let's wait until we solve the issue with node's transitioning.

issue got addressed in https://github.com/crowbar/crowbar-core/pull/889

PR #889, opened 2016-12-14
upgrade: open transition api for crowbar_join

MaximilianMeister:
otherwise it can't set the node status


jsuchome:

PR #890, opened 2016-12-14
upgrade: Ensure cinder controller runs shutdown of services as last node

jsuchome:
As part of services shutdown, we're deleting cinder services from database.
For that, we need them to be stopped everywhere.

Also, we need until this step finishes before continuing with next steps.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
@@ -1,9 +1,10 @@
 #!/bin/bash
 #
-# This script starts the upgrade of the node server to the latest version of
-# system and Cloud product.
+# This script stops unnecessary OpenStack services on the nodes before we dive
+# deeper into the real upgrade.
 
 LOGFILE=/var/log/crowbar/node-upgrade.log
+UPGRADEDIR=/var/lib/crowbar/upgrade
 mkdir -p "`dirname "$LOGFILE"`"
 exec >>"$LOGFILE" 2>&1
 
@@ -28,7 +29,9 @@ done
 <% end %>
 
 # stop cinder-volume at all nodes so the service could be removed from the database later
-/etc/init.d/openstack-cinder-volume stop
+if test -e /etc/init.d/openstack-cinder-volume; then
+    /etc/init.d/openstack-cinder-volume stop
+fi
 
 <% else %>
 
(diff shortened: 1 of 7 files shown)


chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
```diff
@@ -64,4 +72,13 @@ for service in cinder-volume cinder-scheduler; do
     done
 done
 
+ret=$?
+if [ $ret != 0 ] ; then
+    echo "Error occured during deleting cinder services"
+    echo $ret > $UPGRADEDIR/crowbar-shutdown-services-before-upgrade-failed
```
Itxaka: Nothing is removing this file on the script. Shouldn't we try to remove it at the start of it? If not, that means that if it fails one, it will be reported as failing forever (If the api uses this file to report the status)

If it doesnt use it, then Im not sure the point of the file?

jsuchome: Good point

jsuchome: I've fixed it here, but we have to make sure it is deleted in other scripts as well: https://trello.com/c/QrxUXWGe/212-ensure-the-failed-files-are-deleted-when-upgrade-bash-scripts-are-starting

jsuchome:
[Wait until service step ends](https://trello.com/c/6pNOniEM/197-wait-until-service-step-ends)

jsuchome:
@rhafer regarding separate script, you are most likely right, it would be better. I thought about it, and then didn't do it :-(

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -219,12 +219,31 @@ def services
...
+          if node.roles.include?("cinder-controller") &&
+              (!node.roles.include?("pacemaker-cluster-member") || node["pacemaker"]["founder"])
+            cinder_nodes.push node
+          end
+          cmd = node.shutdown_services_before_upgrade
           next if cmd[0] == 200
           errors.push(cmd[1])
         end
 
+        cinder_nodes.each do |node|
```
rhafer: Doesn't this just need to happen on one of the cinder nodes?

Itxaka: Should always be only one node in the list no? due to the check on line 224? I dont have a current installed cloud to test this so Im just eyeballing it.

jsuchome: I would expect the list will have just one item, yes. Can't there be more clusters and thus more controllers in different clusters? (Probably not)

jsuchome:
@rhafer I think I addressed your concerns here

PR #892, opened 2016-12-14
queue: Only dequeue if there is a proposal to dequeue

rsalevsky:


rsalevsky:
[P1: bsc#1014835 - L3: SOC6: various errors in crowbar production log](https://trello.com/c/LBR7JSQJ/11-bsc-1014835-l3-soc6-various-errors-in-crowbar-production-log)

PR #893, opened 2016-12-14
upgrade: Remove the indication of previous script failure

jsuchome:
To enable running scripts again after the failure, we need
to remove failure indications when executing the script.


jsuchome:
[Ensure the -failed files are deleted when upgrade bash scripts are starting](https://trello.com/c/QrxUXWGe/212-ensure-the-failed-files-are-deleted-when-upgrade-bash-scripts-are-starting)

PR #895, opened 2016-12-14
Keep less ohai data

vuntz:
While I don't have any hard proof yet, I believe the bigger the node data is, the slower chef (server & solr) is going to be.

@rhafer cc'ing you as I'd prefer to have your opinion on whether it's okay or not to drop all these interfaces from the ohai data.

### DIFF
--- a/chef/cookbooks/ohai/recipes/default.rb
+++ b/chef/cookbooks/ohai/recipes/default.rb
@@ -108,4 +108,14 @@ end
 # need counters
 o.data.delete("counters")
 
+# drop relatively big attributes that we know we won't use
+o.data.delete("etc")
+o.data["kernel"].delete("modules")
+
+# duplicates the cpu data
+o.data["dmi"].delete("processor")
+# when looking at cpu data, we're happy looking at the first one only; removing
+# the others avoids having tons of useless information when having many cores
+(1..(o.data["cpu"]["total"] - 1)).each { |n| o.data["cpu"].delete(n.to_s) }
+
 node.automatic_attrs.merge! o.data
--- a/chef/cookbooks/ohai/recipes/default.rb
+++ b/chef/cookbooks/ohai/recipes/default.rb
@@ -95,4 +95,17 @@ end
 
 o = Ohai::System.new
 o.all_plugins
+
+# drop virtual interfaces, to not overload chef
+virtual_intfs = ["tap", "qbr", "qvo", "qvb", "brq"]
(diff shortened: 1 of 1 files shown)


aspiers:
Funny you should come up with that 2 days after I pointed [this article](http://engineering.voxer.com/2013/03/22/chef-part-2-performance/) out to @dirkmueller which says the same thing.  However the [referenced bug report](http://tickets.opscode.com/browse/CHEF-1694) only talks about `*:*` searches, so I'm not sure if it's the same thing.

vuntz:
> [...] although even 5000 lines of attributes is not so much [...]

Well, it's moving from 35k to 5k. But again, I don't have a great overview of the impact yet; we need to test on a production cloud (which is planned).

vuntz:
Oh, and for completeness, while I hadn't seen the article you linked to, the next step is indeed to get rid of searches. I was playing with something at https://github.com/crowbar/crowbar-core/compare/master...vuntz:databagconfig?expand=1 -- but I hadn't considered the option of doing just one big search at the beginning... (that wouldn't help for sync marks, but that's another story)

chef/cookbooks/ohai/recipes/default.rb
```diff
@@ -95,4 +95,19 @@
 
 o = Ohai::System.new
 o.all_plugins
+
+# drop virtual interfaces, to not overload chef
+virtual_intfs = ["tap", "qbr", "qvo", "qvb", "brq", "ovs"]
+o.data["network"]["interfaces"].each_key do |intf|
+  if virtual_intfs.include?(intf.slice(0..2))
+    o.data["network"]["interfaces"].delete(intf)
+    o.data["counters"]["network"]["interfaces"].delete(intf)
```
vuntz: Actually, it was some mini-optimization (slightly confusing), but doesn't matter because of... (next comment)

chef/cookbooks/ohai/recipes/default.rb
```diff
@@ -95,4 +95,19 @@
...
+  end
+end
+o.data["counters"]["network"]["interfaces"].each_key do |intf|
```
vuntz: Agreed, done

vuntz:
> Do you think it makes sense to disable some plugins like lua, php, mono from the beginning via the client.rb?

I considered it, but it doesn't make a big difference, tbh:

- in terms of speed, ohai is already fast enough
- in terms of output, that's only a few attributes, nothing big

PR #896, opened 2016-12-14
Use threads to acquire/release locks in apply_role

vuntz:
This doesn't have a huge impact (as opposed to fixing cookbooks), but still some nice improvements (something like 2 minutes -> 1 minute for the non-chef code).

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1459,29 +1459,53 @@ class ServiceObject
 
   private
 
+  THREAD_POOL_SIZE = 20
+
   def release_chef_locks(locks)
     return if locks.empty?
 
-    locks.each_slice(20) do |lock_batch|
-      threads = []
-      lock_batch.each do |lock|
-        threads.push(Thread.new { lock.release })
+    queue = Queue.new
+    locks.each { |l| queue.push l }
+
+    workers = (1...THREAD_POOL_SIZE).map do
+      Thread.new do
+        loop do
+          begin
+            lock = queue.pop(true)
+          rescue ThreadError
+            break
+          end
+
+          lock.release
+        end
       end
-      logger.debug "Waiting for #{threads.length} unlock threads to finish..."
-      ThreadsWait.all_waits(threads)
     end
+
+    logger.debug "Waiting for #{THREAD_POOL_SIZE} unlock threads to finish..."
+    workers.map(&:join)
   end
 
   def lock_and_wait_for_chef_clients(nodes, lock_owner, lock_reason)
     locks = []
     errors = {}
+
+    return [locks, errors] if nodes.empty?
+
     locks_mutex = Mutex.new
     errors_mutex = Mutex.new
 
-    nodes.each_slice(20) do |node_batch|
-      threads = {}
-      node_batch.each do |node|
-        thread = Thread.new do
+    queue = Queue.new
+    nodes.each { |n| queue.push n }
+
+    workers = (1...THREAD_POOL_SIZE).map do
+      Thread.new do
+        loop do
+          begin
+            node = queue.pop(true)
+          rescue ThreadError
+            break
+          end
+
           begin
             lock = Crowbar::Lock::SharedNonBlocking.new(
               logger: @logger,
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/service_object.rb
```diff
@@ -1478,6 +1463,58 @@ def wait_for_chef_daemons(node_list)
 
   private
 
+  def release_chef_locks(locks)
+    return if locks.empty?
+
+    locks.each_slice(20) do |lock_batch|
```
aspiers: Please can this `20` be changed to a constant?  I guess `THREAD_POOL_SIZE` or similar?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1478,6 +1463,58 @@ def wait_for_chef_daemons(node_list)
...
+    end
+  end
+
+  def lock_and_wait_for_chef_clients(nodes, lock_owner, lock_reason)
+    locks = []
+    errors = {}
+    locks_mutex = Mutex.new
+    errors_mutex = Mutex.new
+
+    nodes.each_slice(20) do |node_batch|
```
aspiers: Use the constant here too.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1478,6 +1463,58 @@ def wait_for_chef_daemons(node_list)
...
+          rescue Crowbar::Error::LockingFailure => e
+            errors_mutex.synchronize { errors[node] = e.message }
+          end
+        end
+        threads[thread] = node
+      end
+
+      # wait for all running threads and collect the ones with a non-zero return value
+      logger.debug "Waiting for #{threads.keys.length} lock threads to finish..."
+      ThreadsWait.all_waits(threads.keys)
```
aspiers: Doesn't this mean that if just one node in the first batch was super slow for some reason then it would block all subsequent batches?  Or am I reading this wrong?  I would have thought that [something like this](https://blog.engineyard.com/2014/ruby-thread-pool) might be more efficient.

crowbar_framework/app/models/service_object.rb
```diff
@@ -925,6 +925,9 @@ def apply_role(role, inst, in_queue, bootstrap = false)
     apply_locks = []
     applying_nodes = []
 
+    # Cache some node attributes to avoid useless node reloads
+    node_attr_cache = {}
```
aspiers: Wouldn't it make more sense for this to live in `NodeObject`?  If we still need to ensure that `find_node_by_name` etc. retain their non-caching behaviour by default then we could introduce new `find_node_by_name_cached` type of methods which wrap around it with the caching behaviour enabled.  +1 for caching in general though ;-)

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -582,7 +582,7 @@ def bootstrap_ensure_proposal(bc, override_attr_path)
 
     unless active_proposals.include?(id)
       @logger.debug("Bootstrap: applying proposal for #{bc}.#{id}")
-      answer = service.proposal_commit(id, false, false, true)
+      answer = service.proposal_commit(id, in_queue: false, validate_after_save: false, bootstrap: true)
```
aspiers: I agree with the dog ;-)  But +100 for the legibility improvements :-)

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -922,7 +922,8 @@ def proposal_update
           @proposal["attributes"][params[:barclamp]] = JSON.parse(params[:proposal_attributes])
           @proposal["deployment"][params[:barclamp]] = JSON.parse(params[:proposal_deployment])
           @service_object.save_proposal!(@proposal)
-          answer = @service_object.proposal_commit(params[:name])
+          # validation already happened on save
+          answer = @service_object.proposal_commit(params[:name], validate: false, validate_after_save: false)
```
aspiers: Again the dog is right.  But sounds like a great idea to de-duplicate validation!  How did you think of that?

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -923,7 +923,9 @@ def proposal_update
           @proposal["deployment"][params[:barclamp]] = JSON.parse(params[:proposal_deployment])
           @service_object.save_proposal!(@proposal)
           # validation already happened on save
-          answer = @service_object.proposal_commit(params[:name], validate: false, validate_after_save: false)
+          answer = @service_object.proposal_commit(params[:name],
```
aspiers: Bah OK, you fixed the hound complaints as a separate commit after introducing some new violations :-)  You can see how that causes confusion when reviewing commit by commit.  This is [one of my pet peeves](http://blog.adamspiers.org/2015/03/24/why-and-how-to-correctly-amend-github-pull-requests/#sec-2) which I was explaining to @matelakat earlier :-)

crowbar_framework/app/models/service_object.rb
vuntz: @aspiers ideally yes, but it's more intrusive and I wanted to optimize the apply_role case in a backportable way.

crowbar_framework/app/models/service_object.rb
vuntz: Ah yes, indeed. Will rework that.

vuntz:
@aspiers thread pool implemented

dirkmueller:
do not merge before M7 is out

vuntz:
FWIW, the move to threads create some issues with autoloading. See http://stackoverflow.com/questions/36382752/handling-circular-dependency-in-rails-while-using-threads/36383451#36383451 -- will fix that when I come back.

vuntz:
I moved the non-threading improvements to https://github.com/crowbar/crowbar-core/pull/959

aspiers:
@vuntz wrote:
> I moved the non-threading improvements to #959

By "moved" you mean you also removed those commits from this PR?  Because now I don't see any overlap between the two PRs at all, which confused the hell out of me at first.

vuntz:
> > I moved the non-threading improvements to #959
> By "moved" you mean you also removed those commits from this PR?

Yes.

vuntz:
Completed in https://github.com/crowbar/crowbar-core/pull/1266

PR #897, opened 2016-12-15
upgrade: Let the progress library track substeps and current node

jsuchome:
We need this for detailed information during the last part of upgrade,
upgrading cloud nodes.

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade_status.rb
+++ b/crowbar_framework/lib/crowbar/upgrade_status.rb
@@ -52,7 +52,10 @@ module Crowbar
         # substep is needed for more complex steps like upgrading the nodes
         current_substep: nil,
         # current node is relevant only for the nodes_upgrade step
-        current_node: nil
+        current_node: nil,
+        # number of nodes still to be upgraded
+        remaining_nodes: nil,
+        upgraded_nodes: nil
       }
       # in 'steps', we save the information about each step that was executed
       @progress[:steps] = upgrade_steps_6_7.map do |step|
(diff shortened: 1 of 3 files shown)


jsuchome:

crowbar_framework/spec/lib/crowbar/upgrade_status_spec.rb
```diff
@@ -254,5 +254,24 @@ def new_status
...
+
+    it "saves and checks current node data" do
+      expect(subject.current_substep).to be_nil
+      expect(subject.progress[:current_node]).to be nil
+      expect(subject.progress[:remaining_nodes]).to be nil
+
+      expect(subject.save_substep(:controllers)).to be true
+      expect(subject.current_substep).to eql :controllers
+      expect(subject.progress).to_not be_empty
+      current_node = {
```
MaximilianMeister: i'd just define it on top as a global variable:

```ruby
let(:current_node) do
  {
    name: "controller1",
    ip: "1.2.3.4",
    state: "post-upgrade"
  }
end
```

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -135,6 +137,27 @@ def finished?
       current_step == upgrade_steps_6_7.last
     end
 
+    def save_current_node(node_data = {})
```
MaximilianMeister: maybe call `node_data` `node_progress` ?

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -135,6 +137,27 @@ def finished?
...
+      ::Crowbar::Lock::LocalBlocking.with_lock(shared: false, logger: @logger, path: lock_path) do
+        progress[:current_node] = node_data
+        save
+      end
+    end
+
+    def save_remaining(remaining = 0)
```
MaximilianMeister: i'd call it more explicitly `save_remaining_nodes`

crowbar_framework/lib/crowbar/upgrade_status.rb
jsuchome: I'm not sure I understand. I think all the node related data fit into the same hash...

MaximilianMeister: i didn't mean that, it was more the naming of the variable

jsuchome: Ah, well. I'm still not sure what do we want to store there.

And there are some static data (name, IP), which are really nothing like "progress"

PR #899, opened 2016-12-15
delayed_job: require delayed_job when starting crowbar service

MaximilianMeister:


MaximilianMeister:
gating will fail until https://build.suse.de/request/show/125531 is closed

MaximilianMeister:
[(5) perform more upgrade actions in the background](https://trello.com/c/qjZy2nuP/202-5-perform-more-upgrade-actions-in-the-background)

MaximilianMeister:
I've changed the pid directory to `/run/tmpfiles.d` thanks to @dirkmueller suggestion.
https://github.com/SUSE-Cloud/automation/pull/1537 is also merged, lets see what the next gating says...

MaximilianMeister:
`/run/tmpfiles.d` is only writable by root, so I have now placed the pid files to `crowbar_framework/tmp/pids` which is also used by puma to store the pid files

rsalevsky:
Would be good to have gating here.

PR #900, opened 2016-12-15
Fix proposal commit

rsalevsky:


rsalevsky:
[P1: bsc#1014835 - L3: SOC6: various errors in crowbar production log](https://trello.com/c/LBR7JSQJ/11-p1-bsc-1014835-l3-soc6-various-errors-in-crowbar-production-log)

crowbar_framework/app/models/service_object.rb
```diff
@@ -569,7 +569,11 @@ def proposal_commit(inst, in_queue = false, validate_after_save = true, bootstra
         # Make sure we unmark the wall
         prop.reload
         prop["deployment"][@bc_name]["crowbar-committing"] = false
-        prop.latest_applied = (response.first == 200)
+        prop.latest_applied = if response.blank?
+          false
+        else
+          response.first == 200
+        end
```
vuntz: I'm fine with this; alternatively, we could initialize ` response = [500, "Internal Error -- Should not happen"]` instead of `response = nil`

PR #901, opened 2016-12-15
network_service: Log backtrace in error case

rsalevsky:
Logging the backtrace makes it easier to debug issues like:
```
Error finding address: undefined method `[]' for nil:NilClass`
```


rsalevsky:
[P2: bsc#1014835 - L3: SOC6: various errors in crowbar production log](https://trello.com/c/LBR7JSQJ/11-p2-bsc-1014835-l3-soc6-various-errors-in-crowbar-production-log)

PR #902, opened 2016-12-15
upgrade: Let the progress library track substeps and current node

jsuchome:
We need this for detailed information during the last part of upgrade,
upgrading cloud nodes.

(cherry picked from commit 6befd7c2e3bcd372fc3835a8d3e9ed3cd2bbddb9)

Backport of https://github.com/crowbar/crowbar-core/pull/897


jsuchome:

skazi0:
For the record: this backport is not strictly needed as the substeps are currently used only after admin server is upgraded. It's added here to keep things in sync and avoid problems with future changes.

PR #903, opened 2016-12-16
crowbar_register: handle storage-only deploys (bsc#993475)

tserong:
For SES-only deploys, install ses-release instead of
suse-openstack-cloud-release.  Also, make the messaging slightly
more generic.


dirkmueller:
@tserong could you rebase this change so that the service checks are retriggered? they're currently too old so that I can retrigger them outside this

tserong:
No problem, done.

PR #906, opened 2016-12-16
Implement (and use) a search with cached results for cookbooks

vuntz:
Most of the time, we're happy with node attributes that are not older than the start of the chef-client run. So we can cache things.

~~The last commit looks huge, but it's mostly an indentation change after removing a useless test, so you may want to use `w=1` for reviewing.~~

### DIFF
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -187,7 +187,6 @@ temporary_dhcp = {}
 nodes = node_search_with_cache("*:*")
 fqdns = []
 nodes.each do |n|
-  n = Node.load(n.name)
   fqdns.push(n[:fqdn])
   cname = n["crowbar"]["display"]["alias"] rescue nil
   cname = nil unless cname && ! cname.empty?
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -183,13 +183,12 @@ cluster_zone[:hosts] = Mash.new
 # As DHCP addresses can be re-used, we make sure to use the one node which is
 # the most recent; this requires two passes
 temporary_dhcp = {}
-# Get the config environment filter
-#env_filter = "dns_config_environment:#{node[:dns][:config][:environment]}"
-env_filter = "*:*" # Get all nodes for now.  This is a hack around a timing issue in ganglia.
 # Get the list of nodes
-nodes = search(:node, "#{env_filter}")
+nodes = node_search_with_cache("*:*")
+fqdns = []
 nodes.each do |n|
   n = Node.load(n.name)
+  fqdns.push(n[:fqdn])
   cname = n["crowbar"]["display"]["alias"] rescue nil
   cname = nil unless cname && ! cname.empty?
   base_name_no_net = n[:fqdn].chomp(".#{node[:dns][:domain]}")
@@ -247,8 +246,8 @@ search(:crowbar, "id:*_network").each do |network|
   next unless network.key?("allocated_by_name")
   net_name=network[:id].gsub(/_network$/, "").gsub("_","-")
   network[:allocated_by_name].each_key do |host|
-    if search(:node, "fqdn:#{host}").size > 0 or not host.match(/.#{node[:dns][:domain]}$/)
-      #this is node in crowbar terms or it not belong to our domain, so lets skip it
+    if !host.match(/.#{node[:dns][:domain]}$/) || fqdns.include?(host)
+      # this is node in crowbar terms or it not belong to our domain, so lets skip it
       next
     end
     base_name=host.chomp(".#{node[:dns][:domain]}")
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -90,7 +90,7 @@ execute "enable netfilter for bridges" do
   subscribes :run, resources(cookbook_file: "modprobe-bridge.conf"), :delayed
 end
 
-provisioner_instance = node[:provisioner][:config][:environment].gsub(/^provisioner-config-/, "") rescue "default"
+provisioner_instance = CrowbarHelper.get_proposal_instance(node, "provisioner", "default")
 provisioner = node_search_with_cache("roles:provisioner-server", provisioner_instance).first
 conduit_map = Barclamp::Inventory.build_node_map(node)
 Chef::Log.debug("Conduit mapping for this node:  #{conduit_map.inspect}")
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -90,7 +90,8 @@ execute "enable netfilter for bridges" do
   subscribes :run, resources(cookbook_file: "modprobe-bridge.conf"), :delayed
 end
 
-provisioner = search(:node, "roles:provisioner-server")[0]
+provisioner_instance = node[:provisioner][:config][:environment].gsub(/^provisioner-config-/, "") rescue "default"
+provisioner = node_search_with_cache("roles:provisioner-server", provisioner_instance).first
 conduit_map = Barclamp::Inventory.build_node_map(node)
 Chef::Log.debug("Conduit mapping for this node:  #{conduit_map.inspect}")
 route_pref = 10000
--- a/chef/cookbooks/network/recipes/switch_config.rb
+++ b/chef/cookbooks/network/recipes/switch_config.rb
@@ -50,7 +50,7 @@ admin_ip = node.address.addr
 
 switch_config={}
 
-search(:node, "*:*").each do |a_node|
+node_search_with_cache("*:*").each do |a_node|
   node_map = Chef::Recipe::Barclamp::Inventory.build_node_map(a_node)
   node_map.each do |conduit, conduit_info|
     if_list = conduit_info["if_list"]
--- a/chef/cookbooks/provisioner/recipes/base.rb
+++ b/chef/cookbooks/provisioner/recipes/base.rb
@@ -220,7 +220,7 @@ if node[:platform_family] == "suse" && !node.roles.include?("provisioner-server"
   admin_ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(provisioner_server_node, "admin").address
   web_port = provisioner_server_node[:provisioner][:web_port]
 
-  ntp_instance = node[:ntp][:config][:environment].gsub(/^ntp-config-/, "") rescue "default"
+  ntp_instance = CrowbarHelper.get_proposal_instance(node, "ntp", "default")
   ntp_servers = node_search_with_cache("roles:ntp-server", ntp_instance)
   ntp_servers_ips = ntp_servers.map { |n| Chef::Recipe::Barclamp::Inventory.get_network_by_type(n, "admin").address }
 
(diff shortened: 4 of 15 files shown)


vuntz:
Let me drop the last commit, it creates too much hound noise for now.

vuntz:
So I know the `rescue` is not good style, but it makes things more readable here. Any suggestion?

rsalevsky:
@vuntz The only idea what I have is to reimplement the ruby 2.3.0 dig function (http://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig) in ruby and use this. We have often issues with NilClasses so I guess it would be helpful.

vuntz:
@rsalevsky oh, the dig method is interesting. I guess we could add a mixin for it. But probably something for later...

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -247,8 +245,8 @@ def make_zone(zone)
   next unless network.key?("allocated_by_name")
   net_name=network[:id].gsub(/_network$/, "").gsub("_","-")
   network[:allocated_by_name].each_key do |host|
-    if search(:node, "fqdn:#{host}").size > 0 or not host.match(/.#{node[:dns][:domain]}$/)
-      #this is node in crowbar terms or it not belong to our domain, so lets skip it
+    if !host.match(/.#{node[:dns][:domain]}$/) || fqdns.include?(host)
```
matelakat: it would have read better if we kept the order of statements.

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -247,8 +245,8 @@ def make_zone(zone)
...
+      # this is node in crowbar terms or it not belong to our domain, so lets skip it
```
matelakat: I don't quite understand this comment TBH, maybe I'm lacking context. Does it want to say that the node either belongs to crowbar or is out of our domain?

chef/cookbooks/utils/libraries/search.rb
```diff
@@ -48,5 +37,58 @@ def get_instance(query)
...
+      #  2) Same cookbook: node[@cookbook_name][:config][:environment]
+      env = if !bc_instance.nil?
+        "#{barclamp}-config-#{bc_instance}"
+      elsif node[barclamp] && node[barclamp][:config] && (barclamp == cookbook_name)
+        node[barclamp][:config][:environment]
+      elsif !node[cookbook_name].nil? &&
+          !node[cookbook_name]["#{barclamp}_instance"].nil? &&
+          !node[cookbook_name]["#{barclamp}_instance"].empty?
+        "#{barclamp}-config-#{node[cookbook_name]["#{barclamp}_instance"]}"
+      end
```
matelakat: What will be the value for `env` if that failed on both `elif` tests? Will it be `nil` ?

chef/cookbooks/network/recipes/default.rb
```diff
@@ -90,7 +90,8 @@
   subscribes :run, resources(cookbook_file: "modprobe-bridge.conf"), :delayed
 end
 
-provisioner = search(:node, "roles:provisioner-server")[0]
+provisioner_instance = node[:provisioner][:config][:environment].gsub(/^provisioner-config-/, "") rescue "default"
+provisioner = node_search_with_cache("roles:provisioner-server", provisioner_instance).first
```
matelakat: This is the second time I see these two lines - maybe it's worth extracting to the utils module?

chef/cookbooks/network/recipes/switch_config.rb
```diff
@@ -50,7 +50,7 @@ def setup_interface(switch_config, a_node, conduit, switch_name, interface_id )
 
 switch_config={}
 
-search(:node, "*:*").each do |a_node|
+node_search_with_cache("*:*").each do |a_node|
```
matelakat: The same applies here as earlier, when we used `node_search_with_cache` with a query that does not match `/^(roles|recipes):(\w*).*$/`

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -70,7 +70,8 @@
 
 # Find provisioner servers and include them.
 provisioner_server_node = nil
-search(:node, "roles:provisioner-server AND provisioner_config_environment:#{node[:provisioner][:config][:environment]}") do |n|
+provisioners = node_search_with_cache("roles:provisioner-server")
```
matelakat: is this search equivalent with the one that it replaces? we had more conditions there. Again, I might lack some very basic chef knowledge here.

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -219,7 +220,8 @@
   admin_ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(provisioner_server_node, "admin").address
   web_port = provisioner_server_node[:provisioner][:web_port]
 
-  ntp_servers = search(:node, "roles:ntp-server")
+  ntp_instance = node[:ntp][:config][:environment].gsub(/^ntp-config-/, "") rescue "default"
+  ntp_servers = node_search_with_cache("roles:ntp-server", ntp_instance)
```
matelakat: The same pattern here as earlier.
```ruby
provisioner_instance = node[:provisioner][:config][:environment].gsub(/^provisioner-config-/, "") rescue "default"
provisioner = node_search_with_cache("roles:provisioner-server", provisioner_instance).first
```

chef/cookbooks/provisioner/recipes/dhcp_update.rb
```diff
@@ -1,6 +1,7 @@
 admin_ip = Barclamp::Inventory.get_network_by_type(node, "admin").address
 
-dns_servers = search(:node, "roles:dns-server").map do |n|
+dns_instance = node[:dns][:config][:environment].gsub(/^dns-config-/, "") rescue "default"
+dns_servers = node_search_with_cache("roles:dns-server", dns_instance).map do |n|
```
matelakat: Same search pattern

chef/cookbooks/provisioner/recipes/setup_base_images.rb
```diff
@@ -476,7 +476,8 @@
       # Add base OS install repo for suse
       node.set[:provisioner][:repositories][os][arch]["base"] = { "baseurl=#{install_url}" => true }
 
-      ntp_servers = search(:node, "roles:ntp-server")
+      ntp_instance = node[:ntp][:config][:environment].gsub(/^ntp-config-/, "") rescue "default"
+      ntp_servers = node_search_with_cache("roles:ntp-server", ntp_instance)
```
matelakat: Here again

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -61,9 +61,9 @@ def find_node_boot_mac_addresses(node, admin_data_net)
 pxecfg_subdir = "bios/pxelinux.cfg"
 uefi_subdir = "efi"
 
-nodes = search(:node, "*:*")
+nodes = node_search_with_cache("*:*")
```
matelakat: Again, not sure what `*.*` will cause.

chef/cookbooks/repos/recipes/default.rb
```diff
@@ -15,8 +15,11 @@
 
 return if node[:platform_family] == "suse" || node[:platform_family] == "windows"
 
-provisioners = search(:node, "roles:provisioner-server")
-provisioner = provisioners[0] if provisioners
+# the attribute always exist as this is run as part of the provisioner-base role
+provisioner_instance = node[:provisioner][:config][:environment].gsub(/^provisioner-config-/, "")
+provisioners = node_search_with_cache("roles:provisioner-server", provisioner_instance)
```
matelakat: same pattern, but without `rescue` - is that an issue, a bug?

chef/cookbooks/resolver/recipes/default.rb
```diff
@@ -18,8 +18,8 @@
 # limitations under the License.
 #
 
-env_filter = " AND dns_config_environment:#{node[:dns][:config][:environment]}"
-nodes = search(:node, "roles:dns-server#{env_filter}")
+dns_instance = node[:dns][:config][:environment].gsub(/^dns-config-/, "")
+nodes = node_search_with_cache("roles:dns-server", dns_instance)
```
matelakat: good old pattern again, but without `rescue`

chef/cookbooks/utils/libraries/search.rb
vuntz: Yes, it will be `nil`.

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -183,13 +183,11 @@ def make_zone(zone)
 # As DHCP addresses can be re-used, we make sure to use the one node which is
 # the most recent; this requires two passes
 temporary_dhcp = {}
-# Get the config environment filter
-#env_filter = "dns_config_environment:#{node[:dns][:config][:environment]}"
-env_filter = "*:*" # Get all nodes for now.  This is a hack around a timing issue in ganglia.
 # Get the list of nodes
-nodes = search(:node, "#{env_filter}")
+nodes = node_search_with_cache("*:*")
```
vuntz: No, it will just mean we don't use an additional filter string, and so we really only search for `"*:*"`.

chef/cookbooks/bind9/recipes/default.rb
vuntz: I'm just moving the test which is less expensive first (it matters a bit when we have 100s of nodes).

chef/cookbooks/network/recipes/switch_config.rb
vuntz: See https://github.com/crowbar/crowbar-core/pull/906#discussion_r93328570

chef/cookbooks/provisioner/recipes/base.rb
vuntz: Yes, `node_search_with_cache` adds the additional condition.

chef/cookbooks/provisioner/recipes/update_nodes.rb
vuntz: See https://github.com/crowbar/crowbar-core/pull/906#discussion_r93328570

chef/cookbooks/repos/recipes/default.rb
vuntz: It's not a bug, see comment above the line :-)

chef/cookbooks/network/recipes/default.rb
vuntz: I'm fine extracting the first line to a helper, but I don't think it makes sense to replace the two lines with a one-liner as this becomes too much voodoo if we do that, imho.

I'll add a patch tomorrow for the small helper.

vuntz: Done, see new commit.

chef/cookbooks/utils/libraries/search.rb
```diff
@@ -48,5 +36,40 @@ def get_instance(query)
...
+
+    def crowbar_filter_env(query)
+      # All cookbooks encode the barclamp name as the role name prefix, thus we can
+      # simply grab it from the query (e.g. BC 'keystone' for role 'keystone-server'):
+      barclamp = /^\w*:(\w*).*$/.match(query)[1]
+
+      # There are two conventions to filter by barclamp proposal:
+      #  1) Other barclamp cookbook: node[@cookbook_name][$OTHER_BC_NAME_instance]
+      #  2) Same cookbook: node[@cookbook_name][:config][:environment]
+      env = if node[barclamp] && node[barclamp][:config] && (barclamp == cookbook_name)
```
aspiers: +1 for keeping refactorings separate.

PR #907, opened 2016-12-16
transition: Reduce number of node searches

rsalevsky:
**TODO:**
- [ ] Remove TODOs and FIXMEs
- [ ] Reduce node saves (doing it once at the end?)
- [ ] The return value makes no real sense due to we have the name anyway


rsalevsky:
[(5) #crowbar Create node object that is backed by database](https://trello.com/c/EsVoyO2G/36-5-crowbar-create-node-object-that-is-backed-by-database)

vuntz:
Interesting; that may very well speed up the remaining slow bits I see here when applying (since they're related to transitioning to applying/ready).

PR #918, opened 2016-12-20
upgrade: Use .rpmnew configs for OpenStack services

toabctl:
The packages in Newton have a new configuration structure. Instead of
using the /etc/$project/$project.conf file, the different projects
have a /etc/$project/$project.conf.d/ directory which should be used.
Because package upgrades do not override the config files modified by Crowbar,
we need to reset theses files once.
If a user afterwards changes the files again, we do not override theses then.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.erb
@@ -30,6 +30,19 @@ cleanup()
     exit 2
 }
 
+# check if a .rpmnew file exists and use that
+function default_package_config()
+{
+    local config_file=$1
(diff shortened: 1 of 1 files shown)


toabctl:
This is not tested in a real upgrade job. Will that happen automatically with that PR?

jsuchome:
OK, this can work, I guess. Easier than adapting the chef recipes code, I guess...

jsuchome:
> This is not tested in a real upgrade job. Will that happen automatically with that PR?

Nope. But you can try to build it with the parameters of this job https://ci.suse.de/view/Cloud/view/Cloud7/job/cloud-mkcloud7-job-upgrade-nondisruptive-ha-x86_64/

PR #919, opened 2016-12-20
upgrade: don't lock the progress file when reading only the status

MaximilianMeister:
during the upgrade we will use a lot of polling to /api/upgrade
for the status, therefore we need as minimal impact/cost as
necessary

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -20,7 +20,11 @@ module Api
   class Upgrade < Tableless
     class << self
       def status
-        ::Crowbar::UpgradeStatus.new.progress
+        ::Crowbar::UpgradeStatus.new(
+          Rails.logger,
+          "/var/lib/crowbar/upgrade/6-to-7-progress.yml",
+          false
+        ).progress
       end
 
       def checks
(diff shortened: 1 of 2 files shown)


jsuchome:
Is it really such a problem? Between polling calls there should be at least second anyway...

MaximilianMeister:
> You could even potentially have a situation where the file is being loaded whilst it is being rewritten by a writer, in which case you'd get a corrupt blob of JSON which the parser would barf on, causing a crash.

I guess this would not lead to crash, at least not the backend. it would cause an exception for this one request. but you're right that wouldn't be correct behavior.

> Did you do any profiling to prove this is really a bottleneck?

I didn't claim that this is a bottleneck, as you can see in the commit message I mentioned minimizing cost and impact. I haven't done any detailed profiling, no. i just observed the logs with a server where crowbar-ui is polling every second, and i felt it might make a difference, especially as we are now looking on every corner for runtime improvements.

> And anyway, surely the cost of this is orders of magnitude smaller than the cost of the whole Rails request pipeline

sure there are other areas that are much more memory intensive, but the polling is done through most of the upgrade process, where we haven't done any memory profiling yet either, so from this (rather ignorant) perspective, every LOC adds more load.

jsuchome:
> especially as we are now looking on every corner for runtime improvements.

I do not think this is problem in the upgrade case. After all, when we do this polling, there's only upgrade work running and this is mostly only about taking care of one node at a time.

MaximilianMeister:
> I do not think this is problem in the upgrade case. After all, when we do this polling, there's only upgrade work running and this is mostly only about taking care of one node at a time.

ok. closing then...

PR #920, opened 2016-12-20
Move NodeObject partly to Database

rsalevsky:
Not finished!

The goal is that crowbar isn't accessing any data from chef and only syncs data to chef is needed.


crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -423,7 +422,7 @@ def initiate_nodes_upgrade
     pxecfg_subdir = "bios/pxelinux.cfg"
 
     upgrade_nodes.each do |node|
-      boot_ip_hex = node["crowbar"]["boot_ip_hex"]
+      boot_ip_hex = node.crowbar["crowbar"]["boot_ip_hex"]
```
vuntz: Are you sure about that one?

vuntz: Ah, I see it elsewhere too, so I guess it's intended.

vuntz:
@rsalevsky I'd suggest to create a separate PR that does the `Node.find_node_by_name` -> ` Node.find_by_name` change as well as the `node[:crowbar_call]` -> `node.crowbar_wall` change. We should be able to merge that fast, and that would make this PR smaller/easier to review/less prone to conflicts.

rsalevsky:
We don't do this anymore.

PR #921, opened 2016-12-20
upgrade:Track the sub steps and nodes status in the progress library

jsuchome:
This is not extended to the upgrade of compute nodes, as relevant PR is not merged yet. I think it could be solved as a followup.

### DIFF
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -34,7 +34,7 @@ module Api
 
     def pre_upgrade
       if execute_and_wait_for_finish("/usr/sbin/crowbar-pre-upgrade.sh", 300)
-        save_node_state("Pre upgrade script run was successful.")
+        Rails.logger.info("Pre upgrade script run was successful.")
         return true
       end
       false
@@ -42,7 +42,7 @@ module Api
 
     def os_upgrade
       if execute_and_wait_for_finish("/usr/sbin/crowbar-upgrade-os.sh", 600)
-        save_node_state("Package upgrade was successful.")
+        Rails.logger.info("Package upgrade was successful.")
         return true
       end
       false
@@ -50,7 +50,7 @@ module Api
 
     def router_migration
       if execute_and_wait_for_finish("/usr/sbin/crowbar-router-migration.sh", 600)
-        save_node_state("Router migration was successful.")
+        Rails.logger.info("Router migration was successful.")
         return true
       end
       false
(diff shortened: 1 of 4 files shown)


jsuchome:

crowbar_framework/app/models/api/node.rb
```diff
@@ -136,9 +133,24 @@ def disable_pre_upgrade_attribute_for(name)
...
+        ip: @node.public_ip,
+        state: state,
+        role: role
+      )
+      if state == "upgraded"
+        progress = status.progress
+        remaining = progress[:remaining_nodes] - 1
+        upgraded = progress[:upgraded_nodes] + 1
+        ::Crowbar::UpgradeStatus.new.save_nodes(
+          upgraded, remaining < 0 ? 0 : remaining
```
MaximilianMeister: can `remaining` be smaller than 0?

jsuchome: Probably couldn't. If this were a last node, remaining would be 1 here, and lowered to 0, and the function would not be called next time.

I think.

MaximilianMeister: it looks like that, that's why i was asking. but why do we need this check then?

jsuchome: Probably because I did not think it though the first time ;-)

jsuchome:
BTW, this is the status after upgrading controller nodes:

```
root@crowbar:/opt/dell # crowbarctl upgrade status
+--------------------------------+--------------------------------------+
| Status                         | Value                                |
+--------------------------------+--------------------------------------+
| current_step                   | nodes_db_dump                        |
| current_substep                | computes                             |
| current_node.name              | d52-54-77-77-77-01.vo6.cloud.suse.de |
| current_node.alias             | data1                                |
| current_node.ip                | 10.165.197.2                         |
| current_node.state             | upgraded                             |
| current_node.role              | controller                           |
| remaining_nodes                | 2                                    |
| upgraded_nodes                 | 2                                    |
| steps.upgrade_prechecks.status | passed                               |
| steps.upgrade_prepare.status   | passed                               |
| steps.admin_backup.status      | passed                               |
| steps.admin_repo_checks.status | passed                               |
| steps.admin_upgrade.status     | passed                               |
| steps.database.status          | passed                               |
| steps.nodes_repo_checks.status | passed                               |
| steps.nodes_services.status    | passed                               |
| steps.nodes_db_dump.status     | pending                              |
| steps.nodes_upgrade.status     | pending                              |
| steps.finished.status          | pending                              |
+--------------------------------+--------------------------------------+
```

MaximilianMeister:
`save_error_state` is also still a bit misleading

jsuchome:
> would you mind dropping the save_upgrade_state method as well and replace it with Rails.logger 
> statements?

@MaximilianMeister I'd like to address those in followup PR, the way it is now is not harmful.

PR #922, opened 2016-12-20
upgrade: If upgrade step is out of order, tell what should be correct

jsuchome:


crowbar_framework/lib/crowbar/error/upgrade_status.rb
```diff
@@ -29,8 +29,9 @@ def initialize(step_name = "")
     end
 
     class StartStepOrderError < StandardError
-      def initialize(step_name = "")
-        super("Start of step #{step_name} requested in the wrong order")
+      def initialize(step_name = "", correct_step = "")
+        super("Start of step '#{step_name}' requested in the wrong order. " \
+        "Correct next step is '#{correct_step}'.")
```
MaximilianMeister: what if `correct_step` (the default) is empty? the expression wouldn't make sense anymore. I think you need to construct the help string under some conditions

jsuchome: Well, the concern is the same when `step_name` is empty, right? 
But I'm always passing some values to `StartStepOrderError`, so it it a problem? Maybe we should just drop the default values then?

MaximilianMeister: > Well, the concern is the same when step_name is empty, right?

`Start of step requested in the wrong order.` is still a correct sentence.
`Correct next step is .` is obviously incomplete.

why not just:
```ruby
      def initialize(step_name = "", correct_step = "")
        msg = "Start of step '#{step_name}' requested in the wrong order. "
        msg += "Correct next step is '#{correct_step}'." unless correct_step.empty?
        super(msg)
      end
```

jsuchome: Ok. But still, there should not be empty value passed under any circumstance...

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -217,6 +217,15 @@ def step_allowed?(step)
       false
     end
 
+    # The only case when current step is not the step that should be executed
+    # is when it is already running. In that case, return the next step.
+    def next_step_to_execute
+      step = current_step
+      return step unless running? step
+      i = upgrade_steps_6_7.index step
+      upgrade_steps_6_7[i + 1] || upgrade_steps_6_7.last
```
skazi0: Please add a unit test for this edge case.

jsuchome: It actually cannot happen. Last step (`:finished` does not move further when it becomes `current_step`, so this method exits on line 224.

PR #924, opened 2016-12-21
Make applying proposal slightly faster

vuntz:
When there are many nodes in a proposal, things can start to be slow. Part of it can easily be avoided (by not reloading nodes all the time and skipping useless bits).

Backport of https://github.com/crowbar/crowbar-core/pull/959

(I dropped the backport of https://github.com/crowbar/crowbar-core/pull/896 as it's a bit too intrusive)


pgonin:
We experienced a ~30s improvement (without ~6:30) with this patch on SAP-like QA platform (with 60 compute nodes)

rsalevsky:
Gating failed in tempest but seams unrelated.

PR #925, opened 2016-12-21
don't unload ipmi modules

dirkmueller:
The pattern we see on the SUSE QA team hardware is that on discovery image run, which loads and unloads ipmi modules, every once in a while the bmc suddenly stops responding. Since we can not remove the need for loading ipmi modules, and unloading modules is sort of not really well tested, lets not do that and see if it makes a difference.


vuntz:
TBH, I don't even know why we remove the modules. I'd be +1 for this (just not sure about the ipmi.service, I don't see one?)

PR #928, opened 2016-12-22
upgrade: Workaround incorrect handling of empty responses in puma

skazi0:
Puma in versions < v2.12.3 is improperly generating chunked responses
if the content is empty. The empty responses usually come from head()
calls. This workaround disables chunked encoding for head() responses.

### DIFF
--- a/crowbar_framework/app/controllers/application_controller.rb
+++ b/crowbar_framework/app/controllers/application_controller.rb
@@ -42,6 +42,14 @@ class ApplicationController < ActionController::Base
     Rails.cache.fetch(:sanity_check_errors).empty?
   }
 
+  # workaround for incorrect handling of chunking for empty responses in puma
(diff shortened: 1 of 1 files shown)


skazi0:
[bsc#1016724: Incorrect handling of empty responses in puma](https://trello.com/c/MlZz4nKg/234-bsc-1016724-incorrect-handling-of-empty-responses-in-puma)

rsalevsky:
We updated the package to 2.16.0, can you check if this fixes the problem?

skazi0:
I tested the Cloud7 side and the problem doesn't occur anymore.
**Note that the same problem exists on Cloud6 side but it didn't manifest itself yet.**

dirkmueller:
we'll also update to the new version in cloud6, so that should be fine imho.

PR #934, opened 2016-12-22
upgrade: Added basic time stamps to the scripts executed by upgrade

jsuchome:
The output of the scripts looks like this: http://pastebin.nue.suse.com/18770/src


jsuchome:
[Print time stamps while logging from upgrade scripts (into /var/log/crowbar/node-upgrade.log)](https://trello.com/c/8HItaUt6/240-print-time-stamps-while-logging-from-upgrade-scripts-into-var-log-crowbar-node-upgrade-log)

MaximilianMeister:
what about replacing all echos with a log function that always logs a timestamp as well?

jsuchome:
@mmnelemane sounds good

jsuchome:
I know, some library would be good for it... I just did not want to think how should it be done and where should it belong...

rsalevsky:
Needs a rebase

PR #935, opened 2016-12-22
upgrade: Rework router migration

rhafer:
This is still very much work in progress. Especially the migration for subsequent nodes is largely untested. Just to give an overview where things are heading. Also there is a small conflict with compute node work in #875 (but that should be easy to solve)

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
@@ -10,36 +10,30 @@ UPGRADEDIR=/var/lib/crowbar/upgrade
 mkdir -p "`dirname "$LOGFILE"`"
 exec >>"$LOGFILE" 2>&1
 
+hostname=$1
+if [[ "$2" = "delete-ns" ]]; then
+    delete_ns="--ssh-delete-namespace"
+else
+    delete_ns=""
+fi
+
 echo "Executing $BASH_SOURCE"
 
 set -x
 
 mkdir -p $UPGRADEDIR
 
-if [[ -f $UPGRADEDIR/crowbar-router-migration-ok ]] ; then
-    echo "router migration actions were already successfully executed"
-    exit 0
-fi
-
 set +x
 source /root/.openrc
 set -x
 
-upgraded_agent=$(neutron agent-list --binary neutron-l3-agent --host $(hostname) -f value -c id)
-routers=$(neutron router-list -f value -c id)
-for router in $routers; do
-    agent=$(neutron l3-agent-list-hosting-router $router -f value -c id)
-    old_host=$(neutron agent-show $agent -f value -c host)
+/usr/bin/neutron-ha-tool --l3-agent-evacuate $hostname --wait-for-router $delete_ns
+ret=$?
+if [ $ret != 0 ] ; then
+    echo "Failed to evacuate l3 agent on host: $hostname"
+    echo $ret > $UPGRADEDIR/crowbar-router-migration-failed
+    exit $ret
+fi
 
-    # delete old router assignment to agent
-    neutron l3-agent-router-remove $agent $router
-
-    # assign router to agent in upgraded node
-    neutron l3-agent-router-add $upgraded_agent $router
-
-    # delete old routers namespaces. Unassignment didn't delete the namespaces
-    ssh $old_host ip netns delete qrouter-$router
-
-done
 
 touch $UPGRADEDIR/crowbar-router-migration-ok
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -316,25 +316,25 @@ module Api
 
       # Method for upgrading first node of the cluster
       # other_node_name argument is the name of any other node in the same cluster
-      def upgrade_first_cluster_node(node_name, other_node_name)
-        node_api = Api::Node.new node_name
+      def upgrade_first_cluster_node(node, other_node)
+        node_api = Api::Node.new node.name
         return true if node_api.upgraded?
-        other_node_api = Api::Node.new other_node_name
+        other_node_api = Api::Node.new other_node.name
         node_api.save_node_state("controller")
-        save_upgrade_state("Starting the upgrade of node #{node_name}")
+        save_upgrade_state("Starting the upgrade of node #{node.name}")
         # upgrade the first node
         return false unless node_api.upgrade
 
         # Explicitly mark the first node as cluster founder
         # and in case of DRBD setup, adapt DRBD config accordingly.
-        unless Api::Pacemaker.set_node_as_founder node_name
-          save_error_state("Changing the cluster founder to #{node_name} has failed")
+        unless Api::Pacemaker.set_node_as_founder node.name
+          save_error_state("Changing the cluster founder to #{node.name} has failed")
           return false
         end
         # remove pre-upgrade attribute, so the services can start
-        return false unless other_node_api.disable_pre_upgrade_attribute_for node_name
+        return false unless other_node_api.disable_pre_upgrade_attribute_for node.name
         # delete old pacemaker resources (from the node where old pacemaker is running)
-        return false unless delete_pacemaker_resources other_node_name
+        return false unless delete_pacemaker_resources other_node.name
         # start crowbar-join at the first node
         return false unless node_api.post_upgrade
         return false unless node_api.join_and_chef
@@ -343,18 +343,19 @@ module Api
         node_api.save_node_state("controller", "upgraded")
       end
 
-      def upgrade_next_cluster_node(node_name)
-        node_api = Api::Node.new node_name
+      def upgrade_next_cluster_node(node)
+        node_api = Api::Node.new node.name
         return true if node_api.upgraded?
         node_api.save_node_state("controller")
-        save_upgrade_state("Starting the upgrade of node #{node_name}")
+        save_upgrade_state("Starting the upgrade of node #{node.name}")
+
         return false unless node_api.upgrade
         return false unless node_api.post_upgrade
         return false unless node_api.join_and_chef
         # Remove pre-upgrade attribute _after_ chef-client run because pacemaker is already running
         # and we want the configuration to be updated first
         # (disabling attribute causes starting the services on the node)
-        return false unless node_api.disable_pre_upgrade_attribute_for node_name
+        return false unless node_api.disable_pre_upgrade_attribute_for node.name
         node_api.save_node_state("controller", "upgraded")
       end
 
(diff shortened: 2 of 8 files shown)


rhafer:

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -407,24 +410,24 @@ def upgrade_drbd_cluster(cluster)
...
+            drbd_slave = drbd_node
           end
         end
-        return false if drbd_slave.empty?
+        return false if drbd_slave.nil?
 
         return false unless upgrade_first_cluster_node drbd_slave, drbd_master
 
-        return false unless upgrade_next_cluster_node drbd_master
+        return false unless upgrade_next_cluster_node drbd_master, drbd_slave
```
abel-navarro: inside `upgrade_next_cluster_node` we [already call](https://github.com/crowbar/crowbar-core/pull/935/files#diff-33ff816d38c460e2a65a4918bf6f94d0R377) `upgrade_first_cluster_node` (line above).

abel-navarro:
actually the >2 node cluster work was already addressed here: #779

crowbar_framework/app/models/api/upgrade.rb
rhafer: @abel-navarro Not sure what you mean by that. Please elaborate.

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -25,21 +27,6 @@ set +x
...
-    neutron l3-agent-router-remove $agent $router
-
-    # assign router to agent in upgraded node
-    neutron l3-agent-router-add $upgraded_agent $router
-
-    # delete old routers namespaces. Unassignment didn't delete the namespaces
-    ssh $old_host ip netns delete qrouter-$router
-
-done
+/usr/bin/neutron-ha-tool --l3-agent-evacuate $hostname --wait-for-router
```
rhafer: Yeah, that's one reason why I added the WIP label.

crowbar_framework/app/models/api/upgrade.rb
abel-navarro: @rhafer sorry my mistake. I didn't expand the sections on the PR diff and mistakenly assigned a call to the wrong function. Alles klar.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1123,6 +1123,16 @@ def wait_for_script_to_finish(script, seconds, args = [])
     end
   end
 
+  # Removes the -ok/-failed files that might have been created by a script
+  # running via "wait_for_script_to_finish"
+  def delete_script_exit_files(script)
+    base = "/var/lib/crowbar/upgrade/" + File.basename(script, ".sh")
+    ok_file = base + "-ok"
+    failed_file = base + "-failed"
```
jsuchome: It would probably make sense to have a method that would return `[ ok_file, failed_file ]` for given script. We're using this mechanism already on 3 places in this model...

jsuchome:
Would it be possible to squash the last commit to the commit which broke the testsuite? 
Or is it rather fixing the tests that were broken by design and not failing?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -445,6 +451,28 @@ def delete_pacemaker_resources(node_name)
         end
       end
 
+      # Evacuate all routers away from the specified network node to other
+      # available network nodes. The evacuation procedure it started on the
+      # specified controller node
+      def evacuate_network_node(controller, network_node, delete_namespaces = false)
```
jsuchome: So the `controller` node is the target node for the routers, right? Wouldn't it be better name?
(Feel free to discard this comment, if network experts think it is better this way, than it likely is)

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -10,36 +10,23 @@ UPGRADEDIR=/var/lib/crowbar/upgrade
...
-    neutron l3-agent-router-remove $agent $router
-
-    # assign router to agent in upgraded node
-    neutron l3-agent-router-add $upgraded_agent $router
-
-    # delete old routers namespaces. Unassignment didn't delete the namespaces
-    ssh $old_host ip netns delete qrouter-$router
-
-done
+/usr/bin/neutron-ha-tool --l3-agent-evacuate $hostname --wait-for-router $delete_ns
```
jsuchome: Could this `neutron-ha-tool` invocation fail? 

Currently the only possible failure of this whole script is the timeout...

jsuchome:
So, I've tested this manually, and noticed only few packet drops during the first invocation of the router migration script. Here's the relevant node-upgrade.log from the first node (edited for better readability): http://pastebin.nue.suse.com/18778/src

crowbar_framework/app/models/node_object.rb
rhafer: Hm, true. Are you ok if I address that in a separate PR?

crowbar_framework/app/models/api/upgrade.rb
rhafer: The controller node is actually not the target. Well, it might be some cases (the standard single-two-node-cluster setup). But in most other cases it isn't.

rhafer:
> Would it be possible to squash the last commit to the commit which broke the testsuite?

Sure.

> Or is it rather fixing the tests that were broken by design and not failing?

It's a mixture of both actually. But I guess it makes sense to squash it with the other commit.

crowbar_framework/app/models/node_object.rb
jsuchome: Yes

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Ah, ok

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
rhafer: Fixed

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -10,36 +10,30 @@ UPGRADEDIR=/var/lib/crowbar/upgrade
...
-
-    # assign router to agent in upgraded node
-    neutron l3-agent-router-add $upgraded_agent $router
-
-    # delete old routers namespaces. Unassignment didn't delete the namespaces
-    ssh $old_host ip netns delete qrouter-$router
+/usr/bin/neutron-ha-tool --l3-agent-evacuate $hostname --wait-for-router $delete_ns
+ret=$?
+if [ $ret != 0 ] ; then
+    echo "Failed to evacuate l3 agent on host: %hostname"
```
jsuchome: `$hostname`

rhafer: fixed

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -451,6 +457,28 @@ def delete_pacemaker_resources(node_name)
...
+        args = [network_node]
+        args << "delete-ns" if delete_namespaces
+        controller.wait_for_script_to_finish(
+          "/usr/sbin/crowbar-router-migration.sh", 600, args
```
matelakat: This hardwired `600` looks strange, although I understand that this was the timeout originally for the other script. Given that we are using `--wait-for-router`, execution time can increase hugely. I am not sure we need to address it here, but at some point this hardwired value could be a real pain. Best would be to make the timeout's value depending on the number of routers. This is something that we might want to put inside neutron-ha-tool

crowbar_framework/spec/models/api/upgrade_spec.rb
```diff
@@ -379,10 +394,6 @@
       allow_any_instance_of(Api::Node).to receive(:router_migration).and_return(true)
```
matelakat: Do we need this line? Is it referring to the same method that we just removed?

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -10,36 +10,30 @@ UPGRADEDIR=/var/lib/crowbar/upgrade
...
+    delete_ns=""
+fi
+
 echo "Executing $BASH_SOURCE"
 
 set -x
 
 mkdir -p $UPGRADEDIR
 
-if [[ -f $UPGRADEDIR/crowbar-router-migration-ok ]] ; then
```
matelakat: Why are we removing this part? If I understand it correcly, if the migration is successful, crowbar will remove the files. I looked for occurances of `crowbar-router-migration-failed` and `crowbar-router-migration-ok`, but found none, only this file creates them, then we remove them from crowbar. What is the purpose of these files?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -322,6 +322,8 @@ def upgrade_first_cluster_node(node, other_node)
         other_node_api = Api::Node.new other_node.name
         node_api.save_node_state("controller")
         save_upgrade_state("Starting the upgrade of node #{node.name}")
+        return false unless evacuate_network_node(node, node["hostname"])
```
matelakat: Why don't we force namespace cleanup with the first cluster node?

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
jsuchome: The scripts are executed in background (over ssh) and the caller is checking for presence of these files to realize how did they end.

crowbar_framework/app/models/api/upgrade.rb
rhafer: All services are still up and running and we don't need the ssh step in that case.

crowbar_framework/app/models/api/upgrade.rb
rhafer: Yeah, I am not really happy with that either. And 10 minutes is clearly not enough for a somewhat bigger setup.
We will need to rethink a few other timeouts here as well.

crowbar_framework/spec/models/api/upgrade_spec.rb
rhafer: Hm, right. I overlooked that.

rhafer:
> Could you amend the typos, and also remove the references to the removed function from the tests? Also I have some questions inline. I am a bit confused about the role of the -ok -failed files, but I might lack some context.

@matelakat I think I addressed most of the concerns you raised. I believe the timeout thing needs a somewhat broader discussion and the solution should not only over the router_migration step. Are you ok if I create a followup item in the backlog for it? For the namespace deletion issue we just discussed in mumble I created: https://bugzilla.suse.com/show_bug.cgi?id=1018582 as that needs to be addressed in neutron-ha-tool.

matelakat:
> @matelakat I think I addressed most of the concerns you raised. I believe the timeout thing needs a somewhat broader discussion and the solution should not only over the router_migration step. Are you ok if I create a followup item in the backlog for it? For the namespace deletion issue we just discussed in mumble I created: https://bugzilla.suse.com/show_bug.cgi?id=1018582 as that needs to be addressed in neutron-ha-tool.

Coo, thanks!

PR #938, opened 2016-12-23
upgrade: enable ptf repo during the noderepochecks (bsc#1015827)

MaximilianMeister:
the ptf repo is not tracked under any feature so we need to enable
it by hand.

enabling it as a feature would not make sense as it may not be
available, and the api response would show it as missing and
make the repocheck appear as failed. Also conceptually the ptf
repo does not seem to be a feature.

### DIFF
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -175,13 +175,19 @@ module Api
         features = []
         features.push(addon)
         architectures = node_architectures
+        platform = Api::Upgrade.target_platform(platform_exception: addon)
+
+        # as the ptf repo is not registered as a feature we need to enable it manually
+        provisioner_service = ProvisionerService.new(Rails.logger)
+        architectures.values.flatten.uniq.each do |architecture|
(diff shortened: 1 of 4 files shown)


MaximilianMeister:
[bsc#1015827: Ensure 'repocheck nodes' activates also PTF repo](https://trello.com/c/AsHC54vW/217-bsc-1015827-ensure-repocheck-nodes-activates-also-ptf-repo)

crowbar_framework/app/models/api/node.rb
```diff
@@ -162,13 +162,19 @@ def repocheck(options = {})
         features = []
         features.push(addon)
         architectures = node_architectures
+        platform = Api::Upgrade.target_platform(platform_exception: addon)
+
+        # as the ptf repo is not registered as a feature we need to enable it manually
+        architectures.values.flatten.uniq.each do |architecture|
+          provisioner_service = ProvisionerService.new(Rails.logger)
```
vuntz: Put this line outside of the loop.

MaximilianMeister: good point. thanks fixed

PR #941, opened 2017-01-02
crowbar: Don't create systemd unit for crowbar

rhafer:
crowbar.service is part of the package and installed in
/usr/lib/systemd/system/ nowadays.


rhafer:
Removing the file during upgrades is handled here: https://github.com/crowbar/crowbar/pull/2297

PR #943, opened 2017-01-03
Authn core

toabctl:


chef/cookbooks/provisioner/recipes/setup_base_images.rb
```diff
@@ -159,6 +159,7 @@
 if node[:platform_family] == "suse"
```
jgrassler: The commit message says "load on non SUSE systems", but this code says otherwise. Which one is correct?

toabctl: @jgrassler this is the 2nd commit. The first commit is only useful for non-SUSE systems

jgrassler: Ah, that clears it up, thanks (I only looked at the summary diff earlier).

PR #946, opened 2017-01-03
upgrade: Perform node upgrade in the background

MaximilianMeister:


MaximilianMeister:
[(5) perform more upgrade actions in the background](https://trello.com/c/qjZy2nuP/202-5-perform-more-upgrade-actions-in-the-background)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -273,26 +271,34 @@ def nodes
           remaining = NodeObject.find(
             "state:crowbar_upgrade AND NOT run_list_map:ceph_*"
           ).size
-          ::Crowbar::UpgradeStatus.new.save_nodes(0, remaining)
+          status.save_nodes(0, remaining)
```
jsuchome: Hm, I'm not sure if this is gonna work. I think I was using the original object in my first tests, but then went to creating new one because the data were rewritten from some other source (I think from `api/nodes`)

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -136,19 +136,21 @@ def services
...
+      errors: {
+        nodes_upgrade: {
+          data: e.message,
+          help: I18n.t("api.upgrade.nodes.help.default")
         }
-      }, status: :unprocessable_entity
-    end
+      }
+    }, status: :unprocessable_entity
+  end
```
jsuchome: Seems to be one `end` too many

crowbar_framework/app/models/api/upgrade.rb
MaximilianMeister: ok, that might cause some confusion indeed

jsuchome:
Hm, executed this way, it requires (yet non-existent) `nodes_db_dump` step to be run first. Could you please add fake action for `nodes_db_dump`?

MaximilianMeister:
> Hm, executed this way, it requires (yet non-existent) nodes_db_dump step to be run first. Could you please add fake action for nodes_db_dump?

@jsuchome added

MaximilianMeister:
@jsuchome  the `nodes_db_dump` fakery is now done in the api that `crowbarctl` can make use of it https://github.com/crowbar/crowbar-openstack/pull/721

PR #947, opened 2017-01-03
Start moving away from search queries to static config items for cookbooks

vuntz:
The static config items are actually data bag items (for now).

The main benefits are:

- performance improvement (as a search query can be expensive)
- going in the direction of not depending on centralized chef-server
- in some cases, simplification of the cookbooks

The rake task will be used from barclamp_install.rb so that the config items get regenerated when the code is updated (so that there's no need to reapply a barclamp to get an up-to-date config item with the new fields).

~~Marking as WIP as I'd like to gather feedback; but this seems to work quite well.~~

### DIFF
--- a/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+++ b/chef/cookbooks/barclamp/libraries/barclamp_library.rb
@@ -377,5 +377,48 @@ module BarclampLibrary
         end
       end
     end
+
+    class Config
+      def self.node=(node)
+        @node = node
+      end
+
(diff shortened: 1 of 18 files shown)


vuntz:
IMHO, this is ready for reviews (& merging, if people are fine with this).

vuntz:
Oh, we'll want https://github.com/crowbar/crowbar/pull/2302 at the same time.

vuntz:
@dirkmueller Added two commits to fix the remaining hound comments.

PR #949, opened 2017-01-03
provisioner: Remove one level of indentation in update_nodes recipe

vuntz:
There's an "if" test which is useless.

This looks big, but use https://github.com/crowbar/crowbar-core/pull/949/files?w=1 to review. Also, there are many hound comments (which are valid), but they are actually not introduced by this change and I didn't want to fix all of them as this would make things less reviewable.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -132,8 +132,10 @@ node_search_with_cache("*:*").each do |mnode|
   admin_mac_addresses = find_node_boot_mac_addresses(mnode, admin_data_net)
   admin_ip_address = admin_data_net.nil? ? mnode[:ipaddress] : admin_data_net.address
 
-  case
-  when (new_group == "delete")
+  ####
+  # First deal with states that don't require PXE booting
+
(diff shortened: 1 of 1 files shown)


vuntz:
For the record, I'm not going to fix the other hound violations here: they weren't introduced by the change and I'm not agreeing these should be fixed.

nicolasbock:
@vuntz The trick with appending `?w=1` is very neat! Where did you find it? I have been trying to find a UI option that would turn this on but so far haven't found it.

vuntz:
> @vuntz The trick with appending ?w=1 is very neat! Where did you find it?

It's probably one of the first thing I looked for when doing reviews in github ;-) So no idea how I found it as it's been a long time for me using it...

nicolasbock:
@vuntz Thanks for sharing :)

aspiers:
@nicolasbock @vuntz There's also https://github.com/jerone/UserScripts/tree/master/Github_Commit_Whitespace which is a more convenient version of the same. I recommend you (re-)read my [September post to `research@`](http://mailman.suse.de/mlarch/SuSE/research/2016/research.2016.09/msg00301.html) :-)

PR #950, opened 2017-01-04
upgrade: after failure and restart of failed step, continue from the same point

jsuchome:
- for DRBD - slave node gets changed, so if first node upgrade failed, next attempt will pick the other one as first ... this means e.g. that router migration will be done again, but from another node... - should be fixed by 4th commit

- most of the upgrade scripts are skipped when there's attempt to run them second time, but `reboot_and_wait` is not ... and after reboot, who starts the pacemaker if initial crowbar_join was already done (and thus is now skipped?) - should be fixed by 3rd commit

### DIFF
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -102,6 +102,12 @@ module Api
 
     # Reboot the node and wait until it comes back online
     def reboot_and_wait
+      rebooted_file = "/var/lib/crowbar/upgrade/crowbar-node-rebooted-ok"
+      if @node.file_exist? rebooted_file
+        Rails.logger.info("Node was already rebooted after the package upgrade.")
+        return true
+      end
+
       ssh_status = @node.ssh_cmd("/sbin/reboot")
       if ssh_status[0] != 200
         raise_upgrade_error("Failed to reboot the machine. Could not ssh.")
@@ -109,6 +115,7 @@ module Api
 
       wait_for_ssh_state(:down, "reboot")
       wait_for_ssh_state(:up, "come up")
+      @node.run_ssh_cmd("touch #{rebooted_file}")
     end
 
     def upgraded?
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -118,10 +118,6 @@ module Api
       @node.run_ssh_cmd("touch #{rebooted_file}")
     end
 
-    def upgraded?
-      @node.file_exist? "/var/lib/crowbar/upgrade/node-upgraded-ok"
-    end
-
     # Do the complete package upgrade of one node
     def upgrade
       prepare_repositories
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -365,7 +365,7 @@ module Api
       # other_node_name argument is the name of any other node in the same cluster
       def upgrade_first_cluster_node(node, other_node)
         node_api = Api::Node.new node.name
-        return true if node_api.upgraded?
+        return true if node.upgraded?
         other_node_api = Api::Node.new other_node.name
         node_api.save_node_state("controller")
         save_upgrade_state("Starting the upgrade of node #{node.name}")
@@ -393,7 +393,7 @@ module Api
       def upgrade_next_cluster_node(node, founder)
         evacuate_network_node(founder, node["hostname"], true)
         node_api = Api::Node.new node.name
-        return true if node_api.upgraded?
+        return true if node.upgraded?
         node_api.save_node_state("controller")
         save_upgrade_state("Starting the upgrade of node #{node.name}")
 
@@ -426,32 +426,59 @@ module Api
         )
         if drbd_nodes.empty?
           save_upgrade_state("There's no DRBD-based node in cluster #{cluster}")
-          return true
+          return
         end
 
         # First node to upgrade is DRBD slave. There might be more resources using DRBD backend
         # but the Master/Slave distribution might be different among them.
         # Therefore, we decide only by looking at the first DRBD resource we find in the cluster.
-        drbd_slave = nil
-        drbd_master = nil
+        #
+        # But we have to make sure that when this method is invoked for a second time
+        # (probably after previous failure), same order of nodes is selected as in the first case.
+        # Looking at the DRBD state could be problematic, because master and slave are normally
+        # switched during the upgrade. So in such case we have to look at the data we have about
+        # the upgraded nodes.
+        first = nil
+        second = nil
 
-        drbd_nodes.each do |drbd_node|
-          cmd = "LANG=C crm resource status ms-drbd-{postgresql,rabbitmq}\
-          | sort | head -n 2 | grep \\$(hostname) | grep -q Master"
-          out = drbd_node.run_ssh_cmd(cmd)
-          if out[:exit_code].zero?
-            drbd_master = drbd_node
-          else
-            drbd_slave = drbd_node
+        nodes_processed = drbd_nodes.select(&:upgraded?)
+        if nodes_processed.size == drbd_nodes.size
+          save_upgrade_state("All nodes in cluster #{cluster} have already been upgraded.")
+          return
+        end
+
+        # if no node is fully upgraded already, check if the upgrade was started on any of the nodes
+        if nodes_processed.empty?
+          nodes_processed = drbd_nodes.select(&:upgrading?)
+        end
+
+        if nodes_processed.empty?
+          # No DRBD node upgrade has started yet, so let's pick the first/second as slave/master
+          drbd_nodes.each do |drbd_node|
+            cmd = "LANG=C crm resource status ms-drbd-{postgresql,rabbitmq}\
+            | sort | head -n 2 | grep \\$(hostname) | grep -q Master"
+            out = drbd_node.run_ssh_cmd(cmd)
+            if out[:exit_code].zero?
+              second = drbd_node
+            else
+              # this is DRBD slave
+              first = drbd_node
+            end
           end
+        else
+          # If one node is already fully upgraded, we can continue with the other one.
+          # If one node is being upgraded (and we know none is fully upgraded),
+          # we can continue with that one as it is still the first node.
+          first = nodes_processed.first # there is only one item here anyway
+          second = drbd_nodes.detect { |n| n.name != first.name }
         end
 
-        if drbd_slave.nil? || drbd_master.nil?
-          raise_upgrade_error("Unable to detect drb master and/or slave nodes")
+        if first.nil? || second.nil?
+          raise_upgrade_error("Unable to detect DRBD master and/or slave nodes")
         end
 
-        upgrade_first_cluster_node drbd_slave, drbd_master
-        upgrade_next_cluster_node drbd_master, drbd_slave
+        upgrade_first_cluster_node first, second
+        upgrade_next_cluster_node second, first
 
         save_upgrade_state("Nodes in DRBD-based cluster successfully upgraded")
       end
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -343,13 +343,11 @@ module Api
         drbd_nodes = ::Node.find("drbd_rsc:*")
         return upgrade_drbd_clusters unless drbd_nodes.empty?
 
-        founder = ::Node.find(
-          "state:crowbar_upgrade AND pacemaker_founder:true"
-        ).first
+        founder = ::Node.find("pacemaker_founder:true").first
         cluster_env = founder[:pacemaker][:config][:environment]
 
         non_founder = ::Node.find(
-          "state:crowbar_upgrade AND pacemaker_founder:false AND " \
+          "pacemaker_founder:false AND " \
           "pacemaker_config_environment:#{cluster_env}"
         ).first
 
@@ -357,7 +355,7 @@ module Api
 
         # upgrade the rest of nodes in the same cluster
         ::Node.find(
-          "state:crowbar_upgrade AND pacemaker_config_environment:#{cluster_env}"
+          "pacemaker_config_environment:#{cluster_env}"
         ).each do |node|
           upgrade_next_cluster_node node.name, founder.name
         end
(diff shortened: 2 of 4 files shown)


jsuchome:

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -356,21 +356,19 @@ def upgrade_controller_nodes
...
 
-        founder = NodeObject.find(
-          "state:crowbar_upgrade AND pacemaker_founder:true"
-        ).first
+        founder = NodeObject.find("pacemaker_founder:true").first
         cluster_env = founder[:pacemaker][:config][:environment]
 
         non_founder = NodeObject.find(
-          "state:crowbar_upgrade AND pacemaker_founder:false AND " \
+          "pacemaker_founder:false AND " \
```
MaximilianMeister: MINOR: you could now combine the 2 lines into one

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -402,24 +399,52 @@ def upgrade_drbd_cluster(cluster)
...
-          | sort | head -n 2 | grep \\$(hostname) | grep -q Master"
-          out = drbd_node.run_ssh_cmd(cmd)
-          if out[:exit_code].zero?
-            drbd_master = drbd_node.name
+        first = ""
+        second = ""
+
+        # But we have to make sure that when this method is invoked for a second time
+        # (probably after previous failure), same order of nodes is selected as in the first case.
+        # Looking at the DRBD state could be problematic, becase master and slave are normally
```
MaximilianMeister: MINOR: s/becase/because

crowbar_framework/app/models/api/upgrade.rb
jsuchome: No, it's still above 80 lines :-(

crowbar_framework/app/models/api/node.rb
```diff
@@ -100,6 +100,14 @@ def wait_for_ssh_state(desired_state, action)
 
     # Reboot the node and wait until it comes back online
     def reboot_and_wait
+      target_platform = @node.target_platform
+      version = target_platform.split("-").last
+      out = @node.run_ssh_cmd("grep #{version} /etc/os-release")
```
Itxaka: minor nitpick, if you dont care about the output, just use `-q` switch for grep

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -356,21 +356,19 @@ def upgrade_controller_nodes
...
+        founder = NodeObject.find("pacemaker_founder:true").first
         cluster_env = founder[:pacemaker][:config][:environment]
 
         non_founder = NodeObject.find(
-          "state:crowbar_upgrade AND pacemaker_founder:false AND " \
+          "pacemaker_founder:false AND " \
           "pacemaker_config_environment:#{cluster_env}"
         ).first
 
         return false unless upgrade_first_cluster_node founder.name, non_founder.name
```
rhafer: if the founder is in state ```ready``` here already, it indicates that everything including ```node_api.join_and_chef``` succeeded. Shouldn't we just skip those steps then in ```upgrade_first_cluster_node```?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -356,21 +356,19 @@ def upgrade_controller_nodes
...
 
         # upgrade the rest of nodes in the same cluster
         NodeObject.find(
-          "state:crowbar_upgrade AND pacemaker_config_environment:#{cluster_env}"
+          "pacemaker_config_environment:#{cluster_env}"
         ).each do |node|
           return false unless upgrade_next_cluster_node node.name
```
rhafer: Same here, if the state of ```node``` is already ```ready``` it seem a large part of ```upgrade_next_cluster_node``` can be skipped.

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Yes, there's a check for `node_api.upgraded?` at the start of  `upgrade_first_cluster_node`

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Yes, there's a check for `node_api.upgraded?` at the start of `upgrade_next_cluster_node`

crowbar_framework/app/models/api/upgrade.rb
rhafer: Sure, but that's only set after all step of the upgrade where completed. There are case where the node can be in ```ready``` but ```upgraded?``` still returns false. E.g. when ```disable_pre_upgrade_attribute_for``` failed in ```upgrade_next_cluster_node```

jsuchome: OK, I can check the state of node, that's true. So let's wait until https://github.com/crowbar/crowbar-core/pull/935 is merged

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -402,24 +399,52 @@ def upgrade_drbd_cluster(cluster)
...
+        first = ""
+        second = ""
+
+        # But we have to make sure that when this method is invoked for a second time
+        # (probably after previous failure), same order of nodes is selected as in the first case.
+        # Looking at the DRBD state could be problematic, because master and slave are normally
+        # switched during the upgrade. So let's look at the data we have about the upgraded nodes.
+        status = ::Crowbar::UpgradeStatus.new
+        # current_node is only set when the upgrade of a node starts, and is empty at the start.
+        current_node = status.progress[:current_node] || {}
```
jsuchome: There might be a problem here, as just discussed with @rhafer : this `current_node` might not belong to the DRBD cluster we're trying to upgrade. We should check for the nodes state at the beggining of the method.

crowbar_framework/app/models/api/node.rb
```diff
@@ -131,7 +140,10 @@ def upgrade
         return false
       end
 
-      reboot_and_wait
+      unless reboot_and_wait
+        save_error_state("Error while rebooting the node")
+        return false
```
MaximilianMeister: the return can be dropped if we merge #960 as `save_error_state` will raise an exception and not reach that point ever

jsuchome: True. This one is waiting for merge of #935 anyway (so now we can add #960 to the list)

crowbar_framework/app/models/api/upgrade.rb
jsuchome: should be solved by that check for `all_upgraded`

jsuchome:
@rhafer I think your comments were addressed

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -341,15 +341,17 @@ def upgrade_first_cluster_node(node, other_node)
...
-        node_api = Api::Node.new node.name
         return true if node.upgraded?
+        node_api = Api::Node.new node.name
         node_api.save_node_state("controller")
-        save_upgrade_state("Starting the upgrade of node #{node.name}")
 
-        node_api.upgrade
-        node_api.post_upgrade
-        node_api.join_and_chef
+        unless node.state == "ready"
```
rhafer: Nitpick. There is ```node.ready?``` :)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -401,29 +401,72 @@ def upgrade_drbd_cluster(cluster)
...
+              second = drbd_node
+            else
+              # this is DRBD slave
+              first = drbd_node
+            end
+          end
+        else
+          # In this branch, we know that upgrade of DRBD nodes was already started but failed
+          # and the method was executed again. Let's find out which node went first/second initialy
+          # and stick to that order.
```
rhafer: Somehow I think we should add a big warning here (and maybe in the commit message as well) that this part of the code (i.e. the following ~15 lines) heavily depend on the fact that the drbd cluster is the first one to be upgraded. I think for now it's ok to depend on that, but we need to aware of it.

jsuchome: Done

jsuchome:
@rhafer I've enhanced the comment, and also slightely modified the logic, see usage of `drbd_names` - if it contains a node from other cluster, we do not mind.

rhafer:
@jsuchome The longer I look at that drbd commit, the more I think we could make our lives much easier by setting an attribute on the node that indicates whether that node is currently being upgraded (e.g. as the first thing in ```node.upgrade```).
We already have the indication of a successfully upgraded node via the "node-upgraded-ok" file (It might makes sense to switch that to a node attribute as well, currently nothing seems be requiring that file on the node directly).

If we had something like ```node.upgrading?``` I believe the ```upgrade_drbd_cluster``` could look a lot simpler and we would no longer rely on the side effects that require drbd to be upgraded first.

If both of the drbd nodes have ```node.upgraded?``` we'd not need to do anything. If only one node has ```node.upgraded?``` that would become "first". If no node has ```node.upgraded?``` we'd need to check ```node.upgrading?``` if there's a node that has it (there should be either one or zero) that would be "first" otherwise we use the old logic. Am I missing anything? Do you think that makes sense?

The currents code logic is probably not too different from that, but I believe the code would be better readable with such a change.

jsuchome:
Yes, it makes sense. I do not think the code would be that much different, the main point would be replacing of that ugly check for `upgraded_nodes.zero?` which really breaks when there's different order of clusters.

crowbar_framework/app/models/node.rb
```diff
@@ -856,6 +856,21 @@ def upgraded?
...
+
+  def set_upgraded
+    run_ssh_cmd(
+      "touch /var/lib/crowbar/upgrade/node-upgraded-ok && " \
+      "rm -f /var/lib/crowbar/upgrade/node-upgrading"
+    )
+  end
+
+  def set_upgrading
+    run_ssh_cmd("touch /var/lib/crowbar/upgrade/node-upgrading")
```
rhafer: What's the reason for using files instead a node attribute?

crowbar_framework/app/models/node.rb
```diff
@@ -850,6 +850,27 @@ def file_exist?(file)
...
```
rhafer: What's a reason for using files for the upgrade states? Instead of a node attribute?

jsuchome: I'm trying to avoid `node.save`, don't you think it is slower than ssh command? (Also, I would go directly for attributes, but as we are already using files during the upgrade, this feels like consistent with it)

rhafer: > I'm trying to avoid node.save, don't you think it is slower than ssh command? 

I know that chef's db is slow, but do you really think it's that slow? Using the database is certainly cleaner.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -417,32 +417,63 @@ def upgrade_drbd_cluster(cluster)
...
+          node_upgraded = drbd_node.upgraded?
+          upgrading = drbd_node if drbd_node.upgrading?
+          first = drbd_node if node_upgraded
+          all_upgraded &&= node_upgraded
+        end
+        if all_upgraded
+          save_upgrade_state("All nodes in cluster #{cluster} have already been upgraded.")
+          return
+        end
+
```
rhafer: Couldn't the selection be simplified to something like this?

```
nodes_processed = drbd_nodes.select{ |n| n.upgraded? }
if nodes_processed.size == drbd_nodes.size
    # All cluster nodes already upgraded we're done
    ---> everything upgraded return
end

# if no node is fully upgraded already, check if the upgrade was started on anyof the nodes
nodes_processed = drbd_nodes.select{ |n| n.upgrading? } if nodes_processed.empty?

if !nodes_processed.empty?
    first = nodes_processed[0]
    second = drbd_nodes.select{ |n| n.name != first.name }
else
   ---> set first and second based on the master/slave state as previously
```

I'm not sure if ```nodes_processed``` is the best naming here. But ```nodes_upgraded_or_upgrading```  seemed a bit long :)

jsuchome: ok

jsuchome:
@rhafer done

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -417,32 +417,59 @@ def upgrade_drbd_cluster(cluster)
...
 
-        if drbd_slave.nil? || drbd_master.nil?
-          raise_error_state("Unable to detect drb master and/or slave nodes")
+        if first.nil? || second.nil?
+          raise_error_state("Unable to detect DRBD master and/or slave nodes")
         end
 
-        upgrade_first_cluster_node drbd_slave, drbd_master
-        upgrade_next_cluster_node drbd_master, drbd_slave
+        upgrade_first_cluster_node first, second
```
rhafer: Is this better? ```upgrade_first_cluster_node(first, second) unless first.upgraded?```

jsuchome: Why? upgraded? is the first check inside `upgrade_first_cluster_node`

jsuchome:
I've adapted DRBD commit message

crowbar_framework/app/models/api/upgrade.rb
rhafer: Ok. Didn't check there. I guess then it's not worth it.

jsuchome:
@Itxaka did you retract your review, or was it removed automatically?

Itxaka:
@jsuchome retracted as there were many changes post-review and wanted to go over it again

crowbar_framework/app/models/node.rb
```diff
@@ -850,6 +850,22 @@ def file_exist?(file)
...
+    return false unless upgrade_state == "upgraded"
+    Rails.logger.info("Node #{@node.name} was already upgraded.")
+    true
+  end
+
+  def upgrading?
+    @node["crowbar_wall"]["node_upgrade_state"] == "upgrading"
+  end
+
+  def upgrade_state=(state)
```
Itxaka: Is the `=` in the function name expected? I would expect that not, but this is ruby so who knows :)

jsuchome: It is, look at other (old) methods in `node.rb`.
It was actually Hound which forced me to do it this way, I had `set_upgrade_state` method before.

Itxaka: woah

jsuchome:
rebased

jsuchome:
Looks like I had to extract one commit into https://github.com/crowbar/crowbar-core/pull/976 due to some incorrect merges...

PR #952, opened 2017-01-04
Create event hook infrastructure

vuntz:
There are several cases where we have events that should trigger some activity in some other part of Crowbar. For instance:

 - when the public name of a node is saved, it may impact the endpoint of an OpenStack service
 - when the public name of the VIP of haproxy is changed, it impacts the endpoint of OpenStack services
 - when the keystone proposal is applied, we may want to reapply all proposals that depend on keystone

What we need for this is the ability to notify about the events in the rails application and then to dispatch the notifications to hooks listen listening to them that will decide if some action should be triggered.

The main reason we didn't have this in the past is that we likely don't want to do that in the foreground of the rails application. But now that we have delayed_job, we can send the notifications and run the hooks in the background.

In this commit, we add the simple infrastructure about notifications and hooks:

 - the events are defined with a name and a hash that contains the details of the event. The structure of the hash depends on the event.

 - we have two events that are generated and notified: node_changed (when a node is saved and some specific attributes have been changed) and proposal_applied (when a proposal is successfully applied).

 - a simple dispatcher exists that simply connects the hooks to the events.

 - the hooks only exist for service objects for the time being; a service object simply needs to have a event_hook method to register the hook, and will need to filter for the events it cares about. The signature of event_hook is as follows: def event_hook(role, event, details)

It could be argued that the hooks should be registered for some specific events (hence moving the filter to the event dispatcher), but it's not worth the complexity for now.

We could add many more events but for the time being, we only add the events that we know are useful.

### DIFF
--- a/crowbar_framework/lib/crowbar/event_dispatcher.rb
+++ b/crowbar_framework/lib/crowbar/event_dispatcher.rb
@@ -1,0 +1,48 @@
+#
+# Copyright 2017, SUSE
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+module Crowbar
+  class EventDispatcher
+    class << self
+      def trigger_hooks(event, details)
+        BarclampCatalog.barclamps.keys.each do |barclamp|
+          begin
+            cls = ServiceObject.get_service(barclamp)
+          rescue NameError
+            # catalog may contain barclamps which don't have services
+            next
+          end
+
+          next unless cls.method_defined?(:event_hook)
+
+          service = cls.new(Rails.logger)
+
+          proposals = Proposal.where(barclamp: barclamp)
+          proposals.each do |proposal|
+            role = proposal.role
+            next if role.nil?
+            begin
+              service.event_hook(role, event, details)
+            rescue StandardError => e
+              raise "Error while executing event hook of #{barclamp} for #{event}: #{e.message}"
+            end
+          end
+        end
+      end
+      handle_asynchronously :trigger_hooks
+    end
+  end
+end
--- a/crowbar_framework/app/models/node.rb
+++ b/crowbar_framework/app/models/node.rb
@@ -16,6 +16,7 @@
 #
 
 require "chef/mixin/deep_merge"
+require "set"
 require "timeout"
 require "open3"
 
@@ -37,6 +38,7 @@ class Node < ChefObject
     end
     # deep clone of @role.default_attributes, used when saving node
     @attrs_last_saved = @role.default_attributes.deep_dup
+    @attributes_changed = Set.new
     @node = node
   end
 
@@ -67,7 +69,10 @@ class Node < ChefObject
   def availability_zone=(value)
     @node["crowbar_wall"] ||= {}
     @node["crowbar_wall"]["openstack"] ||= {}
-    @node["crowbar_wall"]["openstack"]["availability_zone"] = value
+    if @node["crowbar_wall"]["openstack"]["availability_zone"] != value
+      @node["crowbar_wall"]["openstack"]["availability_zone"] = value
+      @attributes_changed.add("availability_zone")
+    end
   end
 
   def intended_role
@@ -257,6 +262,7 @@ class Node < ChefObject
   def update_public_name(value)
     unless value.nil?
       crowbar["crowbar"]["public_name"] = value
+      @attributes_changed.add("public_name")
     end
   end
 
@@ -619,6 +625,20 @@ class Node < ChefObject
     @attrs_last_saved = @role.default_attributes.deep_dup
 
     Rails.logger.debug("Done saving node: #{@node.name} - #{crowbar_revision}")
+
+    unless @attributes_changed.empty?
+      Rails.logger.debug("Notifying about saved node: #{@node.name}")
+
+      details = {
+        node: @node.name,
+        attributes: @attributes_changed
+      }
+      Crowbar::EventDispatcher.trigger_hooks(:node_changed, details)
+
+      @attributes_changed = Set.new
+
+      Rails.logger.debug("Done notifying about saved node: #{@node.name}")
+    end
   end
 
   def destroy
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1372,6 +1372,13 @@ class ServiceObject
     proposal.save unless roles_to_remove.empty?
 
     update_proposal_status(inst, "success", "")
+
+    details = {
+      barclamp: @bc_name,
+      instance: inst
+    }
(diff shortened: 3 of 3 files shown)


crowbar_framework/app/models/node_object.rb
```diff
@@ -259,6 +259,7 @@ def initialize(node)
     end
     # deep clone of @role.default_attributes, used when saving node
     @attrs_last_saved = @role.default_attributes.deep_dup
+    @notify_on_save = {}
```
matelakat: This variable really holds the notifications to be sent on the next save, if understand it correctly. Also looking at the point where the notifications are sent out, it seems, that we are not using the value in the hash, only the keys. I think a better name for this variable might help, and also making it an array/set would also help. I would name it something like: `on_save_events`. Thus the protocol would look like: the client adds/queues up items to the `on_save_events` collection, and those are being sent and flushed on save.

crowbar_framework/app/models/node_object.rb
```diff
@@ -479,6 +483,7 @@ def validate_public_name(value, unique_check = true)
   def update_public_name(value)
     unless value.nil?
       crowbar["crowbar"]["public_name"] = value
+      @notify_on_save["public_name"] = true
```
matelakat: It might be a stupid question: Are we sure that `update_public_name` is only called whenever the value of it has changed? I'm not seeing the same check-if-changed mechanism as earlier.

crowbar_framework/app/models/node_object.rb
```diff
@@ -835,6 +840,20 @@ def _remove_elements_from_node(old, new, from_node)
     @attrs_last_saved = @role.default_attributes.deep_dup
 
     Rails.logger.debug("Done saving node: #{@node.name} - #{crowbar_revision}")
+
+    unless @notify_on_save.empty?
+      Rails.logger.debug("Notifying about saved node: #{@node.name}")
+
+      details = {
+        node: @node.name,
+        attributes: @notify_on_save.keys
```
matelakat: As I mentioned earlier, we only send the keys, not the values. So we might want to send the values as well, or use a single set as mentioned earlier.

crowbar_framework/app/models/node_object.rb
vuntz: See https://github.com/crowbar/crowbar-core/pull/952/files#diff-496ddf4ffc5694a5fcdddbff8206dc8dR441 (there's also `force_public_name=` where the check is missing, but I think it's fine to still send the event in that case since it's forced).

crowbar_framework/app/models/node_object.rb
vuntz: I can move to a set, sure. FWIW, in ruby, a set is implemented (behind the scenes) exactly the way I did, so won't be too much different but I guess it's slightly more readable.

vuntz:
@matelakat I addressed your comments. I also renamed the small library to EventDispatcher as Notify was too generic...

crowbar_framework/lib/crowbar/event_dispatcher.rb
```diff
@@ -0,0 +1,44 @@
...
+# limitations under the License.
+#
+
+module Crowbar
+  class EventDispatcher
+    class << self
+      def trigger_hooks(event, details)
+        BarclampCatalog.barclamps.keys.each do |barclamp|
+          begin
+            cls = eval("#{barclamp.camelize}Service")
```
aspiers: Calculating the class name and obtaining the class surely belongs in methods in `BarclampCatalog`.  Doing it here sounds like privacy violation / feature envy.

crowbar_framework/lib/crowbar/event_dispatcher.rb
```diff
@@ -0,0 +1,44 @@
...
+
+          proposals = Proposal.where(barclamp: barclamp)
+          proposals.each do |proposal|
+            role = proposal.role
+            next if role.nil?
+            service.event_hook(role, event, details)
+          end
+        end
+      end
+      handle_asynchronously :trigger_hooks
```
aspiers: Should be a blank line before this.

aspiers: Oh, maybe not.  It's kind of like a decorator.

crowbar_framework/lib/crowbar/event_dispatcher.rb
```diff
@@ -0,0 +1,44 @@
...
+
+          next unless cls.method_defined?(:event_hook)
+
+          service = cls.new(Rails.logger)
+
+          proposals = Proposal.where(barclamp: barclamp)
+          proposals.each do |proposal|
+            role = proposal.role
+            next if role.nil?
+            service.event_hook(role, event, details)
```
vuntz: We need some exception catching here.

crowbar_framework/app/models/node_object.rb
```diff
@@ -289,7 +291,10 @@ def availability_zone
   def availability_zone=(value)
     @node["crowbar_wall"] ||= {}
     @node["crowbar_wall"]["openstack"] ||= {}
-    @node["crowbar_wall"]["openstack"]["availability_zone"] = value
+    if @node["crowbar_wall"]["openstack"]["availability_zone"] != value
```
matelakat: @vuntz as I mentioned it in the "eliminate saves" PR earlier, this pattern might worth extracting to a helper, something like: `set_value_and_call_function_if_changed` or I believe ruby must have better ways to express it.

crowbar_framework/lib/crowbar/event_dispatcher.rb
Itxaka: Isnt the way of doing this better with Kernel.const_get ?*


*not a ruby expert

crowbar_framework/app/models/node_object.rb
```diff
@@ -835,6 +841,20 @@ def _remove_elements_from_node(old, new, from_node)
...
+    unless @on_save_events.empty?
+      Rails.logger.debug("Notifying about saved node: #{@node.name}")
+
+      details = {
+        node: @node.name,
+        attributes: @on_save_events
+      }
+      Crowbar::EventDispatcher.trigger_hooks(:node_changed, details)
+
+      @on_save_events = Set.new
```
toabctl: Can this be called in parallel by different puma threads? If so isn't this a race and we could miss save events when cleaning the Set?

vuntz: The node object is not shared among puma threads, so it should be fine. When we create threads ourselves, we should be careful of course and use proper locking mechanisms, but it's not specific to this change.

crowbar_framework/lib/crowbar/event_dispatcher.rb
vuntz: @Itxaka you're right.

vuntz: We actually have `ServiceObject.get_service` exactly for this...

crowbar_framework/app/models/node_object.rb
```diff
@@ -259,6 +260,7 @@ def initialize(node)
     end
     # deep clone of @role.default_attributes, used when saving node
     @attrs_last_saved = @role.default_attributes.deep_dup
+    @attributes_changed_for_event = Set.new
```
aspiers: As discussed could you rename this to `@attributes_changed` and then add a comment explaining what it's used for?

aspiers:
Well, of course the merge conflict and Travis failures have to be fixed first :)

vuntz:
@aspiers rebased

vuntz:
Completed in https://github.com/crowbar/crowbar-core/pull/1308

PR #955, opened 2017-01-05
upgrade: node upgrade cleanup

MaximilianMeister:


jsuchome:
Also, I like saving the error message on error, but don't we end with multiple attempts of ending the step (one from `save_error_state`, one from `nodes`) ?

jsuchome:

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -557,8 +555,11 @@ def save_upgrade_state(message = "")
       end
 
       def save_error_state(message = "")
```
jsuchome: What if we'd just raise an exception (with `message`) from `save_error_state`, catch it in `api/upgrade.rb` (nodes method) and call `end_step` once we catch the exception?

Than we could remove all those checks for `false` when calling methods from `api/node.rb` ...

jsuchome:

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Something like https://github.com/crowbar/crowbar-core/pull/960

MaximilianMeister: i dropped the last commit now, so we can base #960 on top

MaximilianMeister:
rebased

MaximilianMeister:
@rhafer there was a commit that has been removed in the meantime, because it is now covered by one of @jsuchome PR's, I changed the title now

MaximilianMeister:
rebased again

MaximilianMeister:
rebased again

PR #956, opened 2017-01-05
[upgrade] Use block-migration when needed

rhafer:
When nova is deployed without shared storage we need to use
--block-migrate for live migrating instances away from compute nodes.


jsuchome:

PR #958, opened 2017-01-06
upgrade: Check the node state after first chef-client run

jsuchome:


jsuchome:
[Correctly detect problems after crowbar_join run](https://trello.com/c/rtD9mKQX/247-correctly-detect-problems-after-crowbar-join-run)

PR #959, opened 2017-01-06
Make applying proposal slightly faster

vuntz:
When there are many nodes in a proposal, things can start to be slow. Part of it can easily be avoided (by not reloading nodes all the time and skipping useless bits).

This is the safe part of https://github.com/crowbar/crowbar-core/pull/896


vuntz:
Rebased after NodeObject -> Node change.

PR #960, opened 2017-01-06
upgrade: Raise exceptions on node upgrade errors and end the step

jsuchome:
The exception should provide good error message. Once catched,
we must end the upgrade step with failure.

This is how the progress file looks when the error happens in prepare-repositories script: http://pastebin.suse.de/18858

### DIFF
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -24,59 +24,64 @@ module Api
     def execute_and_wait_for_finish(script, seconds)
       Rails.logger.info("Executing #{script} at #{@node.name}...")
       @node.wait_for_script_to_finish(script, seconds)
-      true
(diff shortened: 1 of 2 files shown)


jsuchome:

crowbar_framework/app/models/api/node.rb
```diff
@@ -171,8 +159,8 @@ def save_node_state(role, state = "upgrading")
     end
 
     def save_error_state(message = "")
-      # FIXME: save the error to global status
       Rails.logger.error(message)
+      raise message
```
MaximilianMeister: now that it just logs and raises an exception, the `save_error_state` name is also a bit misleading, as it actually saves in the next method

jsuchome: Good point, I've renamed it

PR #961, opened 2017-01-09
Rename NodeObject to Node

rsalevsky:
This PR got split out of #920 to avoid endless rebases which happens when we do any change to the current NodeObject or call it.

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -80,7 +80,7 @@ module Api
 
       def ceph_status
         {}.tap do |ret|
-          ceph_node = NodeObject.find("roles:ceph-mon AND ceph_config_environment:*").first
+          ceph_node = ::Node.find("roles:ceph-mon AND ceph_config_environment:*").first
           return ret if ceph_node.nil?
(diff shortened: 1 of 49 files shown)


rsalevsky:
Just for the record: I will not fix the hound issues because they got not introduced by this PR.

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -80,7 +80,7 @@ def addons
 
       def ceph_status
         {}.tap do |ret|
-          ceph_node = NodeObject.find("roles:ceph-mon AND ceph_config_environment:*").first
+          ceph_node =  ::Node.find("roles:ceph-mon AND ceph_config_environment:*").first
```
vuntz: You have two spaces before `::Node` here.

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -94,7 +94,7 @@ def ceph_status
       def compute_resources_status
         {}.tap do |ret|
           ["kvm", "xen"].each do |virt|
-            compute_nodes = NodeObject.find("roles:nova-compute-#{virt}")
+            compute_nodes =  ::Node.find("roles:nova-compute-#{virt}")
```
vuntz: And here too.

PR #962, opened 2017-01-09
Backup errors

skazi0:
Currently any error/exception during backup creation propagates up the stack and causes 500 response with not explanation plus traceback in the logs. In the UI this triggers generic error message:
zrzut ekranu z 2017-01-09 14-11-06

With these changes, failure during backup creation will look like this:
zrzut ekranu z 2017-01-09 13-58-49

### DIFF
--- a/crowbar_framework/app/models/api/backup.rb
+++ b/crowbar_framework/app/models/api/backup.rb
@@ -174,6 +174,9 @@ class Api::Backup < ActiveRecord::Base
     self.version = ENV["CROWBAR_VERSION"]
     self.size = path.size
     self.migration_level = ActiveRecord::Migrator.current_version
+  rescue StandardError => e
+    errors.add(:base, I18n.t("backups.index.create_backup_failed", msg: e.message))
+    false
   end
 
   def save_archive
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -373,7 +373,14 @@ class Api::UpgradeController < ApiController
         false,
         admin_backup: @backup.errors.full_messages.first
       )
-      render json: { error: @backup.errors.full_messages.first }, status: :unprocessable_entity
+      render json: {
+        errors: {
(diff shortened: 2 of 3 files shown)


jsuchome:
> failure during backup creation will look like this: ...

... as opposed to?

skazi0:
[(3) UI: Generic error handling (admin backup)](https://trello.com/c/KwAOzuC2/205-3-ui-generic-error-handling-admin-backup)

crowbar_framework/app/models/api/backup.rb
```diff
@@ -174,6 +174,9 @@ def create_archive
     self.version = ENV["CROWBAR_VERSION"]
     self.size = path.size
     self.migration_level = ActiveRecord::Migrator.current_version
+  rescue Exception => e
```
jsuchome: I think hound is fine with `rescue StandardError`

Itxaka:
I think the PR explanation could be better. This standardises the error format from the API so its like the other errors reported by the API, so we dont need to do a lot of manipulation on the UI side to show the same error style across different api calls, right @skazi0 ?

skazi0:
What @Itxaka said is just one part of this. The common format is still not used in all places and one of commits is here to improve this a bit. The other thing is general catching of errors and returning them to clients.

crowbar_framework/app/models/api/backup.rb
skazi0: Yeah, I updated this but I'm not sure if this is still not too wide. @jsuchome, @MaximilianMeister any hints on this?

jsuchome: I think this is what we are checking for other cases

MaximilianMeister:
@skazi0 can you prepare the backport?

skazi0:
@MaximilianMeister yes I'll do it

skazi0:
[Save upgrade errors in consistent format (bsc#1019298)](https://trello.com/c/OujhW7wm/256-save-upgrade-errors-in-consistent-format-bsc-1019298)

PR #963, opened 2017-01-09
upgrade: handle failures on save (bsc#1018110)

MaximilianMeister:
raise a custom exception when save fails, otherwise we might continue
silently

### DIFF
--- a/crowbar_framework/lib/crowbar/error/upgrade_status.rb
+++ b/crowbar_framework/lib/crowbar/error/upgrade_status.rb
@@ -41,5 +41,11 @@ module Crowbar
         super("Step #{step_name} cannot be finished, as it is not running")
       end
     end
+
+    class SaveUpgradeStatusError < StandardError
+      def initialize(msg)
+        super("Exception during saving the status file: #{msg}")
+      end
+    end
   end
 end
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -103,7 +103,8 @@ class Api::UpgradeController < ApiController
       }, status: :unprocessable_entity
     end
   rescue Crowbar::Error::StartStepRunningError,
-         Crowbar::Error::StartStepOrderError => e
+         Crowbar::Error::StartStepOrderError,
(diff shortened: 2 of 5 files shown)


MaximilianMeister:
[(2) S53P2 bsc#1018110 - Upgrade status library save() failures not handled](https://trello.com/c/S2piNMx2/245-2-s53p2-bsc-1018110-upgrade-status-library-save-failures-not-handled)

crowbar_framework/lib/crowbar/error/upgrade_status.rb
```diff
@@ -41,5 +41,11 @@ def initialize(step_name = "")
         super("Step #{step_name} cannot be finished, as it is not running")
       end
     end
+
+    class SaveUpgradeStatusError < StandardError
+      def initialize(msg = "")
```
Itxaka: why give the value an empty default? Isnt the value always necessary?

MaximilianMeister: wow you're quick :) you're right i'll change it

jsuchome:
Could you also adapt the clients?

crowbar_framework/lib/crowbar/error/upgrade_status.rb
```diff
@@ -41,5 +41,11 @@ def initialize(step_name = "")
         super("Step #{step_name} cannot be finished, as it is not running")
       end
     end
+
+    class SaveUpgradeStatusError < StandardError
+      def initialize(msg)
+        super("Exception during saving the status file: #{msg}")
```
jsuchome: Sounds kind of strange. What about *Exception when saving...* ?

MaximilianMeister: hrm, i don't find it that strange. i just took it as it is. btw you introduced it here https://github.com/crowbar/crowbar-core/commit/814b2372120a4237bf1cbb3040230f5fb574fff4 ;-)

jsuchome: Yes, that looks like me. If you think it's correct English, just keep it.

PR #966, opened 2017-01-10
upgrade: Explicitely start nova-compute service after the upgrade

jsuchome:
This is temporary workaround for bsc#1016302 (service might not
be running after initial chef-client run). It should be removed
once the proper patch is released.


jsuchome:

jsuchome:
After first tests with https://review.openstack.org/#/c/416928/4 , it seems that situation might be even more complicated so I'd prefer to merge this temporary solution now.

PR #968, opened 2017-01-10
Node upgrade fixes

jsuchome:


jsuchome:
@rhafer yes, I've already removed it from #950

PR #969, opened 2017-01-10
upgrade: Raise and catch upgrade specific Exceptions during upgrade

jsuchome:
Also, end the upgrade step for non-upgrade error, so it is not
stuck as running.

This should fix  https://bugzilla.suse.com/show_bug.cgi?id=1019179

Requires https://github.com/crowbar/crowbar-core/pull/968


jsuchome:
[https://bugzilla.suse.com/show_bug.cgi?id=1019179](https://trello.com/c/mt2RTWk1/255-https-bugzilla-suse-com-show-bug-cgi-id-1019179)

PR #971, opened 2017-01-11
upgrade: Handle asynchronous case of (bsc#1018110)

MaximilianMeister:
mark an error during save and perform a query during the status
api call to see if an error exists

### not tested yet

@skazi0 this would need to be tested in combination with the UI polling i guess.

### DIFF
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -16,6 +16,7 @@
 
 class Api::UpgradeController < ApiController
   skip_before_filter :upgrade
+  before_filter :async_save_fail
 
   api :GET, "/api/upgrade", "Show the Upgrade progress"
   header "Accept", "application/vnd.crowbar.v2.0+json", required: true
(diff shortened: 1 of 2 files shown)


crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -180,6 +180,20 @@ def save
       true
     rescue StandardError => e
       @logger.error("Exception during saving the status file: #{e.message}")
+      # clear old errors to make sure there is only one type of that error present
+      # this is necessary to query for such an error through the :async_save_fail before_filter
+      # in the UpgradeController
+      error = Api::Error.find_by(
```
jsuchome: Do we have access to this from all usages of progress library (e.g. from crowbar-init)?

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -540,4 +541,17 @@ def dummy_crowbar_init_api_upgrade_connect
...
   end
+
+  def async_save_fail
+    error = Api::Error.find_by(error: "Crowbar::Error::SaveUpgradeStatusError")
+    return unless error
+    render json: {
+      errors: {
+        error.caller => {
+          data: error.message,
+          help: I18n.t("api.upgrade.#{error.caller}.help.default")
```
jsuchome: Hm, default message for a step will not be helpful here, I'm afraid. I think we need to advice to do something about the progress file.

crowbar_framework/lib/crowbar/upgrade_status.rb
MaximilianMeister: Good point, no we don't.

MaximilianMeister:
we might reopen this at a later point. but we decided not to continue the effort for now, which means leaving the async cases as they are

PR #972, opened 2017-01-11
upgrade: Save the upgrade errors in consistent format

jsuchome:
When the synchronous API call fails, it saves error in a form
of 'data' and 'help' touple. We need to do it in all other cases
as well.

This should be a fix for  https://bugzilla.suse.com/show_bug.cgi?id=1019298

Here's the result of new error output, cli: http://pastebin.nue.suse.com/19090/src and json: http://pastebin.nue.suse.com/19094/src

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -173,9 +173,13 @@ module Api
               missing_repo_arch = v[:repos].keys.first.to_sym
               v[:repos][missing_repo_arch][:missing]
             end.flatten.compact.join(", ")
-            upgrade_status.end_step(
+            ::Crowbar::UpgradeStatus.new.end_step(
               false,
-              adminrepocheck: "Missing repositories: #{missing_repos}"
+              adminrepocheck: {
+                data: "Missing repositories: #{missing_repos}",
+                help: "Fix the repository setup for the Admin server before " \
+                  "you continue with the upgrade"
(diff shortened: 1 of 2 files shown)


jsuchome:
[https://bugzilla.suse.com/show_bug.cgi?id=1019298](https://trello.com/c/OujhW7wm/256-https-bugzilla-suse-com-show-bug-cgi-id-1019298)

jsuchome:
@skazi0 does the example output fit your needs?

skazi0:
@jsuchome looks correct. This will handle errors from the background job, right? Not just the triggering call.

jsuchome:
@skazi0 this only changes the way the errors are saved in the progress file. And the status call (invoked for background API's) will return it.

jsuchome:
@skazi0 yes, I plan to extend it (just like @MaximilianMeister suggested), but first I wanted to know if you are able to test it with some real code...

jsuchome:
Note: I'm actually not sure if those zypper errors during adminrepocheck are correctly saved to status file...

skazi0:
If you mean "in prechecks" then these should be fixed in #987

jsuchome:
> If you mean "in prechecks" then these should be fixed in #987

No, I mean adminrepocheck, see https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/api/upgrade.rb#L133 and others

PR #973, opened 2017-01-11
upgrade: Remove -failed indication right before starting the script (again)

jsuchome:
The scripts are deleting this file when they start. But if something
would slow them too much, crowbar could detect the presence of -failed file
from previous failed execution and assume the script has failed.


jsuchome:
> I guess it means we can remove ...

Yes. From all the scripts actually.

PR #975, opened 2017-01-11
Fixes for compute nodes upgrade

jsuchome:


crowbar_framework/app/models/node.rb
```diff
@@ -302,6 +302,13 @@ def ready?
     state === "ready"
   end
 
+  # Right after node upgrade, chef-client is not yet running thus the state
+  # might not be updated for some time.
+  def ready_after_upgrade?
+    s = state
+    s == "ready" || s == "noupdate"
```
rhafer: I guess ```["ready", "noupdate"].include? state``` would be a bit more rubyish.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -579,22 +579,23 @@ def upgrade_compute_nodes(virt)
           next if n.upgraded?
           node_api = Api::Node.new n.name
           node_api.save_node_state("compute")
+          hostname = n[:hostname] || n.name
```
rhafer: I think a small comment why ```hostname``` instead of ```name``` is preferrable here (or on the commit message), would help.

crowbar_framework/app/models/node.rb
jsuchome: I did not want to add `state` there, as it is a method, no need to invoke it twice

rhafer: I am pretty sure that doesn't happen when you use ```include?``` as the return value of the ```state``` method will be passed as parameter to the ```include?``` function.

crowbar_framework/app/models/node.rb
```diff
@@ -302,6 +302,13 @@ def ready?
     state === "ready"
   end
 
+  # Right after node upgrade, chef-client is not yet running thus the state
+  # might not be updated for some time.
+  def ready_after_upgrade?
+    s = state
```
rhafer: You don't really need this. ```state``` will be called just once.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -579,20 +579,23 @@ def upgrade_compute_nodes(virt)
           next if n.upgraded?
           node_api = Api::Node.new n.name
           node_api.save_node_state("compute")
-          unless n.ready?
-            live_evacuate_compute_node(controller, n.name)
+          hostname = n[:hostname] || n.name
```
rhafer: > Nova knows nodes by the hostnames, and does not find a node when FQDN is provided.

So if the FQDN does not work for nova. Why should we fall back to it at all?

jsuchome:
@rhafer reworked again

PR #977, opened 2017-01-11
network: Enable rx/tx offloading by default (bsc#1015266)

vuntz:
This was disabled for bsc#910298 which mentioned a "wellknown vlan
offloading bug". However, with the newer kernel that we have now, this
should hopefully be better.

https://bugzilla.suse.com/show_bug.cgi?id=1015266

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -376,8 +376,8 @@ Nic.nics.each do |nic|
   end
 
   unless nic.kind_of?(Nic::Vlan) or nic.kind_of?(Nic::Bond)
-    nic.rx_offloading = node["network"]["enable_rx_offloading"] || false
-    nic.tx_offloading = node["network"]["enable_tx_offloading"] || false
+    nic.rx_offloading = node["network"]["enable_rx_offloading"] || true
+    nic.tx_offloading = node["network"]["enable_tx_offloading"] || true
   end
 
   if ifs[nic.name]["mtu"]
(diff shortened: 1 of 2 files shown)


vuntz:
@dirkmueller you were the one mentioning the well-known vlan offloading bug. Do you have more details about it so we can check if it's fixed?

dirkmueller:
Afaik the openvswitch/vlan with virtio offloading problems are still not fixed .. i'd be happy to flip the default if sp2 is good

rhafer:
> Afaik the openvswitch/vlan with virtio offloading problems are still not fixed 

@dirkmueller Which problem is that? Do you have a bug id at hand? I found these offloading related issues additional to what is already mentioned in the commit. None of the really matches "openvswitch/vlan with virtio" but maybe I am missing something:

https://bugzilla.suse.com/show_bug.cgi?id=884706
https://bugzilla.suse.com/show_bug.cgi?id=853370
https://bugzilla.suse.com/show_bug.cgi?id=854093

Appartently at least the first two are fixed. Not sure about the last one.

abel-navarro:
There were also problems with [TSO offloading for nova-docker](https://bugs.launchpad.net/nova-docker/+bug/1519357). I'm reluctant to enable offloading by default.

vuntz:
> There were also problems with TSO offloading for nova-docker. I'm reluctant to enable offloading by default.

We dropped nova-docker support, so that helps here ;-)

vuntz:
@dirkmueller it's all green (had to re-trigger the HA job, though); how do you feel about taking this now? I agree it's slightly risky, but at least it's an easy revert if we realize late it causes issues.

dirkmueller:
Lets give it a whirl in M8

PR #978, opened 2017-01-12
upgrade: During the retry of nodes upgrade, stick to the original order

jsuchome:
Do not process the nodes that have been already upgraded.
If there's a node for which we already started the upgrade, continue with this
one next time we retry the upgrade method.

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -344,21 +344,30 @@ module Api
         return upgrade_drbd_clusters unless drbd_nodes.empty?
 
         founder = ::Node.find("pacemaker_founder:true").first
-        cluster_env = founder[:pacemaker][:config][:environment]
+        cluster = founder[:pacemaker][:config][:environment]
 
-        non_founder = ::Node.find(
+        non_founder_nodes = ::Node.find(
           "pacemaker_founder:false AND " \
-          "pacemaker_config_environment:#{cluster_env}"
-        ).first
+          "pacemaker_config_environment:#{cluster}"
+        )
+        non_founder_nodes.select! { |n| !n.upgraded? }
 
-        upgrade_first_cluster_node founder, non_founder
+        if founder.upgraded? && non_founder_nodes.empty?
+          save_upgrade_state("All nodes in cluster #{cluster} have already been upgraded.")
+          return
+        end
+
+        upgrade_first_cluster_node founder, non_founder_nodes.first
+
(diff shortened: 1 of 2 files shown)


jsuchome:
[restart upgrade of compute nodes from the point where it has failed](https://trello.com/c/oFBClp7T/258-restart-upgrade-of-compute-nodes-from-the-point-where-it-has-failed)

jsuchome:
FIXME: while I'm in it, order of non-drbd controller nodes should be fixed as well

crowbar_framework/spec/models/api/upgrade_spec.rb
```diff
@@ -472,6 +473,58 @@
       expect(subject.class.nodes).to be_a(Delayed::Backend::ActiveRecord::Job)
     end
 
+    it "during the upgrade of controller nodes, detect that they are upgraded" do
```
jsuchome: Not true any more

jsuchome:

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -344,21 +344,30 @@ def upgrade_controller_nodes
...
+        cluster = founder[:pacemaker][:config][:environment]
 
-        non_founder = ::Node.find(
+        non_founder_nodes = ::Node.find(
           "pacemaker_founder:false AND " \
-          "pacemaker_config_environment:#{cluster_env}"
-        ).first
+          "pacemaker_config_environment:#{cluster}"
+        )
+        non_founder_nodes.select! { |n| !n.upgraded? }
```
abel-navarro: you are changing the `non_founder_nodes` set with this call. I'd prefer the node sets are immutable. In practice that would mean to have a new variable for `non_upgraded_nodes`. It will work the same way, but semantically I find it makes more sense.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -546,6 +555,15 @@ def upgrade_compute_nodes(virt)
           return
         end
 
+        # remove upgraded compute nodes
+        compute_nodes.select! { |n| !n.upgraded? }
```
abel-navarro: same comment as with `non_founder_nodes`. I'd prefer you assign the new set to a new variable, like `non_upgraded_compute_nodes`.

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Why? I found this one of few cases where changing the actual array makes sense.

PR #979, opened 2017-01-12
upgrade: Catch status exceptions during prechecks on cloud7

MaximilianMeister:
if the upgrade status is called at a later step we don't want to
get a 500 error from the API, but instead render the usual json
error response


jsuchome:
> if the upgrade status

You mean upgrade status or *upgrade prechecks*?

MaximilianMeister:
> You mean upgrade status or upgrade prechecks?

oops, prechecks i meant

PR #981, opened 2017-01-12
upgrade: rename internal steps to match crowbarctl subcommands

MaximilianMeister:
this improves the UX as the user can read from the status more easily
which step comes next


MaximilianMeister:
[rename crowbarctl upgrade command arguments to match the steps provided by the progress library](https://trello.com/c/g9Tk7kzj/257-rename-crowbarctl-upgrade-command-arguments-to-match-the-steps-provided-by-the-progress-library)

MaximilianMeister:
with the latest commit there is now a cross dependency to https://github.com/crowbar/crowbar-client/pull/138

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -22,32 +22,32 @@ class Api::UpgradeController < ApiController
...
+      "repocheck_crowbar": {
         "status": "passed",
         "errors": {}
       },
-      "admin_upgrade": {
+      "admin": {
         "status": "failed",
         "errors": {
-          "admin_upgrade": {
+          "admin": {
```
skazi0: The error keys don't need to match the steps. `admin` seems too generic as an error code. Same for other error codes below.

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -22,32 +22,32 @@ class Api::UpgradeController < ApiController
...
         "status": "passed",
         "errors": {}
       },
-      "admin_repo_checks": {
+      "repocheck_crowbar": {
         "status": "passed",
         "errors": {}
       },
-      "admin_upgrade": {
+      "admin": {
```
skazi0: Why is this `admin` not `crowbar` like in the crowbarctl?

skazi0: I see you propose to rename `crowbar` to `admin` also in crowbarctl. If so, then `backup_crowbar` is not renamed here.

MaximilianMeister: it will be renamed, see https://github.com/crowbar/crowbar-client/pull/138

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: i'd prefer to keep all of them in sync, just imagine an `upgrade_` prefix for all of the steps and it makes more sense. explicitly saying `upgrade` is not necessary IMO as we are in the upgrade by context already

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: > I see you propose to rename crowbar to admin also in crowbarctl. If so, then why backup_crowbar is not renamed?

not sure if i fully understand the question. but the logic behind `backup_crowbar` is `crowbarctl upgrade backup crowbar`

skazi0: Before it was `crowbarctl upgrade crowbar` and now it's `crowbarctl upgrade admin`. Why not do the same for backup?

crowbar_framework/app/controllers/api/upgrade_controller.rb
skazi0: OK, this will require some additional changes on the UI side (mostly in translations).

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: ah now i get it... makes sense somehow.
but on the other side having crowbar and openstack in the backup subcommand also makes sense, plus there is also node data in there, not just admin related stuff...

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: should i address this within this PR? then i need to know more or less where to make the changes

jsuchome:
Looks good after initial testing. Could you please prepare Cloud6 version, so we can merge them closely together (without having upgrade broken for a longer time)?

skazi0:
OK, lets merge this whole thing and see how it works.

PR #983, opened 2017-01-12
Add ping api endpoint for crowbar agent

felixsch:
To add the information when a node was seen last time a endpoint
is needed where the nodes can ping that they are alive.

This endpoint must be available without authentification!

__This pull request is based on https://github.com/crowbar/crowbar-core/pull/920 so no merge is possibly yet!__


crowbar_framework/db/migrate/20170105095251_add_last_seen_to_node.rb
```diff
@@ -0,0 +1,6 @@
+class AddLastSeenToNode < ActiveRecord::Migration
+  def change
+    add_column :nodes, :last_seen, :datetime
```
rsalevsky: I would call it `seen_at` which is similar to the rails defaults like created_at or updated_at.

crowbar_framework/app/controllers/api/crowbar_controller.rb
```diff
@@ -101,4 +101,28 @@ def upgrade
   def maintenance
     render json: ::Crowbar::Checks::Maintenance.updates_status
   end
+
+  api :PUT, "/api/crowbar/ping", "Mark a node as seen recently"
+  header "Accept", "application/vnd.crowbar.v2.0+json", required: true
+  param :hostname, String, desc: "Hostname of the caller", required: true
+  api_version "2.0"
+  def ping
+    node = ::Node.find_by(name: ping_params)
```
rsalevsky: `::Node.find_by_name(ping_params)` also this should go to the nodes_controller

PR #985, opened 2017-01-13
[stable/3.0] upgrade: Save the upgrade errors in consistent format

jsuchome:
When the synchronous API call fails, it returns the error in a form
of 'data' and 'help' touple. We should save it the same way in the
progress file, especially because for aynchronous format, the content
of progress file is returned by the status API call.

(cherry picked from commit 44f7a55da5b79412480846d5de80603b3ab21559)

Backport of https://github.com/crowbar/crowbar-core/pull/972

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -154,7 +154,7 @@ module Api
             end.flatten.compact.join(", ")
             ::Crowbar::UpgradeStatus.new.end_step(
               false,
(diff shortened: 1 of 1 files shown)


jsuchome:
[Save upgrade errors in consistent format (bsc#1019298)](https://trello.com/c/OujhW7wm/256-save-upgrade-errors-in-consistent-format-bsc-1019298)

jsuchome:
As mentioned by @skazi0 , output from prechecks still looks different than others, we should look into that more closely...

PR #990, opened 2017-01-16
upgrade: Orchestrate the upgrade of more controller clusters

jsuchome:
- [x] evacuate network node routers in time
- [x] delete resource from network cluster in time
- [x] set the correct order of clusters for upgrade
- [x] check that services are stopped in all clusters 
- [x] test the whole thing

Requirs https://github.com/crowbar/crowbar-openstack/pull/780

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
@@ -35,6 +35,9 @@ set +x
 source /root/.openrc
 set -x
 
+
+zypper --non-interactive install openstack-neutron-ha-tool
+
 /usr/bin/neutron-ha-tool --l3-agent-evacuate $hostname --wait-for-router $delete_ns
 ret=$?
 if [ $ret != 0 ] ; then
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -393,29 +393,50 @@ module Api
 
       protected
 
+      # If there's separate network cluster, we have touch it before we start upgrade of other
+      # nodes, specificaly we need to evacuate the network routers from the first network node.
+      def prepare_network_node(network_node)
+        # TODO: do something to network node
+        # TODO: do not repeat it!
+        true
+      end
+
       #
       # controller nodes upgrade
       #
       def upgrade_controller_clusters
-        ::Node.find(
-          "pacemaker_founder:true"
-        ).each do |founder|
+        network_node = ::Node.find(
+          "pacemaker_founder:true AND " \
+          "run_list_map:neutron-network AND NOT " \
+          "run_list_map:neutron-server"
+        ).first
+        prepare_network_node(network_node) unless network_node.nil?
+
+        # Now we must upgrade the clusters in the correct order:
+        # 1. data, 2. API, 3. network
+        cluster_founders = ::Node.find("pacemaker_founder:true")
+
+        sorted_founders = cluster_founders.sort do |n1, n2|
+          first_data = n1[:run_list_map].key? "database-server"
+          first_api = n1[:run_list_map].key? "keystone-server"
+          second_net = n2[:run_list_map].key? "neutron-network"
+          first_data || (first_api && second_net) ? -1 : 1
+        end
+        sorted_founders.each do |founder|
           cluster_env = founder[:pacemaker][:config][:environment]
-          upgrade_cluster cluster_env
+          upgrade_cluster founder, cluster_env
         end
       end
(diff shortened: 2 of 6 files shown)


jsuchome:
[(8) S54P6 (S49P16) Orchestrate upgrade of controller nodes in HA setup (upgrade step 12 part 2 -- HA/multiple clusters)](https://trello.com/c/MquyXrTn/252-8-s54p6-s49p16-orchestrate-upgrade-of-controller-nodes-in-ha-setup-upgrade-step-12-part-2-ha-multiple-clusters)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -362,15 +362,55 @@ def nodes
 
       protected
 
+      # If there's separate network cluster, we have touch it before we start upgrade of other
+      # nodes, specificaly we need to evacuate the network routers from the first network node.
+      def prepare_network_node(network_node)
+        return if network_node.upgraded?
+
+        evacuate_network_node(network_node, network_node)
```
jsuchome: @rhafer is it important which node is the *controller node* in this call? 
If it were not `network_node`, I'd need to make sure `neutron-ha-tool` is installed. But why would it be important? At least in this first invocation of `prepare_network_node`, all nodes are still running Cloud6

jsuchome: The only other point I can think of is that *controller (API) nodes* have `.openrc`. For network node i have to find some way of bringing it.

rhafer: Yeah, it reallly shouldn't matter much (apart from the .openrc). I guess the important part is that the neutronclient on the node that it is running on is on the same API level as the neutron-server (and even that might not be a real issue).

jsuchome: @rhafer Thanks, so I've updated it to locate some controller node and install `neutron-ha-tool`. So far no checking for version of such node (maybe it's enough to ensure whole `prepare_network_node` is run only once)

jsuchome:
Hm. I've just found out that with current implementation, `evacuate_network_node` will work in the initial step (`prepare_network_node`), but it will fail later when we proceed with upgrading network cluster and invoke it from `upgrade_first_cluster_node`/`upgrade_next_cluster_node`. (It will fail because the lack of credentials file at network nodes)

It should be possible to skip the invocation somehow from `upgrade_first_cluster_node` where it is not needed (right because we've already evacuated from `prepare_network_node`, but what about the other case? I think we can go back to providing `.openrc` to network nodes (and current state of https://github.com/crowbar/crowbar-openstack/pull/758 is not good) or we can search for real controller node (even in different cluster) in `upgrade_next_cluster_node` and passing that one to `evacuate_network_node`.

@rhafer what do you think? (I'm bit in favor of searching for controller node)

jsuchome:
So, using controller node from different cluster will not work as `neutron-ha-tool` needs ssh access to the nodes and the authorized_keys are distributed only among the nodes of the same cluster... so we're back to the point where we must execute router migration from the network node(s)

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -35,6 +35,9 @@ set +x
 source /root/.openrc
 set -x
 
+
+zypper --non-interactive install openstack-neutron-ha-tool
```
jsuchome: This probably is not needed any more (but does not hurt, I guess)

jsuchome:
YES:

```
root@crowbar:/opt/dell/crowbar_framework/app/models/api # crowbarctl upgrade status
+--------------------------------+--------------------------------------+
| Status                         | Value                                |
+--------------------------------+--------------------------------------+
| current_step                   | finished                             |
| current_substep                | computes                             |
| current_node.name              | d52-54-77-77-77-09.vg1.cloud.suse.de |
| current_node.alias             | compute1                             |
| current_node.ip                | 192.168.253.90                       |
| current_node.state             | upgraded                             |
| current_node.role              | compute                              |
| remaining_nodes                | 0                                    |
| upgraded_nodes                 | 10                                   |
| steps.prechecks.status         | passed                               |
| steps.prepare.status           | passed                               |
| steps.backup_crowbar.status    | passed                               |
| steps.repocheck_crowbar.status | passed                               |
| steps.admin.status             | passed                               |
| steps.database.status          | passed                               |
| steps.repocheck_nodes.status   | passed                               |
| steps.services.status          | passed                               |
| steps.backup_openstack.status  | passed                               |
| steps.nodes.status             | passed                               |
| steps.finished.status          | pending                              |
+--------------------------------+--------------------------------------+
```

PR #994, opened 2017-01-16
provisioner: Fix order of creating groups and users

nicolasbock:
The user `qemu` can not be created unless the group `kvm` was created
beforehand.

Thanks goes to Bryan Gartner <bryan.gartner@suse.com> for pointing this out.

### DIFF
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -225,10 +225,10 @@ fi
 
 # for making HA on shared NFS backend storage work
 add_group glance 200
-add_user glance 200 glance
 add_group qemu 201
-add_user qemu 201 kvm
 add_group kvm 202
+add_user glance 200 glance
+add_user qemu 201 kvm
 
 # Check that we're really on the admin network
 # --------------------------------------------


chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -225,10 +225,10 @@ fi
 
 # for making HA on shared NFS backend storage work
 add_group glance 200
-add_user glance 200 glance
 add_group qemu 201
```
sidthe: is this one actually needed? or is it a copy paste error?

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -225,10 +225,10 @@ fi
...
-add_user qemu 201 kvm
 add_group kvm 202
+add_user glance 200 glance
+add_user qemu 201 kvm
```
sidthe: wouldn't it make sense to keep group/user add order so that the relationship is easily visible?

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
nicolasbock: User `qemu` is in two groups, `qemu` and `kvm`.

nicolasbock: Whether we need that I am not sure :) I have to check the rest of the code.

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
nicolasbock: That's what I though at first too, but line 231 depends on line 229 and simply moving the `add_user ... kvm` line up would break the order in the uid/gid values. So for aesthetic reasons I opted to order by uid/gid values and grouped by `add_{user,group}` commands :smile:.

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
sidthe: Looks like this is mirrored in autoyast as well, and the qemu group is also there

nicolasbock:
HA gating is marked as failing because our CI is overloaded. The job never ran. This change is not related to HA and the non-HA job passed. I'll merge it.

PR #995, opened 2017-01-17
upgrade: Fix handling of multiple zypper prompts

skazi0:
There can be more than one `<prompt>` element in zypper xml output.
In such cases, `Hash.from_xml()` returns array of hashes instead of
a hash for those subtrees.

In this particular case, the simplest solution is to show only first
prompt as the user needs to complete the actions in sequence anyway
when running it in the terminal.

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -162,18 +162,23 @@ module Api
             }
           end
 
-          unless zypper_stream["prompt"].nil?
+          prompt = zypper_stream["prompt"]
+          unless prompt.nil?
+            # keep only first prompt for easier formatting
+            prompt = prompt.first if prompt.is_a?(Array)
+
             upgrade_status.end_step(
               false,
               repocheck_crowbar: {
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -138,11 +138,15 @@ def adminrepocheck
             }
           end
 
-          unless zypper_stream["prompt"].nil?
+          prompt = zypper_stream["prompt"]
+          unless prompt.nil?
+            # keep only first prompt for easier formatting
+            prompt = prompt.first if prompt.is_a?(Array)
```
MaximilianMeister: can't we just concatenate the prompts? the formatting will be already kind of ugly i assume?

skazi0: We could, assuming that `text` will always be the only field we use.

jsuchome:
@MaximilianMeister needs rebase

skazi0:
Doing close & reopen trick to force triggering of travis checks

PR #996, opened 2017-01-17
retry when trying to evacuate l3 agent

jsuchome:


chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -31,16 +31,20 @@ set -x
 mkdir -p $UPGRADEDIR
 rm -f $UPGRADEDIR/crowbar-router-migration-failed
 
-set +x
-source /root/.openrc
-set -x
+if [[ -f /usr/bin/neutron-ha-tool ]] ; then
```
abel-navarro: if we don't have `neutron-ha-tool` should we fail?

jsuchome: It probably means the script was executed on the wrong node. So maybe yes, failure seems more correct... but then the 1st commit is not needed as the failure will happen anyway.

jsuchome:
@abel-navarro I've removed the first commit then

jsuchome:
[bsc#1016943 - neutron-ha-tool: migrating dvr router fails on first try](https://trello.com/c/D8MqR7mr/264-bsc-1016943-neutron-ha-tool-migrating-dvr-router-fails-on-first-try)

jsuchome:
Hm, I'm not sure if by retrying we're not losing the chance to report some real error, like http://pastebin.nue.suse.com/19670/src

rhafer:
> Hm, I'm not sure if by retrying we're not losing the chance to report some real error

Yeah, it seems that "-r" will hide some types of failures currenty :(. E.g. if the l2-agent (ovs-agent, linuxbridge-agent) is not running on the target node, and the waiting for the router ports fails it will not retry to migrate that router, because it's already gone from the source node.

PR #997, opened 2017-01-17
upgrade: Check if a feature/addon is deployed, not enabled

MaximilianMeister:
if we only check if it is enabled on the repository level, we will
have a false result within the repochecks in case a repository is
disabled, but necessary to have, because it has been deployed.

so if an ha repo becomes unavailable at some point during the upgrade
it will not even show up in the repochecks anymore, because it gets
disabled on the repository level once it is not available anymore.

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -73,7 +73,7 @@ module Api
       def addons
         [].tap do |list|
           ["ceph", "ha"].each do |addon|
-            list.push(addon) if addon_installed?(addon) && addon_enabled?(addon)
+            list.push(addon) if addon_installed?(addon) && addon_deployed?(addon)
           end
         end
       end
@@ -81,7 +81,7 @@ module Api
       def ceph_status
         {}.tap do |ret|
           ceph_node = ::Node.find("roles:ceph-mon AND ceph_config_environment:*").first
-          return ret if ceph_node.nil?
+          next unless ceph_node
           ssh_retval = ceph_node.run_ssh_cmd("LANG=C ceph health 2>&1")
           unless ssh_retval[:stdout].include? "HEALTH_OK"
             ret["errors"] = [
@@ -140,10 +140,14 @@ module Api
         false
       end
 
-      def addon_enabled?(addon)
-        Api::Node.repocheck(addon: addon)[addon]["available"]
+      def addon_deployed?(addon)
+        case addon
+        when "ceph"
+          ::Node.find("roles:ceph-mon AND ceph_config_environment:*").any?
+        when "ha"
+          ::Node.find("pacemaker_founder:true AND pacemaker_config_environment:*").any?
+        end
       end
-
     end
   end
 end
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -57,12 +57,14 @@ module Api
             )
           }
 
-          ceph_status = Api::Crowbar.ceph_status
-          ret[:ceph_healthy] = {
-            required: true,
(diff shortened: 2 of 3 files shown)


MaximilianMeister:
@jsuchome i accidentally named the upgrade job without nodes `hacloud=1`, so i was overriding the full upgrade job...

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -140,10 +140,14 @@ def addon_installed?(addon)
...
       end
 
-      def addon_enabled?(addon)
-        Api::Node.repocheck(addon: addon)[addon]["available"]
+      def addon_deployed?(addon)
+        case addon
+        when "ceph"
+          ::Node.find("roles:ceph-mon AND ceph_config_environment:*").any?
+        when "ha"
+          NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*").any?
```
rsalevsky: NodeObject is obsolete

MaximilianMeister: @rsalevsky then you forgot to update it in `crowbar-ha`

MaximilianMeister:
@jsuchome btw all gatings were green

jsuchome:
It seems to make good sense. However I'm bit unsure why we didn't do it from start. Wasn't there some good reason for that?

jsuchome:
Note: I'm currently not sure why do we need it in Cloud6...

MaximilianMeister:
> Note: I'm currently not sure why do we need it in Cloud6...

the prechecks are using the addons method as well to do the ha/ceph checks or not

> It seems to make good sense. However I'm bit unsure why we didn't do it from start. Wasn't there some good reason for that?

at least i haven't mentioned any reason in https://github.com/crowbar/crowbar-core/commit/ad9a800f5dc35a9ee1dbb4851752aff201f34ddf :-(
but the main thing IIRC was that the addons are always installed, so there was another check necessary to ensure the presence of the addon, and I probably took the repocheck just because it was my first thought.

jsuchome:
> the prechecks are using the addons method as well to do the ha/ceph checks or not

Hm, you mean e.g. https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/api/upgrade.rb#L65 ? So now we have the same check in `addons` as in `ceph_status` ... maybe `ceph_status` should be called inside that `if` block

MaximilianMeister:
> Hm, you mean e.g. https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/api/upgrade.rb#L65 ?

yes

> So now we have the same check in addons as in ceph_status ... maybe ceph_status should be called inside that if block

good point, we can make use of the new method in `ceph_status`

jsuchome:
What about `ha_presence_check`?

MaximilianMeister:
> What about ha_presence_check?

seems that we are always doing the `ha_presence` check, there is no condition in the prechecks

jsuchome:
Ah, OK, that's true

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -73,15 +73,15 @@ def version
...
+            list.push(addon) if addon_installed?(addon) && addon_deployed?(addon)
           end
         end
       end
 
       def ceph_status
         {}.tap do |ret|
           ceph_node = ::Node.find("roles:ceph-mon AND ceph_config_environment:*").first
-          return ret if ceph_node.nil?
+          next unless ceph_node
```
jsuchome: Is this the same as `retururn ret`?

MaximilianMeister: yes, it is. it's just exiting the block. we have it in other places as well, that's why i changed it

PR #998, opened 2017-01-17
node: Rename find_node_by_name to find_by_name

rsalevsky:
This is change is part of the switch to a ActiveRecord based Node
object.

### DIFF
--- a/crowbar_framework/app/models/node.rb
+++ b/crowbar_framework/app/models/node.rb
@@ -1628,7 +1628,7 @@ class Node < ChefObject
       end
     end
 
-    def find_node_by_name(name)
+    def find_by_name(name)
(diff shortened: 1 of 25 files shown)


crowbar_framework/app/models/node.rb
```diff
@@ -1647,8 +1647,13 @@ def find_node_by_name(name)
       end
     end
 
+    def find_node_by_name(name)
+      Rails.logger.warn("find_node_by_name is depricated, please use find_by_name!")
```
jsuchome: Typo: *depricated*

rsalevsky:
@MaximilianMeister Correct, that's the reason why I linked the old method to the new one with a deprecation warning. So we have to change it there as well but it's not super urgent and we break nothing.

rsalevsky:
Rebased after merge of #999

PR #1000, opened 2017-01-17
upgrade: Remove obsolete code from 5-6 upgrade

jsuchome:
I'm not yet sure if this is correct, we might want to reuse some of that db_backup code...

But I admit - mostly I wanted to reach PR no. 1000 ...


jsuchome:
@MaximilianMeister actually your backup code is totally new and different, so I think we could proceed with this removal, what do you think?

PR #1001, opened 2017-01-17
node: Unify crowbar data access

rsalevsky:


rsalevsky:
@vuntz The behavior of how we access the crowbar data seams strange to me. Sometimes you need node.crowbar["crowbar"] but sometimes you only need node.crowbar for doing the same thing. Also why do we use the data from the role and not directly from the node?

PR #1002, opened 2017-01-18
ohai: common ohai attribute for libvirt guest uuid

jsuchome:
This adds a new attribute node[:crowbar_ohai][:libvirt][:guest_uuid]
which contains the UUID of the libvirt instance that the node is running
in. The purpose is to have a common place for this attribute independent
of the node's architecture (currently x86 ist storing this somewhere in
the "dmi" subtree and s390x in "s390x").

(cherry picked from commit 7548957de6b76c2760b76a72bd8bbfe2f48e738c)

### DIFF
--- a/chef/cookbooks/ohai/files/default/plugins/linux/crowbar_libvirt.rb
+++ b/chef/cookbooks/ohai/files/default/plugins/linux/crowbar_libvirt.rb
@@ -1,0 +1,21 @@
+provides "crowbar_ohai"
+
+require_plugin "kernel"
+require_plugin "dmi"
+require_plugin "linux::s390x"
+
+libvirt_uuid = nil
+
+if kernel[:machine] == "s390x"
+  if s390x[:system][:manufacturer] == "KVM"
+    libvirt_uuid = s390x[:system][:uuid]
+  end
+else
+  manufacturer = dmi[:system] ? dmi[:system][:manufacturer] : "unknown"
+  if ["Bochs", "QEMU"].include? manufacturer
+    libvirt_uuid = dmi[:system][:uuid]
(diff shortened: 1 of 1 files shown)


jsuchome:
Note: We do not really **need** this in Cloud6, but Cloud7 is using this ohai data already in migration scripts and it breaks the 6-7 upgrade

rhafer:
I think it would be good to update the commit message to mention that this is purely indented to make the cloud 6 -> 7 upgrade work.

@vuntz And I would welcome some feedback  if you think it's the right fix. Or if we should adapt prepare_stonith_attributes() in master to be able to handle a missing [:libvirt][:guest_uuid] attribute and fall back to the old name?

rhafer:
See here for more details: https://github.com/crowbar/crowbar-ha/pull/163#discussion_r96606130

vuntz:
> @vuntz And I would welcome some feedback if you think it's the right fix. Or if we should adapt prepare_stonith_attributes() in master to be able to handle a missing [:libvirt][:guest_uuid] attribute and fall back to the old name?

Hrm, actually, I would think it's not the right fix (and https://github.com/crowbar/crowbar-ha/pull/175 is) as during the upgrade, when this code is run on schema migration, there's no way that the ohai plugin is installed on nodes (since they're still technically Cloud 6 nodes).

vuntz:
Oh wait, this is backporting to Cloud 6, so I guess that would work.

rhafer:
Yeah, actually both fixes should work. I am still undecided which one is nicer ;)

jsuchome:
@rhafer you mean uglier :-)

PR #1003, opened 2017-01-18
puma: Fix puma version to 2.16.x

rsalevsky:
We updated puma to 2.16.0 in the product, let's do the same for testing.


rsalevsky:
Ready to merge.

PR #989, opened 2017-01-16
upgrade: Map features to repository names

MaximilianMeister:
this is useful information for any client (UI/crowbarctl)

### crowbarctl
```bash
root@crowbar:/opt/dell/crowbar_framework # crowbarctl upgrade repocheck nodes
+----------------------------------+--------------------------------+
| Status                           | Value                          |
+----------------------------------+--------------------------------+
| ha.available                     | true                           |
| ha.repos                         | SLE12-SP2-HA-Pool              |
|                                  | SLE12-SP2-HA-Updates           |
| os.available                     | false                          |
| os.repos                         | SLES12-SP2-Pool                |
|                                  | SLES12-SP2-Updates             |
|                                  | SLES12-SP2-Pool                |
| os.errors.missing.x86_64         | SLES12-SP2-Pool                |
| os.errors.inactive.x86_64        | SLES12-SP2-Pool                |
| openstack.available              | true                           |
| openstack.repos                  | SUSE-OpenStack-Cloud-7-Pool    |
|                                  | SUSE-OpenStack-Cloud-7-Updates |
+----------------------------------+--------------------------------+
```

### Raw json
```json
{
  "ha": {
    "available": true,
    "repos": [
      "SLE12-SP2-HA-Pool",
      "SLE12-SP2-HA-Updates"
    ],
    "missing_repos": {
    }
  },
  "os": {
    "available": false,
    "repos": [
      "SLES12-SP2-Pool",
      "SLES12-SP2-Updates"
    ],
    "errors": {
      "missing": {
        "x86_64": [
          "SLES12-SP2-Pool"
        ]
      },
      "inactive": {
        "x86_64": [
          "SLES12-SP2-Pool"
        ]
      }
    }
  },
  "openstack": {
    "available": true,
    "repos": [
      "SUSE-OpenStack-Cloud-7-Pool",
      "SUSE-OpenStack-Cloud-7-Updates"
    ],
    "missing_repos": {
    }
  }
}
```

PR #1006, opened 2017-01-19
upgrade: Adapt admin repository check

MaximilianMeister:
this is not a straight backport, so I haven't cherry-picked it

https://github.com/crowbar/crowbar-core/pull/989/commits/24302a8e48b2e63da9d88894d4a1faf00ea89540

partial backport of https://github.com/crowbar/crowbar-core/pull/989

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -162,26 +162,38 @@ module Api
           os_available = repo_version_available?(products, "SLES", "12.2")
           ret[:os] = {
             available: os_available,
-            repos: {}
+            repos: [
+              "SLES12-SP2-Pool",
+              "SLES12-SP2-Updates"
+            ],
+            errors: {}
           }
-          ret[:os][:repos][admin_architecture.to_sym] = {
-            missing: ["SLES-12-SP2-Pool", "SLES12-SP2-Updates"]
-          } unless os_available
+          unless os_available
+            ret[:os][:errors][admin_architecture.to_sym] = {
+              missing: ret[:os][:repos]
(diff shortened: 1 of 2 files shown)


jsuchome:
With this patch, crowbartcl output looks like this: http://pastebin.nue.suse.com/19818/src 

When the repository is missing, it is actually listed twice, once as required and once as missing. I would think seeing it as missing should be enough (and if none is missing, no need to show any repo). What do the others think?

jsuchome:
Also, we're listing **Cloud** repo now.
Is it really necessary when there's **SUSE-OpenStack-Cloud-7-Pool** ? @vuntz I'm kind of confused here, could you shed some light ?

skazi0:
[crowbarctl should show nice names for missing repositories](https://trello.com/c/xr2UOZyN/261-crowbarctl-should-show-nice-names-for-missing-repositories)

vuntz:
> Also, we're listing Cloud repo now.
> Is it really necessary when there's SUSE-OpenStack-Cloud-7-Pool ? @vuntz I'm kind of confused here, could you shed some light ? (This also applies to #989 which has nicer example output)

https://trello.com/c/4YsHrqUW/270-5-simplify-repo-setup

In short, right now, I think we require the Cloud repo during install; but it's actually not needed if Pool is there, indeed.

MaximilianMeister:
> In short, right now, I think we require the Cloud repo during install; but it's actually not needed if Pool is there, indeed.

updated, thanks for the hint

MaximilianMeister:
fixed the commit hash in the commit message, as there was a minor change (comment) on the cloud7 side

PR #1007, opened 2017-01-19
upgrade: Enable running prechecks also out of order

jsuchome:
This was (partially) removed by https://github.com/crowbar/crowbar-core/commit/f88889317a7937150f4d8cce2aceb60a7c946319

Ideally, we should not need it and user should not be allowed to run the prechecks later when some other steps have already been run. But UI might need to do it in some special cases


jsuchome:
[bsc#1020899  Upgrade UI blocked after failed prepare and page refresh](https://trello.com/c/HOL2F6PL/2206-bsc-1020899-upgrade-ui-blocked-after-failed-prepare-and-page-refresh)

PR #1011, opened 2017-01-20
upgrade: Implement openstack backup creation

MaximilianMeister:
1. check size of database and compare with free space on system
2. create a postgresql dump and save it to the admin server


MaximilianMeister:
[(3) Implement API for providing backup of OpenStack database during upgrade (part of upgrade step 9-10)](https://trello.com/c/DFaqwV10/116-3-implement-api-for-providing-backup-of-openstack-database-during-upgrade-part-of-upgrade-step-9-10)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -866,6 +936,18 @@ def repo_version_available?(products, product, version)
       def admin_architecture
         ::Node.admin_node.architecture
       end
+
+      #
+      # general helpers
+      #
+      def run_cmd(*args)
```
jsuchome: Don't we have something like this elsewhere (in some library, or other model)?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -324,6 +324,76 @@ def services
...
+        unless db_size[:exit_code].zero?
+          Rails.logger.error(
+            "Failed to check size of OpenStack database: #{db_dump[:stdout_and_stderr]}"
+          )
+          raise ::Crowbar::Error::UpgradeDatabaseSizeError.new(
+            db_size[:stdout_and_stderr]
+          )
+        end
+
+        free_space = run_cmd("df -B1 -l --output='avail' / | tail -n1")
```
jsuchome: Not sure if the command can be translated, but I'd rather make sure to pass LANG=C

jsuchome: I'm also not sure if the last item is always the one we need to know... in one of my tests this is giving the result for tmpfs

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -324,6 +324,76 @@ def services
...
+            free_space[:stdout_and_stderr]
+          )
+        end
+        if free_space[:stdout_and_stderr].strip.to_i < db_size[:stdout_and_stderr].strip.to_i
+          Rails.logger.error("Not enough free disk space to create the OpenStack database dump")
+          raise ::Crowbar::Error::UpgradeNotEnoughDiskSpaceError.new
+        end
+
+        Rails.logger.debug("Creating OpenStack database dump")
+        db_dump = run_cmd("PGPASSWORD=#{pass} pg_dumpall -h #{host} -U #{user} > #{dump_location}")
```
jsuchome: It was said that this could take very long time... should we think about backgroudning the whole action?

crowbar_framework/app/models/api/upgrade.rb
MaximilianMeister: we can exclude `tmpfs` and `devtmpfs`

crowbar_framework/app/models/api/upgrade.rb
MaximilianMeister: it's already handled asynchronously

crowbar_framework/app/models/api/upgrade.rb
MaximilianMeister: in crowbar-init ...

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -324,6 +324,76 @@ def services
...
+        unless db_dump[:exit_code].zero?
```
jsuchome: Is it possible that some broken dump is created in this case? If so, we should probably try to delete it...

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -324,6 +324,76 @@ def services
...
+          )
+        end
+
+        true
+      rescue ::Crowbar::Error::UpgradeNotEnoughDiskSpaceError => e
+        ::Crowbar::UpgradeStatus.new.end_step(
+          false,
+          backup_openstack: {
+            data: e.message,
+            help: "Make sure you have enough diskspace to store the OpenStack database dump."
```
jsuchome: *diskspace* isn't a word  I believe :-)

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Ah, I missed that, OK

jsuchome:
I was wondering if we couldn't move all those upgrade errors into one file/module...

MaximilianMeister:
> I was wondering if we couldn't move all those upgrade errors into one file/module...

agree that it slowly makes sense to do that. I'll update it on Monday

MaximilianMeister:
@jsuchome updated

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -340,6 +340,80 @@ def services
...
+          }
+        )
+      rescue ::Crowbar::Error::UpgradeFreeDiskSpaceError,
+             ::Crowbar::Error::UpgradeDatabaseSizeError,
+             ::Crowbar::Error::UpgradeDatabaseDumpError => e
+        ::Crowbar::UpgradeStatus.new.end_step(
+          false,
+          backup_openstack: {
+            data: e.message,
+            help: "Please refer to the error message in the response."
```
jsuchome: I'm wondering if, in the case of such generic "help texts", we shouldn't rather omit passing "help" at all. AFAIK from the UI point the "help" key is optional, right @skazi0 ?

crowbar_framework/lib/crowbar/error/upgrade.rb
```diff
@@ -0,0 +1,54 @@
...
+module Crowbar
+  module Error
+    # generic upgrade
+    class UpgradeError < StandardError
+    end
+
+    # cancel upgrade
+    class UpgradeCancelError < StandardError
+      def initialize(step_name = "")
+        super("Not possible to cancel the upgrade at the step #{step_name}")
```
jsuchome: What do you think about making these messages translatable? We do it e.g. in case of prechecks...

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -340,6 +340,80 @@ def services
...
+          Rails.logger.error(
+            "Failed to check size of OpenStack database: #{db_size[:stdout_and_stderr]}"
+          )
+          raise ::Crowbar::Error::UpgradeDatabaseSizeError.new(
+            db_size[:stdout_and_stderr]
+          )
+        end
+
+        free_space = run_cmd(
+          "LANG=C df -x 'tmpfs' -x 'devtmpfs' -B1 -l --output='avail' / | tail -n1"
```
jsuchome: Are you sure checking `/` is correct? Most likely it is, but cannot the target directory be located at different partition?
Maybe we should check first...

MaximilianMeister: I guess we could just add the path `/var/lib/crowbar/backup` there to be sure

jsuchome: Yes, that might be enough

crowbar_framework/lib/crowbar/error/upgrade.rb
MaximilianMeister: e.g. in case of the upgrade status errors we can't do that, as we require the library also from other places that don't have the translation rendering available from Rails. so i'd keep it non translated, at least for consistency with the other exceptions

crowbar_framework/app/models/api/upgrade.rb
MaximilianMeister: true, it can be omitted

crowbar_framework/lib/crowbar/error/upgrade.rb
```diff
@@ -0,0 +1,54 @@
...
+      end
+    end
+
+    # openstack backup
+    class UpgradeNotEnoughDiskSpaceError < StandardError
+      def initialize
+        super("Not enough space on to create a OpenStack database dump.")
```
jsuchome: *Not enough space on* ? On a disk? Or in the directory we're checking?

MaximilianMeister: oops, where did the words go? :-)

crowbar_framework/lib/crowbar/error/upgrade.rb
jsuchome: OK... I guess

crowbar_framework/lib/crowbar/error/upgrade.rb
```diff
@@ -0,0 +1,54 @@
...
+    class UpgradeCancelError < StandardError
+      def initialize(step_name = "")
+        super("Not possible to cancel the upgrade at the step #{step_name}")
+      end
+    end
+
+    # openstack backup
+    class UpgradeNotEnoughDiskSpaceError < StandardError
+      def initialize
+        super("Not enough space in /var/lib/crowbar/upgrade to create a OpenStack database dump.")
```
jsuchome: *an OpenStack* ?

jsuchome: It would be nicer to give that path as an argument... in case we allow some changes later.

crowbar_framework/lib/crowbar/error/upgrade.rb
```diff
@@ -0,0 +1,54 @@
...
+      end
+    end
+
+    class UpgradeFreeDiskSpaceError < StandardError
+      def initialize(message = "")
+        super("Cannot determine free disk space. #{message}")
```
jsuchome: This is shown as
``steps.backup_openstack.errors.backup_openstack.data | Cannot determine free disk space. 8544944128``

Please wrap the size into some text.

jsuchome: Ah, OK, it's stdout + stderr, so I guess it's fine this way.

crowbar_framework/lib/crowbar/error/upgrade.rb
MaximilianMeister: done

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -345,6 +345,80 @@ def services
...
+        free_space = run_cmd(
+          "LANG=C df -x 'tmpfs' -x 'devtmpfs' -B1 -l --output='avail' #{crowbar_lib_dir} | tail -n1"
+        )
+        unless free_space[:exit_code].zero?
+          Rails.logger.error("Cannot determine free disk space: #{free_space[:stdout_and_stderr]}")
+          raise ::Crowbar::Error::UpgradeFreeDiskSpaceError.new(
+            free_space[:stdout_and_stderr]
+          )
+        end
+        if free_space[:stdout_and_stderr].strip.to_i < db_size[:stdout_and_stderr].strip.to_i
```
dirkmueller: I'd add some extra pessimism factor to it (like 10% more than database size or something like that)..

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -345,6 +345,80 @@ def services
...
+          Rails.logger.error("Not enough free disk space to create the OpenStack database dump")
+          raise ::Crowbar::Error::UpgradeNotEnoughDiskSpaceError.new("#{crowbar_lib_dir}/upgrade")
+        end
+
+        Rails.logger.debug("Creating OpenStack database dump")
+        db_dump = run_cmd(
+          "PGPASSWORD=#{psql[:pass]} pg_dumpall -h #{psql[:host]} -U #{psql[:user]} > #{dump_path}"
```
dirkmueller: here we could in theory pipe it through gzip to reduce disk impact..

MaximilianMeister:
> I can't really review this to be honest, but from what I could understand it looks good to me. some nitpicks / comments inline

@dirkmueller thanks for the review. i'll address the enhancements in a follow up as we have some other work that depends on this PR

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -345,6 +345,80 @@ def services
...
       handle_asynchronously :services
 
+      def openstackbackup
+        ::Crowbar::UpgradeStatus.new.start_step(:backup_openstack)
+
+        crowbar_lib_dir = "/var/lib/crowbar"
+        dump_path = "#{crowbar_lib_dir}/upgrade/6-to-7-openstack_dump.sql"
+        if File.exist?(dump_path)
+          Rails.logger.debug("OpenStack backup already exists. Skipping...")
+          return
```
skazi0: what about end_step?

jsuchome: Good point, something for followup

PR #1013, opened 2017-01-20
upgrade: Unify calls to UpgradeStatus.start_step

MaximilianMeister:


jsuchome:
What about the `end_step`s?

MaximilianMeister:
> What about the end_steps?

I'd like to keep them in the model (for consistency) as this is somehow mandatory for the background methods

jsuchome:
Well, that seems little inconsistent :-(

skazi0:
If I understand correctly, there are two parts here: 
1. using new instance of status on every start and end.
2. moving some calls between the controllers and models

The first part seems fine to me as it should limit some race condition windows but I'm not sure about the second part. Models and controllers are well defined for "data manipulation" APIs but in our case it's not that clear and I don't have any opinion on where those calls belong.

MaximilianMeister:
I'll try to rework this PR, by first including `best_method` in the `checks` method, that probably enables to call the `start_step` in the model, so we will be consistent as well

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -59,29 +60,39 @@ def checks
...
+            "none"
+          elsif !ret[:checks].any? { |_id, c| (c[:required] || !c[:required]) && !c[:passed] }
+            "non-disruptive"
+          elsif !ret[:checks].any? do |_id, c|
+            (c[:required] && !c[:passed]) && (!c[:required] && c[:passed])
+          end
+            "disruptive"
+          end
+
           return ret unless upgrade_status.current_step == :prechecks
```
jsuchome: Do we really need that new :checks submap for the `best_method` evaluation? For example here we're returning different `ret` than expected before...

MaximilianMeister: the `checks` subtree has been rendered in the controller before, now it's just here in the model

jsuchome: Thanks

jsuchome:
Why this is good: apart from the cleanup, reworking of `best_method` means that the prechecks are not called twice.

PR #1015, opened 2017-01-20
[stable/3.0] Use crowbar-pacemaker provider for apache service for HA

jsuchome:
apache is utilized by various OpenStack services, e.g. horizon.
Just like other services, we should let it be handled by Pacemkaer
when it is part of HA configuration.

This is an alternative to https://github.com/crowbar/crowbar-ha/pull/176


vuntz:
See https://github.com/crowbar/crowbar-ha/pull/176#issuecomment-274467602

PR #1016, opened 2017-01-23
upgrade: Non-requried precheck errors in status

skazi0:
The errors coming from prechecks which were `required: false` were
not stored in the status file. This change removes the filtering so
that all of the errors are stored.


skazi0:
This change would cause the step to fail because of non-required prechecks failures but we want to be able to continue (in disruptive mode). Closing.

jsuchome:
Thanks

PR #1018, opened 2017-01-23
apache2: Do not define apache2 service twice (bsc#1021106)

vuntz:
This causes the magic to override the LWRP in crowbar-pacemaker to not
fully work, as it only overrides the second definition, and not the
first one. Which results in the :enable action to always be done for the
apache2 service, even when pacemaker is used.

https://bugzilla.suse.com/show_bug.cgi?id=1021106


jsuchome:
[apache service not configured correctly with HA (not handled by pacemaker)](https://trello.com/c/1b5iNWF7/273-apache-service-not-configured-correctly-with-ha-not-handled-by-pacemaker)

PR #1025, opened 2017-01-24
service_object: Use rails logger directly

rsalevsky:


rsalevsky:
Rebased

vuntz:
@rsalevsky while I'm fine with it, I'm slightly concerned that it will make backporting / forward-porting changes with the sap-oc bits more complicated. What do you think?

rsalevsky:
@vuntz Yes, it will make it a little bit less straight forward to backport patches. But the conflict will be very easy to fix.

PR #1026, opened 2017-01-24
barclamp_model: Drop dummy barclamp_model

rsalevsky:


aspiers:
@rsalevsky You forgot to delete the corresponding code in the Crowbar repository:

- https://github.com/crowbar/crowbar/blob/master/scripts/barclamp_create.rb
- various bits of https://github.com/crowbar/crowbar/blob/master/scripts/barclamp_mgmt_lib.rb

aspiers:
And for the record, I would have preferred this skeleton to be fixed rather than deleted.  Just this morning we had someone asking "how do I create a skeleton barclamp?" on `#crowbar` IRC.

aplanas:
@aspiers for that this is some documentation that @jgrassler is writing. Maybe he can comment if is in some use to maintain this code and update after is updated

aspiers:
@aplanas It seems you are not on `#crowbar` so you missed the discussion we already had about this :)

aplanas:
@aspiers indeed I am not in `#crowbar`. What was the conclusion of the discussion? some agreement about this PR?

rsalevsky:
@aspiers If someone agrees to maintain this and @jgrassler has the opinion that this useful then I prefer to create a new Github repo or move it to the crowbar repo but this place was in my opinion wrong.

jgrassler:
Much like people on here missed the IRC discussion I only discovered this discussion once Adam's link on https://github.com/crowbar/crowbar/pull/2294 pointed me here, sorry :-)

That being said I'll still add my two cents: I think just the documentation is fine for now, because it doesn't bitrot quite so easily. People tend to simply discard a script once they discover it has bitrotted to the point of no longer running. They might just submit a pull request fixing the part of the documentation that no longer works or at least complain about the part where they needed to deviate from the One True to somebody who then fixes the documentation.

PR #1027, opened 2017-01-24
crowbar: Log puma std{out,err}

nicolasbock:
For debugging we should log puma's std{out,err}.

### DIFF
--- a/crowbar_framework/config/puma.rb
+++ b/crowbar_framework/config/puma.rb
@@ -51,3 +51,9 @@ end
 ].each do |name|
   FileUtils.mkdir_p File.join(ROOT, name)
 end
+
+stdout_redirect(
+  "/var/log/crowbar/production.log",
+  "/var/log/crowbar/production.log",
+  true
+)


nicolasbock:
It's not quite doing what I was hoping for. It's also logging the normal log output that's already in production.log.

nicolasbock:
@rsalevsky Maybe you can help here. I was really only trying to log things such as failures in barclamps that prevent crowbar from going into `custom` view.

rsalevsky:
@nicolasbock What do you mean with custom view? Can you explain in more detail what you want to achieve?

nicolasbock:
@rsalevsky sure: A barclamp has a `raw` or a `custom` view showing the raw json or some buttons and fields. When I updated a barclamp yesterday I kept getting the `raw` view and couldn't switch to the other one. I didn't find any error messages in `/var/log/crowbar/production.log`. Finally @sayalilunkad and @Itxaka pointed out to me on `#cloud` that I should rather look in `/var/log/messages`, and indeed, the error messages indicating a missing translation was right there. I think we should be logging those puma errors in `/var/log/crowbar` so that they are easier to find. Unfortunately the changes I made here seem to also log normal log messages in the new log file.

Itxaka:
@nicolasbock should we just direct everything to /var/log/crowbar/production.log? I meant, stdout and stderror are both valuable and having them together into a single file seems helpful, so I would really like to have them in the same place, unless this makes the normal logging appear twice.

nicolasbock:
@Itxaka Ok, that's a good point. Let's do that then.

aspiers:
@Itxaka @nicolasbock Not a -1 but I just wanted to point out that `lnav` significantly reduces the advantages of having things merged into the same log file, and of course there is a downside to the merging too.

nicolasbock:
@Itxaka I think @aspiers has a good point there. Using `lnav` is actually very simple and it definitely beats using `less` :wink:.

Itxaka:
Sure, I also agree but in /var/log/messages there is a lot of messages that we are not interested into, plus the log rotation can be different than the production.log so this could lead to a loss of valuable logs. So its more about preservation and clarity(so **end users** know that all log messages from crowbar are in there) than easy of use, because if not could just dump everything into /var/log/messages and forget about it ;)

nicolasbock:
@aspiers Well, that's a good question. Are you thinking more in terms of how apache splits its logs?

crowbar_framework/config/puma.rb
```diff
@@ -51,3 +51,6 @@
 ].each do |name|
   FileUtils.mkdir_p File.join(ROOT, name)
 end
+
+stdout_redirect "/var/log/crowbar/production.log",
+  "/var/log/crowbar/production.log", true
```
rsalevsky: The code change is fine, only a small style fix would be nice to have.

```
stdout_redirect(
  "/var/log/crowbar/production.log",
  "/var/log/crowbar/production.log",
  true
)
```

nicolasbock: done

nicolasbock:
@vuntz I was just about to press the `merge` button :smile:

PR #1029, opened 2017-01-25
upgrade: Save the backup location info in the progress file

jsuchome:
Requires https://github.com/crowbar/crowbar-core/pull/1011

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade_status.rb
+++ b/crowbar_framework/lib/crowbar/upgrade_status.rb
@@ -152,14 +152,14 @@ module Crowbar
       ].include?(current_step) && !running?(:admin)
     end
 
-    def save_crowbar_backup(backup_location = "")
+    def save_crowbar_backup(backup_location)
       ::Crowbar::Lock::LocalBlocking.with_lock(shared: false, logger: @logger, path: lock_path) do
         progress[:crowbar_backup] = backup_location
         save
       end
     end
 
-    def save_openstack_backup(backup_location = "")
+    def save_openstack_backup(backup_location)
       ::Crowbar::Lock::LocalBlocking.with_lock(shared: false, logger: @logger, path: lock_path) do
         progress[:openstack_backup] = backup_location
         save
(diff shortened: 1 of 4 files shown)


jsuchome:
[(3) Implement API for providing backup of OpenStack database during upgrade (part of upgrade step 9-10)](https://trello.com/c/DFaqwV10/116-3-implement-api-for-providing-backup-of-openstack-database-during-upgrade-part-of-upgrade-step-9-10)

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -55,7 +55,10 @@ def initialize_state
         current_node: nil,
         # number of nodes still to be upgraded
         remaining_nodes: nil,
-        upgraded_nodes: nil
+        upgraded_nodes: nil,
+        # locations of the backups took during the upgrade
```
skazi0: taken?

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -149,6 +152,20 @@ def cancel_allowed?
       ].include?(current_step) && !running?(:admin)
     end
 
+    def save_crowbar_backup(backup_location = "")
```
MaximilianMeister: i'd not allow an empty backup_location here, or put some reasonable default

jsuchome: Well, `reasonable default` here is strange. Library itself should not know about that path if if it wasn't provided with the call, there's no reason to guess...

jsuchome: And if it is empty string, that's nothing wrong from the library POV.

MaximilianMeister: it's just when you don't submit any parameter the path will be empty, isn't that wrong? why not just enforce the path by not giving a default?

skazi0: +1 on not giving the default. This is a clear indication to the caller that something needs to be passed.

jsuchome:
@MaximilianMeister @skazi0 removed default value

PR #1031, opened 2017-01-25
upgrade: cleanup and catch unforeseen exceptions

MaximilianMeister:
What we need to test here:

- correct handling of progress lib errors (step out of order, step already running)
- correct handling of real upgrade errors (e.g. failure during some node action in nodes step)
- correct handling of non-upgrade errors (syntax errors, calling methods on nil objects etc.)

### DIFF
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -83,6 +83,8 @@ class Api::UpgradeController < ApiController
   api_version "2.0"
   error 422, "Failed to prepare nodes for Crowbar upgrade"
   def prepare
+    ::Crowbar::UpgradeStatus.new.start_step(:prepare)
+
     if Api::Upgrade.prepare(background: true)
       head :ok
     else
@@ -95,9 +97,9 @@ class Api::UpgradeController < ApiController
         }
       }, status: :unprocessable_entity
     end
-  rescue Crowbar::Error::StartStepRunningError,
-         Crowbar::Error::StartStepOrderError,
-         Crowbar::Error::SaveUpgradeStatusError => e
+  rescue ::Crowbar::Error::StartStepRunningError,
+         ::Crowbar::Error::StartStepOrderError,
+         ::Crowbar::Error::SaveUpgradeStatusError => e
     render json: {
       errors: {
         prepare: {
@@ -113,12 +115,12 @@ class Api::UpgradeController < ApiController
   api_version "2.0"
   error 422, "Failed to stop services on all nodes"
   def services
+    ::Crowbar::UpgradeStatus.new.start_step(:services)
     Api::Upgrade.services
     head :ok
-  rescue Crowbar::Error::StartStepRunningError,
-         Crowbar::Error::StartStepOrderError,
-         Crowbar::Error::EndStepRunningError,
-         Crowbar::Error::SaveUpgradeStatusError => e
+  rescue ::Crowbar::Error::StartStepRunningError,
+         ::Crowbar::Error::StartStepOrderError,
+         ::Crowbar::Error::SaveUpgradeStatusError => e
     render json: {
       errors: {
         services: {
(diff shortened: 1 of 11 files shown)


MaximilianMeister:
[(2) S54P3 Catch exceptions crashing crowbar and end the upgrade step with failure](https://trello.com/c/dVrsUTHe/272-2-s54p3-catch-exceptions-crashing-crowbar-and-end-the-upgrade-step-with-failure)

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -291,7 +303,8 @@ def noderepocheck
   rescue Crowbar::Error::StartStepRunningError,
          Crowbar::Error::StartStepOrderError,
          Crowbar::Error::EndStepRunningError,
-         Crowbar::Error::SaveUpgradeStatusError => e
+         Crowbar::Error::SaveUpgradeStatusError,
+         StandardError => e
```
jsuchome: No, I think you need to handle `StandardError` separately and re-raise it. Otherwise we do not have full trace in the production.log

jsuchome: Also, the point of catching StandardError is to end the step with failure, and this is not done when you are catching it here from controller. As a result, if there's a syntax error in the **model** code of nodesrepocheck, not only the error won't be reraised (so I just know `e.message` which is not enough), but also there's no code that would end the nodesrepocheck step and it is still marked as running

MaximilianMeister: true, will move it to the model

MaximilianMeister:
@jsuchome updated

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -341,12 +386,18 @@ def services
         else
           ::Crowbar::UpgradeStatus.new.end_step
         end
+      rescue StandardError => e
+        ::Crowbar::UpgradeStatus.new.end_step(
+          false,
+          services: {
+            data: e.message
```
jsuchome: Could you add the usual `help: "Crowbar has failed. Check /var/log/crowbar/production.log for details."` here?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -392,23 +444,23 @@ def openstackbackup
...
             help: "Make sure you have enough disk space to store the OpenStack database dump."
           }
         )
-      rescue ::Crowbar::Error::UpgradeFreeDiskSpaceError,
-             ::Crowbar::Error::UpgradeDatabaseSizeError,
-             ::Crowbar::Error::UpgradeDatabaseDumpError => e
+      rescue ::Crowbar::Error::Upgrade::FreeDiskSpaceError,
+             ::Crowbar::Error::Upgrade::DatabaseSizeError,
+             ::Crowbar::Error::Upgrade::DatabaseDumpError => e
         ::Crowbar::UpgradeStatus.new.end_step(
```
jsuchome: I think you are forgetting to rescue `StandardError` here

MaximilianMeister: it's because i started the PR when the openstackbackup PR was still not merged. fixed

crowbar_framework/lib/crowbar/error/upgrade.rb
```diff
@@ -19,35 +19,44 @@ module Error
...
+        end
       end
-    end
 
-    class UpgradeDatabaseSizeError < StandardError
-      def initialize(message = "")
-        super("Cannot determine size of OpenStack databases. #{message}")
+      class DatabaseSizeError < UpgradeError
+        def initialize(message = "")
+          super("Cannot determine size of OpenStack databases. #{message}")
```
jsuchome: BTW, here should be singular (database)

MaximilianMeister: actually there are multiple databases, one for each cinder, glance, nova etc. that's why i chose the plural here

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -142,12 +137,12 @@ def services
   # This is gonna initiate the upgrade of all nodes.
   # The method runs asynchronously, so there's a need to poll for the status and possible errors
   def nodes
+    ::Crowbar::UpgradeStatus.new.start_step(:nodes)
```
jsuchome: I think there's still `start_step` in the model that needs to be removed

MaximilianMeister: removed

jsuchome:
So I *think* it works correctly now... but as it is quite complicated and large changed, I'd rather wait for more tests...

PR #1033, opened 2017-01-26
upgrade: Fix no_live_migration check without nova

skazi0:
If nova-controller role is not applied correctly, prechecks were
failing with 500 response and status was left at running.

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -113,7 +113,7 @@ module Api
           )
         end
         nova = NodeObject.find("roles:nova-controller").first
-        ret[:no_live_migration] = true unless nova["nova"]["use_migration"]
+        ret[:no_live_migration] = true if nova && !nova["nova"]["use_migration"]
         ret
       end
 


crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -113,7 +113,7 @@ def compute_status
           )
         end
         nova = NodeObject.find("roles:nova-controller").first
-        ret[:no_live_migration] = true unless nova["nova"]["use_migration"]
```
MaximilianMeister: we could also just return early here if there is no nova

jsuchome: Yes, but that's the same

skazi0: If we return early there would be no indication at all that something is missing and technically if there's no nova at all, live migration is not configured in nova for sure :)

The real question is: should there be a separate error code for not having nova at all or can we rely on the one for live migration?

jsuchome: If there is no nova, then there is no error and upgrade will just be faster as we do not need to do the live-migration. I think it's good enough to have checked like the one in PR

skazi0: But with this PR the `no_live_migration` will be reported if there's no nova. If I understand correctly this is not the right logic.

jsuchome: I'm always confused by the logic after unless. So it's probably better to just `return ret if nova.nil?` as @MaximilianMeister originally suggested

skazi0: After inverting the logic it doesn't look bad imo. And we got rid of `unless` :)

skazi0: @MaximilianMeister @jsuchome please recheck the logic

PR #1036, opened 2017-01-26
upgrade: Indicate running upgrade process

jsuchome:
The file can be used by crowbar so the non-upgrade related
requests are blocked during the upgrade.

Fixes bsc#1016923


jsuchome:
[bsc#1016923: crowbar disabled too early](https://trello.com/c/vZnBi0Vv/2145-bsc-1016923-crowbar-disabled-too-early)

PR #1038, opened 2017-01-26
upgrade: Restart the newly upgraded services on compute nodes

jsuchome:
It's better then just starting, in case we need to run the script
again after a failure.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
@@ -92,25 +92,25 @@ zypper --non-interactive up <%= @neutron_agent %>
 crudini --set /etc/neutron/neutron-openvswitch-agent.conf.d/200-crowbar-upgrade.conf ovs ovsdb_interface vsctl
 crudini --set /etc/neutron/neutron-openvswitch-agent.conf.d/200-crowbar-upgrade.conf ovs of_interface ovs-ofctl
 <% end %>
-systemctl start <%= @neutron_agent %>
+systemctl restart <%= @neutron_agent %>
 
 <% if @l3_agent %>
 zypper --non-interactive up <%= @l3_agent %>
-systemctl start <%= @l3_agent %>
+systemctl restart <%= @l3_agent %>
 <% end %>
 
 <% if @metadata_agent %>
 zypper --non-interactive up <%= @metadata_agent %>
-systemctl start <%= @metadata_agent %>
+systemctl restart <%= @metadata_agent %>
 <% end %>
 
 log "Starting nova-compute service"
-systemctl start openstack-nova-compute.service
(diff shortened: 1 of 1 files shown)


chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -92,25 +92,25 @@ zypper --non-interactive up <%= @neutron_agent %>
 crudini --set /etc/neutron/neutron-openvswitch-agent.conf.d/200-crowbar-upgrade.conf ovs ovsdb_interface vsctl
 crudini --set /etc/neutron/neutron-openvswitch-agent.conf.d/200-crowbar-upgrade.conf ovs of_interface ovs-ofctl
 <% end %>
-systemctl start <%= @neutron_agent %>
+systemctl restart <%= @neutron_agent %>
```
aplanas: silly question, but this script will be executed only in non-ha?

jsuchome: Yes, we do not expect neutron agents running on compute nodes. (but there's no guard against it yet)

jsuchome: Ah, I did not meant neutron agents themselves actually, but services that should be handled by pacemaker. Is it correct to deploy `neutron-network` role to compute-ha node?

(Speaking of which, we do not have any code actually thinking about compute-ha)

rhafer: > Is it correct to deploy neutron-network role to compute-ha node

In general not. But there will be at least an l2-agent running on the compute node. In the DVR case there will also be the l3-agent. (In the compute node-HA case I assume those will be managed by pacemaker, but you're right, we have not looked at compute node HA for upgrades yet)

jsuchome: I assume the point @aplanas was raising is that we should not restart the service that is handled by pacemaker. For other cases, restarting the agent service as (it is currently in this script) should be correct

rhafer: > I assume the point @aplanas was raising is that we should not restart the service that is handled by pacemaker

Yes, and that's something we need to worry about as soon as we start working on the compute node HA upgrade.  (It's probably worth to add a checklist item to the associated trello card).
For now I think this PR is ok to be merged.

jsuchome: Added to https://trello.com/c/2qIRkJMO/2060-s49p16-orchestrate-compute-ha-upgrade-of-controller-nodes-in-ha-setup-upgrade-step-12-part-2-ha-compute-ha

PR #1039, opened 2017-01-26
Make initial registration against chef server more robust

dirkmueller:


dirkmueller:
@nicolasbock yeah, ok. we can lower the --retry, as 10 minutes 30 times is quite long indeed.

PR #1041, opened 2017-01-27
network: Manually set datapath-ids on the ovs-bridges (bsc#1022074)

rhafer:
Openvswitch by default sets the datapath-id based on the MAC address of
the bridges. If the underlying interfaces share the same MAC address
(i.e.  when using eth0 for br-fixed and eth0.300 for br-public) we will
endup with duplicate datapath-ids, which break neutron-openvswitch-agent.

https://bugzilla.suse.com/show_bug.cgi?id=1022074

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -140,6 +140,17 @@ def kill_nic(nic)
   end
 end
 
+require "securerandom"
+def get_datapath_id_for_ovsbridge(bridge)
+  node.set["network"]["ovs_datapath_ids"] = {} if node["network"]["ovs_datapath_ids"].nil?
+  unless node["network"]["ovs_datapath_ids"][bridge]
+    datapath_id = SecureRandom.hex(8)
+    node.set["network"]["ovs_datapath_ids"][bridge] = datapath_id
+    Chef::Log.info("Generated datapath_id #{datapath_id} for ovsbridge #{bridge}")
+  end
+  node["network"]["ovs_datapath_ids"][bridge]
+end
+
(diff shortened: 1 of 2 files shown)


chef/cookbooks/network/recipes/default.rb
```diff
@@ -275,6 +286,10 @@ def kill_nic(nic)
       Chef::Log.info("Creating OVS bridge #{bridge} for network #{network.name}")
       Nic::OvsBridge.create(bridge)
     end
+
+    datapath_id = get_datapath_id_for_ovsbridge(bridge)
+    br.datapath_id = datapath_id unless br.datapath_id == datapath_id
```
dirkmueller: can this actually happen? it seems this code is only executed with network.add_ovs_bridge set, which is hopefully done only once per bridge?

dirkmueller:
nit: commit message should start with "network:" not "[network]"

chef/cookbooks/network/recipes/default.rb
rhafer: ```network.add_ovs_bridge``` is always ```true``` during when a specific network needs an ovs_bridge (even if that bridge is already there). It's backed by a node attribute that is set e.g. by the neutron barclamp (https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/app/models/neutron_service.rb#L372)

I've thought about adding the datapath_id code to the ```else``` branch a few lines above to have it only executed when the bridge is created. But that would mean we would never set the id during the upgrade from Cloud 6. So I opted for the current approach

aspiers:
For the record, here's how to manually work around the problem this PR fixes, in case it's required on an older cloud:

    ovs-vsctl set Bridge br-fixed other_config:datapath-id=1234567812345678
    rcnetwork restart

PR #1042, opened 2017-01-27
Adding pull request template

nkrinner:
Adding a template for pull requests to guide contributors to writing
helpful pull request messages.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,12 @@
+Please help potential reviewers to understand this pull request and speed
+up the process by writing a meaningful pull request message.
+
+Answering the following questions can help, but is optional.
+
+**Why is this change necessary?**
+
+**How does it address the issue?**
+
+**Is there additional information worth sharing like links to a Trello
+card, bug references, testing advice or dependencies to other pull
+requests?**


nkrinner:
https://github.com/blog/2111-issue-and-pull-request-templates

vuntz:
What happens when a PR with only one commit is created? So far, the commit message ended up as PR description. Is it still going to be the case with a template?

nkrinner:
@aspiers: I try to not overload the template, but have added a line to remind about additional information like bug references etc.

@vuntz: With a single-commit pr the commit message will be appended after the template lines.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,10 @@
+Please help potential reviewers to understand this pull request and speed up 
+the process by answering the following questions, where appropriate.
+
+# Why was this pull request created?
```
dirkmueller: What's this supposed to ask for? 'because there was a trello card'? Or 'xy crashes when using z'?

I hope it's the latter, but then I'd describe this as "Why is this change necessary?"

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,10 @@
...
+
+# How is this achieved?
```
dirkmueller: By changing the code?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,10 @@
...
+
+# How to test this?
```
dirkmueller: Really? We require textual description of tests again instead of automated tests?

aspiers: Yes, that makes sense to me, so that QA know what to test.  CI very often does not test the change, because unfortunately hardly any of us (except maybe Mate) are in the habit of writing tests for our changes.

dirkmueller: I'm happy to be proven wrong, but we didn't have resources to do a manual per-commit/PR validation.

.github/PULL_REQUEST_TEMPLATE.md
aspiers: Yes, it's the latter, and your suggestion sounds like better text.

.github/PULL_REQUEST_TEMPLATE.md
aspiers: Obviously the goal of this section is to give a TL;DR technical overview of what the commits do.

.github/PULL_REQUEST_TEMPLATE.md
aspiers: Firstly this is not about per-commit validation, since it's a description of how to test the PR as a whole.

Secondly, it doesn't need to be per-PR validation. Even if maintenance QA aren't validating *individual* PRs, surely they are validating batches of changes which include bugfixes from PRs, therefore they still need to know how to validate each fix?  Or am I misunderstanding something, and QA actually don't test any of our bugfixes?

aspiers:
> First of all adding this to a pull request seems wrong to me, i really want all of that in the git commit message.

You seem to be assuming that this is an "exclusive or" situation, i.e. *either* we put that info in the PR description, *or* we put it in the commit messages.  But I don't think that's right - we need information about the "why" and the "how" of all changes at *both* the commit level *and* the PR level.  The commit messages should go into more technical detail, and the PR description should give a higher-level overview, to give reviewers context about what they are about to review (or even to allow people to decide that this is not a good topic for them to review, or that the approach is fundamentally wrong, even before they waste time looking at individual commits).

If the PR only has a single commit then IMHO it's fine for the PR description to have the exact same text as the commit message.  And that's what github offers by default anyway.

Also often the PR description doesn't need to be overkill - no long essay required, just a quick summary of the why and the how.  As you pointed out, the commit messages should contain any required details, so there's no need to duplicate all the details.

> The pr should be more about dependencies to other prs, user or documentation side effects, reviewer, stakeholder and priority clarification.

Yes I totally agree that stuff should be in the PR description.  But it *must* have a "why this change is necessary" section.  That's the bit which in our team we most often forget to include, and it's the most important bit.

> Although I suggest to make use of previous research on this topic rather than reinventing the wheel. Like for example: https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message

I strongly agree with almost all of that article, although in practice the 50 character limit is too small.  But the article doesn't really apply very well to PRs, for example the line-wrapping point is the exact opposite, since it's much better to let GitHub do line-wrapping of PR descriptions.

.github/PULL_REQUEST_TEMPLATE.md
dirkmueller: Yeah, but thats why we link bugs to a change (the bug should have reproducer instructions for QAM). Anyway, it just feels duplicated information, but I'm not objecting, just raising an opinion :)

dirkmueller:
> Yes I totally agree that stuff should be in the PR description.  But it *must* have a "why this change is necessary" section.  That's the bit which in our team we most often forget to include, and it's the most important bit.

I think we're in agreement. Just make it more explicit what you're asking for in the template, like link to bugreport, trellocard, FATE whatever, so that you have a *guide* on what is inspected rather than asking a unspecific question which can be answered in a million different ways. 

In a very extreme case the PR template would be machine readable json with an exact schematics. I'm not proposing that to be done, just as a potential vision of how I'd ideally like to be (combined with a PR dashboard that trends those PRs according to that PR description. Big data yeah!). 

Anyway, I'm not blocking this Review, I'm just raising an opinion. I'm ok when its merged as it is, just giving thoughts on it.  I agree it doesn't have to be perfect on the first try, and I don't want to cause an endless discussion.. I thought its easy to give feedback  :)

nkrinner:
Addressed comments and reworded the template.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,12 @@
+Please help potential reviewers to understand this pull request and speed
+up the process by writing a meaningful pull request message.
+
+Answering the following questions can help, but is optional.
+
+# Why is this change nesessary?
+
+# How does it address the issue?
+
+# Is there addtitional information worth sharing like links to a Trello
```
rsalevsky: The markdown is broken here.

aspiers: Also "additional" typo

nkrinner:
Fixed markdown

PR #1045, opened 2017-01-30
upgrade: Mark successfull router migration on the target node

jsuchome:
That way we should know that there's no need to run the script again.

This is trying to solve the problem where we restart the upgrade after the failure in the step where routers were already evacuated, neutron-server was stopped, but node upgrade did not reach the point where the new resource is configured.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
@@ -35,17 +35,23 @@ set +x
 source /root/.openrc
 set -x
 
+if ssh $hostname "test -e $UPGRADEDIR/crowbar-router-migrated" ; then
+    log "l3 agent was already evacuated from host $hostname"
+    touch $UPGRADEDIR/crowbar-router-migration-ok
+    exit 0
+fi
 
 zypper --non-interactive install openstack-neutron-ha-tool
 
 /usr/bin/neutron-ha-tool --l3-agent-evacuate $hostname --wait-for-router $delete_ns
 ret=$?
 if [ $ret != 0 ] ; then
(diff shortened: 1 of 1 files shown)


jsuchome:
[Restarting after failed nodes upgrade often fails in router migration](https://trello.com/c/FqsAPBHD/287-restarting-after-failed-nodes-upgrade-often-fails-in-router-migration)

chef/cookbooks/crowbar/templates/default/crowbar-router-migration.sh.erb
```diff
@@ -35,17 +35,23 @@ set +x
 source /root/.openrc
 set -x
 
+if ssh $hostname "test -e $UPGRADEDIR/crowbar-router-migrated" ; then
+    log "l3 agent was already evacuated from host $hostname"
+    touch $UPGRADEDIR/crowbar-router-migration-ok
+    exit 0
+fi
```
rhafer: It feels somehow wrong for me to do this from within here. Have you considered doing it from the rails model in ```
evacuate_network_node``` (e.g. around here: https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/api/upgrade.rb#L776)

jsuchome: Yes, I did.
I was not sure which way it is better, but here it at least felt correct to create that variant of -ok file (`crowbar-router-migrated`) as it seems kind of consistent with other scripts. And if creating of file is here, why not checking for it's presence... (it was already established that we have ssh access to other nodes of the cluster)

rhafer: As said, it looks a little weird to do it this way here. But if other reviewers are ok with it am willing to +1 it as well.

PR #1047, opened 2017-01-30
upgrade: follow up for openstack backup creation

MaximilianMeister:
follow up for https://github.com/crowbar/crowbar-core/pull/1011

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -400,7 +400,7 @@ module Api
 
       def openstackbackup
         crowbar_lib_dir = "/var/lib/crowbar"
-        dump_path = "#{crowbar_lib_dir}/upgrade/6-to-7-openstack_dump.sql"
+        dump_path = "#{crowbar_lib_dir}/upgrade/6-to-7-openstack_dump.sql.gz"
         if File.exist?(dump_path)
           Rails.logger.debug("OpenStack backup already exists. Skipping...")
           return
@@ -437,7 +437,8 @@ module Api
 
         Rails.logger.debug("Creating OpenStack database dump")
         db_dump = run_cmd(
-          "PGPASSWORD=#{psql[:pass]} pg_dumpall -h #{psql[:host]} -U #{psql[:user]} > #{dump_path}"
+          "PGPASSWORD=#{psql[:pass]} pg_dumpall -h #{psql[:host]} -U #{psql[:user]} | " \
+            "gzip > #{dump_path}"
         )
         unless db_dump[:exit_code].zero?
           Rails.logger.error(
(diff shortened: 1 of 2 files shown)


jsuchome:
So first commit could be removed, right?

MaximilianMeister:
> So first commit could be removed, right?

true

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -348,9 +348,10 @@ def openstackbackup
         ::Crowbar::UpgradeStatus.new.start_step(:backup_openstack)
 
         crowbar_lib_dir = "/var/lib/crowbar"
-        dump_path = "#{crowbar_lib_dir}/upgrade/6-to-7-openstack_dump.sql"
+        dump_path = "#{crowbar_lib_dir}/upgrade/6-to-7-openstack_dump.sql.tar.gz"
```
jsuchome: Shouldn't we put it under `/var/lib/crowbar/backup/` , just like the crowbar (admin) backup?

MaximilianMeister: done in the latest commit

MaximilianMeister:
[#followup address dirk's comments for the openstack backup](https://trello.com/c/zE2RyPHg/284-followup-address-dirk-s-comments-for-the-openstack-backup)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -400,9 +400,10 @@ def services
 
       def openstackbackup
         crowbar_lib_dir = "/var/lib/crowbar"
-        dump_path = "#{crowbar_lib_dir}/upgrade/6-to-7-openstack_dump.sql"
+        dump_path = "#{crowbar_lib_dir}/backup/6-to-7-openstack_dump.sql.tar.gz"
```
skazi0: this is not `.tar.gz` but rather a plain `.gz`.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -400,9 +400,10 @@ def services
...
         if File.exist?(dump_path)
           Rails.logger.debug("OpenStack backup already exists. Skipping...")
+          ::Crowbar::UpgradeStatus.new.end_step
```
skazi0: I'm not sure if this is the right logic. Silently skipping the backup if the file already exists? I'd rather make it like the crowbar backup that the filename is passed from the UI and contains some timestamp so that there can be more than one without conflict.

MaximilianMeister: hmm i think we should rather keep it as 1 backup only, if we allow multiple backups we might run into some unforeseeable disk space issues, as the backups can be quite large (GB) in size

crowbar_framework/app/models/api/upgrade.rb
MaximilianMeister: thanks, fixed

crowbar_framework/app/models/api/upgrade.rb
skazi0: Then maybe it would be better to fail this step if the backup exists already?

skazi0: ...or at least log the message as a warning not debug.

PR #1048, opened 2017-01-30
upgrade: partially backport #1031

MaximilianMeister:
https://github.com/crowbar/crowbar-core/pull/1031

### DIFF
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -148,6 +148,15 @@ class Api::UpgradeController < ApiController
         }
       }
     }, status: :unprocessable_entity
+  rescue StandardError => e
+    ::Crowbar::UpgradeStatus.new.end_step(
+      false,
+      backup_crowbar: {
+        data: e.message,
+        help: "Crowbar has failed. Check /var/log/crowbar/production.log for details."
+      }
+    )
+    raise e
   ensure
     @backup.cleanup unless @backup.nil?
   end
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -22,6 +22,8 @@ class Api::UpgradeController < ApiController
   end
 
   def prepare
+    ::Crowbar::UpgradeStatus.new.start_step(:prepare)
+
     if Api::Upgrade.prepare(background: true)
       head :ok
     else
@@ -34,9 +36,9 @@ class Api::UpgradeController < ApiController
         }
       }, status: :unprocessable_entity
     end
-  rescue Crowbar::Error::StartStepRunningError,
-         Crowbar::Error::StartStepOrderError,
-         Crowbar::Error::SaveUpgradeStatusError => e
+  rescue ::Crowbar::Error::StartStepRunningError,
+         ::Crowbar::Error::StartStepOrderError,
+         ::Crowbar::Error::SaveUpgradeStatusError => e
     render json: {
       errors: {
         prepare: {
@@ -99,10 +101,7 @@ class Api::UpgradeController < ApiController
     else
       render json: check
     end
-  rescue Crowbar::Error::StartStepRunningError,
-         Crowbar::Error::StartStepOrderError,
-         Crowbar::Error::EndStepRunningError,
-         Crowbar::Error::SaveUpgradeStatusError => e
+  rescue Crowbar::Error::UpgradeError => e
     render json: {
       errors: {
         repocheck_crowbar: {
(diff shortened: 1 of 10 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -121,6 +121,19 @@ def checks
             upgrade_status.end_step
           end
         end
+      rescue ::Crowbar::Error::StartStepRunningError,
+             ::Crowbar::Error::StartStepOrderError,
+             ::Crowbar::Error::SaveUpgradeStatusError => e
+        raise ::Crowbar::Error::UpgradeError.new(e.message)
+      rescue StandardError => e
+        ::Crowbar::UpgradeStatus.new.end_step(
```
jsuchome: Here we should probably do explicit check if the step is running. Maybe it should be in all these exceptions, but in prechecks it's especially needed because it can be run without actually starting the step...

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -21,11 +21,9 @@ def show
     render json: Api::Upgrade.status
   end
 
-  def update
-    head :not_implemented
-  end
-
   def prepare
+    ::Crowbar::UpgradeStatus.new.start_step(:prepare)
```
jsuchome: You have it twice, once in model, once in controller

crowbar_framework/app/models/api/upgrade.rb
MaximilianMeister: good point, but i guess it doesn't need to be in all the steps, as they are always running implicitly once you hit this exception. so i guess your suggestion applies for the prechecks here.

jsuchome: Yes, just do it for prechecks

MaximilianMeister: i've added this as a separate commit now, as we'll need to forward-port it

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: thanks, must have been rebasing mistake... fixed

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -103,10 +101,7 @@ def adminrepocheck
     else
       render json: check
     end
-  rescue Crowbar::Error::StartStepRunningError,
-         Crowbar::Error::StartStepOrderError,
-         Crowbar::Error::EndStepRunningError,
-         Crowbar::Error::SaveUpgradeStatusError => e
+  rescue Crowbar::Error::UpgradeError => e
```
skazi0: You should catch SystemError here too. It's caught in the model and will set failed status but the response will be still generic 500.

crowbar_framework/app/models/api/upgrade.rb
skazi0: IMO this should be done in all cases. Otherwise if you have something running fine and you try to start it again it will set the status to failed, but the step is still running. In case of 'running' status the response should say 'failed' but the status should still say 'running'. For async steps the trigger endpoint should not succeed in such cases.

jsuchome: @skazi0  If you try to start running step again, you should get `Crowbar::Error::StartStepRunningError` and no setting status to failed

crowbar_framework/app/controllers/api/upgrade_controller.rb
MaximilianMeister: thanks, fixed

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -64,6 +64,15 @@ def upgrade!
             message: msg
           }
         end
+      rescue StandardError => e
```
jsuchome: Here you are not caching usual exceptions from the library, so if I start the step in the incorrect order, it ends in this block with `StandardError` which is wrong

skazi0: This will also catch StartStepRunningError etc. which should bubble up to the controller. If we leave this here, `start_step()` when status is `running` will turn the status to `failed`.

MaximilianMeister:
[(2) S54P3 Catch exceptions crashing crowbar and end the upgrade step with failure](https://trello.com/c/dVrsUTHe/272-2-s54p3-catch-exceptions-crashing-crowbar-and-end-the-upgrade-step-with-failure)

skazi0:
@MaximilianMeister make sure to forwardport the last changes

MaximilianMeister:
> make sure to forwardport the last changes

that's why i already included them as separate commits ;-)

PR #1050, opened 2017-01-31
upgrade: Mark successfull router migration on the target node

jsuchome:
That way we should know that there's no need to run the script again.

An alternative to https://github.com/crowbar/crowbar-core/pull/1045


jsuchome:
[Restarting after failed nodes upgrade often fails in router migration](https://trello.com/c/FqsAPBHD/287-restarting-after-failed-nodes-upgrade-often-fails-in-router-migration)

PR #1052, opened 2017-01-31
[3.0] upgrade: Implement a dummy cloud 7 api

MaximilianMeister:
when a client on cloud 6 requests an upgrade route that only exists
in cloud 7, a StartStepOrder exception should be thrown


MaximilianMeister:
[#crowbarctl running an upgrade cloud7 only step on cloud6 fails badly](https://trello.com/c/zPnN2lP8/289-crowbarctl-running-an-upgrade-cloud7-only-step-on-cloud6-fails-badly)

PR #1059, opened 2017-02-02
Don't do platform checks against clusters in JS validation

tserong:
Found while investigating https://bugzilla.suse.com/show_bug.cgi?id=956582
If a role specifies a platform constraint, the JavaScript validation will
try to enforce this when assigning a cluster to that role.  But, clusters
don't actually have platform information attached, so the check will always
fail (this isn't a problem for platform *exclusion* tests, which effectively
work by accident).  This commit skips all platform checks for clusters,
which seems to be the approach taken by the backend code in service_object.rb
(violates_platform_constraint? and violates_exclude_platform_constraint?)

### DIFF
--- a/crowbar_framework/vendor/assets/javascripts/jquery/nodeList.js
+++ b/crowbar_framework/vendor/assets/javascripts/jquery/nodeList.js
@@ -370,37 +370,43 @@
         }
       }
 
-      if (constraints.platform !== undefined && constraints.platform) {
-        var platforms = $.map(constraints.platform, function(c_version, c_platform) {
-          return (platform === c_platform) && self.equalOrMatchVersion(platform_version, c_version);
-        });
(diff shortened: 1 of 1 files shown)


tserong:
Seriously? Can we ignore Hound here please? All I did was add `if (!cluster && !remotes) {` and indent a block...

tserong:
@MaximilianMeister not sure if you're the most appropriate person to review or not, please suggest alternates if not :-)

MaximilianMeister:
> Seriously? Can we ignore Hound here please? All I did was add if (!cluster && !remotes) { and indent a block...

you were just the first lucky one to change some JS code after introducing hound, I'd just ignore it too ;-)

> not sure if you're the most appropriate person to review or not, please suggest alternates if not :-)

probably not, but the change looks good to me. nevertheless maybe @vuntz could have a quick look?

PR #1060, opened 2017-02-02
upgrade: Enable different modes for upgrading the nodes

jsuchome:
'all' triggers the upgrade of all nodes
'controllers' upgrades only controller nodes
<NAME> upgrades the node identified by NAME

### DIFF
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -140,28 +140,9 @@ class Api::UpgradeController < ApiController
   # The method runs asynchronously, so there's a need to poll for the status and possible errors
   def nodes
     if params[:component]
-      case params[:component]
-      when "all"
-        ::Crowbar::UpgradeStatus.new.start_step(:nodes)
-        Api::Upgrade.nodes
-        head :ok
-      when "controllers"
-        render json: {
-          errors: {
-            nodes: {
-              data: "Not implemented yet."
-            }
-          }
-        }, status: :not_implemented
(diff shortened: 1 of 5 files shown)


jsuchome:
@rhafer @MaximilianMeister 

So, what I see as the main problem is that I'm not using the status library correctly.

I thought I would just start the "nodes" step when user initiates the controllers upgrade. But after these are upgraded, the step stays as 'running' and does not indicate to user that it's possible to continue with compute nodes. Well, there's something, the value of "substep" at least.

However the bigger problem is, what to do if upgrade of one compute node fails - should we mark the whole step as failed?

To me it seems like we need to extend the meaning of substeps, so they could be running/passed/failed as well. Or is there any better idea?

jsuchome:
BTW this is not tested, but I believe it should work for "all" case. (As seems to be confirmed by passing unit tests)

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -140,28 +140,15 @@ def services
...
-            }
-          }
-        }, status: :not_implemented
-      end
+      # FIXME: we need to omit start of the step if:
+      # - it is running
+      # - controller substep is marked as finished (and there is no compute substep yet)
+      # - params[:component] is name of the node
+      # OR we could introduce new value for step status and mark the nodes step status as
+      # waiting/paused/* after the controllers are upgraded
```
jsuchome: OR just do the start (for single node option )only when compute substep has failed

jsuchome:

PR #1061, opened 2017-02-02
upgrade: Delete running indicator after cancel

MaximilianMeister:


jsuchome:
The file is already deleted when you call cancel, see https://github.com/crowbar/crowbar-core/pull/1043

MaximilianMeister:
already done in the upgrade status library

PR #1062, opened 2017-02-02
upgrade: Enable node upgrade status api

MaximilianMeister:
this lists all upgraded and all non upgraded nodes

## NOTE
~~`crowbarctl` PR will follow~~

here it is https://github.com/crowbar/crowbar-client/pull/147

## example response
```bash
{
  "upgraded": [
  ],
  "not_upgraded": [
    "d52-54-77-77-77-03.emil.suse.de",
    "d52-54-77-77-77-01.emil.suse.de",
    "d52-54-77-77-77-02.emil.suse.de",
    "d52-54-77-77-77-04.emil.suse.de"
  ]
}
```

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -23,6 +23,19 @@ module Api
         ::Crowbar::UpgradeStatus.new.progress
       end
 
+      def node_status
+        not_upgraded = if ::Crowbar::UpgradeStatus.new.pending?(:prepare) || \
+            ::Crowbar::UpgradeStatus.new.running?(:prepare)
+          ::Node.all.reject(&:admin?).map(&:name)
+        else
+          ::Node.find("state:crowbar_upgrade").map(&:name)
+        end
+        {
+          upgraded: ::Node.all.reject(&:admin?).select(&:ready_after_upgrade?).map(&:name),
+          not_upgraded: not_upgraded
+        }
+      end
+
       #
       # prechecks
       #
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -24,11 +24,10 @@ module Api
       end
 
       def node_status
-        not_upgraded = if ::Crowbar::UpgradeStatus.new.pending?(:prepare) || \
-            ::Crowbar::UpgradeStatus.new.running?(:prepare)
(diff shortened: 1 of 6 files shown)


jsuchome:
Do we really need a backport here?

MaximilianMeister:
> Do we really need a backport here?

yes, as we will have the same `crowbarctl` on cloud6

MaximilianMeister:

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -23,6 +23,13 @@ def status
         ::Crowbar::UpgradeStatus.new.progress
       end
 
+      def node_status
+        {
+          upgraded: ::Node.all.reject(&:admin?).select(&:ready_after_upgrade?).map(&:name),
+          not_upgraded: ::Node.find("state:crowbar_upgrade").map(&:name)
```
skazi0: are we sure that this will give us all nodes which are not upgraded yet?

MaximilianMeister: i guess only when it's executed after the `prepare` step. @jsuchome is this an issue?

jsuchome: Hm, you are right that before `prepare` (heck, even before `nodes`) it does not make sense.
Maybe we could just check for the current step here?

MaximilianMeister: should be fixed now

skazi0: what if something gets out-of-sync somehow? iirc I had some cases where node switched to 'no-data' and then when it returned it was 'ready' not 'upgrade'... but I might be wrong.

jsuchome: That "no data" state should be covered by `:ready_after_upgrade?` I think

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -23,6 +23,19 @@ def status
         ::Crowbar::UpgradeStatus.new.progress
       end
 
+      def node_status
+        not_upgraded = if ::Crowbar::UpgradeStatus.new.pending?(:prepare) || \
```
jsuchome: Seems like `passed(step)?` method in the library would be handy...

MaximilianMeister: done

PR #1063, opened 2017-02-02
upgrade: Adapt the swift configuration on compute nodes

jsuchome:
Before is the compute node fully upgraded, we upgrade only selected
services that are needed for instances live-migration.

The configuration needs to be adapted as well, to be able to talk
to upgraded services at controller nodes.


jsuchome:
[Can we have swift-storage at compute nodes?](https://trello.com/c/IvtWHvT2/290-can-we-have-swift-storage-at-compute-nodes)

PR #1066, opened 2017-02-04
[stable/3.0] Enable node upgrade status api 

jsuchome:
Partial backport of https://github.com/crowbar/crowbar-core/pull/1062 - node status should be just always empty here


jsuchome:
@MaximilianMeister sorry, but as it is a port of your PR, I've dared to put you here...

PR #1067, opened 2017-02-04
upgrade: Extend running? method to check if any step is running

jsuchome:


crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -130,8 +130,14 @@ def end_step(success = true, errors = {})
       end
     end
 
+    # Check if given step is running.
+    # Without argument, check if any step is running
     def running?(step_name = nil)
-      step = progress[:steps][step_name || current_step]
+      if step_name.nil?
+        return progress[:steps].select { |_, s| s[:status] == :running }.any?
```
Itxaka: only a nit, but I find it good to use decent naming, even if they are gonna be discarded, when doing one liners as to be more clear on a first look. This had me looking at it for a good 30 seconds trying to understand what that strange `|_` was until I got that its a var name that its not gonna be used.

jsuchome: I might agree with you, but Hound would not :-)
It would point exactly to the fact that the variable is not used and I should use `_`

crowbar_framework/lib/crowbar/error/upgrade_status.rb
```diff
@@ -24,7 +24,9 @@ def initialize(step_name = "")
 
     class StartStepRunningError < StandardError
       def initialize(step_name = "")
-        super("The step #{step_name} has already been started")
+        msg = "Some step is already running."
```
skazi0: Not sure what is the recommended way of doing such things in ruby but I'd switch to a regular `if ... else ...` to make this part clear.

crowbar_framework/lib/crowbar/upgrade_status.rb
skazi0: @Itxaka this seems to be a common practice in ruby. I hit it already a couple of times.

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -93,8 +93,8 @@ def start_step(step_name)
           @logger.warn("The step #{step_name} doesn't exist")
           raise Crowbar::Error::StartStepExistenceError.new(step_name)
         end
-        if running? step_name
-          @logger.warn("The step has already been started")
+        if running?
+          @logger.warn("Some step is already running.")
           raise Crowbar::Error::StartStepRunningError.new(step_name)
```
skazi0: This exception would say that `step_name` is running while the check is actually for "any" step.

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -161,7 +167,7 @@ def cancel_allowed?
         :backup_crowbar,
         :repocheck_crowbar,
         :admin
-      ].include?(current_step) && !running?(:admin)
+      ].include?(current_step) && !running?
```
skazi0: good change! :)

crowbar_framework/lib/crowbar/upgrade_status.rb
rhafer: IIRC the recommendation is to prefix the variable name with ```_``` if it is unused. (https://github.com/bbatsov/ruby-style-guide#underscore-unused-vars)

crowbar_framework/lib/crowbar/upgrade_status.rb
jsuchome: That's why I changed the exception message as well.

crowbar_framework/lib/crowbar/upgrade_status.rb
jsuchome: Hm, that would probably be better...

crowbar_framework/lib/crowbar/upgrade_status.rb
skazi0: But shouldn't the exception be thrown without an argument?

jsuchome: Hm, you are right, I should probably remove the argument from the exception completely

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -136,8 +136,14 @@ def end_step(success = true, errors = {})
       end
     end
 
+    # Check if given step is running.
+    # Without argument, check if any step is running
     def running?(step_name = nil)
-      step = progress[:steps][step_name || current_step]
+      if step_name.nil?
+        return progress[:steps].select { |_key, s| s[:status] == :running }.any?
```
Itxaka: yeah, much nicer!

PR #1069, opened 2017-02-06
upgrade: Save 'node_upgrade_state' into the role, not node

jsuchome:
If node_upgrade_state is saved into the node object, it could get
rewritten from chef-client running at the node.


jsuchome:
[Nodes upgrade marked as finished after single node 'upgraded' twice](https://trello.com/c/Dqw04leb/294-nodes-upgrade-marked-as-finished-after-single-node-upgraded-twice)

PR #1070, opened 2017-02-06
Increase timeouts for upgrade related actions

jsuchome:
Not nice, I know, but I've met some timeouts  already several times, so let's raise them.

Ideally, we would make them configurable.


jsuchome:
We have now https://trello.com/c/ivFOALtv/2257-make-the-timeouts-configurable

PR #1071, opened 2017-02-07
upgrade: Fix the check for upgrade finish

jsuchome:


crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -126,7 +126,7 @@ def end_step(success = true, errors = {})
           status: success ? :passed : :failed,
           errors: errors
         }
-        if finished? && success
+        if upgrade_steps_6_7.last && success
```
skazi0: This seems to be wrong. You check for the last step name instead of current_step being the last one.

I would actually do it differently:
something like:
```
if finished?
...
end
```

and:
```
def finished?
  passed(upgrade_steps_6_7.last)
end
```

`progress[:steps][current_step][:status]` is set just one line above the `finished?` check. You don't even need to check `success` as it's already used to set the `:status`.

wdyt?

jsuchome: oh, this is indeed wrong

PR #1072, opened 2017-02-07
[3.0] Uses World Wide Name as most persistent name for disks (bsc#1022302)

bhdn:
Uses World Wide Name as most persistent name for disks

Cherry-picked from 5697817479c0a7a71805d5935e09a46d95aaa62c


rsalevsky:
Hound is wrong in this case, no need to fix it.

PR #1073, opened 2017-02-07
upgrade: Change the check for list of upgraded/not upgraded nodes

jsuchome:
Original ready_after_upgrade? is not 100% correct, there are still some
later node actions after this is set. So let's check node.upgraded?
However, this does only return correct results during the upgrade, so
additional check if full upgrade is finished is needed.

Requires https://github.com/crowbar/crowbar-core/pull/1071

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -24,13 +24,24 @@ module Api
       end
 
       def node_status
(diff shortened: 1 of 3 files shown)


jsuchome:
[Check status nodes should check for node.upgraded? and not upgraded is the rest](https://trello.com/c/CXhBojPJ/302-check-status-nodes-should-check-for-node-upgraded-and-not-upgraded-is-the-rest)

PR #1076, opened 2017-02-08
Ceph prechecks: New texts, format and a version check

jsuchome:


jsuchome:
[(2) S55P7 Ceph upgrade: let the prechecks check the SES version](https://trello.com/c/b4urENSV/222-2-s55p7-ceph-upgrade-let-the-prechecks-check-the-ses-version)

jsuchome:
@skazi0 I've changed a return format a bit, to keep up with some other ones. This will probably require some change in crowbar-ui ...

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -99,16 +99,23 @@ def health_check
...
-          unless ssh_retval[:stdout].include? "HEALTH_OK"
-            ret["errors"] = [
-              "ceph cluster health check failed with #{ssh_retval[:stdout]}"
-            ]
-          end
+        ret = {}
+        ceph_node = ::Node.find("roles:ceph-mon AND ceph_config_environment:*").first
+        return ret if ceph_node.nil?
+
+        ssh_retval = ceph_node.run_ssh_cmd("LANG=C ceph health 2>&1")
```
tserong: This probably wants to be `LANG=C ceph --connect-timeout 5 health 2>&1"`; that way it'll wait five seconds to get the status then die. Otherwise if the MONs aren't quorate, this command will hang forever.

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -99,16 +99,23 @@ def health_check
...
+        unless ssh_retval[:stdout].include? "HEALTH_OK"
```
tserong: I have once concern about this (which needn't stop this PR going in, it can be a separate issue if you like) -- if I have a cluster with too many (or two few) PGs per OSD, that is otherwise functional, `ceph health` will print something like `HEALTH_WARN too many PGs per OSD (624 > max 300)`. That's a condition that shouldn't prevent an upgrade IMO, because the cluster is otherwise functional, and it's not necessarily possible to even fix that case without destroying pools or adding more OSDs.

jsuchome: Well, what other command should I use for a health check result of which should prevent the upgrade?

crowbar_framework/app/models/api/crowbar.rb
jsuchome: What does it print if it dies? Or does it at least exit with non-zero?

skazi0:
If I read it correctly, the only changes relevant for the UI are in "error codes". Is that correct?

jsuchome:
> If I read it correctly, the only changes relevant for the UI are in "error codes". Is that correct?

Yes, that's what I meant. (And actually, this should be usable from Cloud6 only, I forgot to start there)

jsuchome:
@tserong your comment is hidden after later changes, so let me restate it here: What better health check should I use to cover your WARN scenario?

tserong:
> @tserong your comment is hidden after later changes, so let me restate it here: What better health check should I use to cover your WARN scenario?

I'm honestly not sure what the best way to deal with this is. The too many/few PGs per OSD warning can be suppressed by tweaking ceph.conf, but that'd mean restarting the MONs (not a good idea for a health check). Those parameters can also be changed at runtime. This might be a bad idea, but you could do this:

````
ceph tell 'mon.*' injectargs '--mon_pg_warn_min_per_osd 0'
ceph tell 'mon.*' injectargs '--mon_pg_warn_max_per_osd 0'
ceph health
````

That'd make it ignore those two warnings and give HEALTH_OK. Of course, those warnings would then remain suppressed, which might not be desirable.

Maybe someone else has an idea... @smithfarm, @jan--f, @theanalyst, @kmroz we're trying to find out if a ceph cluster is basically healthy and stable (everything is up, nothing is dead or dying), before letting the user upgrade from SES2.1->4.0 and/or before upgrading SOC. But we don't care about HEALTH_WARN for things like too many/few PGs per OSD (and maybe some other stuff, now that I think of it, like legacy tunables?) Any ideas how best to achieve this?

**Comment by jsuchome** [comment]
@tserong OK, let's make it separate PR once we decide on a way of testing the health.

File: crowbar_framework/app/models/api/crowbar.rb [hunk 3]
```diff
@@ -99,16 +100,31 @@ def health_check
...
           end
+          return ret
         end
+        # ceph --version
+        # SES2.1:
+        # ceph version 0.94.9-93-g239fe15 (239fe153ffde6a22e1efcaf734ff28d6a703a0ba)
+        # SES4:
+        # ceph version 10.2.4-211-g12b091b (12b091b4a40947aa43919e71a318ed0dcedc8734)
+        ssh_retval = ceph_node.run_ssh_cmd("LANG=C ceph --version | cut -d ' ' -f 3")
+        ret[:old_version] = true if ssh_retval[:stdout] < "10"
```
skazi0: do we really want to compare strings here?

jsuchome: Why not? What could possibly go wrong?

It's also covered by unit tests, btw.

skazi0: For those two cases in the comment above this should be fine but if you somehow get e.g. version 2.0 it would end up as newer than 10.0

Itxaka: This is compared alphabetically I think, so its prone to explode. Please do not do numerical comparisons with strings, everytime you do a kitten dies.
```
$ irb
 :001 > "9" < "10"
 => false 
 :002 > "10" < "10"
 => false 
 :003 > "11" < "10"
 => false 
```

jsuchome: Hm, that's true, but it cannot be "2.0". We should really have only these versions.

Would you like me to change the check to e.g. `.ssh_retval[:stdout].split("-").first.to_f < 10.2`

smithfarm: could just change `if ssh_retval[:stdout] < "10"` to `if int(ssh_retval[:stdout]) < 10` ?

smithfarm: sorry, `int(ssh_retval[:stdout].split('.')[0]`

smithfarm:
> That'd make it ignore those two warnings and give HEALTH_OK. Of course, those warnings would then remain suppressed, which might not be desirable.

I think that's what we're looking at, but we could:

1. find out what the actual pg_min, pg_max values are
2. set them to zero
3. get the health (untainted by pg_min, pg_max)
4. reinstate the values

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -99,16 +100,31 @@ def health_check
...
           end
+          return ret
         end
+        # ceph --version
+        # SES2.1:
+        # ceph version 0.94.9-93-g239fe15 (239fe153ffde6a22e1efcaf734ff28d6a703a0ba)
+        # SES4:
+        # ceph version 10.2.4-211-g12b091b (12b091b4a40947aa43919e71a318ed0dcedc8734)
+        ssh_retval = ceph_node.run_ssh_cmd("LANG=C ceph --version | cut -d ' ' -f 3")
+        ret[:old_version] = true if ssh_retval[:stdout].split("-").first.to_f < 10.2
```
skazi0: You don't need the split. `to_f` will parse the string until it makes sense and then return (e.g.`"2-3".to_f = 2.0`).

jsuchome: ok, coming with next version...

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -89,7 +89,8 @@ def addons
       def health_check
         ret = {}
         unready = []
-        NodeObject.find_all_nodes.each do |node|
+        # We are ignoring the ceph nodes, at they should already be in crowbar_upgrade state
```
rhafer: typo s/at/as

jsuchome:
> Are you considering to implement @smithfarm's suggestiong for the health check?

I would rather address it in separate PR (as stated before). If possible I'd rather have one command for a health check and if that is not possible, maybe longer discussion will be needed.

PR #1077, opened 2017-02-08
[stable/3.0] Enable preparing only ceph nodes for the upgrade

jsuchome:


jsuchome:

jsuchome:
@rhafer adapted

PR #1078, opened 2017-02-08
provisioner: set blocksize option on tftp server to 1024

Itxaka:
It seems that a variety of tftp clients can be picky about
the options, so there is a mechanism to restrict what
options the server offers to the clients to increase compatibility
with a wide variety of clients.

The qemu uefi tftp client doesnt seem to work very well negotiating the
blocksize. Forcing it to be 1024 makes it work.

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/tftp.service.erb
+++ b/chef/cookbooks/provisioner/templates/default/tftp.service.erb
@@ -3,4 +3,4 @@ Description=Tftp Server
 
 [Service]
 Type=simple
-ExecStart=/usr/sbin/in.tftpd -u tftp -s <%= @tftproot %> -m /etc/tftpd.conf -L -a <%= @admin_ip %>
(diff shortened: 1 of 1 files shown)


Itxaka:
[(5) make uefi job work](https://trello.com/c/nUO30l24/461-5-make-uefi-job-work)

bmwiedemann:
typo in commit message. blksize stands for block-size, not bulk

aplanas:
I am OK in jenkins is OK

vuntz:
Could it possibly slow down the tftp transfer?

Itxaka:
@vuntz yes, according to https://tools.ietf.org/html/rfc2348 it would be around 42% slower if it automatically goes to the default 512 block size.

Im still playing with it to see if its the negotiation or the size what makes it fail. pxe clients are a pita!

Itxaka:
@vuntz Nice! Looks like a block size of 1024 works flawlessly and that should only be 10% lower than the default 1428!

Itxaka:
@bmwiedemann @vuntz went over the pxe booting without issues on the CI :D

vuntz:
@Itxaka it's probably okay in this case, but please remember that we're frozen right now, so we shouldn't merge changes at the moment.

PR #1079, opened 2017-02-08
[stable/3.0] Extra page for preparing ceph nodes for the upgrade

jsuchome:
Before the Cloud upgrade, ceph nodes must be manually upgraded to
the latest version of SES. Before that, we need to move them to
crowbar_upgrade state.
There is no UI for SES upgrade, but we can provide simple UI for
the preparation that is needed for both SES and following SOC upgrades.

Requires https://github.com/crowbar/crowbar-core/pull/1077

ceph_pre_upgrade

### DIFF
--- a/crowbar_framework/app/controllers/ceph_pre_upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/ceph_pre_upgrade_controller.rb
@@ -18,6 +18,13 @@ class CephPreUpgradeController < ApplicationController
     ceph_nodes = get_ceph_nodes
     @prepared = nodes_prepared(ceph_nodes)
     @no_ceph_nodes = ceph_nodes.empty?
+
+    respond_to do |format|
+      format.json do
+        render json: { prepared: @prepared, no_ceph_nodes: @no_ceph_nodes }
+      end
+      format.html
+    end
   end
 
   def prepare
@@ -41,12 +48,19 @@ class CephPreUpgradeController < ApplicationController
       status = :unprocessable_entity
     end
 
-    if status == :ok
-      flash[:notice] = success_msg
-    else
-      flash[:alert] = error_msg
+    respond_to do |format|
+      format.json do
+        render json: status
+      end
+      format.html do
+        if status == :ok
+          flash[:notice] = success_msg
+        else
+          flash[:alert] = error_msg
+        end
+        redirect_to ceph_pre_upgrade_url
+      end
     end
-    redirect_to ceph_pre_upgrade_url
   end
 
   private
--- a/crowbar_framework/app/controllers/ceph_pre_upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/ceph_pre_upgrade_controller.rb
@@ -15,8 +15,9 @@
 #
 class CephPreUpgradeController < ApplicationController
   def index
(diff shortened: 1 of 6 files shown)


jsuchome:

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -558,6 +559,17 @@ en:
       active: 'Active'
       active_repo_tooltip: 'Managed nodes will be automatically configured to use repositories only if they are marked as active'
 
+  ceph_pre_upgrade:
+    index:
+      title: 'Preparation of Ceph Nodes'
+      hint: |
+        Some explanatory text why this is needed.
+        This might take more lines.
```
tserong: How about something like this?

> The upgrade from SUSE Enterprise Storage 2.1 to 4 must be performed as outlined in the [SUSE Enterprise Storage 4 documentation](docserv.suse.de/documents/Storage_4/ses4-admin/html/cha.ceph.upgrade.html). For this to work, the Ceph nodes must first be put into a prepared state, so that Crowbar will leave them alone for the duration of the upgrade.

jsuchome:
Adding @dirkmueller and @vuntz here, to make sure this is known and approved solution

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -559,6 +560,21 @@ en:
       active: 'Active'
       active_repo_tooltip: 'Managed nodes will be automatically configured to use repositories only if they are marked as active'
 
+  ceph_pre_upgrade:
+    index:
+      title: 'Preparation of Ceph Nodes'
+      hint_html: |
+        The upgrade from SUSE Enterprise Storage 2.1 to 4 must be performed as outlined in the <a href="https://www.suse.com/documentation/ses-4/book_storage_admin/data/ceph_upgrade_to4.html">SUSE Enterprise Storage 4 documentation</a>.<br>
+        For this to work, the Ceph nodes must first be put into a prepared state, so that Crowbar will leave them alone for the duration of the upgrade of both SUSE Enterprise Storage and SUSE OpenStack Cloud products.
```
vuntz: s/leave them alone/stop configuring them/

crowbar_framework/app/controllers/ceph_pre_upgrade_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def prepare
+    status = :ok
+    error_msg = ""
+    prepared = nodes_prepared(get_ceph_nodes)
+
+    begin
+      service_object = CrowbarService.new(Rails.logger)
+      if prepared
```
vuntz: I probably wouldn't have this API act as a toggle; seems better to me to have the behavior change depending on a parameter passed to the API call.

jsuchome: I just haven't find a way how to pass a parameter with a submit button...

skazi0: You can add hidden field in the form e.g.: `%input{type: "hidden", name: "prepare_action", value: @prepared ? 'revert' : 'prepare'}`. This will be available as `params[:prepare_action]` inside the controller.

PR #1080, opened 2017-02-08
upgrade: Move "crowbar_upgrade_step" attribute to node role

rhafer:
Similar to the previous "node_upgrade_state" change this is done to
prevent a race condition with a chef-client running on the node that
might overwrite "crowbar_wall".

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -24,7 +24,7 @@
 
 return unless node[:platform_family] == "suse"
 
-upgrade_step = node["crowbar_wall"]["crowbar_upgrade_step"] || "none"
+upgrade_step = node["crowbar_upgrade_step"] || "none"
 
 Chef::Log.info("Current upgrade step: #{upgrade_step}")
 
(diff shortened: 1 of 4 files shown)


rhafer:
This is largely untested and requires changes on the cloud 6 branch as well as crowbar-ha.

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -24,7 +24,7 @@
 
 return unless node[:platform_family] == "suse"
 
-upgrade_step = node["crowbar_wall"]["crowbar_upgrade_step"] || "none"
+upgrade_step = node["crowbar_upgrade_step"] || "none"
```
jsuchome: Don't we need `node["crowbar"]` here?

nicolasbock: @jsuchome I think you are right.

jsuchome: No, this is actualy correct...

rhafer:
[nodes lose "crowbar_upgarde_step" attribute due to a chef-client run](https://trello.com/c/PBAjwFrL/310-nodes-lose-crowbar-upgarde-step-attribute-due-to-a-chef-client-run)

jsuchome:
For the record: that "upgrade" gating job cannot pass right now, as it would require additional 2 PR's (one on Cloud6 side) as well as pending fix for crowbarctl.

PR #1081, opened 2017-02-08
upgrade: Move "crowbar_upgrade_step" attribute to node role

rhafer:
This is a partial backport of
https://github.com/crowbar/crowbar-core/pull/1080 to stable/3.0

Note: This should not change any pieces related to the upgrade from
tex to stable/3.0 to avoid breaking that. That's why the there are still
a lot of reference to crowbar_wall["crowbar_upgrade_step"] left.


rhafer:
[nodes lose "crowbar_upgarde_step" attribute due to a chef-client run](https://trello.com/c/PBAjwFrL/310-nodes-lose-crowbar-upgarde-step-attribute-due-to-a-chef-client-run)

PR #1082, opened 2017-02-08
deobfuscate the network conduit resolver

aspiers:
This is a forward-port of the first commit of #575 (893bfbb5) in an attempt to make this code even clearer.  #575 never got merged because it took too long to review it, and in the meantime all of this code got significantly rewritten during extraction of the conduit resolver code into a library to be shared between Chef and Crowbar.

Since I don't have much clue what happened during the rewrite, I'm relying on @vuntz and particularly @rhafer to check that the comments I've ported over to here are all correct.

The other (fix) part of #575 was superseded by #803.

### DIFF
--- a/crowbar_framework/lib/crowbar/conduit_resolver.rb
+++ b/crowbar_framework/lib/crowbar/conduit_resolver.rb
@@ -19,7 +19,13 @@ module Crowbar
   module ConduitResolver
     ### Public methods
 
-    ## Bus order for this node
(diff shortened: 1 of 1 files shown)


dirkmueller:
[bsc#990745 - Installing new node; fails to start install](https://trello.com/c/uwPEX4Qp/107-bsc-990745-installing-new-node-fails-to-start-install)

PR #1084, opened 2017-02-09
network: Set MTU for VLAN parent interface (bsc#1024279)

toabctl:
If a physical interface has VLAN(s) configured with a custom MTU set but
the underlying physical network is not used directly and has no MTU in
its config, use the MTU from the VLAN interface for the parent device.
If there are multiple VLANs, use the highest MTU for the parent.

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -393,6 +393,37 @@ Nic.nics.each do |nic|
     nic.tx_offloading = node["network"]["enable_tx_offloading"] || true
   end
 
+  # Do some MTU checks/configuration for the parent of the vlan nic
+  if nic.is_a?(Nic::Vlan)
+    # 1) validate that the parent of vlan nic is not used by a network directly
+    # and the requested mtu for the parent is lower than the vlan nic mtu.
+    # That would not work and setting the mtu on the vlan would fail.
+    networks_using_parent = if_mapping.select { |net, ifaces| ifaces.include? nic.parent }.keys
+    networks_using_parent.each do |net_name|
+      net = Barclamp::Inventory.get_network_by_type(node, net_name)
+      unless net.use_vlan
+        nic_parent = Nic.new nic.parent
+        if nic_parent.mtu.to_i < ifs[nic.name]["mtu"].to_i
+          msg = "#{nic.name} wants mtu #{ifs[nic.name]["mtu"]} but network #{net_name} " \
+                "using the parent nic #{nic.parent} wants a lower mtu #{net.mtu}. " \
+                "This network mtu configuration is invalid."
+          Chef::Log.fatal(msg)
+          raise msg
+        end
+      end
+    end
+    # 2) set the mtu for the parent if needed
+    if !ifs[nic.parent].key? "mtu" || (ifs[nic.parent]["mtu"].to_i < ifs[nic.name]["mtu"].to_i)
+      # we want the highest mtu to end up in the ifcfg-$parent config
+      ifs[nic.parent]["mtu"] = ifs[nic.name]["mtu"]
+      parent_nic = Nic.new(nic.parent)
+      Chef::Log.info("vlan #{nic.name} wants mtu #{ifs[nic.name]["mtu"]} but " \
+                     "parent #{parent_nic.name} wants no/lower mtu. Set mtu "\
+                     "for #{parent_nic.name} to #{ifs[nic.parent]['mtu']}")
+      parent_nic.mtu = ifs[nic.parent]["mtu"]
+    end
+  end
+
   if ifs[nic.name]["mtu"]
     nic.mtu = ifs[nic.name]["mtu"]
   end


chef/cookbooks/network/recipes/default.rb
```diff
@@ -393,6 +393,20 @@ def get_datapath_id_for_ovsbridge(bridge)
     nic.tx_offloading = node["network"]["enable_tx_offloading"] || true
   end
 
+  # if the nic is a vlan and the physical parent is not used, also set
+  # the mtu from the vlan nic for the parent
+  if nic.kind_of?(Nic::Vlan)
+    unless ifs[nic.parent].key? "mtu"
```
vuntz: Shouldn't we ignore the mtu setting if it's lower than the one for the vlan interface?

chef/cookbooks/network/recipes/default.rb
```diff
@@ -393,6 +393,20 @@ def get_datapath_id_for_ovsbridge(bridge)
...
+      parent_nic = Nic.new(nic.parent)
+      # if there are multiple vlan's on the parent nic, use the highest mtu
+      if parent_nic.mtu < ifs[nic.name]["mtu"].to_i
+        Chef::Log.info("vlan #{nic.name} wants mtu #{nic.mtu} but parent #{parent_nic.name} " \
+                       "has no explicit mtu. Set mtu for #{parent_nic.name} to #{ifs[nic.name]['mtu']}")
```
rhafer: I think you need to make sure here, that ```parent_nic``` is not used by another network directly. It should be possible by looking at ```if_mappings```. It looks like this:

```
{
   "<networkname>" => [ "list", "of", "interfaces"],
   "<othernet>" => [ .... ]
}
```

As the list of interface always contains a complete chain that is used by the network e.g. [ eth0, eth0.300, br-public] you need to find all networks that contain ```nic.parent``` and have ```use_vlan != true```  (e.g. by looking at ```Barclamp::Inventory.get_network_by_type(node, <networkname>).use_vlan```

If there is such a device an its mtu is smailler that the mtu of the vlan we're currently looking at, I think we need to error out.

chef/cookbooks/network/recipes/default.rb
```diff
@@ -393,6 +393,20 @@ def get_datapath_id_for_ovsbridge(bridge)
...
+        parent_nic.mtu = ifs[nic.name]["mtu"]
```
rhafer: In order to have the new mtu also appear in the config files you need to set ```ifs[nic.parent]["mtu"] = ifs[nic.name]["mtu"]``` as well.

chef/cookbooks/network/recipes/default.rb
toabctl: Good point. But it would be nice to validate things like this **before** we apply the config. If a user sets an MTU of 1500 for a physical interface and sets a higher MTU for the vlan network, the configuration is wrong.
Are we doing already some network validation somewhere?

chef/cookbooks/network/recipes/default.rb
toabctl: Done

chef/cookbooks/network/recipes/default.rb
toabctl: Done

chef/cookbooks/network/recipes/default.rb
```diff
@@ -393,6 +393,36 @@ def get_datapath_id_for_ovsbridge(bridge)
...
+          msg = "#{nic.name} wants mtu #{nic.mtu} but network #{net_name} " \
+                "using the parent nic #{nic.parent} wants a lower mtu #{net.mtu}." \
+                "This network mtu configuration is invalid."
+          Chef::Log.fatal(msg)
+          raise msg
+        end
+      end
+    end
+    # 2) set the mtu for the parent if needed
+    if !ifs[nic.parent].key? "mtu" || ifs[nic.parent]["mtu"].to_i < ifs[nic.name]["mtu"].to_i
```
vuntz: Stupid question: do we really need the `to_i` here?

toabctl:
> Stupid question: do we really need the to_i here?

not sure but I do not want to compare strings. I guess it does not hurt, right?

chef/cookbooks/network/recipes/default.rb
```diff
@@ -393,6 +393,36 @@ def get_datapath_id_for_ovsbridge(bridge)
...
+      # we want the highest mtu to end up in the ifcfg-$parent config
+      ifs[nic.parent]["mtu"] = ifs[nic.name]["mtu"]
+      parent_nic = Nic.new(nic.parent)
+      Chef::Log.info("vlan #{nic.name} wants mtu #{nic.mtu} but parent #{parent_nic.name} " \
+                     "wants no/lower mtu. Set mtu for #{parent_nic.name} to " \
+                     "#{ifs[nic.parent]['mtu']}")
```
toabctl: I would need to escape the double quotes so I prefer this style here

chef/cookbooks/network/recipes/default.rb
```diff
@@ -393,6 +393,36 @@ def get_datapath_id_for_ovsbridge(bridge)
...
+          msg = "#{nic.name} wants mtu #{nic.mtu} but network #{net_name} " \
+                "using the parent nic #{nic.parent} wants a lower mtu #{net.mtu}." \
+                "This network mtu configuration is invalid."
+          Chef::Log.fatal(msg)
+          raise msg
+        end
+      end
+    end
+    # 2) set the mtu for the parent if needed
+    unless ifs[nic.parent].key? "mtu" || (ifs[nic.parent]["mtu"].to_i < ifs[nic.name]["mtu"].to_i)
```
vuntz: The move to `unless` means the second part of the test should be reverted, doesn't it?

toabctl: true. reverted to ```if !```. I like that more

vuntz:
>> Stupid question: do we really need the to_i here?
>
> not sure but I do not want to compare strings. I guess it does not hurt, right?

It doesn't hurt, but shouldn't be needed. Anyway, no big deal.

chef/cookbooks/network/recipes/default.rb
vuntz: Actually you don't need. But we had a thread on this a couple of weeks ago, and we agreed on this style.

toabctl:
Rebased (without any code change)

vuntz:
@toabctl CI is failing with: "eth0.200 wants mtu 8900 but network public using the parent nic eth0 wants a lower mtu 1500.This network mtu configuration is invalid."

chef/cookbooks/network/recipes/default.rb
```diff
@@ -393,6 +393,36 @@ def get_datapath_id_for_ovsbridge(bridge)
...
   end
 
+  # Do some MTU checks/configuration for the parent of the vlan nic
+  if nic.is_a?(Nic::Vlan)
+    # 1) validate that the parent of vlan nic is not used by a network directly
+    # and the requested mtu for the parent is lower than the vlan nic mtu.
+    # That would not work and setting the mtu on the vlan would fail.
+    networks_using_parent = if_mapping.select { |net, ifaces| ifaces.include? nic.parent }.keys
+    networks_using_parent.each do |net_name|
+      net = Barclamp::Inventory.get_network_by_type(node, net_name)
```
toabctl: @vuntz could it be that I need to get the network via ```Barclamp::Inventory.get_network_definition``` ? Using ```get_network_by_type``` might not return the correct network but in the case of this PR I want to get the correct network definition even if the network is not used on the node.

vuntz: We shouldn't hit the code if the network is not used on the node, or am I missing something?

On IRC you mentioned use_vlan which is not as expected, but `neutron_service.rb` in the rails app is setting this.

toabctl:
@rhafer @vuntz could you please review again?

chef/cookbooks/network/recipes/default.rb
```diff
@@ -393,6 +393,37 @@ def get_datapath_id_for_ovsbridge(bridge)
...
+    end
+    # 2) set the mtu for the parent if needed
+    if !ifs[nic.parent].key? "mtu" || (ifs[nic.parent]["mtu"].to_i < ifs[nic.name]["mtu"].to_i)
+      # we want the highest mtu to end up in the ifcfg-$parent config
+      ifs[nic.parent]["mtu"] = ifs[nic.name]["mtu"]
+      parent_nic = Nic.new(nic.parent)
+      Chef::Log.info("vlan #{nic.name} wants mtu #{ifs[nic.name]["mtu"]} but " \
+                     "parent #{parent_nic.name} wants no/lower mtu. Set mtu "\
+                     "for #{parent_nic.name} to #{ifs[nic.parent]['mtu']}")
+      parent_nic.mtu = ifs[nic.parent]["mtu"]
```
vuntz: I'm really confused. On line 406, the test is: `if nic_parent.mtu.to_i < ifs[nic.name]["mtu"].to_i` and no we do `parent_nic.mtu = ifs[nic.parent]["mtu"]`.

Something doesn't make sense to me. Doesn't it just work here because there's no network definition that uses `parent_nic` directly?

rhafer: Yes. The cases where ```parent_nic``` is used directly (e.g. by the admin network) should be caught by the checks in part one, I think.

nvin-krja:
This is already merged but I'll still comment here. While setting parent MTU, we should scan all the child vlans for a parent and use the max MTU value configured on vlans. Otherwise it fails if we first apply a vlan with lower MTU and then apply a vlan with higher MTU for same parent.

PR #1085, opened 2017-02-09
Register crowbar v2 api specific mime type (bsc#1023834)

rhafer:
This should make crowbar properly accept request with the v2 specific
Content-Type: application/vnd.crowbar.v2.0+json

As e.g. done by the crowbarctl upgrade subcommands. This is a followup
fix for:
https://github.com/crowbar/crowbar-client/commit/812c8ec0339cce9dda1a5fab3eaeff6cc1297454

As before that crowbarctl was actually sending all requests urlencoded.

### DIFF
--- a/crowbar_framework/config/initializers/register_json_mime_type.rb
+++ b/crowbar_framework/config/initializers/register_json_mime_type.rb
@@ -1,0 +1,20 @@
+#
+# Copyright 2017, SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+api_mime_types = ["application/vnd.crowbar.v2.0+json", "text/x-json"]
+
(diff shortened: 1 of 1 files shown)


rhafer:
This should fix: https://github.com/crowbar/crowbar-client/pull/151

Itxaka:
```
 1) Api::BackupsController with a failed backups API request doesn't find a crowbar backup
     Failure/Error: expect(response.body).to eq("Not found")
       
       expected: "Not found"
            got: "{\"error\":\"Not found\"}"
       
       (compared using ==)
     # ./spec/controllers/api/backups_controller_spec.rb:105:in `block (3 levels) in <top (required)>'```

rsalevsky:
Our idea was based on the fact that we wanted to avoid to change the mime-type because this could cause issues in other areas. I have to look into this properly to say if this is wrong or right. But based on the fact how it was conceived and my feeling I would say this is the wrong way.

rhafer:
> Our idea was based on the fact that we wanted to avoid to change the mime-type because this could cause issues in other areas. 

Which are those?

> I have to look into this properly to say if this is wrong or right. But based on the fact how it was conceived and my feeling I would say this is the wrong way.

An alternative is to change crowbarctl to not send that contenttype:
https://github.com/crowbar/crowbar-client/blob/master/lib/crowbar/client/util/apiversion.rb#L36

@rsalevsky Do you think that's any better?

rsalevsky:
@rhafer I can't remember properly but we spoke about the crowbar cmd-line tool and customer scripts which are using the API.

https://github.com/crowbar/crowbar-client/blob/master/lib/crowbar/client/util/apiversion.rb#L36 looks wrong to me, AFIK we only wanted to set the accept header.

rhafer:
closing in favor of https://github.com/crowbar/crowbar-client/issues/152

PR #1086, opened 2017-02-09
upgrade: Do not touch ceph nodes during services shutdown

jsuchome:
Ceph nodes should already be upgraded, there's no need to
mark an upgrade step there and it would be wrong to shut down
their services.

TODO: start chef-client & crowbar_join on ceph nodes, move them to ready state -missing, will be done in a followup


jsuchome:

PR #1088, opened 2017-02-10
upgrade: Precheck for Cloud6 repositories

jsuchome:
Before the upgrade starts we need to check if Cloud6 and SOC6 repos
are correctly in place so user can get the latest updates.

UPDATE: By using the same method as for "admin repo check", we're actually checking for products, not repositories. That's actually https://bugzilla.suse.com/show_bug.cgi?id=999946
This should be fixed (for all usages of the method), I'd propose in some later PR....

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -181,7 +181,124 @@ module Api
           : {}
       end
 
+      def maintenance_updates_check
+        initial_repocheck = check_repositories("6")
+
+        # These are the zypper failures
+        if initial_repocheck.key? :error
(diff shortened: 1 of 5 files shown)


jsuchome:
[(2) S55P10 add precheck to Ensure that the Cloud6 related repos are correctly enabled before starting upgrade](https://trello.com/c/eAB7rzuq/305-2-s55p10-add-precheck-to-ensure-that-the-cloud6-related-repos-are-correctly-enabled-before-starting-upgrade)

jsuchome:
Wow, this code is complicated. 
@skazi0 as you've also worked on it, I'd like to kindly ask you to look at this and offer your comments...

PR #1090, opened 2017-02-10
upgrade: Allow upgrading all the rest of nodes after controllers

jsuchome:
When controller nodes have been upgraded previously,
whole nodes step was not actually finished, just a substep.
It make sense then to upgrade the rest with 'all'

### DIFF
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -145,12 +145,17 @@ class Api::UpgradeController < ApiController
   # The method runs asynchronously, so there's a need to poll for the status and possible errors
   def nodes
     if params[:component]
+      upgrade_status = ::Crowbar::UpgradeStatus.new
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -145,12 +145,17 @@ def services
...
   def nodes
     if params[:component]
+      upgrade_status = ::Crowbar::UpgradeStatus.new
+      substep = upgrade_status.current_substep
+      status = upgrade_status.current_substep_status
       if ["all", "controllers"].include? params[:component]
-        ::Crowbar::UpgradeStatus.new.start_step(:nodes)
+        # When controller nodes have been upgraded previously,
+        # whole 'nodes' step was not actually finished, just a substep.
+        # It make at this time to upgrade the rest with 'all'.
```
jsuchome: Ah, typo.

jsuchome:

PR #1092, opened 2017-02-13
upgrade: Fix showing missing repositories in prechecks

jsuchome:
1st commit is about the fix of https://github.com/crowbar/crowbar-core/pull/1088
Surprisingly, this was not properly checked when merging .

@skazi0 with 2nd commit, we actually can make the check somewhat usable even with the products - if user did not clean Cloud7/SP2 repos after cancel, this should show an error in next precheck.

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -199,11 +199,25 @@ module Api
           return { repositories_missing: missing_repos }
         end
 
+        # Now check if new (Cloud7) products are not yet enabled
+        next_version_repocheck = check_repositories("7")
+        if next_version_repocheck.key? :error
+          return { zypper_errors: next_version_repocheck[:error] }
+        end
+
+        if next_version_repocheck.any? { |_k, v| v[:available] }
+          available = next_version_repocheck.collect do |k, v|
+            next unless v[:available]
+            v[:repos]
+          end.flatten.compact.join(", ")
+          return { repositories_too_soon: available }
+        end
+
         updates_status = ::Crowbar::Checks::Maintenance.updates_status
         updates_status.empty? ? {} : { maintenance_updates: updates_status }
       end
 
-      def check_repositories(soc_version)
+      def check_repositories(soc_version, end_step_on_error = false)
         sp = soc_version == "6" ? "12.1" : "12.2"
         sp_version = soc_version == "6" ? "SP1" : "SP2"
         upgrade_status = ::Crowbar::UpgradeStatus.new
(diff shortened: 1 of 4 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -278,11 +278,19 @@ def maintenance_updates_check_errors(check)
         if check[:repositories_missing]
           ret[:repositories_missing] = {
             data: I18n.t("api.upgrade.prechecks.repos_missing.error",
-              missing: check[:repositories_missing].join(", ")),
+              missing: check[:repositories_missing]),
```
skazi0: wouldn't it be better to remove the join from the other end? other checks seem to  have join right before language code interpolation.

jsuchome: But on 'the other end', the join is used also when doing adminrepocheck, so having join there seems to be consistent when analyzing the output of check_repositories method

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -191,19 +191,33 @@ def maintenance_updates_check
...
+          end.flatten.compact.join(", ")
+          return { repositories_too_soon: available }
+        end
+
         updates_status = ::Crowbar::Checks::Maintenance.updates_status
         updates_status.empty? ? {} : { maintenance_updates: updates_status }
       end
 
-      def check_repositories(soc_version)
+      def check_repositories(soc_version, end_step_on_error)
```
skazi0: default value missing.

jsuchome: fixing...

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -278,11 +278,19 @@ def maintenance_updates_check_errors(check)
...
             help: I18n.t("api.upgrade.prechecks.repos_missing.help")
           }
         end
 
+        if check[:repositories_too_soon]
+          ret[:repositories_too_soon] = {
+            data: I18n.t("api.upgrade.prechecks.repos_too_soon.error",
+              missing: check[:repositories_too_soon]),
```
skazi0: placeholder in the language file is `too_soon` not `missing`.

jsuchome: true

jsuchome:
[(2) S55P10 add precheck to Ensure that the Cloud6 related repos are correctly enabled before starting upgrade](https://trello.com/c/eAB7rzuq/305-2-s55p10-add-precheck-to-ensure-that-the-cloud6-related-repos-are-correctly-enabled-before-starting-upgrade)

PR #1093, opened 2017-02-13
Disable all current repos before starting the upgrade

rhafer:
As part of the node upgrade we're replacing all the product repositories
with the ones from the new product version. To be able to cleanup
correctly we need disable the old product repositories in crowbar (i.e.
remove the associated data bags) before the repository configuration
files (/etc/crowbar/repos.yml and ./crowbar_framework/config/repos-*.yml)
are replaced.

If the upgrade is canceled, we re-enable them.


rhafer:
[(2) S55P2 Deactivate old repos during the upgrade](https://trello.com/c/GLSLmAhN/306-2-s55p2-deactivate-old-repos-during-the-upgrade)

PR #1095, opened 2017-02-14
delayed_jobs: Enable traceback logging

Itxaka:
We didnt have traceback logging enabled for
delayed jobs. This enables it and also filters
the SQL insert/update for them from the crowbar
log as to not pollute the log with uneeded and
unreadable lines

### DIFF
--- a/crowbar_framework/config/initializers/delayed_job_config.rb
+++ b/crowbar_framework/config/initializers/delayed_job_config.rb
@@ -14,12 +14,24 @@
 # limitations under the License.
 #
 
+module Delayed
+  class Worker
+    def handle_failed_job_with_logging(job, error)
+      handle_failed_job_without_logging(job, error)
+      Rails.logger.error("Error processing job ID #{job.id}: #{error.message} "\
+        "#{error.backtrace.join("\n")}")
+    end
+    alias_method_chain :handle_failed_job, :logging
+  end
+end
+
 Delayed::Worker.destroy_failed_jobs = false
 Delayed::Worker.delay_jobs = !Rails.env.test?
 Delayed::Worker.raise_signal_exceptions = :term
 Delayed::Worker.max_attempts = 1
-if Rails.env.production?
-  Delayed::Worker.logger = Logger.new(File.join(ENV["CROWBAR_LOG_DIR"], "background_jobs.log"))
+log_file = if Rails.env.production?
+  File.join(ENV["CROWBAR_LOG_DIR"], "background_jobs.log")
 else
-  Delayed::Worker.logger = Logger.new(File.join(Rails.root, "log", "background_jobs.log"))
+  File.join(Rails.root, "log", "background_jobs.log")
 end
+Delayed::Worker.logger = Logger.new(log_file)
--- a/crowbar_framework/config/initializers/filter_sql_log.rb
+++ b/crowbar_framework/config/initializers/filter_sql_log.rb
@@ -24,6 +24,10 @@ module ActiveRecord
         Rails.logger.debug(
           "SQL: INSERT/UPDATE transaction: filtering query to not expose potentially sensitive data"
         )
+      elsif event.payload[:sql] =~ /^(INSERT INTO|DELETE FROM|UPDATE) "delayed_jobs".*/
+        Rails.logger.debug(
+          "SQL: delayed_jobs transaction: filtering query to not clutter the logs"
+        )
       else
         old_sql(event)
       end


Itxaka:
[(3) delayed_job: improve logging when background task failed](https://trello.com/c/NqqP6cVv/2198-delayed-job-improve-logging-when-background-task-failed)

crowbar_framework/config/initializers/delayed_job_config.rb
```diff
@@ -14,6 +14,18 @@
 # limitations under the License.
 #
 
+# Enable proper traceback logging of delayed jobs
+class Delayed
```
jsuchome: Looks like this needs to be `module`

jsuchome:
Cool, now I have a nice info in `background_jobs.log`. 
But would it be possible to add some info about the failure into the `production.log` as well?

crowbar_framework/config/initializers/delayed_job_config.rb
Itxaka: ah yes, good catch, is a leftover of a hound complaint, it was class `Delayed::Worker` before!

Itxaka:
@jsuchome May be, but do we want to? production.log is part of crowbar, crowbar-jobs is a different entity altogether, so not sure if the logs should be mixed.

I could filter and rewrite the SQL update statement as to have some info of a job failed into the production.log maybe?

crowbar_framework/config/initializers/delayed_job_config.rb
```diff
@@ -14,6 +14,18 @@
 # limitations under the License.
 #
 
+# Enable proper traceback logging of delayed jobs
+Module Delayed
```
jsuchome: `module Delayed`

jsuchome:
@Itxaka I'm not sure. I'd be happy with just a notification that some background job has failed and specific log contains the details... if it is not too difficult.

Otherwise, we'll need to update our universal error message that says `Crowbar has failed. Check /var/log/crowbar/production.log for details.`

crowbar_framework/config/initializers/delayed_job_config.rb
Itxaka: argh! :+1:

Itxaka:
@jsuchome sounds doable, as we are already filtering the update messages, we can prolly just extract and rewrite the info the the log in a more nicer way. Let me check!

crowbar_framework/config/initializers/delayed_job_config.rb
```diff
@@ -14,6 +14,18 @@
 # limitations under the License.
 #
 
+# Enable proper traceback logging of delayed jobs
+module Delayed
+  class Worker
+    def handle_failed_job_with_logging(job, error)
+      handle_failed_job_without_logging(job, error)
+      Delayed::Worker.logger.error(error.message)
+      Delayed::Worker.logger.error(error.backtrace.join("\n"))
```
rsalevsky: In crowbar we use always something like: `Delayed::Worker.logger.error("Error processing job: #{error.message} #{error.backtrace.join("\n")}")`

Do you think it makes sense do to the same here?

Itxaka: done!

Itxaka:
@jsuchome Updated.

It will now post the following into the production.log

```
D, [2017-02-14T14:49:31.025497 #12804:0x0000000252e660] ERROR -- delayed_job Object#Whatever with ID 40 failed. See the delayed_job log for more details.
```

crowbar_framework/config/initializers/delayed_job_config.rb
```diff
@@ -14,6 +14,20 @@
 # limitations under the License.
 #
 
+# Enable proper traceback logging of delayed jobs
+module Delayed
+  class Worker
+    def handle_failed_job_with_logging(job, error)
+      handle_failed_job_without_logging(job, error)
+      Rails.logger.error("delayed_job #{job.name} with ID #{job.id} failed. "\
+        "See the delayed_job log for more details.")
```
jsuchome: The log file is called `background_jobs.log` (I do not care if you rename it or change the text here, btw)

Itxaka:
@rhafer so you would like to have the information duplicated? Or Im understanding that request wrongly?

I think that having **any** delayed_job errors in the production.log file is wrong, as they are separated components, they should have separated log files. It would be the same as having the X-api and X-engine services from openstack log into the same file. One uses the other and they need each other, but they are separate services with separate responsibilities and logging.

Maybe Im not understanding what rails+delayed_job does in this case, but isn't it like django+celery? One its like the emitter of jobs, and the other is the engine which executes them?

crowbar_framework/config/initializers/delayed_job_config.rb
```diff
@@ -14,12 +14,27 @@
...
 #
 
+module Delayed
+  class Worker
+    if Rails.env.production?
+      @log_file = File.join(ENV["CROWBAR_LOG_DIR"], "background_jobs.log")
+    else
+      @log_file = File.join(Rails.root, "log", "background_jobs.log")
+    end
+    @@logger = Logger.new(@log_file)
```
Itxaka: I believe this needs to be a class var as its shared around all classes and makes no sense to have copies around, plus I dont know other way of overriding it properly to use the same `log_file` without having to check for the `Rails.env.production`

crowbar_framework/config/initializers/delayed_job_config.rb
```diff
@@ -14,12 +14,27 @@
 # limitations under the License.
 #
 
+module Delayed
+  class Worker
+    if Rails.env.production?
```
Itxaka: This is so ugly, what a crappy check. sure, 2 vars instead of one but really? The code needs to be read by actual persons, not robots. :angry: 
```
@log_file = if Rails.env.production?
                  File.join(ENV["CROWBAR_LOG_DIR"], "background_jobs.log")
                else 
                  File.join(Rails.root, "log", "background_jobs.log")
                end
```

Itxaka:
Updated so it uses the proper log file location + logs the job ID so its easier to track down the job from which the error came.

jsuchome:
Hm, does not seem to be properly initialized:

``ERROR -- delayed_job Api::Upgrade.nodes_without_delay with ID 5 failed. Check  for more details.``

rhafer:
> so you would like to have the information duplicated? Or Im understanding that request wrongly?

I'd like to have the errors have them easily locate-able. And if it is easy to correlated them with a certain action in the crowbar app that's also a plus. The jobs are submitted by the crowbar app and in many cases the app actually polls for success, why is it wrong to have the crowbar app log an meaningful error message in such a case?
Logging "please check this other log file" just doesn't strike as very user friendly here (provided that there is an easy way to log a more useful error). But as I already said it's certainly better that logging the SQL INSERT/UPDATE statement from delayed_jobs in production.log. That's why I actually didn't -1 this change.

Itxaka:
> I'd like to have the errors have them easily locate-able. And if it is easy to correlated them with a certain action in the crowbar app that's also a plus. The jobs are submitted by the crowbar app and in many cases the app actually polls for success, why is it wrong to have the crowbar app log an meaningful error message in such a case?
Logging "please check this other log file" just doesn't strike as very user friendly here (provided that there is an easy way to log a more useful error). But as I already said it's certainly better that logging the SQL INSERT/UPDATE statement from delayed_jobs in production.log. That's why I actually didn't -1 this change.

How about logging the task ID when creating the task from the crowbar app into production.log?* That way should be easier to search and locate the appropriate log lines by searching for the task ID. Would that work for you?


*Im not sure if thats possible :D

Itxaka:
> How about logging the task ID when creating the task from the crowbar app into production.log?* That way should be easier to search and locate the appropriate log lines by searching for the task ID. Would that work for you?

doesnt seem that easy without adding way too much code :(

The log will log the job.name being called, for example `Api::Upgrade.nodes_without_delay` so that should be good enough to know where it comes from, I guess.

crowbar_framework/config/initializers/delayed_job_config.rb
```diff
@@ -14,12 +14,27 @@
...
 #
 
+module Delayed
+  class Worker
+    @@log_file = if Rails.env.production?
+      File.join(ENV["CROWBAR_LOG_DIR"], "background_jobs.log")
+    else
+      File.join(Rails.root, "log", "background_jobs.log")
+    end
+    @@logger = Logger.new(@@log_file)
```
rsalevsky: Why you don't keep the old way? I don't understand this change and the code looks unclean. Please explain.

crowbar_framework/config/initializers/delayed_job_config.rb
```diff
@@ -14,12 +14,27 @@
...
+
+    def handle_failed_job_with_logging(job, error)
+      handle_failed_job_without_logging(job, error)
+      Rails.logger.error("delayed_job #{job.name} with ID #{job.id} failed. "\
```
rsalevsky: Why Rails.logger? Can you please explain?

crowbar_framework/config/initializers/filter_sql_log.rb
```diff
@@ -24,6 +24,8 @@ def sql(event)
         Rails.logger.debug(
           "SQL: INSERT/UPDATE transaction: filtering query to not expose potentially sensitive data"
         )
+      elsif event.payload[:sql] =~ /^(INSERT INTO|DELETE FROM|UPDATE) "delayed_jobs".*/
+        # Dont log SQL queries from delayed_jobs as they have their own log file
```
rsalevsky: Can you add in which log they get logged?

crowbar_framework/config/initializers/delayed_job_config.rb
Itxaka: So it goes to the rails app so somebody checking only the production.log can see that a background job failed thus pointing them to the background job log.

For me, it makes no sense to log from here into a different service but it was requested. `¯\_(ツ)_/¯`

crowbar_framework/config/initializers/filter_sql_log.rb
Itxaka: Depending on the environment they might get logged to different places so I dont like to hardcode stuff in comments that can change in the future. Leads to incongruity in comments vs reality.

crowbar_framework/config/initializers/delayed_job_config.rb
Itxaka: Basically because I need to access the log_file var which is set up as part of the environment, and I didnt want to check the env twice (one for the logger, one for the class log message)

crowbar_framework/config/initializers/delayed_job_config.rb
```diff
@@ -14,12 +14,27 @@
 # limitations under the License.
 #
 
+module Delayed
+  class Worker
+    @@log_file = if Rails.env.production?
```
Itxaka: I dont know why but it didnt work on my tests if I was using a class instance var. Probably was doing something wrong here, so I need to recheck :/

Itxaka:
This seems to be not really in the scope of what the card is asking now that I re-read it. I went with the traceback onto the background jobs log as it looks to me the more logical way of separation between the 2 services, but the card is asking for the traceback to be on the normal crowbar log.

I'll rework this to accommodate to the original task.

crowbar_framework/config/initializers/delayed_job_config.rb
rsalevsky: Yes, I agree this makes no sense we should keep them separated.

crowbar_framework/config/initializers/filter_sql_log.rb
rsalevsky: Makes sense agreed.

jsuchome:
I'd be fine with current status, are there any other requests? See http://pastebin.nue.suse.com/22022/src

jsuchome:
I just wanted to ask you about it! :)

PR #1100, opened 2017-02-15
upgrade: Bring ceph nodes back to ready state

jsuchome:
Ceph nodes were upgraded independently, we need to make them ready again.

When the upgrade is finished, all nodes should be ready. UPDATE: Actually we might do it earlier, e.g. before all other nodes.

Executes crowbar_join on ceph  nodes.

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -547,13 +547,15 @@ module Api
 
         # initialize progress info about nodes upgrade
         if remaining.nil?
-          remaining = ::Node.find(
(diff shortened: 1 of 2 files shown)


jsuchome:

jsuchome:
We might need to enhance error reporting. These nodes are not being shown as 'current nodes being upgraded', so user might be surprised if an error happens...

jsuchome:
So we might want to 

1. mark some progress that such nodes are being handled
2. do it actually earlier, e.g. at the first step of the 'nodes' step, not the last one

jsuchome:
Tested manually, seems to work.

PR #1101, opened 2017-02-16
upgrade: Save the upgrade mode into the progress library

jsuchome:

We need to track the upgrade mode that is decided at the start of upgrade, so we can later do different upgrade actions for disruptive and non-disruptive upgrade.

Saves the info into the file that we are already using for similar purposes

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -122,6 +122,8 @@ module Api
             "disruptive"
           end
 
+          ::Crowbar::UpgradeStatus.new.save_upgrade_mode(ret[:best_method])
+
           return ret unless upgrade_status.current_step == :prechecks
 
           # transform from this:
@@ -143,9 +145,9 @@ module Api
                    reduce({}, :merge)
 
           if errors.any?
-            upgrade_status.end_step(false, errors)
+            ::Crowbar::UpgradeStatus.new.end_step(false, errors)
           else
-            upgrade_status.end_step
+            ::Crowbar::UpgradeStatus.new.end_step
           end
         end
       rescue ::Crowbar::Error::StartStepRunningError,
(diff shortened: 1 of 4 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -143,7 +143,8 @@ def checks
                    reduce({}, :merge)
 
           if errors.any?
-            upgrade_status.end_step(false, errors)
+            ::Crowbar::UpgradeStatus.new.save_upgrade_mode(ret[:best_method])
```
skazi0: why do we save it only if there are errors?

jsuchome: Oh, it should go to the other branch :-)

jsuchome:
> Does it make sense to also provide an API (and crowbarctl support) to force a certain mode?

Good point. Maybe for another PR, because we need to decide where should such API go, if we want UI change (probably not) and maybe something else.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -143,9 +143,10 @@ def checks
                    reduce({}, :merge)
 
           if errors.any?
-            upgrade_status.end_step(false, errors)
+            ::Crowbar::UpgradeStatus.new.end_step(false, errors)
           else
-            upgrade_status.end_step
+            ::Crowbar::UpgradeStatus.new.save_upgrade_mode(ret[:best_method])
```
skazi0: this will not work when re-running prechecks after one successful run (it will return in line 125 above)

jsuchome:
[Implement disruptive upgrade (non-HA, one compute node, etc.)](https://trello.com/c/2B0jUWrs/1927-implement-disruptive-upgrade-non-ha-one-compute-node-etc)

PR #1103, opened 2017-02-16
upgrade: Before saving updated progress, read the latest from file

jsuchome:

We need to make writing of progress library atomic.

While saving, we load the data under the same lock as is used for writing.

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade_status.rb
+++ b/crowbar_framework/lib/crowbar/upgrade_status.rb
@@ -71,12 +71,6 @@ module Crowbar
       save
     end
 
-    def load!
-      ::Crowbar::Lock::LocalBlocking.with_lock(shared: true, logger: @logger, path: lock_path) do
-        @progress = YAML.load(progress_file_path.read)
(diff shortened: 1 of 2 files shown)


jsuchome:
[(3) Fix UpgradeStatus atomicity](https://trello.com/c/fo2uoWO6/2197-3-fix-upgradestatus-atomicity)

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -69,9 +69,13 @@ def initialize_state
       save
     end
 
+    def load_while_locked
+      @progress = YAML.load(progress_file_path.read)
+    end
+
```
rhafer: I think this should better be a ```protected``` method, right?

jsuchome: actually both load methods

jsuchome: I've moved it in latest update

PR #1107, opened 2017-02-16
upgrade: Check if Swift configuration is compatible with the upgrade

jsuchome:

Newton's Swift does not allow number of replicas bigger than number
of disks.

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -171,6 +171,21 @@ module Api
         ret
       end
 
+      def openstack_check
+        ret = {}
+        # swift replicas check vs. number of disks
+        prop = Proposal.where(barclamp: "swift").first
+        return ret if prop.nil?
+        replicas = prop["attributes"]["swift"]["replicas"] || 0
+
+        disks = 0
+        NodeObject.find("roles:swift-storage").each do |n|
+          disks += n["swift"]["devs"].size
+        end
+        ret[:too_many_replicas] = replicas if replicas > disks
+        ret
+      end
+
       def compute_status
         ret = {}
         ["kvm", "xen"].each do |virt|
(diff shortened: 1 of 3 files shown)


crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -171,6 +171,22 @@ def ceph_status
         ret
       end
 
+      def swift_check
+        # 1. number of replicas
+        ret = {}
+        prop = Proposal.where(barclamp: "swift").first
+        return ret if prop.nil?
+        replicas = prop["attributes"]["swift"]["replicas"]
```
dirkmueller: You need to multiply by the bumber of zones I thibk as well.

jsuchome: Are you sure? 
2 zones, 2 replicas and 2 swift-storage nodes works for me.
2 zones, 3 replicas, 2 swift-storage zones and I got an error before

jsuchome: https://github.com/openstack/swift/blob/master/swift/common/ring/builder.py#L437

jsuchome:
@skazi0 new precheck -> crowbar-ui update

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -362,6 +369,17 @@ def clusters_health_report_errors(check)
...
+      def swift_check_errors(check)
+        ret = {}
+        if check[:too_many_replicas]
+          ret[:too_many_replicas] = {
+            data: I18n.t("api.upgrade.prechecks.swift_replicas.error",
+              replicas: check[:too_many_replicas]),
+            help: I18n.t("api.upgrade.prechecks.swift_replicas.help")
+          }
+        end
+      end
```
skazi0: you miss one `ret` here

aplanas: indeed

jsuchome:
Merging. @dirkmueller if you think this is wrong, we'll fix it in followup

jsuchome:
[(2) swift issues: Add precheck for replica settings when upgrading from 6 to 7](https://trello.com/c/m6dpfsqO/2245-2-swift-issues-add-precheck-for-replica-settings-when-upgrading-from-6-to-7)

PR #1108, opened 2017-02-17
upgrade: Better check for health of ceph cluster

jsuchome:
Current health check might be too strict. I currently do not know how to make it better. 

Some options were suggested here https://github.com/crowbar/crowbar-core/pull/1076#issuecomment-278481474 and here https://github.com/crowbar/crowbar-core/pull/1076#issuecomment-278771739

@tserong @smithfarm, please take a look and comment

UPDATE: I decided not to change the health check method, it seems to be too complicated. However, I've found several other errors while playing with it...

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -147,12 +147,14 @@ module Api
         ret = {}
         ceph_nodes = NodeObject.find("roles:ceph-* AND ceph_config_environment:*")
         return ret if ceph_nodes.empty?
-        ceph_node = ceph_nodes.first
+        mon_node = NodeObject.find("run_list_map:ceph-mon AND ceph_config_environment:*").first
 
-        ssh_retval = ceph_node.run_ssh_cmd("LANG=C ceph health --connect-timeout 5 2>&1")
+        ssh_retval = mon_node.run_ssh_cmd("LANG=C ceph health --connect-timeout 5 2>&1")
+        # Some warnings do not need to be critical, but we have no way to find out.
+        # So we assume user knows how to tweak cluster settings to show the healthy state.
         unless ssh_retval[:stdout].include? "HEALTH_OK"
           ret[:health_errors] = ssh_retval[:stdout]
-          unless ssh_retval[:stderr].empty?
+          unless ssh_retval[:stderr].nil? || ssh_retval[:stderr].empty?
             ret[:health_errors] += "; " unless ssh_retval[:stdout].empty?
             ret[:health_errors] += ssh_retval[:stderr]
           end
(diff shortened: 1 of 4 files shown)


jsuchome:
[ceph health check: not good enough](https://trello.com/c/p5iuWhJ3/311-ceph-health-check-not-good-enough)

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -149,7 +149,11 @@ def ceph_status
         return ret if ceph_nodes.empty?
         ceph_node = ceph_nodes.first
 
+        # FIXME: do I need to pick ceph-mon node? (It might happen that this is just ceph-osd, without the mon role)
         ssh_retval = ceph_node.run_ssh_cmd("LANG=C ceph health --connect-timeout 5 2>&1")
+        # FIXME: here we need some better check for health:
+        # a) could we ignore HEALTH_WARN in general, thus only checking for HEALTH_ERR?
```
tserong: No. We'll see HEALTH_WARN if, for example, a MON is down and presumably other cases we want to catch.

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -149,7 +149,11 @@ def ceph_status
...
+        # b) do we need some magic to suppress only some kind of warnings, making the cluster return HEALTH_OK?
```
tserong: What @smithfarm suggested can be implemented something like this:

````
#!/usr/bin/env ruby

require 'json'

# `ceph daemon` doesn't need --connect-timeout as it's
# talking directly to the daemon on this node only to
# get a config setting.  If the mon isn't running, this'll
# return something ugly and the JSON parse will throw an
# exception.
warn_max = JSON.parse(`ceph daemon mon.$(hostname) config get mon_pg_warn_max_per_osd`)['mon_pg_warn_max_per_osd']
warn_min = JSON.parse(`ceph daemon mon.$(hostname) config get mon_pg_warn_min_per_osd`)['mon_pg_warn_min_per_osd']

# have to tell *all* the mons the new config value, not
# just the current mon (experimentation indicates using
# `ceph daemon mon.$(hostname) config set ...` doesn't
# always result in `ceph health` noticing the value is
# set yet
`ceph tell 'mon.*' injectargs '--mon_pg_warn_max_per_osd 0' >/dev/null 2>&1`
`ceph tell 'mon.*' injectargs '--mon_pg_warn_min_per_osd 0' >/dev/null 2>&1`

rv = `LANG=C ceph health --connect-timeout 5 2>&1`

# back to previous values
`ceph tell 'mon.*' injectargs '--mon_pg_warn_max_per_osd #{warn_max}' >/dev/null 2>&1`
`ceph tell 'mon.*' injectargs '--mon_pg_warn_min_per_osd #{warn_min}' >/dev/null 2>&1`

if rv.include? "HEALTH_OK"
  print "cluster healthy\n"
else
  print "cluster not healthy\n"
end
````

That looks kinda complicated to me though. I'm going to ask about this upstream and see if anyone knows a better way.

Comment (jsuchome) [inline crowbar_framework/app/models/api/crowbar.rb]: I also do not really like executing this from crowbar ... What if we just leave it as it is and let user do such changes  him/her self once the WARN message is shown?

File: crowbar_framework/app/models/api/crowbar.rb [hunk 3]
```diff
@@ -149,7 +149,11 @@ def ceph_status
         return ret if ceph_nodes.empty?
         ceph_node = ceph_nodes.first
 
+        # FIXME: do I need to pick ceph-mon node? (It might happen that this is just ceph-osd, without the mon role)
```
jsuchome: @tserong What about this one? Do I need to pick ceph-mon node to run that health check command?

crowbar_framework/app/models/api/crowbar.rb
tserong: That might be the best option actually (unless I hear back anything from upstream about an easy/sensible way to do this). We can always document "if you're getting $these_warnings run $these_commands to suppress them for the duration of the upgrade"...

tserong: Nothing immediately useful from upstream (there's a three year old feature request for an [easier-to-process health report](http://tracker.ceph.com/issues/7192))

PR #1109, opened 2017-02-17
upgrade: Add a check for unsupported mixed-roles setup

jsuchome:

If compute node has some controller related role, we should not allow non-disruptive upgrade. It's because with non-disruptive upgrade we want to upgrade control plane first, and only then continue with the compute nodes.

Check for additional OpenStack roles that are assigned to compute nodes.

The list of roles that are really conflicting (from the upgrade process POV) is open to debate.

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -207,6 +207,37 @@ module Api
         ret
       end
 
+      def mixed_roles_check
+        conflicting_roles = [
+          "cinder-controller",
+          "glance-server",
+          "keystone-server",
+          "neutron-server",
+          "neutron-network",
+          "nova-controller",
+          "swift-proxy",
+          "swift-ring-compute",
+          "ceilometer-server",
+          "heat-server",
+          "horizon-server",
+          "manila-server",
+          "trove-server",
+          # FIXME: use this as better check to make sure compute node is not part of cluster?
+          "pacemaker-cluster-member"
+        ]
+        ret = {}
+        ["kvm", "xen"].each do |virt|
+          NodeObject.find("roles:nova-compute-#{virt}").each do |node|
+            conflict = node.roles & conflicting_roles
+            unless conflict.empty?
+              ret[:role_conflicts] ||= {}
+              ret[:role_conflicts][node.name] = conflict
+            end
+          end
+        end
+        ret
+      end
+
       def compute_status
         ret = {}
         ["kvm", "xen"].each do |virt|


jsuchome:
[(3) Add the precheck for compute-controller mixed setup](https://trello.com/c/14imamZ1/2053-3-add-the-precheck-for-compute-controller-mixed-setup)

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -171,6 +171,38 @@ def ceph_status
...
+          "heat-server",
+          "horizon-server",
+          "manila-server",
+          "trove-server",
+          # FIXME: use this as better check to make sure compute node is not part of cluster?
+          "pacemaker-cluster-member",
+        ]
+        ret = {}
+        ["kvm", "xen"].each do |virt|
+          NodeObject.find("roles:nova-compute-#{virt}").each do |node|
```
rsalevsky: NodeObject got obsoleted by Node

jsuchome: This is 3.0

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -183,6 +183,37 @@ def openstack_check
...
+            conflict = node.roles & conflicting_roles
+            unless conflict.empty?
+              ret[:role_conflicts] ||= {}
+              ret[:role_conflicts][node.name] = conflict
```
abel-navarro: where do we handle `:role_conflicts`?

jsuchome: So far nowhere, that still missing. Adding that part is easy, first I wanted to discuss this part (those FIXME's in this PR)

jsuchome:
@rhafer so, this tries to find out controller roles at compute nodes. Question is, is such (mixed) setup broken only for non-disruptive upgrade (that should ideally be goal of this precheck), or even for 'normal' mode?

rhafer:
> so, this tries to find out controller roles at compute nodes. Question is, is such (mixed) setup broken only for non-disruptive upgrade (that should ideally be goal of this precheck), or even for 'normal' mode?

I think it is broken (even for the disruptive upgrade) as long as it will cause the nova-compute node to be upgraded while nova-controller is not running.

I guess that means it's problematic, if nova-compute is deployed on a different node than nova-controller and that nova-compute node also has deployed some openstack barclamp with a run_order < than run_order(nova).

rhafer:
> I think it is broken (even for the disruptive upgrade) as long as it will cause the nova-compute node to be upgraded while nova-controller is not running.

Forgot to add it here: I think for the other cases, i.e. when nova-controller is already running again, it should be ok for the disruptive upgrade.

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -202,6 +202,36 @@ def openstack_check
           )
           ret[:lbaas_v1] = true unless out[:stdout].nil? || out[:stdout].empty?
         end
+      end
+
+      def mixed_roles_check
+        conflicting_roles = [
+          "cinder-controller",
```
rsalevsky: Could we maybe define this is the barclamp config? I'm sure that we will forget this list very fast if which can cause some problems.

jsuchome: Sounds like interesting idea, but I'm afraid some of these dependencies are upgrade only related...

jsuchome:
So I will use this commit in different PR.

PR #1111, opened 2017-02-17
upgrade: Disable cron during the node upgrade

jsuchome:
Some existing cron jobs might try to start services that were
not fully configured yet.


jsuchome:
[Delete old cron jobs before reboot of upgraded node?](https://trello.com/c/tvAr5l9r/321-delete-old-cron-jobs-before-reboot-of-upgraded-node)

PR #1113, opened 2017-02-17
upgrade: Remove all current repositories from the node to upgrade

rhafer:
Since we now disable all repositories in crowbar before starting the
upgrade, we don't have a definitive list of repos to remove anymore.
So we just remove all configured repositories from the node. This
shouldn't do any harm as the nodes' repositories should be managed by
crowbar anyway.

We needed to adapt this because of: https://github.com/crowbar/crowbar-core/pull/1094


jsuchome:
[(2) S55P2 Deactivate old repos during the upgrade](https://trello.com/c/GLSLmAhN/306-2-s55p2-deactivate-old-repos-during-the-upgrade)

PR #1115, opened 2017-02-20
upgrade: Mark the next substep as running

jsuchome:
If substep is immediatelly following after previous successfull
substep, we have to mark it as running (like controller nodes
after ceph ones).


jsuchome:
BTW I broke it in https://github.com/crowbar/crowbar-core/pull/1100

PR #1116, opened 2017-02-20
crowbar: Remove forgotten nodes from proposals

vuntz:
Forgotten nodes are not removed from existing proposals (applied and
non-applied). While the webui removes non-existing nodes when displaying
a proposal, if only the API is used, applying a proposal with such
non-existing nodes may result in failures.

Closes https://github.com/sap-oc/crowbar-core/issues/8


matelakat:
@vuntz shall we merge this, or the CI failure is a blocker?

matelakat:
@aspiers could you weigh in please, we need your help :-)

aspiers:
The CI failure is a standard haproxy failure which has been happening for a while. I doubt the CI even exercises this code path.

PR #1118, opened 2017-02-27
upgrade: Precheck for keystone hybrid backend.

jsuchome:
Hybrid backend is not upgradable (or is it not supported with Cloud7 at all?), so we are adding precheck at the start of the upgrade.


jsuchome:
[(2) Add precheck for keystone hybrid backend](https://trello.com/c/TKuEV0OH/2237-2-add-precheck-for-keystone-hybrid-backend)

PR #1119, opened 2017-02-27
upgrade: Add a precheck for cinder-volume backends

jsuchome:
cinder-volume needs to be configured with HA to enable non-disruptive upgrade

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -220,15 +220,12 @@ module Api
         ret
       end
 
-      # Check for presence of HA setup, which is a requirement for non-disruptive upgrade
-      def ha_presence_check
-        unless addon_installed? "ha"
-          return { errors: [I18n.t("api.upgrade.prechecks.ha_configured.not_installed")] }
-        end
+      # Check for presence and state of HA setup, which is a requirement for non-disruptive upgrade
+      def ha_config_check
+        return { ha_not_installed: true } unless addon_installed? "ha"
         founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
-        founders.empty? ?
-          { errors: [I18n.t("api.upgrade.prechecks.ha_configured.not_configured")] }
-          : {}
+        return { ha_not_configured: true } if founders.empty?
+        {}
       end
 
       def maintenance_updates_check
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -225,6 +225,15 @@ module Api
         return { ha_not_installed: true } unless addon_installed? "ha"
         founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
         return { ha_not_configured: true } if founders.empty?
+
+        # Check if cinder is using correct backend enabling live-migration
+        prop = Proposal.where(barclamp: "cinder").first
+
+        backends = prop["attributes"]["cinder"]["volumes"].select do |volume|
+          backend_driver = volume["backend_driver"]
+          ["local", "raw"].include? backend_driver
+        end
+        return { cinder_wrong_backend: true } unless backends.empty?
         {}
       end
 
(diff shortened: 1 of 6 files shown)


jsuchome:
[(2) Adapt upgrade pre-checks to ensure cinder-volume is HA](https://trello.com/c/I0uNF2Gb/2084-2-adapt-upgrade-pre-checks-to-ensure-cinder-volume-is-ha)

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -201,15 +201,18 @@ def compute_status
...
+        return { ha_not_installed: true } unless addon_installed? "ha"
         founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
-        founders.empty? ?
-          { errors: [I18n.t("api.upgrade.prechecks.ha_configured.not_configured")] }
-          : {}
+        return { ha_not_configured: true } if founders.empty?
+
+        # Check if cinder-volume is highly available
+        prop = Proposal.where(barclamp: "cinder").first
+        elements = prop["deployment"]["cinder"]["elements"]["cinder-volume"] rescue []
```
jsuchome: I think it makes sense here

jsuchome:
@skazi0 New precheck error label for UI...

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -201,15 +201,18 @@ def compute_status
...
+        non_clustered = elements.select { |e| !ServiceObject.is_cluster? e }
```
rhafer: nit: I believe you could use ```elements.any? { |e| !ServiceObject.is_cluster? e }``` here which would return a true/false value.

jsuchome: true

jsuchome: Or not, i need the value of `non_clustered` as a return value anyway....

rhafer: Ah, right. Not sure why I overlooked that, sorry.

jsuchome:
As we've discussed recently, having cinder-volume HA might not actually be what we need for non-disruptive upgrade. 

Maybe we just need to check for role allocation of cinder-volume roles?

Or should we do some specific check for backends? Something like: RADOS backend is always good, even without extra HA? Or, raw backend is always bad and so on.

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -201,15 +201,21 @@ def compute_status
...
+        return { ha_not_configured: true } if founders.empty?
+
+        # Check if cinder is using correct backend enabling live-migration
+        prop = Proposal.where(barclamp: "cinder").first
+
+        backends = prop["attributes"]["cinder"]["volumes"].select do |volume|
+          backend_driver = volume["backend_driver"]
+          ["local", "raw"].include? backend_driver
+        end
+        return { cinder_backend: true } unless backends.empty?
```
Itxaka: minor nit: use a proper var that explains what is checked. cinder_backend = true does not explain what this var represents.

Itxaka:
oh wait, wip label was removed!

jsuchome:
@Itxaka i think I've addressed the issues

jsuchome:
> Good to me once the hound issues are resolved.

I'm reluctant to 'resolve' these. I think it's gonna fail with some other problem when I follow the recommendation and current way is sync with what we already have in those files...

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -220,15 +220,21 @@ def compute_status
...
+        return { ha_not_configured: true } if founders.empty?
+
+        # Check if cinder is using correct backend enabling live-migration
+        prop = Proposal.where(barclamp: "cinder").first
+
+        backends = prop["attributes"]["cinder"]["volumes"].select do |volume|
+          backend_driver = volume["backend_driver"]
+          ["local", "raw"].include? backend_driver
+        end
+        return { cinder_wrong_backend: true } unless backends.empty?
```
Itxaka: :+1:

PR #1120, opened 2017-02-27
upgrade: Precheck for lbaas version

jsuchome:
We only allow lbaas v2 during the upgrade.

UPDATE: Added check if lbaas config is active. If I read https://docs.openstack.org/admin-guide/networking-adv-features.html#basic-load-balancer-as-a-service-operations correctly, pool is needed for all other kind of objects, so I'm only listing pools in the check ... correct me if I'm wrong

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -188,6 +188,20 @@ module Api
         return ret if prop.nil?
         driver = prop["attributes"]["identity"]["driver"] || "sql"
         ret[:keystone_hybrid_backend] if driver == "hybrid"
+
+        # check for lbaas version
+        prop = Proposal.where(barclamp: "neutron").first
+        return ret if prop.nil?
+        if prop["attributes"]["neutron"]["use_lbaas"] &&
+            !prop["attributes"]["neutron"]["use_lbaasv2"]
+
+          # So lbaas v1 is configured, let's find out if it is actually used
+          neutron = NodeObject.find("roles:neutron-server").first
+          out = neutron.run_ssh_cmd(
+            "source /root/.openrc; neutron lb-pool-list -f value -c id"
+          )
+          ret[:lbaas_v1] = true unless out[:stdout].nil? || out[:stdout].empty?
+        end
         ret
       end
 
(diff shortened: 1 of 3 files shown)


jsuchome:
[upgrade: Add precheck for lbaasv2](https://trello.com/c/Zyn65Owv/2274-upgrade-add-precheck-for-lbaasv2)

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -835,6 +835,9 @@ en:
         swift_replicas:
           error: 'The number of replicas is bigger than the number of disks assigned for Swift.'
           help: 'Such configuration is not supported in Swift delivered with SOC 7. Adapt your Swift configuration before proceeding with the upgrade.'
+        lbaas_v1:
+          error: 'Old version of LBaaS is configured.'
+          help: 'Such configuration is not supported during the upgrade. Adapt your Neutron configuration to use LBaaS v2 before you continue.'
```
rhafer: "Such a configuration is no supported for the upgrade. Please switch to LBaaS v2 before you continue"

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -175,14 +175,23 @@ def openstack_check
         ret = {}
         # swift replicas check vs. number of disks
         prop = Proposal.where(barclamp: "swift").first
-        return ret if prop.nil?
-        replicas = prop["attributes"]["swift"]["replicas"] || 0
+        unless prop.nil?
+          replicas = prop["attributes"]["swift"]["replicas"] || 0
```
abel-navarro: this change, although needed, is unrelated to this PR. Overall looks good to me.

jsuchome: Well, it was not needed until this PR. Do you really think there should be separate PR for that?

abel-navarro: no need for a separate PR although a separate commit would help.

jsuchome:
So I've added some kind of check if lbaas config is active. If I read https://docs.openstack.org/admin-guide/networking-adv-features.html#basic-load-balancer-as-a-service-operations correctly, pool is needed for all other kind of objects, so I'm only listing pools in the check ... correct me if I'm wrong @vuntz @rhafer @abel-navarro

PR #1121, opened 2017-02-28
provisioner: Define nameservers in autoyast.xml (bsc#1026837)

vuntz:
After the installation, but before chef is run, DNS resolution won't
work without this. This is critical in case the repositories are
configured with a DNS name, as this means the nodes won't be able to be
automatically configured (since packages won't be installable).

https://bugzilla.suse.com/show_bug.cgi?id=1026837

requests?**


Itxaka:
[(3) S56P3 bsc#1026837 - L3: external smt server: node installation failed due to empty resolv.conf](https://trello.com/c/Y53TcdMX/475-3-s56p3-bsc-1026837-l3-external-smt-server-node-installation-failed-due-to-empty-resolv-conf)

PR #1126, opened 2017-03-01
Disruptive upgrade

jsuchome:

Disruptive upgrade is needed for users without the setup that allows non-disruptive upgrade

Implements the proposal described at https://etherpad.nue.suse.com/p/cloud-upgrade-disruptive

**For followup** 

- ~~Showing some nice status when doing paralel parts: https://github.com/crowbar/crowbar-core/pull/1133~~ **Merged**
- More paralelization (currently it is done only for OS upgrade of nodes)

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-evacuate-host.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-evacuate-host.sh.erb
@@ -47,7 +47,7 @@ fi
 
 # FIXME: here we need checks for potential errors during the evacuation
 
-while nova list --host "$host" --fields host 2>/dev/null | grep "$host"; do
+while nova list --all-tenants --host "$host" --fields host,status 2>/dev/null | grep -e ACTIVE -e MIGRATING | grep -q "$host"; do
     log "There is still some VM running at $host..."
     sleep 10
 done
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -370,6 +370,32 @@ module Api
         end
       end
 
+      # Check if they are any instances running.
+      # It's necessary for disruptive upgrade, when user has to shut them down manually before
+      # proceeding with the services shutdown.
+      def check_for_running_instances
+        controller = ::Node.find("run_list_map:nova-controller").first
+        if controller.nil?
+          save_upgrade_state("No nova-controller node found.")
+          return
+        end
+        cmd = "openstack server list --all-projects --long " \
+          "--status active -f value -c Host"
+        out = controller.run_ssh_cmd("source /root/.openrc; #{cmd}")
+        unless out[:exit_code].zero?
+          raise_services_error(
+            "Error happened when trying to list running instances.\n" \
+            "Command '#{cmd}' failed at node #{controller.name} " \
+            "with #{out[:exit_code]}."
+          )
+        end
+        return if out[:stdout].nil? || out[:stdout].empty?
+        hosts = out[:stdout].split.uniq.join("\n")
+        raise_services_error(
+          "Following compute nodes still have instances running:\n#{hosts}."
+        )
+      end
+
       #
       # service shutdown
       #
@@ -391,7 +417,7 @@ module Api
           return
         end
 
-        # FIXME: report an error if there is running instances and mode is disruptive
+        check_for_running_instances if upgrade_mode == :normal
 
         # Initiate the services shutdown for all nodes
         errors = []
@@ -435,6 +461,14 @@ module Api
         else
           ::Crowbar::UpgradeStatus.new.end_step
         end
+      rescue ::Crowbar::Error::Upgrade::ServicesError => e
+        ::Crowbar::UpgradeStatus.new.end_step(
+          false,
+          running_instances: {
+            data: e.message,
+            help: "Suspend or stop all instances before you continue with the upgrade."
+          }
+        )
       rescue StandardError => e
         ::Crowbar::UpgradeStatus.new.end_step(
           false,
@@ -735,6 +769,11 @@ module Api
       def raise_node_upgrade_error(message = "")
         Rails.logger.error(message)
         raise ::Crowbar::Error::Upgrade::NodeError.new(message)
+      end
+
+      def raise_services_error(message = "")
+        Rails.logger.error(message)
+        raise ::Crowbar::Error::Upgrade::ServicesError.new(message)
       end
 
       protected
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -1110,12 +1110,12 @@ module Api
 
       def prepare_all_compute_nodes
         ["kvm", "xen"].each do |virt|
-          prepare_compute_nodes virt
+          prepare_compute_nodes virt, upgrade_mode
         end
       end
 
       # Prepare the compute nodes for upgrade by upgrading necessary packages
-      def prepare_compute_nodes(virt)
+      def prepare_compute_nodes(virt, upgrade_mode = :non_disruptive)
         save_upgrade_state("Preparing #{virt} compute nodes for upgrade... ")
         compute_nodes = ::Node.find("roles:nova-compute-#{virt}")
         if compute_nodes.empty?
@@ -1140,12 +1140,14 @@ module Api
             120
           )
           save_upgrade_state("Repositories prepared successfully.")
-          execute_scripts_and_wait_for_finish(
-            compute_nodes,
-            "/usr/sbin/crowbar-pre-upgrade.sh",
-            300
-          )
-          save_upgrade_state("Services on compute nodes upgraded and prepared.")
+          if upgrade_mode == :non_disruptive
+            execute_scripts_and_wait_for_finish(
+              compute_nodes,
+              "/usr/sbin/crowbar-pre-upgrade.sh",
+              300
+            )
+            save_upgrade_state("Services on compute nodes upgraded and prepared.")
+          end
         rescue StandardError => e
           raise_node_upgrade_error(
             "Error while preparing services on compute nodes. " + e.message
@@ -1159,6 +1161,39 @@ module Api
       def upgrade_all_compute_nodes
         ["kvm", "xen"].each do |virt|
           upgrade_compute_nodes virt
+        end
+      end
+
+      def parallel_upgrade_compute_nodes(compute_nodes)
+        save_upgrade_state("Entering disruptive upgrade of compute nodes")
+        save_nodes_state(compute_nodes, "compute", "upgrading")
+        ::Crowbar::UpgradeStatus.new.save_current_node_action("upgrading the packages")
+        begin
+          execute_scripts_and_wait_for_finish(
+            compute_nodes,
+            "/usr/sbin/crowbar-upgrade-os.sh",
+            900
+          )
+          save_upgrade_state("Repositories prepared successfully.")
+        rescue StandardError => e
+          raise_node_upgrade_error(
+            "Error while upgrading compute nodes. " + e.message
+          )
+        end
+        # FIXME: should we paralelize this as well?
+        compute_nodes.each do |node|
+          next if node.upgraded?
+          if node.ready_after_upgrade?
+            save_upgrade_state(
+              "Node #{node.name} is ready after the initial chef-client run."
+            )
+          else
+            node_api = Api::Node.new node.name
+            node_api.save_node_state("compute", "upgrading")
+            node_api.reboot_and_wait
+            node_api.join_and_chef
+          end
+          node_api.save_node_state("compute", "upgraded")
         end
       end
 
@@ -1178,6 +1213,8 @@ module Api
           )
           return
         end
+
+        return parallel_upgrade_compute_nodes(compute_nodes) if upgrade_mode == :normal
 
         controller = fetch_nova_controller
 
@@ -1203,7 +1240,7 @@ module Api
         controller
       end
 
-      def upgrade_one_compute_node(name)
+      def upgrade_one_compute_node(name, upgrade_mode = :non_disruptive)
         controller = fetch_nova_controller
         node = ::Node.find_node_by_name_or_alias(name)
         if node.nil?
@@ -1211,11 +1248,11 @@ module Api
             "No node with '#{name}' name or alias was found. "
           )
         end
-        upgrade_compute_node(controller, node)
+        upgrade_compute_node(controller, node, upgrade_mode)
       end
 
       # Fully upgrade one compute node
-      def upgrade_compute_node(controller, node)
+      def upgrade_compute_node(controller, node, upgrade_mode = :non_disruptive)
         return if node.upgraded?
         node_api = Api::Node.new node.name
         node_api.save_node_state("compute", "upgrading")
@@ -1226,11 +1263,16 @@ module Api
             "Node #{node.name} is ready after the initial chef-client run."
           )
         else
-          live_evacuate_compute_node(controller, hostname)
+          live_evacuate_compute_node(controller, hostname) if upgrade_mode == :non_disruptive
           node_api.os_upgrade
           node_api.reboot_and_wait
           node_api.post_upgrade
           node_api.join_and_chef
+        end
+
+        if upgrade_mode == :normal
+          node_api.save_node_state("compute", "upgraded")
+          return
         end
 
         controller.run_ssh_cmd(
@@ -1294,6 +1336,7 @@ module Api
       # Wait until all scripts at all nodes correctly finish or until some error is detected
       def execute_scripts_and_wait_for_finish(nodes, script, seconds)
         nodes.each do |node|
+          save_upgrade_state("Executing script '#{script}' at #{node.name}")
           ssh_status = node.ssh_cmd(script).first
           if ssh_status != 200
             raise "Execution of script #{script} has failed on node #{node.name}."
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -584,6 +584,14 @@ module Api
         upgrade_status.initialize_state
       end
 
+      def upgrade_pacemaker_cluster(cluster_env)
+        founder = ::Node.find(
+          "pacemaker_founder:true AND " \
+          "pacemaker_config_environment:#{cluster_env}"
+        ).first
+        upgrade_cluster founder, cluster_env
+      end
+
       # Take the list of elements as argument (that could be both nodes and clusters)
       def upgrade_nodes_disruptive(elements_to_upgrade)
         elements_to_upgrade.each do |element|
@@ -591,15 +599,16 @@ module Api
           if ServiceObject.is_cluster?(element)
             cluster = ServiceObject.cluster_name element
             cluster_env = "pacemaker-config-#{cluster}"
-            founder = ::Node.find(
-              "pacemaker_founder:true AND " \
-              "pacemaker_config_environment:#{cluster_env}"
-            ).first
-            upgrade_cluster founder, cluster_env
+            upgrade_pacemaker_cluster cluster_env
           else
-            # if role has single node(s) assigned upgrade those nodes (optionally all at once)
-            # FIXME: could we do it in paralel?
-            upgrade_one_node element
+            node = ::Node.find_by_name(element)
+            if node["run_list_map"].key? "pacemaker-cluster-member"
+              # if the node is part of some cluster, upgrade the whole cluster
+              upgrade_pacemaker_cluster node[:pacemaker][:config][:environment]
+            else
+              # if role has single node(s) assigned upgrade those nodes (optionally all at once)
+              upgrade_one_node element
+            end
           end
         end
 
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -391,6 +391,8 @@ module Api
           return
         end
 
+        # FIXME: report an error if there is running instances and mode is disruptive
+
         # Initiate the services shutdown for all nodes
         errors = []
         upgrade_nodes = ::Node.find("state:crowbar_upgrade AND NOT roles:ceph-*")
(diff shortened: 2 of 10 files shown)


jsuchome:
[(8) S56P5 Implement disruptive upgrade (non-HA, one compute node, etc.)](https://trello.com/c/2B0jUWrs/334-8-s56p5-implement-disruptive-upgrade-non-ha-one-compute-node-etc)

chef/cookbooks/crowbar/templates/default/crowbar-evacuate-host.sh.erb
```diff
@@ -47,7 +47,7 @@ fi
 
 # FIXME: here we need checks for potential errors during the evacuation
 
-while nova list --host "$host" --fields host 2>/dev/null | grep "$host"; do
+while nova list --all-tenants --status active --host "$host" --fields host 2>/dev/null | grep "$host"; do
```
Itxaka: Not sure about status active. What about instances that are shutting off? They are still running until they finally shut off so they should be taken into account?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -535,6 +572,63 @@ def cancel
...
+            to_upgrade |= elements[el]
+          end
+        end
+        to_upgrade
+      end
+
+      def upgrade_disruptive_controllers
+        save_upgrade_state("Entering disruptive upgrade of controller nodes")
+
+        # Go through all OpenStack barclamps (skip "pacemaker" and "openstack") ordered by run_order
```
Itxaka: Is this comment correct? Seems that it actually skips anything NOT on the openstack category?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -535,6 +572,63 @@ def cancel
...
+          if ServiceObject.is_cluster?(element)
+            cluster = ServiceObject.cluster_name element
+            cluster_env = "pacemaker-config-#{cluster}"
+            founder = ::Node.find(
+              "pacemaker_founder:true AND " \
+              "pacemaker_config_environment:#{cluster_env}"
+            ).first
+            upgrade_cluster founder, cluster_env
+          else
+            # if cluster(s) and node(s) assigned to the role are already upgraded, skip to next role
```
Itxaka: Is this done here or its already coded in the `upgrade_one_node` function? `return if node.upgraded?`

chef/cookbooks/crowbar/templates/default/crowbar-evacuate-host.sh.erb
jsuchome: This is actually used during the live migration, where instances are not shutting down, but moving to different host. (So it's maybe confusing to have it in this PR).

Anyway, your comment probably still applies, do you have any idea how to make the check better? We need to make sure that the live-migration was successfully finished and the $host in question does not have any instances running (it could have some not-running though).

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Yes, that's actually solved by `node.upgraded?` check later. (I should remove the comment)

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Here I'm not sure what was the meaning of "openstack" here, there's no barclamp like that... @rhafer ?

(It's true that we are skipping all nodes that do not have any openstack roles assigned here... that was not intentional, but I think it could be solved at the end of upgrade by a check for 'anything that is left')

chef/cookbooks/crowbar/templates/default/crowbar-evacuate-host.sh.erb
Itxaka: Probably get all instances and filter with a whitelist for statuses that we know are troublesome? I cant remember the status of an instance migrating, but its probably not `active` so we could be going forward while there is still instances migrating :/

jsuchome: @Itxaka check the new state, I'm grepping for MIGRATING status now

jsuchome:
Needs rebase after merging https://github.com/crowbar/crowbar-core/pull/1129

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -535,6 +572,81 @@ def cancel
...
+        end
+
+        # For each active proposal go through the roles in element order
+        proposals.each do |name, propmap|
+          proposal = propmap[:prop]
+          elements = proposal["deployment"][name]["elements"]
+          proposal["deployment"][name]["element_order"].flatten.each do |el|
+            # skip nova compute nodes, they will be upgraded separately
+            next if el.start_with?("nova-compute") || elements[el].nil?
+            # skip cinder-volume nodes that also have a nova compute role assigned
```
jsuchome: We might want to skip also `ceilometer-agent` nodes...

crowbar_framework/app/models/api/upgrade.rb
rhafer: > Here I'm not sure what was the meaning of "openstack" here, there's no barclamp like that... @rhafer ?

The ```prop.category != "OpenStack"``` takes information from the ```config/catalog.yml```. Which itself is generated from the various ```<barclampname>.yml``` in the git repositores. There you can define, that a certain barclamp is a "member" of another barclamp. ```prop.category != "OpenStack"``` matches all barclamps that are member of the "OpenStack" barclamp (yes, there is a barclamp called openstack, it doesn't do much and does really provide a UI)

> It's true that we are skipping all nodes that do not have any openstack roles assigned here... that was not intentional, but I think it could be solved at the end of upgrade by a check for 'anything that is left'

What would that be?

jsuchome: So, you're saying that the check is correct or not?

Itxaka: This is not correct or at least it does not sync with the comment
> Go through all OpenStack barclamps (skip "pacemaker" and "openstack") ordered by run_order

This is not skipping the openstack barclamp but its skipping any barclamps that are NOT in the openstack category, like `ntp` or `provisioner` which have the `crowbar` category. Which Im guessing its the end goal here, so the comment should change.

```
irb(main):021:0> bar = []
irb(main):022:0> BarclampCatalog.barclamps.each do |k, v|
irb(main):023:1* p = Proposal.where(barclamp: k).first
irb(main):024:1> next if p.nil?
irb(main):025:1> next if p.category != "OpenStack"
irb(main):026:1> next if k == "pacemaker"
irb(main):027:1> bar.push(k)
irb(main):028:1> end
irb(main):029:0> bar
=> ["nova", "keystone", "rabbitmq", "neutron", "cinder", "glance", "horizon", "database"]
```

jsuchome: > This is not correct or at least it does not sync with the comment

That's what I tried to say earlier https://github.com/crowbar/crowbar-core/pull/1126#discussion_r104113114 . I believe the code is correct, and the comment should go, unless @rhafer (who wrote the original like) disagrees.

jsuchome: @Itxaka oh, so once again, we are skipping barclamp named `openstack`, but that does not mean we are skipping other barclamps in `OpenStack` category

jsuchome:
Rebased after the mode renaming

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -535,6 +572,81 @@ def cancel
...
+            # FIXME: could we do it in paralel?
+            upgrade_one_node element
+          end
+        end
+
+        # and handle those as part of the compute node upgrade
+      end
+
+      # Go through active proposals and return elements with assigned roles in the right order
+      def upgradable_elements_of_proposals(proposals)
```
rhafer: I somehow agree with Hound here that this is a bit to complex. I am currently experimenting a bit with this to see if we could make it a bit simpler.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -535,6 +537,83 @@ def cancel
...
+        # For each active proposal go through the roles in element order
+        proposals.each do |name, propmap|
+          proposal = propmap[:prop]
+          elements = proposal["deployment"][name]["elements"]
+          proposal["deployment"][name]["element_order"].flatten.each do |el|
+            # skip nova compute nodes, they will be upgraded separately
+            next if el.start_with?("nova-compute") || elements[el].nil?
+            # skip some roles if they are assigned to compute nodes
+            if ["cinder-volume", "ceilometer-agent", "swift-storage"].include? el
+              elements[el].each do |node|
```
rhafer: I think this won't work if e.g. ```cinder-volume``` is assigned to a cluster. In that case ```elements[el]``` will contain ```cluster:<name>```.  So it's probably better to expand the elements here. ```ServiceObject.expand_nodes_for_all``` should be able to do that.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -535,6 +537,83 @@ def cancel
...
+        to_upgrade = []
+
+        # First find out the nodes with compute role, for later checks
+        compute_nodes = {}
+        nova_prop = proposals["nova"][:prop] || nil
+        unless nova_prop.nil?
+          elements = nova_prop["deployment"]["nova"]["elements"]
+          elements.each do |role, nodes|
+            next unless role.start_with?("nova-compute")
+            nodes.each { |node| compute_nodes[node] = 1 }
```
rhafer: In compute-ha setups this is not a list of nodes but somethink like ```remotes:<name>```. Should also be expandable using ```ServiceObject.expand_nodes_for_all```.

But then again it seems compute-ha is using a completely different role ..., hm.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -535,6 +537,83 @@ def cancel
...
+      # Go through active proposals and return elements with assigned roles in the right order
+      def upgradable_elements_of_proposals(proposals)
+        to_upgrade = []
+
+        # First find out the nodes with compute role, for later checks
+        compute_nodes = {}
+        nova_prop = proposals["nova"][:prop] || nil
+        unless nova_prop.nil?
+          elements = nova_prop["deployment"]["nova"]["elements"]
+          elements.each do |role, nodes|
```
rhafer: Nit: Why not ```nova_prop["deployment"]["nova"]["elements"].each do |role, nodes|```?

crowbar_framework/app/models/api/upgrade.rb
rhafer: In general this function seems to have an issue with mixed cluster/non-cluster setups. E.g. if keystone is deployed non-HA on a node that is a cluster member for some later barclamp (glance, neutron, cinder, ...) it won't upgrade the whole cluster.

I know this is not really a supported setup (and no sane person would install that it that way), but we might want to deal with it in  the disruptive upgrade path somehow. (Solution would be to load the node objects and check if they are a cluster member).

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -535,6 +537,83 @@ def cancel
...
+        save_upgrade_state("Entering disruptive upgrade of controller nodes")
+
+        # Go through all OpenStack barclamps ordered by run_order
+        proposals = {}
+        BarclampCatalog.barclamps.map do |name, attrs|
+          next if BarclampCatalog.category(name) != "OpenStack"
+          next if ["pacemaker", "openstack"].include? name
+          prop = Proposal.where(barclamp: name).first
+          next if prop.nil? || !prop.active?
+          proposals[name] = { order: BarclampCatalog.run_order(name), prop: prop }
```
rhafer: This still seems a bit weird to be. I think you don't need the order here. Just doing:

```proposals[name] = prop```

 here and sorting it later with:

```proposals = proposals.sort_by { |key, _unused| BarclampCatalog.run_order(key) }.to_h```

should do the trick. (Or sorting the barclamps already before loading the proposal and writing the ```proposal``` hash in a sorted fashion from the beginning as suggested in my initial comment.

File: crowbar_framework/app/models/api/upgrade.rb [hunk 10]
```diff
@@ -535,6 +537,83 @@ def cancel
...
+              end
+            else
+              to_upgrade |= elements[el]
+            end
+          end
+        end
+        to_upgrade
+      end
+
+      def upgrade_disruptive_controllers
```
rhafer: To allign with ```upgrade_nodes_disruptive``` I suggest to rename this to ```upgrade_controllers_disruptive```

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -535,6 +537,83 @@ def cancel
         upgrade_status.initialize_state
       end
 
+      def upgrade_nodes_disruptive(to_upgrade)
```
rhafer: Please add a comment that this is really working with elements and not nodes. (Maybe even rename it).

Also I suggest to rename ```to_upgrade``` to ```elements_to_upgrade```

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -357,6 +357,32 @@ def target_platform(options = {})
...
+          raise_services_error(
+            "Error happened when trying to list running instances.\n" \
+            "Command '#{cmd}' failed at node #{controller.name} " \
+            "with #{out[:exit_code]}."
+          )
+        end
+        return if out[:stdout].nil? || out[:stdout].empty?
+        hosts = out[:stdout].split.uniq.join("\n")
+        raise_services_error(
+          "There are some running instances at nodes\n#{hosts}."
```
rhafer: ```The following compute nodes still have instances running: ....```

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -378,7 +404,8 @@ def services
           return
         end
 
-        # FIXME: report an error if there is running instances and mode is disruptive
+        status = ::Crowbar::UpgradeStatus.new
+        check_for_running_instances if status.upgrade_mode == "disruptive"
```
rhafer: Using just ```upgrade_mode``` should work here.

crowbar_framework/app/models/api/upgrade.rb
jsuchome: true

crowbar_framework/app/models/api/upgrade.rb
jsuchome: will do

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Good point, but can we save it to compute-ha work? Probably some other code parts will be affected as well...

crowbar_framework/app/models/api/upgrade.rb
jsuchome: `cinder-volume` in cluster and compute node at the node that forms this cluster? That's probably broken... :-)

rhafer: "unsupported" yes, But not sure if it's completely broken. And I don't think we currently prevent such a setup from being upgraded. (Maybe we should)

crowbar_framework/spec/models/api/upgrade_spec.rb
```diff
@@ -367,7 +379,175 @@
...
+        suggested_upgrade_mode: :normal
+      )
+      expect(Crowbar::UpgradeStatus.new.upgrade_mode).to be :normal
+
+      allow(Api::Upgrade).to receive(:upgrade_controllers_disruptive).and_return(true)
+      allow(Api::Upgrade).to receive(:prepare_all_compute_nodes).and_return(true)
+
+      allow(Node).to(
+        receive(:find).with("roles:nova-compute-kvm").
+        and_return([Node.find_by_name("testing.crowbar.com")])
```
rhafer: Would be good to really test this with multiple compute nodes. And verify that ```execute_scripts_and_wait_for_finish``` is actually called with multiple nodes passed.

crowbar_framework/spec/models/api/upgrade_spec.rb
```diff
@@ -367,7 +379,175 @@
...
+    it "successfully upgrades compute nodes" do
+      allow(Node).to(
+        receive(:find).with("state:crowbar_upgrade").
+        and_return([Node.find_by_name("testing.crowbar.com")])
+      )
+      allow(Api::Upgrade).to receive(:join_ceph_nodes).and_return(true)
+
+      allow_any_instance_of(Crowbar::UpgradeStatus).to receive(:progress).and_return(
+        suggested_upgrade_mode: :normal
+      )
```
rhafer: Same as above: 
As ```Crowbar::UpgradeStatus.new.upgrade_mode``` is already tested elsewhere. IMO it's ok to change this to:

```
allow_any_instance_of(Crowbar::UpgradeStatus).to receive(:upgrade_status).and_return( :normal )
```

here and delete the next line.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1143,7 +1145,38 @@ def upgrade_all_compute_nodes(upgrade_mode = :non_disruptive)
         end
       end
 
-      def upgrade_compute_nodes(virt)
+      def paralel_upgrade_compute_nodes(compute_nodes)
```
rhafer: Typo: "paralel"

crowbar_framework/app/models/api/upgrade.rb
rhafer: Sure

crowbar_framework/spec/models/api/upgrade_spec.rb
```diff
@@ -367,7 +379,175 @@
...
+    it "successfully upgrades controller nodes" do
+      allow(Node).to(
+        receive(:find).with("state:crowbar_upgrade").
+        and_return([Node.find_by_name("testing.crowbar.com")])
+      )
+      allow(Api::Upgrade).to receive(:join_ceph_nodes).and_return(true)
+
+      allow_any_instance_of(Crowbar::UpgradeStatus).to receive(:progress).and_return(
+        suggested_upgrade_mode: :normal
+      )
```
jsuchome: you mean `allow_any_instance_of(Crowbar::UpgradeStatus).to receive(:upgrade_mode)` right?

jsuchome:
Rebased after the merge of https://github.com/crowbar/crowbar-core/pull/1133. Added a call of `save_nodes_state` into `parallel_upgrade_compute_nodes`

crowbar_framework/spec/models/api/upgrade_spec.rb
rhafer: Oops. Yeah.

crowbar_framework/app/models/api/upgrade.rb
rhafer: it's ```status.upgrade_mode == :normal``` (you also don't really need the ```status``` here as Api::Upgrade provides an ```upgrade_mode``` method. But I am repeating myself ;)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1143,7 +1145,39 @@ def upgrade_all_compute_nodes(upgrade_mode = :non_disruptive)
...
+            "/usr/sbin/crowbar-upgrade-os.sh",
+            900
+          )
+          save_upgrade_state("Repositories prepared successfully.")
+        rescue StandardError => e
+          raise_node_upgrade_error(
+            "Error while upgrading compute nodes. " + e.message
+          )
+        end
+        # FIXME: should we paralelize this as well?
```
rhafer: Yeah, especially having the ```join_and_chef``` part  running in parallel could speed things up quite a bit. But it's ok to address that later IMO, just add it to the backlog.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -585,22 +585,31 @@ def cancel
...
-              "pacemaker_config_environment:#{cluster_env}"
-            ).first
-            upgrade_cluster founder, cluster_env
+            upgrade_pacemaker_cluster cluster_env
           else
-            # if role has single node(s) assigned upgrade those nodes (optionally all at once)
-            # FIXME: could we do it in paralel?
-            upgrade_one_node element
+            node = ::Node.find_by_name(element)
+            if node["run_list_map"].key? "pacemaker-cluster-member"
```
rhafer: IMO this code would better fit into ```upgradable_elements_of_proposals```. Where you would "just" replace the nodenane with the cluster element before inserting it into the result list.

And if you already load the nodes in ```upgradable_elements_of_proposals``` anyway you could also rework the check for computes a bit to also just look at the run_list_map.

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Also, let's keep in mind that for pure compute nodes, chef-client does not take that much time.

rhafer: Well even if it just takes 30 seconds per node (and I am pretty sure it takes longer) it sums about to 25 minutes for 50 minutes. But as I said it's ok to address that later.

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Hm, I'm not sure. I'm mostly working with lists in `upgradable_elements_of_proposals`, so loading a Node there seems like another level of complexity.

jsuchome:
Rebased after merging https://github.com/crowbar/crowbar-core/pull/1136

jsuchome:
Rebased after merge of https://github.com/crowbar/crowbar-core/pull/1134

jsuchome:
> My only concern is, once this gets in we need to remember to kill the manila-service vm for any 
> disruptive  CI jobs ...

Good point, but wasn't @abel-navarro already on it?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -370,6 +370,32 @@ def target_platform(options = {})
         end
       end
 
+      # Check if they are any instances running.
+      # It's necessary for disruptive upgrade, when user has to shut them down manually before
+      # proceeding with the services shutdown.
+      def check_for_running_instances
```
jsuchome: We'll need to add some check for this into our jobs - suspend running VM's before the services step. Including the manila one @Itxaka mentioned.

PR #1128, opened 2017-03-02
upgrade: Look at correct place for proposal attributes

jsuchome:
This was broken in https://github.com/crowbar/crowbar-core/pull/1118


jsuchome:
[(2) S56P13 Add precheck for keystone hybrid backend](https://trello.com/c/TKuEV0OH/341-2-s56p13-add-precheck-for-keystone-hybrid-backend)

PR #1129, opened 2017-03-03
upgrade:Allow overriding the detected upgrade mode

rhafer:
This is needed as we want to provide a way to force a "normal" upgrade on system that would support a "non-disruptive" upgrade.

This PR also include quite a bit of cleanup and fixes for the related unit tests.

As this got a bit more complex than initially expected as decided to split this into two pull requests. This one just changes the status library. The follow up PR will add the new Rest API .

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade_status.rb
+++ b/crowbar_framework/lib/crowbar/upgrade_status.rb
@@ -60,7 +60,7 @@ module Crowbar
         crowbar_backup: nil,
         openstack_backup: nil,
         # disruptive vs. nondisruptive
-        upgrade_mode: nil
+        suggested_upgrade_mode: nil
       }
       # in 'steps', we save the information about each step that was executed
       @progress[:steps] = upgrade_steps_6_7.map do |step|
@@ -70,8 +70,8 @@ module Crowbar
       save
     end
 
-    def upgrade_mode
-      progress[:upgrade_mode]
+    def suggested_upgrade_mode
+      progress[:suggested_upgrade_mode]
     end
 
     def current_substep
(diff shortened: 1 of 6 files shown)


crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -187,6 +202,26 @@ def save_suggested_upgrade_mode(mode)
...
+      end
+    end
+
+    def save_selected_upgrade_mode(mode)
+      ::Crowbar::Lock::LocalBlocking.with_lock(shared: false, logger: @logger, path: lock_path) do
+        load_while_locked
+        if progress[:suggested_upgrade_mode] == :normal && mode != :normal
+          raise ::Crowbar::Error::SaveUpgradeModeError.new(
+            "Upgrade mode '#{mode}' is not possible. " \
+            "Suggested upgrade mode '#{progress[:suggested_upgrade_mode]}'."
```
jsuchome: You could just call `suggested_upgrade_mode` method here.

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -183,10 +198,36 @@ def save_openstack_backup(backup_location)
...
+        end
+        save
+      end
+    end
+
+    def save_selected_upgrade_mode(mode)
+      ::Crowbar::Lock::LocalBlocking.with_lock(shared: false, logger: @logger, path: lock_path) do
+        load_while_locked
+        # It's ok to change the upgrade mode until starting the services step
+        unless pending? :services
```
jsuchome: This part is actually not needed on Cloud6 side (but probably better anyway for consistency...)

rhafer:
repush to address hound issues and @jsuchome's feedback

rhafer:
[(3) S56P4 API for forcing disruptive upgrade mode](https://trello.com/c/XY1ZBnMg/319-3-s56p4-api-for-forcing-disruptive-upgrade-mode)

PR #1131, opened 2017-03-06
upgrade: New API to set/get the current upgrade mode

rhafer:
Base on the recent status library changes it is now possible to request
a certain upgrade mode using POST/GET on /api/upgrade/mode.

### DIFF
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -227,6 +227,27 @@ class Api::UpgradeController < ApiController
     }, status: :unprocessable_entity
   end
 
+  def mode
+    if request.post?
+      Api::Upgrade.upgrade_mode = params[:mode]
+      render json: {}, status: :ok
+    else
+      render json: {
+        mode: Api::Upgrade.upgrade_mode
+      }
+    end
+  rescue ::Crowbar::Error::SaveUpgradeModeError,
+         ::Crowbar::Error::SaveUpgradeStatusError,
+         ::Crowbar::Error::UpgradeError => e
+    render json: {
+      errors: {
+        mode: {
+          data: e.message
+        }
+      }
+    }, status: :unprocessable_entity
+  end
+
   protected
 
   def backup_params
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -30,6 +30,19 @@ module Api
         }
       end
 
+      def upgrade_mode
+        ::Crowbar::UpgradeStatus.new.upgrade_mode
+      end
+
+      def upgrade_mode=(mode)
+        unless ["normal", "non_disruptive"].include? mode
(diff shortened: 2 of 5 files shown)


crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -227,6 +227,28 @@ def openstackbackup
...
+      }
+    end
+  rescue ::Crowbar::Error::SaveUpgradeModeError,
+         ::Crowbar::Error::SaveUpgradeStatusError,
+         ::Crowbar::Error::UpgradeError => e
+    render json: {
+      errors: {
+        mode: {
+          data: e.message,
+          help: "Please refer to the error message in the response."
```
jsuchome: I think we should just drop the `help` part if it does not include anything interesting

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -30,6 +30,17 @@ def node_status
...
 
+      def upgrade_mode
+          ::Crowbar::UpgradeStatus.new.upgrade_mode
+      end
+
+      def upgrade_mode=(mode)
+        unless ["normal", "non_disruptive"].include? mode
+          Rails.logger.debug("Setting upgrade mode #{mode}")
+          ::Crowbar::UpgradeStatus.new.save_selected_upgrade_mode(mode.to_sym)
+        end
```
jsuchome: What if user passes different value? There should be some help message showing allowed values in such case...

crowbar_framework/app/controllers/api/upgrade_controller.rb
skazi0: +1

rhafer: done

crowbar_framework/app/models/api/upgrade.rb
rhafer: True. Added that. And also a unit test for it.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -30,6 +30,18 @@ def node_status
         }
       end
 
+      def upgrade_mode
+        ::Crowbar::UpgradeStatus.new.upgrade_mode
+      end
+
+      def upgrade_mode=(mode)
+        unless ["normal", "non_disruptive"].include? mode
+          raise ::Crowbar::Error::SaveUpgradeModeError, "Invalid upgrade mode #{mode}."
```
jsuchome: Could we please print allowed values here?

rhafer: Hm, I actually added that to crowbar-client already (the PR for that is will follow soon). Buy yeah, I can add it here as well.

rhafer: Done

rhafer:
[(3) S56P4 API for forcing disruptive upgrade mode](https://trello.com/c/XY1ZBnMg/319-3-s56p4-api-for-forcing-disruptive-upgrade-mode)

PR #1133, opened 2017-03-06
upgrade: Save the state of more nodes being upgraded at once

jsuchome:
When an action is being done to set of nodes, we should
indicate the list of nodes being processed via the status library.


jsuchome:
[(8) S56P5 Implement disruptive upgrade (non-HA, one compute node, etc.)](https://trello.com/c/2B0jUWrs/334-8-s56p5-implement-disruptive-upgrade-non-ha-one-compute-node-etc)

jsuchome:
After some discussion, new proposal saves only the number of nodes being upgraded, as the list of nodes could be too long and might not fit into the UI...

PR #1134, opened 2017-03-06
upgrade: Upgrade the nodes that are neither controllers nor compute

jsuchome:
Nodes that do not hold controller role but are not compute nodes
must be upgraded before compute nodes, so live migration
(executed during compute nodes upgrade) can rely on fully upgraded
services.

Currently conflicts with https://github.com/crowbar/crowbar-core/pull/1126 with `upgrade_one_node` method


jsuchome:
[(5) S56P3 Upgrade nodes without controller/compute role](https://trello.com/c/rNyRarfI/301-5-s56p3-upgrade-nodes-without-controller-compute-role)

jsuchome:
This will need some test coverage. But first we need that other PR to merge

PR #1136, opened 2017-03-07
upgrade: Save more detailed information about current upgrade activity

jsuchome:
It seems useful to show more about what we are doing to nodes,
especially when it takes more time.
This extends the previous information saved as part of current_node,
which is marked only at the beginning and end of upgrade.

### DIFF
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -29,6 +29,7 @@ module Api
     end
 
     def pre_upgrade
+      save_node_action("preparing node for the upgrade")
       execute_and_wait_for_finish("/usr/sbin/crowbar-pre-upgrade.sh", 300)
       Rails.logger.info("Pre upgrade script run was successful.")
     rescue StandardError => e
@@ -38,6 +39,7 @@ module Api
     end
 
     def prepare_repositories
+      save_node_action("updating repository configuration")
       execute_and_wait_for_finish("/usr/sbin/crowbar-prepare-repositories.sh", 100)
       Rails.logger.info("Prepare of repositories was successful.")
     rescue StandardError => e
@@ -47,6 +49,7 @@ module Api
     end
 
     def os_upgrade
+      save_node_action("upgrading the packages")
       execute_and_wait_for_finish("/usr/sbin/crowbar-upgrade-os.sh", 900)
       Rails.logger.info("Package upgrade was successful.")
     rescue StandardError => e
@@ -57,6 +60,11 @@ module Api
 
     # Execute post upgrade actions: prepare drbd and start pacemaker
     def post_upgrade
+      if @node["drbd"] && @node["drbd"]["rsc"] && @node["drbd"]["rsc"].any?
+        save_node_action("synchronizing DRBD")
+      else
+        save_node_action("doing post-upgrade cleanup")
+      end
       execute_and_wait_for_finish("/usr/sbin/crowbar-post-upgrade.sh", 600)
       Rails.logger.info("Post upgrade script run was successful.")
     rescue StandardError => e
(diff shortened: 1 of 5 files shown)


jsuchome:
[save more info about node upgrade](https://trello.com/c/YqcXCr5B/344-save-more-info-about-node-upgrade)

jsuchome:
On the other hand, maybe it should go inside `current_mode` hash...

skazi0:
why not use the `state` field in the current_node(s)? is it used for something right now? I only saw `upgrading` and `upgraded` there.

jsuchome:
It could be `state`, yes ... however I'm not sure, `state` is currently used as a binary value, it's either `upgrading` or `upgraded` and in many places we only care about these values (like when we are trying to find out the node which has unfinished upgrade after the restart of failed upgrade step).

So I'd be more in favor of putting extra `action` inside of `current_node(s)`.

skazi0:
OK, then +1 on putting the action in current_node(s).

crowbar_framework/app/models/api/node.rb
```diff
@@ -57,6 +60,7 @@ def os_upgrade
 
     # Execute post upgrade actions: prepare drbd and start pacemaker
     def post_upgrade
+      save_node_action("post-upgrade cleanup")
```
jsuchome: This also includes DRBD sync which takes quite some time. We should rename the action if we know the node is DRBD controller...

jsuchome:
Another update after the chat: as the action is always the same even when we are upgrading more nodes at ones, so it makes sense to keep the action as one global value...

crowbar_framework/app/models/api/node.rb
rhafer: +1

crowbar_framework/app/models/api/node.rb
```diff
@@ -38,6 +39,7 @@ def pre_upgrade
     end
 
     def prepare_repositories
+      save_node_action("preparing repositories")
```
rhafer: How about "upgrading repository configuration"? "preparing repositories" is very generic and to me sound a bit like "doing something" :)

crowbar_framework/app/models/api/node.rb
```diff
@@ -66,6 +70,7 @@ def post_upgrade
     end
 
     def join_and_chef
+      save_node_action("joining crowbar")
```
rhafer: I'd suggest something like: "upgrading configuration and re-joining the crowbar environment"

crowbar_framework/app/models/api/node.rb
```diff
@@ -122,6 +128,7 @@ def reboot_and_wait
       end
 
       wait_for_ssh_state(:down, "reboot")
+      save_node_action("waiting to come up")
```
rhafer: maybe: "wait for node to be back after reboot"

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -909,6 +910,7 @@ def delete_pacemaker_resources(node)
       # available network nodes. The evacuation procedure is started on the
       # specified controller node
       def evacuate_network_node(controller, network_node, delete_namespaces = false)
+        ::Crowbar::UpgradeStatus.new.save_current_node_action("migrating routers")
```
rhafer: I think "evacuating routers" is more precise here.

rhafer:
This of course conflicts with https://github.com/crowbar/crowbar-core/pull/1133/files, not sure which one you want to merge first.

jsuchome:
I think https://github.com/crowbar/crowbar-core/pull/1133 should go first, then I'll rebase this one.

jsuchome:
Rebased after merging https://github.com/crowbar/crowbar-core/pull/1133

PR #1142, opened 2017-03-09
upgrade: Cleanup the current progress at the end of upgrade

jsuchome:
Once the upgrade is finished, there's no point in showing the 'current' information that was used during the upgrade.

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade_status.rb
+++ b/crowbar_framework/lib/crowbar/upgrade_status.rb
@@ -144,8 +144,12 @@ module Crowbar
           errors: errors
         }
         if current_step == upgrade_steps_6_7.last && success
-          # Mark the end of the upgrade process
+          # Mark the end of the upgrade process and cleanup the progress
           FileUtils.rm_f running_file
+          progress[:current_substep] = :end_of_upgrade
+          progress[:current_substep_status] = :finished
+          progress[:current_nodes] = {}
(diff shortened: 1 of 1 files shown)


jsuchome:
[Delete 'current_nodes' when upgrade is finished](https://trello.com/c/j3td7lfi/348-delete-current-nodes-when-upgrade-is-finished)

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -144,8 +144,12 @@ def end_step(success = true, errors = {})
           errors: errors
         }
         if current_step == upgrade_steps_6_7.last && success
-          # Mark the end of the upgrade process
+          # Mark the end of the upgrade process and cleanup the progress
           FileUtils.rm_f running_file
+          progress[:current_substep] = :end_of_upgrade
+          progress[:current_substep_status] = :finished
+          progress[:current_nodes] = {}
+          progress[:current_node_action] = "finished"
```
skazi0: Why not set current action to `nil`?

jsuchome: Could be as well, yes. Would you prefer it?

skazi0: I added additional condition in the UI to not display this last action. I don't mind leaving it as is.

PR #1144, opened 2017-03-10
apply_role: Add progress logging

rsalevsky:
At the moment there is no easy way to watch the progress of applying a
proposal. This change introduces a new logger which logs progress
relevant messages to a seperate log. But also adds this messages as
"PROGRESS" messages to the production log.

Interesting for https://github.com/sap-oc/crowbar-core/issues/6

### DIFF
--- a/crowbar_framework/config/initializers/progress_logger.rb
+++ b/crowbar_framework/config/initializers/progress_logger.rb
@@ -1,0 +1,34 @@
+#
+# Copyright 2017, SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


crowbar_framework/config/initializers/progress_logger.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    log_file = if Rails.env.production?
+      File.join(ENV["CROWBAR_LOG_DIR"], "progress.log")
+    else
+      Rails.root.join("log", "progress.log")
+    end
+
+    logger = Logger.new(log_file)
+    logger.formatter = ::Crowbar::Logger::Formatter.new
+    logger.info(progname)
```
Itxaka: Is this part necessary? Seems like we are logging twice no? Other logger levels do nothing like this so could you clarify why is this needed?

Itxaka: oh, I see, this goes to a separate file + our normal file :+1:

rsalevsky: This part is necessary, but the `add(idx, nil, progname, &block)` could be worth discussing. To log the progress to a separate log is direct customer feedback to allow to track the progress in a easy way.

Itxaka: for me this is great, having on a separate file as per customer feedback + having it on the normal log file to have the full workflow in the same place, thanks!

PR #1146, opened 2017-03-10
Revert "upgrade: Precheck to make sure cinder is using correct backend"

jsuchome:
This reverts commit 7281972c3b4a7d7fbc0d5308708a75c7696d124c.

The commit is actually correct, but currently breaks upgrade CI, as we are not able to deploy setup that would use the correct (non-local) cinder backend. So I'm proposing a revert until we are able to fix the CI job for non-disruptive upgrade.


Itxaka:
Intead of deleting the code, can we not skip it for the moment with a comment to enable it in the future when the problem is fixed?

jsuchome:
@Itxaka that would still mean adding some extra code for that. I think clean revert is nicer and we will use it later anyway, but more time is needed spent on it...

jsuchome:
Reverting anyway. Should be commited again once we fix the CI.

PR #1143, opened 2017-03-10
[stable/3.0] upgrade: Save more detailed information about current upgrade activity

jsuchome:
(cherry picked from commit fd7fc1675e38baa5daed9471dadd5362bcb76811)

Status library related changes ported from https://github.com/crowbar/crowbar-core/pull/1136

PR #1147, opened 2017-03-10
[stable/3.0] upgrade: Cleanup the current progress at the end of upgrade

jsuchome:
(cherry picked from commit 92267107d69ba9ad26054c82e82003c4cc1adc07)

Backport of https://github.com/crowbar/crowbar-core/pull/1142

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade_status.rb
+++ b/crowbar_framework/lib/crowbar/upgrade_status.rb
@@ -144,8 +144,12 @@ module Crowbar
         }
         progress[:steps][current_step][:errors] = errors unless errors.empty?
         if current_step == upgrade_steps_6_7.last && success
-          # Mark the end of the upgrade process
+          # Mark the end of the upgrade process and cleanup the progress
           FileUtils.rm_f running_file
+          progress[:current_substep] = :end_of_upgrade
+          progress[:current_substep_status] = :finished
+          progress[:current_nodes] = {}
+          progress[:current_node_action] = "finished"
         end
         next_step
         save


Itxaka:
Argh, I went to fast, same comment as in https://github.com/crowbar/crowbar-core/pull/1143, this is not needed as a backport I think, as it would be using the crowbar cloud7 version by the time the upgrade is finished.

jsuchome:
@Itxaka you are right it is not strictly needed, but we are trying to keep the library in the same state in cloud6 and cloud7

Itxaka:
> you are right it is not strictly needed, but we are trying to keep the library in the same state in cloud6 and cloud7

Good enough for me. :+1:

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -144,8 +144,12 @@ def end_step(success = true, errors = {})
         }
         progress[:steps][current_step][:errors] = errors unless errors.empty?
         if current_step == upgrade_steps_6_7.last && success
-          # Mark the end of the upgrade process
+          # Mark the end of the upgrade process and cleanup the progress
           FileUtils.rm_f running_file
+          progress[:current_substep] = :end_of_upgrade
+          progress[:current_substep_status] = :finished
+          progress[:current_nodes] = {}
+          progress[:current_node_action] = "finished"
```
rsalevsky: Just a question: Is it intentional to have symbols and strings mixed?

jsuchome: This is correct, yes. We normally have symbols, but `current_node_action` contains really strings (as you could see from its usage from the code ... master, not 3.0)

jsuchome: See e.g. https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/api/upgrade.rb#L1029

rsalevsky: Thanks for the information, makes sense now.

PR #1149, opened 2017-03-10
upgrade: Detect and report that some VM failed to migrate

jsuchome:
Additionally to better checking during the migration itself,
try to offer more information about how to solve the problem.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-evacuate-host.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-evacuate-host.sh.erb
@@ -22,7 +22,7 @@ host=$1
 
 if [ "$host" = "" ]; then
     log "No argument was given! Exiting..."
-    touch $UPGRADEDIR/crowbar-evacuate-host-failed
+    echo 1 > $UPGRADEDIR/crowbar-evacuate-host-failed
     exit 1
 fi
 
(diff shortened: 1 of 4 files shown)


chef/cookbooks/crowbar/templates/default/crowbar-evacuate-host.sh.erb
```diff
@@ -41,17 +41,20 @@ blockmigrate=""
...
 fi
 
-# FIXME: here we need checks for potential errors during the evacuation
-
 while nova list --all-tenants --host "$host" --fields host,status 2>/dev/null | grep -e ACTIVE -e MIGRATING | grep -q "$host"; do
     log "There is still some VM running at $host..."
     sleep 10
 done
 
+if nova list --all-tenants --host "$host" --fields status | grep -q ERROR; then
```
Itxaka: What happens if the customer had an instance in error status from some time ago and never cleaned it up? 
Also, having an instance on error does not necessarily mean that the migration failed. I believe that when migrating an instance, if it fails, it should return to `Active`, or at least my tests turned to that.

Maybe we should track the hosts that were `Active` before the migration and confirm that those were migrated or at least they did not fail?

jsuchome: > What happens if the customer had an instance in error status from some time ago and never 
> cleaned it up?

Good point, but actually I think now user has time to clean it up and restart the step until there is not error instance.

> having an instance on error does not necessarily mean that the migration failed

Hm, true. Any idea for better check? Should we have for "Power State" being migrating? (I'll have to test it again...)

jsuchome: It's `Task State` that is `migrating`. So I think it's gonna stay as `migrating` even when Status changes to `ERROR`.

jsuchome: So what do you think about 
```
nova list --all-tenants --host "$host" --fields status,task_state | grep migrating | grep -q ERROR
```

Itxaka: I just tested and if the host which the instance is moving to suddenly dies, it will keep on migrating both on task_state and status until that node is back. If the destination node is not available then it will return to ACTIVE status on the original node.

So it does not seem like the instance will ever reach ERROR, which kind of makes sense as you dont want an instance which failed to migrate to be on that status, it will fail the migration but not error out.

My original thought was more regarding already existing instances on the host which are on error status and how this could be wrongly reported to the customer. Maybe they tried to launch 15 instances and 4 of them failed, they never cleaned them up.

So with this actual check those instances will be detected and reported as "Error during live-migration" which is not true.

Maybe we should skip this check?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -780,6 +791,11 @@ def raise_node_upgrade_error(message = "")
         raise ::Crowbar::Error::Upgrade::NodeError.new(message)
       end
 
+      def raise_live_migration_error(message = "")
+        Rails.logger.error(message)
+        raise ::Crowbar::Error::Upgrade::LiveMigrationError.new(message)
```
rhafer: Any plans to address this?

jsuchome: No, I would keep our exception classes the way we have them. Cleanup might make sense, but for every one.

chef/cookbooks/crowbar/templates/default/crowbar-evacuate-host.sh.erb
```diff
@@ -39,19 +39,33 @@ blockmigrate="--block-migrate"
 <% else %>
 blockmigrate=""
 <% end %>
+
+$ORIGINAL_INSTANCES=$UPGRADEDIR/original_instances
```
jsuchome: Wrong!

Itxaka: oh wow, small thing, great catch!

crowbar_framework/app/models/api/upgrade.rb
rhafer: Hm, not sure I understand you comment. I don't think it requires changes to the class itself?

jsuchome: Ah, sorry I missed this. I think it would. But maybe I did not pay attention that much

PR #1150, opened 2017-03-10
network: Really run wicked ifreload when ifcfg files change (refix bsc#1011889)

rhafer:
Commit 46998819996 introduced a regression which cause the wicked
ifrelaod to be never actually called. Which reintroduced the original
bug that caused new deployments to loose their network connectivity
after the very first reboot.


nicolasbock:
I second @Itxaka suggestion to add tests for cookbooks!

PR #1152, opened 2017-03-13
backup: Fix upload params

Itxaka:
After crowbar-client >= 3.2.2 the backup params are sent differenty.
We need to account for that and make the params be able to deal with both
ways of sending the params.

To fix the Jenkis CI job https://ci.suse.de/view/Cloud/view/Cloud6/job/cloud-mkcloud6-job-backup-restore-x86_64/

Also to prevent any customers to update crowbar-client on a cloud 6 and fail to be able to do a backup.

It allows for both types of arguments to be send to the backup endpoint. One for crowbar-client >= 3.2.2 and for older versions.

requests?**

### DIFF
--- a/crowbar_framework/app/controllers/backups_controller.rb
+++ b/crowbar_framework/app/controllers/backups_controller.rb
@@ -182,6 +182,6 @@ class BackupsController < ApplicationController
   end
 
   def backup_upload_params
-    params.require(:backup).require(:payload).permit(:file)
+    params.require(:backup).permit(:file)
   end
 end


crowbar_framework/app/controllers/backups_controller.rb
```diff
@@ -182,6 +182,12 @@ def backup_params
   end
 
   def backup_upload_params
-    params.require(:backup).require(:payload).permit(:file)
+    begin
```
Itxaka: implicit over explicit, what a language ¬_¬

Itxaka:
Triggered a cloud6 upgrade job to test that this works.

Maybe we should forwardport this to cloud7? Not sure if there is a requirement of a minimum version of crowbar-client synced to cloud7 so this issue may still happen on cloud7+old crowbar-client version.
cc @vuntz @jsuchome @rsalevsky @dirkmueller

Itxaka:
@rsalevsky thanks, sounds good. chery-pciked + adapted for 3.0

PR #1153, opened 2017-03-13
hound: Switch to global hound configs

rsalevsky:
In the last team meeting we decided to move to global hound configs
which are stored in crowbar/crowbar.


aspiers:
@rsalevsky How does this work?  How does Hound know where to look for the configs?  It would have been helpful to explain that in the commit message.

rsalevsky:
@aspiers 
hound

aspiers:
Ahh I see, thanks.  Yes, that was not obvious at all.  Please include non-obvious stuff in the commit message where possible :)

aspiers:
That's funny, I had totally forgotten about this discussion and was just wondering why the hell there were no hound configs in crowbar-ha any more. I tracked it back to crowbar/crowbar-ha@b1588fd92cf and that did not explain it.  So I had to track that commit back to this PR in order to find the explanation.  I think that shows the importance of having proper commit messages.

We need a document on how to correctly run rubocop locally ...

aspiers:
I found such a document via https://github.com/crowbar/crowbar/pull/2317#issuecomment-292914372!  Thanks a lot to @stefannica for writing https://github.com/crowbar/crowbar/blob/master/doc/barclamp.md#manual-ruby-style-check!  It doesn't work for me, but it's a good start.

PR #1154, opened 2017-03-13
crowbar: Add https support for the crowbar apache server

rsalevsky:
Interesting for https://github.com/sap-oc/crowbar-core/issues/7


rsalevsky:
@cmurphy Yes, definitely but for doing so we have to replace all curl commands in crowbar with crowbarctl commands. Curl is just not smart enough to deal with the different modes (no SSL, SSL insecure, SSL secure).

PR #1155, opened 2017-03-13
upgrade: provide better help for pacemaker issues

Itxaka:
Tries to provide a small help or refer to the HA docs in case of failures
on the cluster resources. Also moves the references of the node names
to the data field and pairs the node name with the error found at it.

In the crowbar backend for the upgrade usually it seems data contain error message and help has additional message. For some cases though, the help part is not helpful at all.

In the "help" part there should probably be something pointing to the possible solution, like reading the troubleshooting section of documentation, or trying the cleanup the resources errors via crm.

Provides a generic+documents to try to help the user fixing the underlying issue.

requests?**

### DIFF
--- a/crowbar_framework/config/locales/crowbar/en.yml
+++ b/crowbar_framework/config/locales/crowbar/en.yml
@@ -797,8 +797,8 @@ en:
           help:
             default: 'Make sure Ceph is healthy'
         clusters_health:
-          crm_failures: '"crm status" failed at nodes %{nodes}.'
-          failed_actions: '"crm status" is showing some failed actions at nodes %{nodes}.'
+          crm_failures: 'Please check that all cluster resources are online or refer to the SUSE High Availability documentation for possible troubleshooting.'
+          failed_actions: 'Please clean the resource history and re−check the current state with "crm_resource -C" or refer to the SUSE High Availability documentation for possible troubleshooting.'
           help:
             default: 'Make sure clusters are healthy'
         no_resources:
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -1464,17 +1464,11 @@ module Api
         failed_actions = check["failed_actions"]
         ret[:clusters_health_crm_failures] = {
           data: crm_failures.values,
-          help: I18n.t(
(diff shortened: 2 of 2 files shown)


Itxaka:
[(2) S56P9 upgrade: Cleanup error messages](https://trello.com/c/k7Kg1D6Q/271-2-s56p9-upgrade-cleanup-error-messages)

jsuchome:
I like the change of the help text. However, are you sure we want to drop that information about the nodes where the failure happened?

jsuchome:
Oh, I see, that's exactly solved in https://github.com/crowbar/crowbar-ha/pull/185! :-)

jsuchome:
OK, still another idea. If you look at other errors, usually we print some kind of error message with short description in data field, and further explanation in help. Looking at these, looks like your added sentences could stand alone as 'help' messages, while the original part (crm has failed...) could be part of data. What do you think?

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -797,8 +797,8 @@ en:
           help:
             default: 'Make sure Ceph is healthy'
         clusters_health:
-          crm_failures: '"crm status" failed at nodes %{nodes}.'
-          failed_actions: '"crm status" is showing some failed actions at nodes %{nodes}.'
+          crm_failures: '"crm status" failed. Please check that all cluster resources are online or refer to the SUSE High Availability documentation for possible troubleshooting.'
+          failed_actions: '"crm status" is showing some failed actions. Please recheck the cluster health with "crm_status -C" or refer to the SUSE High Availability documentation for possible troubleshooting.'
```
skazi0: `crm_status` is not available on my setups and `crm status` has `-C` which sets snmp community, not what you intended I guess.

skazi0: note that I'm checking this on cloud6 side, perhaps the commands are different there.

Itxaka: yeah, need to recheck this as I didnt have the env available when creating the pr, I did eyeball it, so it may be `crm status -D`

Itxaka:
@jsuchome sounds good, I need to test this in a proper env to be able to finalize the proper msg, but moving that part to the `data` or plainly remove it makes sense.

PR #1156, opened 2017-03-14
Cleanup the upgrade code

jsuchome:
Cleanup of some minor stuff, DRY, etc.

Touches these cards 
https://trello.com/c/a630zeyt/347-move-save-substep-inside-the-methods-executing-the-substeps
https://trello.com/c/fGtuEsRe/343-no-comput-enodes-controller-substep-remains-running-atfter-the-whole-upgrade

Not really tested - I think the part with moving those substeps needs some testing to see if I broke something...
UPDATE: Tested manually

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -671,20 +671,35 @@ module Api
         Rails.logger.info("Successfully finished disruptive upgrade of controller nodes")
       end
 
+      def do_controllers_substep(substep)
+        if substep == :ceph_nodes
+          join_ceph_nodes
+          substep = :controller_nodes
+        end
+        if substep == :controller_nodes
+          upgrade_controller_clusters
+          upgrade_non_compute_nodes
+          prepare_all_compute_nodes
+        end
+        ::Crowbar::UpgradeStatus.new.save_substep(substep, :finished)
+      end
+
+      def remaining_nodes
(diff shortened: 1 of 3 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -699,59 +713,32 @@ def nodes(component = "all")
...
-            ::Crowbar::UpgradeStatus.new.end_step
-          end
+        # if all nodes are upgraded, cleanup and end the whole step
+        status = ::Crowbar::UpgradeStatus.new
+        if status.progress[:remaining_nodes].zero?
+          status.save_substep(substep, :finished)
+          finalize_nodes_upgrade
+          status.end_step
         end
+        true
```
Itxaka: Sorry, I cant see where is this used or what is used for :(

jsuchome: Good point, it wasn't there before. I tried to add it to appease my failing testsuite - I still don't see why it is suddenly needed.

I'll need to investigate then...

jsuchome: I was hoping nobody will spot it ;-)

jsuchome: Found it. Testsuite update will follow

Itxaka: > I was hoping nobody will spot it ;-)
:rofl:

PR #1157, opened 2017-03-14
provisioner: fix wrong parameter call

Itxaka:
This call was left without the full namespace, causing the upgrade prepare step to fail.

This seems to block the current CI upgrade jobs from doing the `prepare` step thus making all upgrade jobs unable to continue.

### DIFF
--- a/crowbar_framework/app/models/provisioner_service.rb
+++ b/crowbar_framework/app/models/provisioner_service.rb
@@ -203,7 +203,7 @@ class ProvisionerService < ServiceObject
     all_db.keys.each do |db_name|
       next unless db_name =~ /^repos-.*/
       begin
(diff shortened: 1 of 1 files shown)


Itxaka:
I swear I hit this several times on an upgrade job and changing it fixed it.

jsuchome:
This is weird. Was there some change introduced since @rhafer introduced it? It was working at that time...

jsuchome:
> causing the upgrade prepare step to fail

Are you sure you're fixing it on the correct branch? `prepare_nodes_for_crowbar_upgrade` is called from Cloud6 side

Itxaka:
I think that Im mixing environments here, let me close and recheck

PR #1162, opened 2017-03-15
[stable/3.0] network: Really run wicked ifreload when needed (refix bsc#1011889)

rhafer:
Commit 46998819996 introduced a regression which cause the wicked
ifreload to be never actually called. Which reintroduced the original
bug that caused new deployments to loose their network connectivity
after the very first reboot.


rhafer:
Backport of: https://github.com/crowbar/crowbar-core/pull/1150

PR #1163, opened 2017-03-15
[stable/3.0] Restore ovs bridge settings after reboot (bsc#1029179)

rhafer:
During a clean shutdown wicked will delete any ovs bridges that it feels
responsible for. OTOH, when the node is shutdown uncleanly (e.g. during
a fencing operation) the ovs bridge will not get deleted, which for the
bridges that are touched by the neutron-openvswitch-agent means they
will stay in "secure" mode (without any flows defined) even after
reboot.  This means that the node will be disconnected until
neutron-openvswitch-agent is started. To prevent this we now reset the
bridge's fail-mode before the interface is coming up.


rhafer:
Backport of https://github.com/crowbar/crowbar-core/pull/1161

PR #1165, opened 2017-03-15
upgrade: better error output for wrong step order

Itxaka:
Provide the step that is currently running in the error output
for more clarity

Not necessary but provides a better error message

Provide the step that is currently running in the error output
for more clarity

requests?**

Maybe complementary to https://trello.com/c/OolO2TT3/368-s57p14-bsc-1028950-confusing-message-when-running-crowbarctl-upgrade-repocheck-crowbar

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade_status.rb
+++ b/crowbar_framework/lib/crowbar/upgrade_status.rb
@@ -115,8 +115,8 @@ module Crowbar
           raise Crowbar::Error::StartStepExistenceError.new(step_name)
         end
         if running?
-          @logger.warn("Some step is already running.")
-          raise Crowbar::Error::StartStepRunningError.new
+          @logger.warn("Step #{current_step} is already running.")
+          raise Crowbar::Error::StartStepRunningError.new(current_step)
         end
         unless step_allowed? step_name
           @logger.warn("The start of step #{step_name} is requested in the wrong order")


crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -115,8 +115,8 @@ def start_step(step_name)
           raise Crowbar::Error::StartStepExistenceError.new(step_name)
         end
         if running?
-          @logger.warn("Some step is already running.")
-          raise Crowbar::Error::StartStepRunningError.new
+          @logger.warn("Step #{progress[:current_step]} is already running.")
```
jsuchome: You can directly use `current_step` method here (and bellow).

Itxaka: duuuh :+1:

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -115,8 +115,8 @@ def start_step(step_name)
           raise Crowbar::Error::StartStepExistenceError.new(step_name)
         end
         if running?
-          @logger.warn("Some step is already running.")
-          raise Crowbar::Error::StartStepRunningError.new
+          @logger.warn("Step #{current_step} is already running.")
+          raise Crowbar::Error::StartStepRunningError.new(step_name = current_step)
```
jsuchome: And here i think it's enough to just pass the argument to constructor...

Itxaka: I guess that pythonists (pythoneers?) prefer explicit over implicit :p

skazi0: @Itxaka are you sure this is correct behavior? `running?` without parameters checks *any* step. To get more detailed info you'd need to first call `running?(step_name)`, return with your message and later fallback to the previous code for any step.

PS. if the constructor has one argument, I agree with @jsuchome about simple arg passing.

Itxaka: @skazi0 yes, this is correct. This check is only done when we try to start an step. In case that we try to start an stedp while another one is already on running state we dont care which one is actually running, as there should only be one step running at the time.

skazi0: @Itxaka you're right, I though you were passing `step_name` there not `current_step`.

PR #1166, opened 2017-03-15
upgrade: Fixes for ceph cluster health checks

jsuchome:
(cherry picked from commit 393830d098d33abbc089a9005d21e897a13deb9b)

Backport of https://github.com/crowbar/crowbar-core/pull/1108

We do not actually need ceph health checks on cloud7 side, but we need the check if Ceph (SES) addon is installed, so the repocheck for nodes can correctly activate also SES repositories.


jsuchome:
[S57P9 (3) Address ceph related errors found during Tim's testing](https://trello.com/c/VWhsjVQ0/324-s57p9-3-address-ceph-related-errors-found-during-tim-s-testing)

PR #1167, opened 2017-03-15
Add missing return value

nicolasbock:
The function `Crowbar::ConduitResolver.bus_order` does not return
anything in case the instance variable `@bus_order` is non nil.

### DIFF
--- a/crowbar_framework/lib/crowbar/conduit_resolver.rb
+++ b/crowbar_framework/lib/crowbar/conduit_resolver.rb
@@ -41,9 +41,8 @@ module Crowbar
             break
           end
         end
-
-        result || []
       end
+      @bus_order || []
     end
 
     # Returns an array of interfaces like ["em1", "em2", "em3", "em4"],


nicolasbock:
I am still a little fuzzy on exactly why this wasn't working. I setup a node with 3 interfaces, booting off of the second interface. Using the following `interface_map`:

~~~
{
      "bus_order": [
        "0000:00/0000:00:04",
        "0000:00/0000:00:03",
        "0000:00/0000:00:05"
      ],
      "pattern": "Standard"
}
~~~

I noticed that the nodes could not boot once allocated because the network barclamp insisted on marking the second interface as `deny booting`. The change in this patch fixes it.

nicolasbock:
I am less fuzzy now :smile:. The problem is that `bus_order` didn't return anything in case the instance variable `@bus_order` was non-nil.

nicolasbock:
I updated the title and description of this PR.

vuntz:
I guess I'm missing something obvious: how can `@bus_order` be `nil`? If it's `nil`, then the block is executed and it will be `result || []`, which means it won't be `nil` at the end.

nicolasbock:
@vuntz Thanks for pointing that out, that is a good point. I can't reproduce this problem anymore with the same setup I used before. Once I add the interface map to the network barclamp things work as they are supposed to. I suspect that I didn't enter the map correctly the first time around when I concluded that the whole thing is not working. Oh well...

nicolasbock:
@Itxaka Thanks for the comment. I totally agree that we should have a test for that.

PR #1168, opened 2017-03-15
upgrade: use constants for scripts timeouts

Itxaka:
Have all the timeouts for upgrade related timeouts on the
same location and makes it easier to track and change them at any time

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -19,6 +19,10 @@ require "open3"
 module Api
   class Upgrade < Tableless
     class << self
+      def initialize(attrs = nil)
+        super(attrs)
+        @timeouts = ::Crowbar::UpgradeTimeouts.new
+      end
       def status
         ::Crowbar::UpgradeStatus.new.progress
       end
@@ -435,7 +439,8 @@ module Api
         begin
           unless cinder_node.nil?
             cinder_node.wait_for_script_to_finish(
-              "/usr/sbin/crowbar-delete-cinder-services-before-upgrade.sh", 300
+              "/usr/sbin/crowbar-delete-cinder-services-before-upgrade.sh",
+              @timeouts.values[:delete_cinder_services]
             )
             Rails.logger.info("Deleting of cinder services was successful.")
           end
@@ -1030,7 +1035,8 @@ module Api
       def delete_pacemaker_resources(node)
         save_node_action("deleting old pacemaker resources")
         node.wait_for_script_to_finish(
-          "/usr/sbin/crowbar-delete-pacemaker-resources.sh", 300
+          "/usr/sbin/crowbar-delete-pacemaker-resources.sh",
+          @timeouts.values[:delete_pacemaker_resources]
         )
         Rails.logger.info("Deleting pacemaker resources was successful.")
       rescue StandardError => e
@@ -1060,7 +1066,9 @@ module Api
         args = [hostname]
         args << "delete-ns" if delete_namespaces
         controller.wait_for_script_to_finish(
-          "/usr/sbin/crowbar-router-migration.sh", 600, args
+          "/usr/sbin/crowbar-router-migration.sh",
+          @timeouts.values[:router_migration],
+          args
         )
         Rails.logger.info("Migrating routers away from #{hostname} was successful.")
         network_node.run_ssh_cmd("mkdir -p /var/lib/crowbar/upgrade; touch #{migrated_file}")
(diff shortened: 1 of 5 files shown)


Itxaka:
[(3) S57P13 Make timeouts configurable](https://trello.com/c/ivFOALtv/367-s57p13-make-timeouts-configurable)

Itxaka:
grrrr, stupid rubymine auto indentation

jsuchome:
This is a good step forward.
But do you plan to make it actually configurable by user? Having in the library is nice, but not much different from the end user POV.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -436,7 +436,8 @@ def services
         begin
           unless cinder_node.nil?
             cinder_node.wait_for_script_to_finish(
-              "/usr/sbin/crowbar-delete-cinder-services-before-upgrade.sh", 300
+              "/usr/sbin/crowbar-delete-cinder-services-before-upgrade.sh",
+              ::Crowbar::UpgradeTimeouts.DELETE_CINDER_SERVICES_TIMEOUT
```
skazi0: `.` or `::` ?

crowbar_framework/lib/crowbar/upgrade_timeouts.rb
```diff
@@ -0,0 +1,34 @@
+require "yaml"
+
+module Crowbar
+  class UpgradeTimeouts
+    def values
+      @progress = begin
```
skazi0: why do you call this `@progress`?

Itxaka: I....I dont know. Didnt even notice the name, wth.

Itxaka:
Missing tests for the new stuff!

Itxaka:
Not sure what hound is complaining about, running rubocop locally shows no issues :/

jsuchome:
Do you think we could provide that yaml file with a sample values?

Itxaka:
> Do you think we could provide that yaml file with a sample values?

In any case, that should be done on the package right?

I think that would be ok, but then it turns again into a hassle to track our default timeouts as we need to do a PR against crowbar-core AND a request change to the package to update the default config file, not something that I personally like :/

crowbar_framework/spec/lib/crowbar/upgrade_timeouts_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "spec_helper"
+require "yaml"
+
+describe Crowbar::UpgradeTimeouts do
```
Itxaka: well, duh

jsuchome:
@Itxaka OK, no need for a file, I guess documentation bug report is enough.

crowbar_framework/lib/crowbar/upgrade_timeouts.rb
```diff
@@ -0,0 +1,37 @@
+require "yaml"
+
+module Crowbar
+  class UpgradeTimeouts
+    def values
+      @timeouts_config = begin
+        YAML.load_file("/etc/crowbar/upgrade_timeouts.yml")
```
rhafer: Hm, so this is reloading the yaml file for every ```values``` call. I guess it's not a bit deal as there are not so many calls. But it's not really nice either.

Itxaka: Basically I did it this way so the user does not need to restart crowbar if they add a new value to the config file.

Note that this is done this way because I assume that once rails is started it preloads all this files, so a "static" load_file would only be called once and the user would need to restart the whole server just for this.

If this is not the case, let me know so I can implement it a different way!

rhafer: Hm, you're instantiating the UpgradeTimeouts objects in the ```Api::Node``` and ```Api::Upgrade``` constructors. AFAIU these get recreated with every API request, right? So it should be possible to load that file only once per API request at least.
But I really think that any optimization is currently not needed here.

rhafer: > AFAIU these get recreated with every API request, right?

Hm, actually this seems to be wrong for the Api::Upgrade case. It's all class methods ...

PR #1170, opened 2017-03-16
upgrade: Extend the deployment prechecks

jsuchome:
1. Make really sure controller nodes are HA
2. Make sure compute nodes do not have controller roles.

1+2 are the checks that when failing, still allow "normal" upgrade

3. Make sure nova-compute cannot go before nova-controller

3 is a special case of 2, (as described in https://github.com/crowbar/crowbar-core/pull/1109#issuecomment-285008466), but does not allow any kind of upgrade. So it belongs to extra method with `required = true`

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -227,6 +227,40 @@ module Api
         return { ha_not_installed: true } unless addon_installed? "ha"
         founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
         return { ha_not_configured: true } if founders.empty?
+
+        # Check if roles important for non-disruptive upgrade are deployed in the cluster
+        clustered_roles = [
+          "database-server",
+          "rabbitmq-server",
+          "keystone-server",
+          "glance-server",
+          "cinder-controller",
+          "neutron-server",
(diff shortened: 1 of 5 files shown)


jsuchome:
[S57P5 (3) S56P14 Add the precheck for compute-controller mixed setup](https://trello.com/c/14imamZ1/153-s57p5-3-s56p14-add-the-precheck-for-compute-controller-mixed-setup)

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -291,6 +291,33 @@ def ha_config_check
...
+              b = r.barclamp
+              next if BarclampCatalog.category(b) != "OpenStack"
+              ro = BarclampCatalog.run_order(b)
+              wrong_roles.push role if ro < nova_order
+            end
+            ret = { controller_roles: { node: node.name, roles: wrong_roles } } if wrong_roles.any?
+          end
+        end
+        ret
+      end
```
rhafer: Any chance of a unit test for this function?

jsuchome: it definitely should be there...

jsuchome: And it's there.

PR #1171, opened 2017-03-16
upgrade: Make sure the check ssh output before treating it as string

jsuchome:
Fix for bsc#1029387


jsuchome:
[SOC6>SOC7 upgrade - "crowbarctl upgrade nodes all" - fails with "undefined method `chop' for nil:NilClass"](https://trello.com/c/f6vyk079/374-soc6-soc7-upgrade-crowbarctl-upgrade-nodes-all-fails-with-undefined-method-chop-for-nil-nilclass)

PR #1172, opened 2017-03-17
upgrade: Expose precheck failures to client

skazi0:
When there was some exception during prechecks execution, the info was
stored in the status file but not returned to the client as a response.
Instead a generic 500 response was generated by the framework as a result
of exception propagation.

Now the exceptions are caught in the controller and returned in proper
JSON response.

requests?**
https://trello.com/c/39CCSguR/329-ui-failures-on-crowbar-crash


skazi0:
[UI failures on crowbar crash](https://trello.com/c/39CCSguR/329-ui-failures-on-crowbar-crash)

skazi0:
forward port is not required as Cloud7 had it already implemented in 87f63c05704f2cfa0574510b7b39ba4e1b5ce37d

PR #1175, opened 2017-03-20
upgrade: Fix timeouts initialization

skazi0:
As Api::Upgrade is used only as a singleton, we can't use initialize() to set up the @timeouts variable.

It changes @timeouts to method which will load the timeouts on demand.


skazi0:
An alternative approach was to move @timeouts to class scope and initialize it directly like this:
```
 module Api
   class Upgrade < Tableless
     @timeouts = ::Crowbar::UpgradeTimeouts.new
     class << self
(...)
```
but it would be loaded only once on application start and would require restarting the app after changing timeouts in the config file.

PR #1178, opened 2017-03-20
upgrade: Do not delete remote resources

jsuchome:
It's not necessary to delete remote primitives during the upgrade,
the agents aren't systemd based.  (*EDIT for clarity:* we have to delete primitives for LSB resources to allow upgrading them to `systemd` resources.)


jsuchome:
[(8) S49P16 Orchestrate cloud upgrade with compute-ha setup](https://trello.com/c/2qIRkJMO/395-8-s49p16-orchestrate-cloud-upgrade-with-compute-ha-setup)

PR #1179, opened 2017-03-21
upgrade: Do not allow creating remote related resources before upgrade

jsuchome:
Before compute nodes are upgraded, do not create remote related resources by remote_delegator.

This is an alternative to https://github.com/crowbar/crowbar-openstack/pull/875

Main problem - it takes time before the remote resources are created (after chef-client executed in controller cluster), until then, freshly upgraded compute node does not have nova-compute service running..

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -1158,6 +1158,15 @@ module Api
           return
         end
 
+        compute_nodes.each do |n|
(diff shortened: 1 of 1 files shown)


aspiers:
Can this be closed now that [crowbar/crowbar-openstack#875 (upgrade: Check for remote nodes platform version before creating virtlogd primitive)](https://github.com/crowbar/crowbar-openstack/pull/875) is merged?

jsuchome:
Right

PR #1180, opened 2017-03-22
upgrade: Allow access to docs during upgrade

skazi0:
Documentation served from crowbar should be available during upgrade.


skazi0:
[unblock /docs route during upgrade](https://trello.com/c/JIgqU7yf/380-unblock-docs-route-during-upgrade)

PR #1181, opened 2017-03-22
upgrade: Correctly restart services managed by pacemaker at remote nodes

jsuchome:
We should not manually restart services at remote nodes if these services are managed by pacemaker.

~TODO: after node upgrade, nova-compute is not immediately started (not with chef-client run on the node). It's probably started by chef-client run on cluster node, but I assume we cannot afford that.~

~crm status shows `cl-g-nova-compute` and `remote-d52-54-77-77-01-05` stopped for freshly upgraded node. So it should be possible to just restart these somehow?~

~`crm node ready` might help...~ -> implemented in 4th commit

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -1380,6 +1380,21 @@ module Api
             "Check nova log files at '#{controller.name}' and '#{hostname}'."
           )
         end
+
+        if node[:pacemaker] && node[:pacemaker][:is_remote]
+          out = controller.run_ssh_cmd(
+            "crm --wait node ready remote-#{hostname}",
+            "60s"
+          )
+          unless out[:exit_code].zero?
+            raise_node_upgrade_error(
+              "Starting remote services at '#{hostname}' node has failed. " \
+              "Check /var/log/pacemaker.log at '#{controller.name}' and " \
+              "'#{hostname}' nodes for possible causes."
+            )
+          end
+        end
+
         node_api.save_node_state("compute", "upgraded")
       end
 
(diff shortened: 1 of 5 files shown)


aplanas:
silly question: this is only for compute nodes with pacemaker_remoted service, isn't?

jsuchome:
@aplanas Not a silly question :-) . 
Yes, the code changes are meant for compute nodes managed by pacemaker_remote. I know you've commented on this case earlier, that's why I've asked you for review. But I understand that 2nd commit is really deep in the upgrade code which you might not know.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1140,6 +1140,47 @@ def prepare_all_compute_nodes
         end
       end
 
+      # Restart remote resources at target node from cluster node
+      # This is needed for services (like nova-compute) managed by pacemaker.
+      def restart_remote_resources(controller, node)
+        hostname = node[:hostname]
+        out = controller.run_ssh_cmd(
+          "crm --wait node standby remote-#{hostname}",
```
aplanas: Is there a problem with `crm resource restart`  to restart a service? or for remote this is the only way to do this?

jsuchome: `crm resource restart` was my initial idea. But we would need to know the services, and it's not always the same set (e.g. with openvswitch we have remote services for that as well).

So I've asked yago from HA team and he got an idea that putting remote node into standby node and back restarts all remote services running there.

aplanas: I see, makes sense! Thanks!

jsuchome:
[(8) S49P16 Orchestrate cloud upgrade with compute-ha setup](https://trello.com/c/2qIRkJMO/2060-8-s49p16-orchestrate-cloud-upgrade-with-compute-ha-setup)

PR #1182, opened 2017-03-23
upgrade: Precheck to make sure cinder is using correct backend

jsuchome:
Backend for cinder cannot be raw or local for non-disruptive upgrade.

This reverts commit 26a195da0d22fb99efc127542a2b1e28c8049c72 and brings commit 7281972c3b4a7d7fbc0d5308708a75c7696d124c back.

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -230,6 +230,16 @@ module Api
         founders = NodeObject.find("pacemaker_founder:true AND pacemaker_config_environment:*")
         return { ha_not_configured: true } if founders.empty?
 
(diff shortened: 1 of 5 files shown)


rhafer:
[S57P4 (5) Adapt mkcloud / virtual upgrade jobs to setup and use nfs server for cinder](https://trello.com/c/202p1JSP/360-s57p4-5-adapt-mkcloud-virtual-upgrade-jobs-to-setup-and-use-nfs-server-for-cinder)

jsuchome:
Upgrade gating job passes the precheck, which is what we wanted. It fails on unrelated errors (often with  https://bugzilla.suse.com/show_bug.cgi?id=1030437)

PR #1184, opened 2017-03-23
Upgrade - provide more info, chef recipe fixes...

jsuchome:
Some minor cleanup: 
 - Report more actions during the upgrade
 - Reload the substep that might be changed

### DIFF
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -120,7 +120,9 @@ template "/usr/sbin/crowbar-evacuate-host.sh" do
   group "root"
   action :create
   variables(
-    needs_block_migrate: nova[:nova][:use_shared_instance_storage] ? false : true
+    lazy {
+      { needs_block_migrate: !nova[:nova][:use_shared_instance_storage] }
+    }
   )
   only_if { roles.include? "nova-controller" }
 end
@@ -129,7 +131,7 @@ compute_node = (roles & ["nova-compute-kvm", "nova-compute-xen"]).any?
 cinder_volume = roles.include? "cinder-volume"
 neutron = search(:node, "run_list_map:neutron-server").first
 
-if neutron[:neutron][:networking_plugin] == "ml2"
+if !neutron.nil? && neutron[:neutron][:networking_plugin] == "ml2"
   ml2_mech_drivers = neutron[:neutron][:ml2_mechanism_drivers]
   if ml2_mech_drivers.include?("openvswitch")
     neutron_agent = "openstack-neutron-openvswitch-agent"
@@ -138,7 +140,7 @@ if neutron[:neutron][:networking_plugin] == "ml2"
   end
 end
 
-if neutron[:neutron][:use_dvr]
+if !neutron.nil? && neutron[:neutron][:use_dvr]
   l3_agent = "openstack-neutron-l3-agent"
   metadata_agent = "openstack-neutron-metadata-agent"
 end
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -164,6 +164,7 @@ module Api
     # Disable "pre-upgrade" attribute for given node
     # We must do it from a node where pacemaker is running
     def disable_pre_upgrade_attribute_for(name)
+      save_node_action("disabling pre-upgrade pacemaker attribute")
       hostname = name.split(".").first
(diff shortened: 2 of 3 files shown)


chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -120,7 +120,7 @@
   group "root"
   action :create
   variables(
-    needs_block_migrate: nova[:nova][:use_shared_instance_storage] ? false : true
+    needs_block_migrate: nova.nil? || (nova[:nova][:use_shared_instance_storage] ? false : true)
```
skazi0: why not `nova.nil? || !nova[:nova][:use_shared_instance_storage]` ?

jsuchome: It actually does not matter, because the template is not created because of `only_if { roles.include? "nova-controller" }` clause.

However, if `nova` is nil, the recipe fails during compile phase, which seems to ignore that `only_if` ...

jsuchome: And as for `!nova[:nova][:use_shared_instance_storage]`, seems to make sense, but I'd rather not change the part that is working

Itxaka: I think the failure of the `nova` being nil is because the resource is first compiled and then converged, and in that moment (converge) is when the `only_if` is checked, but the evaluation of `nova` has already been done in the compile phase

You can work around this by using lazy attributes: https://docs.chef.io/resource_common.html#lazy-evaluation


And I agree with @skazi0 here, the `(nova[:nova][:use_shared_instance_storage] ? false : true)` is kind of weird and out of place.

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -129,7 +129,7 @@
 cinder_volume = roles.include? "cinder-volume"
 neutron = search(:node, "run_list_map:neutron-server").first
 
-if neutron[:neutron][:networking_plugin] == "ml2"
+if !neutron.nil? && neutron[:neutron][:networking_plugin] == "ml2"
```
Itxaka: Same as above, if this is causing issues because its executed during the compile phase, maybe we should move it to a function (search for the neutron-server included) and use lazy attribute evaluation so its only executed when its being converged? Maybe a bit too much for this PR but something to consider.

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
jsuchome: `nova[:nova][:use_shared_instance_storage] ? false : true` - I'm not sure why @rhafer put it there, maybe there was some reason I do not see. That's why I prefer not to do it.

As for lazy resources - that seems right, but is it supported in our (quite old) version of chef? I do not see them used anywhere in our recipes.

Itxaka: Stupid chef not having their old documents available ¬_¬

I have used lazy before on chef 10.x so it should work.

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
jsuchome: You are right, but I'd keep this one as it is to avoid larger code changes.

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
jsuchome: I'll adapt it

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
Itxaka: :+1:

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -120,7 +120,9 @@
   group "root"
   action :create
   variables(
-    needs_block_migrate: nova[:nova][:use_shared_instance_storage] ? false : true
+    lazy {
```
Itxaka: shut up hound

jsuchome:
I've tested it once again - the lazy code really gets evaluated only in the next phase i.e. in when `only_if` is also evaluated

PR #1185, opened 2017-03-24
upgrade: Precheck for supported cluster configs

skazi0:
The non-disruptive process supports only limited number of cluster
configurations.

 This precheck is there to verify that the detected configuration matches one of the whitelisted options.

requests?**
https://trello.com/c/of0FMzfd/361-s57p6-5-add-upgrade-precheck-for-unusual-configs-that-might-not-work-for-our-process

zrzut ekranu z 2017-03-24 09-59-04

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -261,14 +261,22 @@ module Api
           "nova"
         ]
         roles_not_ha = []
+        roles_clusters = {}
+        clusters_roles = {}
+        clusters_roles.default = []
         barclamps.each do |barclamp|
           proposal = Proposal.where(barclamp: barclamp).first
           next if proposal.nil?
           proposal["deployment"][barclamp]["elements"].each do |role, elements|
             next unless clustered_roles.include? role
             elements.each do |element|
-              next if ServiceObject.is_cluster?(element)
-              roles_not_ha |= [role]
+              if ServiceObject.is_cluster?(element)
+                # currently roles can't be assigned to more than one cluster
+                roles_clusters[role] = element
+                clusters_roles[element] |= [role]
+              else
+                roles_not_ha |= [role]
+              end
             end
           end
         end
@@ -300,6 +308,48 @@ module Api
             end
           end
         end
+
+        # example inputs:
+        # roles_clusters = {
+        #   "neutron-server": "cluster:cluster1",
+        #   "neutron-network": "cluster:cluster1",
+        #   "database-server": "cluster:cluster2",
+        #   "rabbitmq-server": "cluster:cluster3"
+        # }
+        # clusters_roles = {
+        #   "cluster:cluster1": ["neutron-server", "neutron-network"],
+        #   "cluster:cluster2": ["database-server"],
+        #   "cluster:cluster3": ["rabbitmq-server"]
+        # }
+        deployment_supported =
+          case clusters_roles.length
+          when 0
+            # no clusters, no point complaining as this will be detected by other prechecks
+            true
+
+          when 1
+            # everything on one cluster = no problem
+            true
+
+          when 2
+            # neutron-network in separate cluster
+            true if clusters_roles[roles_clusters["neutron-network"]].length == 1 ||
+                # neutron-network + neutron-server in separate cluster
+                (clusters_roles[roles_clusters["neutron-network"]].length == 2 &&
+                roles_clusters["neutron-network"] == roles_clusters["neutron-server"]) ||
+                # database-server + rabbitmq-server in separate cluster
+                (clusters_roles[roles_clusters["database-server"]].length == 2 &&
+                roles_clusters["database-server"] == roles_clusters["rabbitmq-server"])
+
+          when 3
+            # neutron-network and database-server + rabbitmq-server in separate clusters
+            # rest of *-server roles is implicitly on the third cluster
+            true if clusters_roles[roles_clusters["neutron-network"]].length == 1 &&
+                clusters_roles[roles_clusters["database-server"]].length == 2 &&
+                roles_clusters["database-server"] == roles_clusters["rabbitmq-server"]
+          end
+        ret[:unsupported_cluster_setup] = true unless deployment_supported
+
         ret
       end
 
--- a/crowbar_framework/config/locales/crowbar/en.yml
+++ b/crowbar_framework/config/locales/crowbar/en.yml
@@ -846,6 +846,9 @@ en:
         cinder_wrong_backend:
           error: 'Unsupported cinder backend detected. The Raw Devices and Local File backend cannot be used with non-disruptive upgrade.'
           help: 'For non-disruptive upgrade, cinder cannot use Raw Devices or Local File as a backend. Adapt your Cinder configuration before proceeding with the upgrade.'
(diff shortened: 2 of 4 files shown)


skazi0:
[S57P6 (5) Add upgrade precheck for unusual configs that might not work for our process](https://trello.com/c/of0FMzfd/361-s57p6-5-add-upgrade-precheck-for-unusual-configs-that-might-not-work-for-our-process)

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -290,6 +298,36 @@ def ha_config_check
...
+                roles_clusters["database-server"] == roles_clusters["rabbitmq-server"])
+
+          when 3
+            # neutron-network and database-server + rabbitmq-server in separate clusters
+            # rest of *-server roles is implicitly on the third cluster
+            true if clusters_roles[roles_clusters["neutron-network"]].length == 1 &&
+                clusters_roles[roles_clusters["database-server"]].length == 2 &&
+                roles_clusters["database-server"] == roles_clusters["rabbitmq-server"]
+          end
+        ret[:unsupported_cluster_setup] = true unless deployment_supported
```
jsuchome: Would it be possible to tell user what have we found in the `case` above?

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -290,6 +298,36 @@ def ha_config_check
...
+          when 0
+            # no clusters, no point complaining as this will be detected by other prechecks
+            true
+
+          when 1
+            # everything on one cluster = no problem
+            true
+
+          when 2
+            # neutron-network in separate cluster
```
jsuchome: Could you please write some example values of `clusters_roles` and `roles_clusters` here?

crowbar_framework/app/models/api/crowbar.rb
skazi0: As the `case` is a whitelist I don't see how I can deduce any error information from this.

crowbar_framework/app/models/api/crowbar.rb
skazi0: added example above the `case`

crowbar_framework/app/models/api/crowbar.rb
jsuchome: Maybe just write a different help text for 2 and 3 clusters scenario. I'm not sure it is worth it...

skazi0: I initially had `clusters_roles`, `roles_clusters` and (cluster) `count` in this `ret` but tbh I didn't know what to do with it when it came to the error message. Should I list the whole cluster structure to the user? I think he/she should know this already. Adding cluster count to the error message would make sense maybe for count > 3 but how likely is that?

skazi0: I also tried to make it easy to add more scenarios in the future. Making custom error messages would make this more complicated as the error message might not be relevant for the new scenarios and a message which is too general doesn't bring any benefit.

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -843,6 +843,9 @@ en:
         controller_roles:
           error: 'Found compute node %{node} with controller roles: %{roles}.'
           help: 'It is not possible to upgrade with such setup. These roles cannot be placed on compute node.'
+        unsupported_cluster_setup:
+          error: 'Your cluster/role configuration is not supported by the non-disruptive upgrade process.'
+          help: 'Please refer to the Deployment Guide for list of supported configurations or contact SUSE support.'
```
jsuchome: Please remove that "contact SUSE support" part. I do not believe that we really want that, esp. when the support could only tell user 'Well, your deployment is broken'.

crowbar_framework/app/models/api/crowbar.rb
jsuchome: True, we might want to add more supported setups.

crowbar_framework/config/locales/crowbar/en.yml
skazi0: I thought this was supposed to be the process for adding new supported configurations... wasn't it?

jsuchome: Hm, true, maybe. But I still do not like it in the error message. Maybe someone else should decide ( @vuntz @dirkmueller ?)

crowbar_framework/app/models/api/crowbar.rb
jsuchome: Thanks!

crowbar_framework/spec/models/api/crowbar_spec.rb
```diff
@@ -368,6 +368,156 @@
...
+          )
+        )
+      end
+
+      allow(ServiceObject).to(receive(:is_cluster?).and_return(true))
+
+      expect(subject.class.ha_config_check).to have_key(:unsupported_cluster_setup)
+    end
+
+    it "fails when there are two clusters and roles assignment does not match our patterns" do
```
jsuchome: s/our/supported

crowbar_framework/spec/models/api/crowbar_spec.rb
```diff
@@ -368,6 +368,156 @@
...
+          )
+        )
+      end
+
+      allow(ServiceObject).to(receive(:is_cluster?).and_return(true))
+
+      expect(subject.class.ha_config_check).to eq({})
+    end
+
+    it "fails when there are four clusters" do
```
jsuchome: Please also add a failure case with 3 clusters, thanks.

skazi0: added below

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -290,6 +298,48 @@ def ha_config_check
...
+            true if clusters_roles[roles_clusters["neutron-network"]].length == 1 ||
+                # neutron-network + neutron-server in separate cluster
+                (clusters_roles[roles_clusters["neutron-network"]].length == 2 &&
```
Itxaka: Isnt the check for length a bit superfluous? We only care that network & server are in the same cluster no?

skazi0: My understanding was that the cluster should be dedicated to the network services which means that there can be no other services running there.

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -251,14 +251,22 @@ def ha_config_check
...
         barclamps.each do |barclamp|
           proposal = Proposal.where(barclamp: barclamp).first
           next if proposal.nil?
           proposal["deployment"][barclamp]["elements"].each do |role, elements|
             next unless clustered_roles.include? role
             elements.each do |element|
-              next if ServiceObject.is_cluster?(element)
-              roles_not_ha |= [role]
+              if ServiceObject.is_cluster?(element)
+                # skz: is it possible to have multiple clusters for the same role?
```
rhafer: For the roles listed in ```clustered_roles``` the answer is currently no.

skazi0: updated the comment

PR #1186, opened 2017-03-27
provisioner: disable zypper repo auto-refresh delay

stefannica:
Disable the global zypper repository auto-refresh time
delay by setting repo.refresh.delay to 0 in zypp.conf.

Update the crowbar_register generated script to do the same.

Zypper repos are refreshed on-demand and the updater barclamp
doesn't risk having to work with stale repo contents.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1030462

### DIFF
--- a/chef/cookbooks/provisioner/recipes/base.rb
+++ b/chef/cookbooks/provisioner/recipes/base.rb
@@ -308,6 +308,13 @@ if node[:platform_family] == "suse" && !node.roles.include?("provisioner-server"
   end
 
   if node[:platform_family] == "suse"
+    template "/etc/zypp.conf" do
+      source "zypp.conf.erb"
+      owner "root"
+      group "root"
+      mode "0644"
+    end
+
     # make sure the repos are properly setup
     repos = Provisioner::Repositories.get_repos(node[:platform], node[:platform_version], node[:kernel][:machine])
     for name, attrs in repos
--- a/chef/cookbooks/provisioner/templates/default/zypp.conf.erb
+++ b/chef/cookbooks/provisioner/templates/default/zypp.conf.erb
@@ -1,0 +1,1 @@
+repo.refresh.delay = 0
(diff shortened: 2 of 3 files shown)


stefannica:
[bsc#1030462 - L3: updater barclamp should invoke zypper refresh before applying updates](https://trello.com/c/g31Rwdpr/497-bsc-1030462-l3-updater-barclamp-should-invoke-zypper-refresh-before-applying-updates)

vuntz:
I really wouldn't do that: chef calls zypper many times, and it means that there'd be quite some additional slowness for each chef run since autorefresh would happen so many times (even if it's just about zypper checking that there's no change in the repo files).

stefannica:
Not sure this is a good approach, for the following reasons:
  - there is no per-repository delay setting
  - the remote repo metadata info will be accessed for a repository every time the package manager tool is used and needs to access the repository contents. Although the overhead associated with a single check is low (only the index metadata file is downloaded), these small overheads may sum up if a big list of repositories is used, especially over a high-latency network

stefannica:
Thank you @vuntz, you just confirmed my concerns then. 
Which of the following do you think we should do about this bug then:
  - use a value lower than the default 10 min
  - refresh the PTF repo in the updater barclamp instead
  - document the time limitation and don't touch the code

stefannica:
What do you think, @dirkmueller, should we revert to the initial solution of refreshing the PTF repo in the updater barclamp ? I know you had some reservations about that.

dirkmueller:
is there a way to change refresh delay for just that one repo? I didn't check the code, just asking

stefannica:
@dirkmueller, unfortunately not, zypper only has a global setting for that

stefannica:
Reverting to forcefully refresh the PTF repo during the execution of the updater barclamp. Continued in #1200

PR #1188, opened 2017-03-28
Extend the deployment prechecks

jsuchome:
Forward port of https://github.com/crowbar/crowbar-core/pull/1170


Itxaka:
Is the last commit a cherry-pick as well? It doesnt have a link to the cherry picked commit

jsuchome:
@Itxaka actually it is not... the problem is that I've ported differnt cinder related commit before so I've just added the remaining part as separate one

PR #1189, opened 2017-03-28
upgrade: Remove checking for XEN nodes

jsuchome:
1. Precheck for presence of non-KVM nodes
2. Remove rest of XEN specific searches
3. Deployment check needs to check all compute nodes (relevant for disruptive upgrade)

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -211,14 +211,15 @@ module Api
 
       def compute_status
         ret = {}
-        ["kvm", "xen"].each do |virt|
-          compute_nodes = NodeObject.find("roles:nova-compute-#{virt}")
-          next unless compute_nodes.size == 1
-          ret[:no_resources] ||= []
-          ret[:no_resources].push(
-            "Found only one compute node of #{virt} type; non-disruptive upgrade is not possible"
-          )
+        compute_nodes = NodeObject.find("roles:nova-compute-kvm")
+        if compute_nodes.size == 1
+          ret[:no_resources] =
+            "Found only one KVM compute node; non-disruptive upgrade is not possible"
         end
+        non_kvm_nodes = NodeObject.find(
+          "roles:nova-compute-* AND NOT roles:nova-compute-kvm"
+        ).map(&:name)
+        ret[:non_kvm_computes] = non_kvm_nodes unless non_kvm_nodes.empty?
         nova = NodeObject.find("roles:nova-controller").first
         ret[:no_live_migration] = true if nova && !nova["nova"]["use_migration"]
         ret
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -300,13 +300,11 @@ module Api
           "trove-server"
         ]
         ret = {}
-        ["kvm", "xen"].each do |virt|
-          NodeObject.find("roles:nova-compute-#{virt}").each do |node|
-            conflict = node.roles & conflicting_roles
-            unless conflict.empty?
-              ret[:role_conflicts] ||= {}
-              ret[:role_conflicts][node.name] = conflict
-            end
+        NodeObject.find("roles:nova-compute-kvm").each do |node|
+          conflict = node.roles & conflicting_roles
+          unless conflict.empty?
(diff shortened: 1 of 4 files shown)


jsuchome:
[(3) S58P4 bsc#1029847: Handle non-kvm compute nodes](https://trello.com/c/1Kb1AFzw/376-3-s58p4-bsc-1029847-handle-non-kvm-compute-nodes)

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -823,10 +823,13 @@ en:
           help:
             default: 'Make sure clusters are healthy'
         no_resources:
-          help: 'Make sure you have enough compute nodes so the live migration of instances is possible.'
+          help: 'Make sure you have enough KVM compute nodes so the live migration of instances is possible.'
         no_live_migration:
           error: 'Live migration is not configured in nova barclamp.'
           help: 'Adapt the nova barclamp configuration to support live migration before proceeding with the upgrade.'
+        non_kvm_computes:
+          error: 'Found compute nodes differnt then KVM: %{nodes}.'
```
Itxaka: typo(s)
differnt -> different
then -> than

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -307,26 +307,24 @@ def deployment_check
...
-              # compute node roles are fine
-              next if role.start_with?("nova-compute") || role == "pacemaker-remote"
-              r = RoleObject.find_role_by_name(role)
-              next if r.proposal?
-              b = r.barclamp
-              next if BarclampCatalog.category(b) != "OpenStack"
-              wrong_roles.push role if BarclampCatalog.run_order(b) < nova_order
-            end
-            ret = { controller_roles: { node: node.name, roles: wrong_roles } } if wrong_roles.any?
+        NodeObject.find("roles:nova-compute-*").each do |node|
```
Itxaka: we dont care about kvm/xen differences here?

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -210,18 +210,20 @@ def openstack_check
...
+          }
         end
+        non_kvm_nodes = NodeObject.find(
+          "roles:nova-compute-* AND NOT roles:nova-compute-kvm"
+        ).map(&:name)
+        return { non_kvm_computes: non_kvm_nodes } unless non_kvm_nodes.empty?
         nova = NodeObject.find("roles:nova-controller").first
-        ret[:no_live_migration] = true if nova && !nova["nova"]["use_migration"]
-        ret
+        return { no_live_migration: true } if nova && !nova["nova"]["use_migration"]
```
Itxaka: If there is non kvm nodes in the setup, this code will never be reached, and we still want to report if live_migration is enabled or not right? Or Im am reading this incorrectly?

jsuchome: It does not matter which error we will show for such case. If there is non kvm node, non-disruptive upgrade is not allowed, so there will be no live migration

crowbar_framework/app/models/api/crowbar.rb
jsuchome: I think that this check is for making sure that at least **disruptive** upgrade can work - during which we have an algorithm to pick the nodes order. During disruptive (normal) upgrade mode, we have to upgrade non-kvm nodes as well, but we still need that right order that would be broken if `nova-compute` would go before `nova-controller`.

So this change is actually fixing bug that was in the code already before.

crowbar_framework/config/locales/crowbar/en.yml
jsuchome: fixed

crowbar_framework/app/models/api/crowbar.rb
skazi0: I'd rather show more problems to the user from the beginning. Imagine a situation when there are e.g. 5 problem and user decides to fix them (e.g. add spare compute node for the time of upgrade). With our approach of showing only first error, this can lead to situation where the 5th is not fixable for the user (for whatever reason) and the user will find out about this only after fixing the first 4. Smells like a bad UX.

jsuchome: Well, I can change it. But we do not do it from other methods

PR #1191, opened 2017-03-29
upgrade: Do not treat xen nodes upgradable in non-disruptive mode

jsuchome:
If they are any XEN compute nodes, the would be upgraded in normal
mode only.

Other 2 commits are port of https://github.com/crowbar/crowbar-core/pull/1189


jsuchome:
[(3) S58P4 bsc#1029847: Handle non-kvm compute nodes](https://trello.com/c/1Kb1AFzw/376-3-s58p4-bsc-1029847-handle-non-kvm-compute-nodes)

PR #1193, opened 2017-03-29
upgrade: Update openstack commands for insecure SSL setup

jsuchome:
https://bugzilla.suse.com/show_bug.cgi?id=1031499

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -398,7 +398,7 @@ module Api
           return
         end
 
-        cmd = "openstack server list --all-projects --long " \
+        cmd = "openstack server list --insecure --all-projects --long " \
           "--status active -f value -c Host"
         out = controller.run_ssh_cmd("source /root/.openrc; #{cmd}", "60s")
         unless out[:exit_code].zero?
@@ -1322,13 +1322,12 @@ module Api
           node_api.save_node_state("compute", "upgraded")
           return
         end
-
         controller.run_ssh_cmd(
-          "source /root/.openrc; nova service-enable #{hostname} nova-compute"
+          "source /root/.openrc; nova --insecure service-enable #{hostname} nova-compute"
         )
         out = controller.run_ssh_cmd(
           "source /root/.openrc; " \
-          "nova service-list --host #{hostname} --binary nova-compute " \
+          "nova --insecure service-list --host #{hostname} --binary nova-compute " \
           "| grep -q enabled",
           "60s"
         )
--- a/chef/cookbooks/crowbar/templates/default/crowbar-evacuate-host.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-evacuate-host.sh.erb
@@ -32,7 +32,7 @@ set +x
 source /root/.openrc
 set -x
 
-nova service-disable "$host" nova-compute
+nova --insecure service-disable "$host" nova-compute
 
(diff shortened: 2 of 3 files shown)


jsuchome:
[Fix our CLI tools for insecure SSL setup](https://trello.com/c/HcPgs8sk/397-fix-our-cli-tools-for-insecure-ssl-setup)

jsuchome:
There are still some parts in crowbar.rb that need to be fixed...

Itxaka:
woops, did not see the wip label.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -398,7 +398,14 @@ def check_for_running_instances
           return
         end
 
-        cmd = "openstack server list --all-projects --long " \
+        insecure = ""
+        keystone = ::Node.find("run_list_map:keystone-server").first
+        if keystone[:keystone][:ssl][:insecure] ||
+            (controller[:nova][:ssl][:enabled] && controller[:nova][:ssl][:insecure])
+          insecure = "--insecure"
+        end
```
jsuchome: This block looks weird to me, Don't we have something better in crowbar app to realize we have insecure setup?
Or should I rather check proposals then nodes? @rhafer , @vuntz ?

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -113,6 +113,12 @@
 
 nova = search(:node, "run_list_map:nova-controller").first
 
+unless nova.nil?
+  keystone_settings = KeystoneHelper.keystone_settings(nova, "nova")
```
rhafer: Hm, AFAIK this does a chef search inside to find the keystone node (```roles:keystone-server AND keystone_config_environment:keystone-config-#{instance}")```, this doesn't work in the current stated of upgrade, does it? (If that chef-search does yield anything it will try to fall back to the node it got passed, ```nova``` in your case, which might be wrong

crowbar_framework/app/models/api/upgrade.rb
rhafer: I think ```if keystone[:keystone][:ssl][:insecure] || controller[:nova][:ssl][:insecure]``` should be enough at least that's what we do in the newer recipes.

But checking proposals will not make your life easier I think. They don't carry more information than what's available on the nodes.

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -113,6 +113,12 @@
...
+  nova_insecure = node[:nova][:ssl][:enabled] && node[:nova][:ssl][:insecure]
+  nova_insecure = keystone_settings["insecure"] || nova_insecure
```
rhafer: I think ```nova_insecure = keystone_settings["insecure"] || node[:nova][:ssl][:insecure]``` would be sufficient here. (Well that's ignoring my other comment about keystone_settings being wrong)

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
jsuchome: So I could do explicit search for a keystone node and check `keystone[:keystone][:ssl][:insecure]`, is that right?

skazi0:
Could you update the commit message to be more explicit on why we added --insecure unconditionally.

PR #1195, opened 2017-03-30
upgrade: Do not try to disable non-existing service

jsuchome:
At SP2, systemctl disable complains when disabling service that is not installed.
This is normally not a problem, since we execute this code at SP1 (Cloud6)
nodes only, but it can fail in the corner case when user has Ceph nodes, upgrades them
(so they have SP2 and are staying in the crowbar-upgrade state) and the reapplies
some barclamp (e.g. for fixing some setting required for the Cloud upgrade).

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -48,9 +48,7 @@ when "crowbar_upgrade"
 
   bash "disable_openstack_services" do
     code <<-EOF
-      for i in $(systemctl list-units openstack* --no-legend | cut -d" " -f1) \
-               drbd.service \
-               pacemaker.service;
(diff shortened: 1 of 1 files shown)


jsuchome:
> Or are we talking about barclamps from crowbar-core?

It's actually just the crowbar barclamp and this recipe I'm adapting.

> Isn't it somewhat dangerous to reapply any of those while there are nodes in the crowbar_upgrade 
> state?

Yes, it should not be recommened. But as I wrote, for corner case with SES already upgraded and only ceph nodes in those state, it is possible user wants to do it. I've tested it (just because I did the same mistakte for not having correct nova config before upgrading SES), and this small change enabled it to me.

rhafer:
> I've tested it (just because I did the same mistakte for not having correct nova config before upgrading SES), and this small change enabled it to me.

Maybe I am misunderstanding you, but why would a change to the nova configuration (i.e. editing a setting in the nova barclamp and applying that) trigger a chef-client run on the ceph nodes?

PR #1196, opened 2017-03-30
Fix schema migration for remote PSQL

skazi0:
Schema migration code checked for local postgresql service even if
the actual config used remote database.

Simple check was added for the database host configured in the application.


skazi0:
[Various problems with database connection during upgrade](https://trello.com/c/XrkeGaOC/398-various-problems-with-database-connection-during-upgrade)

PR #1197, opened 2017-03-30
Loosen crowbar/psql service dependencies

skazi0:
With remote postgresql service, it's not possible for crowbar(-jobs) to `Require=` postgresql.

Use `Wants=` which is a less strict alternative.

requests?**
See systemd docs for more details: https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Wants=


skazi0:
[Various problems with database connection during upgrade](https://trello.com/c/XrkeGaOC/398-various-problems-with-database-connection-during-upgrade)

PR #1200, opened 2017-04-03
updater: refresh PTF repository (bsc#1030462) 

stefannica:
Force a refresh for the PTF repository in the updater barclamp
before updating, to make sure that PTF packages are picked up
regardless of the minimum repository auto-refresh delay.

Fixes [bsc#1030462](https://bugzilla.suse.com/show_bug.cgi?id=1030462)

### DIFF
--- a/chef/cookbooks/updater/recipes/default.rb
+++ b/chef/cookbooks/updater/recipes/default.rb
@@ -40,6 +40,11 @@ if !node[:updater].key?(:one_shot_run) || !node[:updater][:one_shot_run]
       zypper_command += " --auto-agree-with-licenses"
     end
 
+    execute "refresh PTF repository" do
+      command "zypper --non-interactive --gpg-auto-import-keys refresh -fr PTF"
+      ignore_failure true
+    end
+
     # Butt-ugly, enhance Chef::Provider::Package::Zypper later on...
     ruby_block "running \"#{zypper_command}\"" do
       block do


stefannica:
[bsc#1030462 - L3: updater barclamp should invoke zypper refresh before applying updates](https://trello.com/c/g31Rwdpr/497-bsc-1030462-l3-updater-barclamp-should-invoke-zypper-refresh-before-applying-updates)

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -40,6 +40,10 @@
       zypper_command += " --auto-agree-with-licenses"
     end
 
+    execute "refresh PTF repository" do
+      command "zypper --non-interactive --gpg-auto-import-keys refresh -fr PTF"
+    end
```
Itxaka: This should have a `only_if` guard to protect in case there is no PTF repo on the node.

stefannica: You're right, I missed that. Done.

dirkmueller:
please append " (bsc#1030462)" at the end of the git summary so that the bug reference gets picked up automatically

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -23,6 +23,11 @@
   node.save
 
   if node[:platform_family] == "suse"
+
+    repos = Provisioner::Repositories.get_repos(node[:platform],
```
dirkmueller: hmm, this gets evaluated in 1st phase already.. are you sure that's safe to do? I'm unsure here.

chef/cookbooks/updater/recipes/default.rb
```diff
@@ -40,6 +45,11 @@
       zypper_command += " --auto-agree-with-licenses"
     end
 
+    execute "refresh PTF repository" do
+      command "zypper --non-interactive --gpg-auto-import-keys refresh -fr PTF"
+      only_if { repos.keys.include? "PTF" }
```
dirkmueller: how likely is it that zypper exits with a non-zero exit code? this is not handled here, but should probably be ignored. also array style command execution might be preferred, although thats just a minor nit.

chef/cookbooks/updater/recipes/default.rb
stefannica: not sure I understand the question. What's a 1st phase ? Anyway, I see the prepare-upgrade-scripts.rb script uses the same function call, so I think it should be safe for a mere update.

chef/cookbooks/updater/recipes/default.rb
stefannica: I ignored the return code, as suggested

chef/cookbooks/updater/recipes/default.rb
dirkmueller: https://coderanger.net/two-pass/

chef/cookbooks/updater/recipes/default.rb
dirkmueller: thanks. we could as well completely remove the only_if {} part then..

chef/cookbooks/updater/recipes/default.rb
stefannica: thanks for the link, it just gave me a big "aha !" moment

chef/cookbooks/updater/recipes/default.rb
stefannica: great idea

PR #1202, opened 2017-04-03
Unclaimed disks: improve handling of multipath devices (bsc#1031065)

bhdn:
dm_multipath devices were being skipped while its paths were being included
in the array of unclaimed devices, allowing the DRBD recipe to claim them
and later fail while running pvcreate.

Introduce two new methods in Disk for checking whether the device (or its
holder) is named as "mpath-*".

### DIFF
--- a/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+++ b/chef/cookbooks/barclamp/libraries/barclamp_library.rb
@@ -150,7 +150,11 @@ module BarclampLibrary
               %x{lsblk #{d.name.gsub(/!/, "/")} --noheadings --output MOUNTPOINT | grep -q -v ^$}
               next if $?.exitstatus == 0
             end
-            d.fixed and not d.claimed?
+            # skip claimed disks and multipath devices held by holders
+            # but include both fixed disks and multipath devices
+            device_type_claimable = d.fixed || d.multipath?
+            in_use = d.claimed? || d.held_by_multipath?
+            device_type_claimable && !in_use
           end
         end
 
@@ -158,6 +162,12 @@ module BarclampLibrary
           all(node).select do |d|
             d.claimed? and d.owner == owner
           end
+        end
+
+        def self.multipath?(device)
+          uuid_path = "/sys/block/#{device}/dm/uuid"
+          return false unless File.exist?(uuid_path)
+          File.open(uuid_path) { |f| f.read(7).start_with?("mpath-") }
         end
 
         # can be /dev/hda, /dev/sda or /dev/cciss/c0d0
@@ -201,6 +211,35 @@ module BarclampLibrary
         def usage
           Chef::Log.error("Usage method for disks is deprecated!  Please update your code to use owner")
           self.owner
+        end
+
+        def multipath?
+          self.class.multipath?(@device)
+        end
+
+        def held_by_multipath?
+          # We need to check if the holders of a device (if it has any)
+          # are multipath-capable, for example:
+          #
+          # root@d52-54-77-77-01-01:~ # multipath -ll
+          # 0QEMU_QEMU_HARDDISK_00002 dm-1 QEMU,QEMU HARDDISK
+          # size=10G features='0' hwhandler='0' wp=rw
+          # -+- policy='service-time 0' prio=1 status=active
+          # - 0:0:0:2 sdc 8:32   active ready running
+          # -+- policy='service-time 0' prio=1 status=enabled
+          # - 0:0:0:3 sdb 8:16   active ready running
+          #
+          # sdb and sdc are paths of dm-1:
+          # root@d52-54-77-77-01-01:~ # ls /sys/block/sdb/holders/
+          # dm-1
+          # root@d52-54-77-77-01-01:~ # ls /sys/block/sdc/holders/
+          # dm-1
+          #
+          # in this case this method should return false for sdb and sdc as we dont want
+          # those disks to appear available, instead we want dm-1 to be made available
+          ::Dir.entries("/sys/block/#{@device}/holders").any? do |holder|
+            self.class.multipath?(holder)
+          end
         end
 
         def fixed
--- a/chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
+++ b/chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
@@ -1,0 +1,64 @@
+require "spec_helper"
+
+require_relative "../../libraries/barclamp_library"
+
+describe BarclampLibrary::Barclamp::Inventory::Disk do
+
+  before(:each) do
+    @chef_run = ::ChefSpec::Runner.new
+    @node = @chef_run.node
+    @node.default[:block_device] = {
+      dm0: { removable: "0" },
+      xvd1: { removable: "0" },
+      xvd2: { removable: "1" }
+    }
+  end
+
+  specify "#unclaimed returns the proper number of unclaimed devices" do
+    a = BarclampLibrary::Barclamp::Inventory::Disk
+    expect(a).to receive(:`).exactly(5).times.and_return(`exit 1`)
+    expect(::File).to receive(:exist?).with("/sys/block/dm0/dm/uuid").and_return(true)
+    expect(::File).to receive(:exist?).with("/sys/block/xvd2/dm/uuid").and_return(false)
+    expect(::File).to receive(:exist?).with("/sys/block/sr0/dm/uuid").and_return(false)
+    expect(::File).to receive(:open).exactly(1).times.with(
+      "/sys/block/dm0/dm/uuid"
+    ).and_yield(StringIO.new("mpath-test"))
+    # return holders
+    expect(::Dir).to receive(:entries).exactly(5).times.and_return([])
+    expect(a.unclaimed(@node).length).to eq(3)
+  end
+
+  describe "multipath features" do
+    specify "#multipath? fails with no device given" do
+      expect { BarclampLibrary::Barclamp::Inventory::Disk.multipath? }.to raise_error(ArgumentError)
+    end
+
+    specify "#multipath returns true if it find the mpath device uuid" do
+      expect(::File).to receive(:exist?).exactly(1).times.and_return(true)
+      expect(::File).to receive(:open).exactly(1).times.and_yield(StringIO.new("mpath-"))
+      expect(
+        BarclampLibrary::Barclamp::Inventory::Disk.multipath?("test")
+      ).to be(true)
+    end
+
+    specify "#multipath returns false if it doesnt find the mpath device uuid" do
+      expect(::File).to receive(:exist?).exactly(1).times.and_return(true)
+      expect(::File).to receive(:open).exactly(1).times.and_yield(StringIO.new("uuid"))
+      expect(
+        BarclampLibrary::Barclamp::Inventory::Disk.multipath?("test")
+      ).to be(false)
+    end
+
+    specify "#held_by_multipath? calls multipath? on each holder" do
+      expect(::Dir).to receive(:entries).exactly(1).times.and_return(["subtest1", "subtest2"])
+      expect(
+        BarclampLibrary::Barclamp::Inventory::Disk
+      ).to receive(:multipath?).exactly(1).times.with("subtest1")
+      expect(
+        BarclampLibrary::Barclamp::Inventory::Disk
+      ).to receive(:multipath?).exactly(1).times.with("subtest2")
+      a = BarclampLibrary::Barclamp::Inventory::Disk.new(@node, "test")
+      expect(a.held_by_multipath?).to be(false)
+    end
+  end
+end
--- a/chef/cookbooks/barclamp/spec/spec_helper.rb
+++ b/chef/cookbooks/barclamp/spec/spec_helper.rb
@@ -1,0 +1,50 @@
+require "chefspec"
+
+ENV["RSPEC_RUNNING"] = "true"
+
+RSpec.configure do |config|
+  # config.mock_with :rspec do |mocks|
+  #   # This option should be set when all dependencies are being loaded
+  #   # before a spec run, as is the case in a typical spec helper. It will
+  #   # cause any verifying double instantiation for a class that does not
+  #   # exist to raise, protecting against incorrectly spelt names.
+  #   mocks.verify_doubled_constant_names = true
+  # end
+
+  # Specify the path for Chef Solo to find cookbooks (default: [inferred from
+  # the location of the calling spec file])
+  this_dir = File.dirname(__FILE__)
+  config.cookbook_path = [
+    File.expand_path("../..", this_dir),
+    File.expand_path("fixtures/cookbooks", this_dir),
+  ]
+
+  # Specify the path for Chef Solo to find roles (default: [ascending search])
+  # config.role_path = '/var/roles'
+
+  # Specify the Chef log_level (default: :warn)
+  config.log_level = ENV["CHEF_LOG_LEVEL"].to_sym if ENV["CHEF_LOG_LEVEL"]
+
+  # Specify the path to a local JSON file with Ohai data (default: nil)
+  # config.path = 'ohai.json'
+
+  # Specify the operating platform to mock Ohai data from (default: nil)
+  config.platform = "suse"
+
+  # Specify the operating version to mock Ohai data from (default: nil)
+  config.version = "12.2"
+
+  # Disable deprecated "should" syntax
+  # https://github.com/rspec/rspec-expectations/blob/master/Should.md
+  config.expect_with :rspec do |c|
+    c.syntax = :expect
+  end
+
+  config.run_all_when_everything_filtered = true
+  config.filter_run focus: true
+end
+
+if ENV["RUBYDEPS"]
+  require "rubydeps"
+  Rubydeps.start
+end
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,26 @@
 language: ruby
 sudo: false
 
-rvm:
-  - 2.1.0
+rvm: 2.1.0
 
-env:
-  - SKIP_CHECKS=yes
+env: SKIP_CHECKS=yes
+
+matrix:
+  include:
+    - gemfile: crowbar_framework/Gemfile
+      script:
+       - cd crowbar_framework
+       - bin/bundle install
+       - bin/rake db:create db:migrate
+       - bundle exec rake spec brakeman:run
+       # ignore rest-client issues, chef 10 requires that
+       - bin/bundle exec bundle-audit update
+       - bin/bundle exec bundle-audit check --ignore CVE-2015-1820 OSVDB-117461
+    - gemfile: chef/cookbooks/barclamp/Gemfile
+      script:
+       - cd chef/cookbooks/barclamp && bundle exec rake
 
 addons:
   apt:
     packages:
       - libarchive-dev
-
-gemfile: crowbar_framework/Gemfile
-
-install:
-  - cd crowbar_framework
-  - bin/bundle install
-  - bin/rake db:create db:migrate
-
-script:
-  - bin/rake spec brakeman:run
-
-  # ignore rest-client issues, chef 10 requires that
-  - bin/bundle exec bundle-audit update
-  - bin/bundle exec bundle-audit check --ignore CVE-2015-1820 OSVDB-117461
--- a/chef/cookbooks/barclamp/.rspec
+++ b/chef/cookbooks/barclamp/.rspec
@@ -1,0 +1,2 @@
+--colour
+--format documentation
--- a/chef/cookbooks/barclamp/Gemfile
+++ b/chef/cookbooks/barclamp/Gemfile
@@ -1,0 +1,9 @@
+source "https://rubygems.org"
+
+gem "rack", "< 2.0.0"
+gem "rake"
+
+group :test, :development do
+  gem "chefspec", "~> 3.0"
+  gem "rspec-expectations", "~> 2.14.0"
+end
--- a/chef/cookbooks/barclamp/Rakefile
+++ b/chef/cookbooks/barclamp/Rakefile
@@ -1,0 +1,6 @@
+task default: "spec"
+
+task :spec do
+  sh "rspec"
+end
+


s-t-e-v-e-n-k:
My only concern with this is there are no tests.

aspiers:
I agree, there really should be tests for this.  @bhdn perhaps you could use https://github.com/crowbar/crowbar-ha/tree/master/chef/cookbooks/pacemaker/spec as a model for how to implement tests for this cookbook library?

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -203,6 +204,17 @@ def usage
           self.owner
         end
 
+        def multipath?(device = @device)
+          uuid_path = "/sys/block/#{device}/dm/uuid"
+          File.exist?(uuid_path) && File.open(uuid_path) { |f| f.read(7).start_with?("mpath-") }
+        end
+
+        def multipath_path?
```
aspiers: This is a little hard to understand at first.  Could you add a comment giving an example of a multipath path and its corresponding holder, to clarify that the holder is multipath?

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -150,7 +150,8 @@ def self.unclaimed(node, include_mounted=false)
               %x{lsblk #{d.name.gsub(/!/, "/")} --noheadings --output MOUNTPOINT | grep -q -v ^$}
               next if $?.exitstatus == 0
             end
-            d.fixed and not d.claimed?
+            next unless d.fixed || d.multipath?
```
aspiers: The logic of these two lines looks correct, but it was quite hard for me to follow.  Please could you add a comment or two?

chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    expect(a).to receive(:`).exactly(4).times.and_return(`exit 1`)
+    # return 3 multipath devices and 1 non-multipath
+    expect(::File).to receive(:exist?).exactly(3).times.and_return(true)
+    expect(::File).to receive(:exist?).exactly(1).times.and_return(false)
+    expect(::File).to receive(:open).exactly(3).times.and_return("mpath-test")
+    expect(::Dir).to receive(:entries).exactly(3).times.and_return([])
+    expect(a.unclaimed(@node).length).to eq(3)
+  end
+end
+
```
Itxaka: :rage4:

chef/cookbooks/barclamp/libraries/barclamp_library.rb
bhdn: As requested by Itxaka, who is now working on it. Given:
```
root@d52-54-77-77-01-01:~ # multipath -ll
0QEMU_QEMU_HARDDISK_00002 dm-1 QEMU,QEMU HARDDISK
size=10G features='0' hwhandler='0' wp=rw
|-+- policy='service-time 0' prio=1 status=active
| `- 0:0:0:2 sdc 8:32   active ready running
`-+- policy='service-time 0' prio=1 status=enabled
  `- 0:0:0:3 sdb 8:16   active ready running
0QEMU_QEMU_HARDDISK_00001 dm-0 QEMU,QEMU HARDDISK
size=10G features='0' hwhandler='0' wp=rw
|-+- policy='service-time 0' prio=1 status=active
| `- 0:0:0:0 sda 8:0    active ready running
`-+- policy='service-time 0' prio=1 status=enabled
  `- 0:0:0:1 sdd 8:48   active ready running
```
`sdb` and `sdc` are paths of `dm-1`.

```
root@d52-54-77-77-01-01:~ # ls /sys/block/sdb/holders/                                                                                                                                  
dm-1
root@d52-54-77-77-01-01:~ # ls /sys/block/sda/holders/
dm-0
root@d52-54-77-77-01-01:~ # ls /sys/block/sdb/holders/
dm-1
root@d52-54-77-77-01-01:~ # ls /sys/block/sdc/holders/
dm-1
root@d52-54-77-77-01-01:~ # ls /sys/block/sdd/holders/
dm-0
```

Itxaka:
@aspiers @s-t-e-v-e-n-k re-review?

aspiers:
@Itxaka I already reviewed - AFAICS my comments are not addressed yet. Please let me know if I missed something.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -203,6 +204,17 @@ def usage
           self.owner
         end
 
+        def multipath?(device = @device)
```
Itxaka: done

chef/cookbooks/barclamp/libraries/barclamp_library.rb
Itxaka: Added as comment

chef/cookbooks/barclamp/libraries/barclamp_library.rb
Itxaka: Not sure what do you want me to explain, I think this is pretty straightforward? It just adds the checks for multipath to the existing checks, as the commit message explains

Itxaka:
@aspiers sorry for that, I thought that I commented + pushed my local changes but I got distracted by a different bug and forgot to do so :o

Itxaka:
Getting sick of hound, inconsistent error reporting causes a mess in the PR.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
aspiers: Thanks @bhdn for this useful information but it will get buried in github never to be found again. That's why I asked for a comment to be embedded in the code to clarify it.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -203,6 +209,30 @@ def usage
...
+          #
+          # 0QEMU_QEMU_HARDDISK_00002 dm-1 QEMU,QEMU HARDDISK
+          # size=10G features='0' hwhandler='0' wp=rw
+          # `-+- policy='service-time 0' prio=1 status=active
+          # `- 0:0:0:2 sdc 8:32   active ready running
+          # `-+- policy='service-time 0' prio=1 status=enabled
+          # `- 0:0:0:3 sdb 8:16   active ready running
+          # sdb and sdc are paths of dm-1:
+          # root@d52-54-77-77-01-01:~ # ls /sys/block/sdb/holders/
+          # dm-1
```
aspiers: This is super helpful - thanks a lot for adding this comment!  Although it looks like some lines got accidentally wrapped.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -160,6 +161,11 @@ def self.claimed(node,owner)
           end
         end
 
+        def self.multipath?(device = @device)
```
aspiers: This looks suspicious to me.  Where would `@device` come from when invoked as a class method?  Please could you add unit tests for this method, both with and without a parameter passed to it?  That would ensure it works in both cases.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -203,6 +209,30 @@ def usage
           self.owner
         end
 
+        def multipath?
+          self.class.multipath?(@device)
+        end
+
+        def multipath_path?
```
aspiers: This method name is a bit ambiguous.  How about `held_by_multipath?` or similar?

chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    @node = @chef_run.node
+    @node.default[:block_device] = {
+      xvd1: {},
+      xvd2: {},
+      xvd3: {},
+      xvd4: {}
+    }
+  end
+
+  it "returns the proper number of unclaimed devices" do
```
aspiers: Awesome! Thanks a lot for adding this!

chef/cookbooks/barclamp/libraries/barclamp_library.rb
aspiers: No sorry, it's really not straightforward to me. (N.B. I'm talking about both these lines, not just one of them.)  I think I managed to understand it last time I looked at it, but looking at it again now, I feel just as confused as I did originally.  There are *four* different things being checked here:

- `d.fixed`
- `d.multipath?`
- `d.claimed?`
- `d.multipath_path?`

And on top of that we have a nice Boolean cocktail of `unless`, `||`, `&&`, and `!`.  So that creates a truth table with 2^4==16 different possibilities, which is not simple to understand.

Some questions which immediately spring to mind:

- Why are two of the conditions used by `next`, and the other two used to form the return value?
- Why are we checking four different things?  The only obvious one is `!d.claimed?` since that corresponds directly to the method name `unclaimed`.  But why are we skipping non-fixed devices, for example?

To be clear: I'm not saying there's anything wrong with the logic, just that it's hard to understand the way it's currently written.  If you really understand it then surely it should be a matter of 60 seconds to add a comment or two which would go a long way to helping others understand.

This is the kind of situation in which @cmurphy's [talk](https://youtu.be/pU-VasVPNAs) seems particularly relevant to me.  I regularly see patches submitted to lots of projects where the submitter assumes that the reviewers have all the necessary context and knowledge to be able to easily review the patch, but this is often not the case.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -203,6 +209,30 @@ def usage
...
+          # size=10G features='0' hwhandler='0' wp=rw
+          # `-+- policy='service-time 0' prio=1 status=active
+          # `- 0:0:0:2 sdc 8:32   active ready running
+          # `-+- policy='service-time 0' prio=1 status=enabled
+          # `- 0:0:0:3 sdb 8:16   active ready running
+          # sdb and sdc are paths of dm-1:
+          # root@d52-54-77-77-01-01:~ # ls /sys/block/sdb/holders/
+          # dm-1
+          # root@d52-54-77-77-01-01:~ # ls /sys/block/sdc/holders/
+          # dm-1
```
aspiers: Not sure what this error is about but I assume that fixing the line wrapping would also fix it.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
bhdn: I put it here as Itxaka is now working on it.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
Itxaka: Woops! Forgot to remove it, good catch!

Also great idea on the unit testing, will add a couple of tests to this new functions instead of only doing the unclaimed method.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
Itxaka: sounds good, will change.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
Itxaka: Seems fair enough to me, let me dig a bit deeper into this and see if:
 - I can simplify it
 - I can add a decent comment to it that its clear enough.

Itxaka:
@s-t-e-v-e-n-k @aspiers ready for a re-review, I think I addressed all concerns and comments regarding the patch!

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -150,6 +150,12 @@ def self.unclaimed(node, include_mounted=false)
               %x{lsblk #{d.name.gsub(/!/, "/")} --noheadings --output MOUNTPOINT | grep -q -v ^$}
               next if $?.exitstatus == 0
             end
+            # we want disks that are multipath but not if they are holders
+            # of a multipath
+            if d.multipath?
```
bhdn: `d.multipath?` and `d.held_by_multipath?` should be mutually exclusive

Itxaka: If that is true, then would we still need to test for `multipath?` ? Couldnt we just check `held_by_multipath?` instead and make the flow easier?

I think I was confused by the different test conditions that were introduced and their combinations as Adam mentioned.

aspiers: Glad I'm not the only one ;-)

bhdn: The original patch:
```
+            next unless d.fixed || d.multipath?
+            !d.claimed? && !d.multipath_path?
```

Suggested:
```
next if d.claimed? || held_by_multipath?
d.fixed || d.multipath?
```

"fixed" devices and multipath must be included; claimed and paths (held) must be skipped.

bhdn: Or maybe `should_be_included = d.fixed || d.multipath?` could help understanding.

bhdn: Calling them "mutually exclusive" was wrong. It may happend that `d.multipath?` and `d.held_by_multipath?` are false.

aspiers: In the suggested version I still don't understand why are two of the conditions used by `next`, and the other two used to form the return value?

bhdn: This is inside `.select` ; using only `next` would result in non-true and the elements would be filtered out. What do you suggest?

aspiers: Why use `next` at all?  Isn't this equivalent?

```ruby
!(d.claimed? || d.held_by_multipath?) &&
d.fixed || d.multipath?
```

or this?

```ruby
next unless d.fixed || d.multipath?
!(d.claimed? || d.held_by_multipath?)
```

I'm not saying they are better, I'm just trying to understand why you're suggesting that particular form.  It seems you're thinking about them as two pairs of two expressions, and maybe there's a good reason for doing that (as suggested by this `should_be_included` idea) but it's not clear to me what that reason is.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -150,7 +150,10 @@ def self.unclaimed(node, include_mounted=false)
               %x{lsblk #{d.name.gsub(/!/, "/")} --noheadings --output MOUNTPOINT | grep -q -v ^$}
               next if $?.exitstatus == 0
             end
-            d.fixed and not d.claimed?
+            # skip claimed disks and multipath holders
```
aspiers: IIUC it's not multipath *holders* which are being skipped here, it's devices *held* by holders.

Itxaka: yeah, Im struggling to find the correct words here as to not make it confusing when someone reads it in the future. devices held by holders seems like also something that could lead to confusing from someone reading it for the first time. Me no good with big words :/

chef/cookbooks/barclamp/libraries/barclamp_library.rb
bhdn: Oh, I suggested this one for no particular reason. So please pick one style that you find best.  Your first proposal looks good style-wise to me, but must be 
```ruby
!(d.claimed? || d.held_by_multipath?) &&
(d.fixed || d.multipath?)
```
otherwise a claimed multipath device could be included in the result.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -203,6 +215,31 @@ def usage
           self.owner
         end
 
+        def multipath?
+          self.class.multipath?(@device)
+        end
+
+        def held_by_multipath?
+          # we need to check if the holders of a device (if it has holders)
+          # are a multipath, for example:
```
s-t-e-v-e-n-k: "are a multipath" should be re-worded, "is a multipath device" perhaps?

chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        BarclampLibrary::Barclamp::Inventory::Disk
+      ).to receive(:multipath?).exactly(1).times.with("subtest1")
+      expect(
+        BarclampLibrary::Barclamp::Inventory::Disk
+      ).to receive(:multipath?).exactly(1).times.with("subtest2")
+      a = BarclampLibrary::Barclamp::Inventory::Disk.new(@node, "test")
+      expect(a.held_by_multipath?).to be(false)
+    end
+  end
+end
```
cmurphy: Are these tests being run automatically somewhere? I don't see them in the travis output.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
aspiers: Yeah, I probably misread this first time round.

chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
aspiers: That's a good question. Maybe an adjustment to `.travis.yml` is needed? Hard to check now, I'm on a plane.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
Itxaka: This is referring to "the holders of a device" so it should be plural no? May be my borked english :)

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -203,6 +215,31 @@ def usage
...
+          # 0QEMU_QEMU_HARDDISK_00002 dm-1 QEMU,QEMU HARDDISK
+          # size=10G features='0' hwhandler='0' wp=rw
+          # -+- policy='service-time 0' prio=1 status=active
+          # - 0:0:0:2 sdc 8:32   active ready running
+          # -+- policy='service-time 0' prio=1 status=enabled
+          # - 0:0:0:3 sdb 8:16   active ready running
```
Itxaka: Good point, I thought I had the command in the first line but seems like I skipped it. Adding it back!

chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
Itxaka: No they are not, let me have a look at the travis file!

Itxaka: Updated!

Itxaka:
@aspiers @bhdn @s-t-e-v-e-n-k @cmurphy all updated, lets get this in :D

bhdn:
I can't vote. But it looks good to me :+1:

chef/cookbooks/barclamp/libraries/barclamp_library.rb
Itxaka: That sounds better, Thanks!!

chef/cookbooks/barclamp/spec/spec_helper.rb
```diff
@@ -0,0 +1,50 @@
...
+  config.log_level = ENV["CHEF_LOG_LEVEL"].to_sym if ENV["CHEF_LOG_LEVEL"]
+
+  # Specify the path to a local JSON file with Ohai data (default: nil)
+  # config.path = 'ohai.json'
+
+  # Specify the operating platform to mock Ohai data from (default: nil)
+  config.platform = "suse"
+
+  # Specify the operating version to mock Ohai data from (default: nil)
+  config.version = "11.3"
```
Itxaka: Done, and tests updated as the mocking of a 12.2 node has extra disks that we need to account (and mock)!

chef/cookbooks/barclamp/libraries/barclamp_library.rb
cmurphy: :+1:

chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
cmurphy: :+1: I see it now

chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    @node.default[:block_device] = {
+      dm0: { removable: "0" },
+      xvd1: { removable: "0" },
+      xvd2: { removable: "1" }
+    }
+  end
+
+  it "unclaimed method returns the proper number of unclaimed devices" do
+    a = BarclampLibrary::Barclamp::Inventory::Disk
+    expect(a).to receive(:`).exactly(5).times.and_return(`exit 1`)
```
cmurphy: What is :` and why would it return a system call?

chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    expect(::File).to receive(:exist?).with("/sys/block/dm0/dm/uuid").and_return(true)
+    expect(::File).to receive(:exist?).with("/sys/block/xvd2/dm/uuid").and_return(false)
+    expect(::File).to receive(:exist?).with("/sys/block/sr0/dm/uuid").and_return(false)
+    expect(::File).to receive(:open).exactly(1).times.with(
+      "/sys/block/dm0/dm/uuid"
+    ).and_yield(StringIO.new("mpath-test"))
+    # return holders
+    expect(::Dir).to receive(:entries).exactly(5).times.and_return([])
+    expect(a.unclaimed(@node).length).to eq(3)
+  end
```
cmurphy: I don't think either this or the held_by_multipath? test below is testing the scenario described in the held_by_multipath? comment, or am I misreading it?

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -150,7 +150,10 @@ def self.unclaimed(node, include_mounted=false)
               %x{lsblk #{d.name.gsub(/!/, "/")} --noheadings --output MOUNTPOINT | grep -q -v ^$}
               next if $?.exitstatus == 0
             end
-            d.fixed and not d.claimed?
+            # skip claimed disks and multipath devices held by holders
+            # but include both fixed disks and multipath devices
```
aspiers: This is certainly getting easier to understand, but the comment doesn't really add much other than just repeat what the code says.  Ideally the comment should explain *why* the code is written that way, not simply what it does.  I would suggest introducing temporary variables to highlight that whilst some devices will *never* be claimable by their inherent nature, whereas others will be claimable unless they're already claimed.  For example:

    device_type_claimable = d.fixed || d.multipath?
    in_use = d.claimed? || d.held_by_multipath?
    device_type_claimable && !in_use

For me, splitting the logic into these two categories makes it clear enough that it essentially becomes self-documenting and doesn't need a comment any more.  Does that make sense?

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -160,6 +163,15 @@ def self.claimed(node,owner)
           end
         end
 
+        def self.multipath?(device)
+          uuid_path = "/sys/block/#{device}/dm/uuid"
+          if File.exist?(uuid_path) && File.open(uuid_path) { |f| f.read(7).start_with?("mpath-") }
```
aspiers: Line is too long.  Also the `true` / `false` in the conditional body lends itself to a rewrite, e.g.

    return false unless File.exist?(uuid_path)
    File.open(uuid_path) { |f| f.read(7).start_with?("mpath-") }

chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    @chef_run = ::ChefSpec::Runner.new
+    @node = @chef_run.node
+    @node.default[:block_device] = {
+      dm0: { removable: "0" },
+      xvd1: { removable: "0" },
+      xvd2: { removable: "1" }
+    }
+  end
+
+  it "unclaimed method returns the proper number of unclaimed devices" do
```
aspiers: Specification header blocks are supposed to read like English, but the `it` doesn't make sense here.  Also `unclaimed method` can be abbreviated to `#unclaimed`:

    specify "#unclaimed returns the proper number of unclaimed devices" do

chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
aspiers: The backtick method runs the given command in a subshell: http://ruby-doc.org/core-2.4.1/Kernel.html#method-i-60

So here it's setting up a mock that will return the same output as `exit 1`.  But that's just the empty string, so I'm not sure what the benefit of the `exit 1` is here.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
Itxaka: Fair enough, will try to clarify it more!

chef/cookbooks/barclamp/libraries/barclamp_library.rb
Itxaka: :+1:

chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
Itxaka: sorry, I did not know you can use the `specify` , will update!

chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
Itxaka: yeah this is wonderful.

So this is mocking the calls that are done to `%x{lsblk #{d.name.gsub(/!/, "/")} --noheadings --output MOUNTPOINT | grep -q -v ^$}` in the `unclaimed` method, but as we check the `$?.exitstatus == 0` I could not find any other proper way of mocking that correctly in a simple way, so I just mock the system call and return an actual `exit 1` and save in complexity :)

> But that's just the empty string

its actually not an empty string due to the special `$?` :O

chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
Itxaka: Not the one in the comment exactly no, I did not use the comment as guide for doing the tests.

Itxaka:
Updated! I have faith this is the latest update for this PR :dancer:

aspiers:
Looks pretty awesome now :-) Thanks!

Itxaka:
Rebased just in case, as the mkcloud failed with something that had nothing to do with this patch, I think.

Itxaka:
mnkcloud run broken due to postgres versions, should be fixed by https://github.com/crowbar/crowbar-openstack/pull/1025/

chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
cmurphy: But the example in the comment is the use case in the bug, yes? So it should have its own test?

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -150,7 +150,11 @@ def self.unclaimed(node, include_mounted=false)
               %x{lsblk #{d.name.gsub(/!/, "/")} --noheadings --output MOUNTPOINT | grep -q -v ^$}
               next if $?.exitstatus == 0
             end
-            d.fixed and not d.claimed?
+            # skip claimed disks and multipath devices held by holders
+            # but include both fixed disks and multipath devices
+            device_type_claimable = d.fixed || d.multipath?
+            in_use = d.claimed? || d.held_by_multipath?
+            device_type_claimable && !in_use
```
cmurphy: Nice, much easier to read

chef/cookbooks/barclamp/spec/libraries/barclamp_library_spec.rb
Itxaka: Indeed. I just did some tests based on the code, never thought to use the example comment as Im not sure if it was the exact case from the bug or a example output.

Itxaka: In case my comment was not clear (its not) I will add a test for this!

Itxaka: Actually, I read this wrong and the `#held_by_multipath` test above this comment is indeed testing exactly the comment!

It should be behaving exactly like the comment says, as the file `/sys/block/#{device}/dm/uuid` wont exit, it should return false

Itxaka:
[P2: bsc#1031065 - L3-Question: crowbar - pacemaker deployment fails with drbd and multipath](https://trello.com/c/8N5tUpaF/219-p2-bsc1031065-l3-question-crowbar-pacemaker-deployment-fails-with-drbd-and-multipath)

PR #1203, opened 2017-04-05
upgrade: Shutdown corosync instead of pacemaker and wait for sbd to stop

jsuchome:
Stopping corosync (unlike stopping pacemaker) initiates also sbd stop.
But it cat take some time to stop sbd so wait until it is really down
before proceeding with other actions.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
@@ -49,15 +49,24 @@ touch $RUNFILE
 trap cleanup INT EXIT
 
 <% if @use_ha %>
-# Shutdown pacemaker so the remaining OpenStack services are stopped
+# Shutdown corosync so the remaining OpenStack services are stopped
 # This is needed so that the zypper dup won't trigger any db migrations on package update
-service pacemaker stop
+service corosync stop
 
 ret=$?
 if [ $ret != 0 ] ; then
-    log "Error occured during shutting down pacemaker service"
+    log "Error occured during shutting down corosync service"
     echo $ret > $UPGRADEDIR/crowbar-pre-upgrade-failed
     exit $ret
+fi
+
+# Make sure sbd is shut down (if it was configured) before proceeding with next actions
+if systemctl --quiet is-enabled sbd 2>/dev/null ; then
+    SBD_PIDFILE=$(systemctl show sbd.service --property PIDFile | sed -n 's/^PIDFile=//p')
+    while [ -e $SBD_PIDFILE ] ; do
+        log "sbd is still active, waiting for shutdown"
+        sleep 1
+    done
 fi
 
 <% else %>


chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -49,17 +49,23 @@ touch $RUNFILE
...
 ret=$?
 if [ $ret != 0 ] ; then
-    log "Error occured during shutting down pacemaker service"
+    log "Error occured during shutting down corosync service"
     echo $ret > $UPGRADEDIR/crowbar-pre-upgrade-failed
     exit $ret
 fi
 
+# Make sure sbd is shut down (if it was configured) before proceeding with next actions
+while systemctl is-active sbd >/dev/null ; do
```
rhafer: I'd prefer ```systemctl --quiet is-active corosync``` instead of the redirection.

I'd also say we need to error out here if sbd is stopped after a certain timeout (1-2minutes?)

jsuchome: There's a timeout for the whole script execution, isn't it enought?

rhafer: Hm, right, I always forget about that :/

jsuchome:
@rhafer added `--quiet`

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -49,17 +49,26 @@ touch $RUNFILE
...
 ret=$?
 if [ $ret != 0 ] ; then
-    log "Error occured during shutting down pacemaker service"
+    log "Error occured during shutting down corosync service"
     echo $ret > $UPGRADEDIR/crowbar-pre-upgrade-failed
     exit $ret
 fi
 
+# Make sure sbd is shut down (if it was configured) before proceeding with next actions
+if systemctl is-enabled sbd ; then
```
rhafer: ```--quiet``` ?

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -49,17 +49,26 @@ touch $RUNFILE
...
+    SBD_PIDFILE=$(sed -n 's/PIDFile=//p' /usr/lib/systemd/system/sbd.service)
```
rhafer: It's a little safer to use use ```s/^PIDFile=//p``` here I think.

And it theory it is possible to override the settings here by configuration in ```/etc/systemd/....```. How about using ```systemctl show sbd.service --property PIDFile | sed -n 's/^PIDFile=//p``` or maybe even ```eval $(systemctl show sbd.service --property PIDFile)```

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
jsuchome: Hm, seems to produce some output even with --quiet when the service is not installed (exit value is correct though)

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
jsuchome: I like it

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -49,17 +49,26 @@ touch $RUNFILE
...
 ret=$?
 if [ $ret != 0 ] ; then
-    log "Error occured during shutting down pacemaker service"
+    log "Error occured during shutting down corosync service"
     echo $ret > $UPGRADEDIR/crowbar-pre-upgrade-failed
     exit $ret
 fi
 
+# Make sure sbd is shut down (if it was configured) before proceeding with next actions
+if systemctl is-enabled sbd 2>/dev/null ; then
```
rhafer: Just redirecting stderr will still produce output stuff on stdout when the service is enabled. So we might still need ```--quiet``` here.

PR #1204, opened 2017-04-05
upgrade: Fix prechecks after prepare

skazi0:
After prepare call, nodes can't be found using roles anymore. Some of the prechecks which rely on this are not prepared to not find the nodes they expect.

Code relying on nodes found by role is guarded with additional condition.

requests?**
Problem was reported in commends under https://bugzilla.suse.com/show_bug.cgi?id=1032377

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -202,10 +202,12 @@ module Api
 
           # So lbaas v1 is configured, let's find out if it is actually used
           neutron = NodeObject.find("roles:neutron-server").first
-          out = neutron.run_ssh_cmd(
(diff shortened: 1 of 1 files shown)


skazi0:
[bsc#1032377 - admin repocheck already running](https://trello.com/c/824vXKAc/403-bsc-1032377-admin-repocheck-already-running)

skazi0:
[bsc#1030067: crowbarctl upgrade prechecks - error 500 when running in wrong order](https://trello.com/c/n4NQKTHB/2316-bsc-1030067-crowbarctl-upgrade-prechecks-error-500-when-running-in-wrong-order)

skazi0:
Forward port is not needed as the problem is inside lbaas precheck which is not present in SOC7.

PR #1205, opened 2017-04-06
upgrade: Log api exceptions to production.log

skazi0:
Exceptions are caught in api endpoints to return proper json responses instead of default 500.html. This disables default exception handling and doesn't log the backtraces to production.log.

The backtraces need to be manualy logged to not lose this information. Already existing function log_exception() was used for logging.

### DIFF
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -56,6 +56,7 @@ class Api::UpgradeController < ApiController
   def prechecks
     render json: Api::Upgrade.checks
   rescue StandardError => e
+    log_exception(e)
     render json: {
       errors: {
         prechecks: {
@@ -89,6 +90,7 @@ class Api::UpgradeController < ApiController
       }
     }, status: :locked
   rescue StandardError => e
+    log_exception(e)
     render json: {
       errors: {
         cancel: {
@@ -114,8 +116,17 @@ class Api::UpgradeController < ApiController
     else
       render json: check
     end
-  rescue Crowbar::Error::UpgradeError,
-         StandardError => e
+  rescue Crowbar::Error::UpgradeError => e
+    render json: {
+      errors: {
+        repocheck_crowbar: {
+          data: e.message,
+          help: I18n.t("api.upgrade.adminrepocheck.help.default")
+        }
+      }
+    }, status: :unprocessable_entity
+  rescue StandardError => e
+    log_exception(e)
     render json: {
       errors: {
         repocheck_crowbar: {


skazi0:
The `log_exception()` output doesn't look exactly like the default rails backtraces but it was there already and all of the needed information is logged so I left it as is.

crowbar_framework/app/controllers/api/crowbar_controller.rb
```diff
@@ -41,6 +41,7 @@ def upgrade
     end
   rescue Crowbar::Error::StartStepRunningError,
          Crowbar::Error::StartStepOrderError => e
+    log_exception(e)
```
jsuchome: Are you sure it belongs here? I think we are able to handle thise Upgrade related exceptions pretty well, and the problematic part is really with `StandardError` only

skazi0: You're right. I initially added logging everywhere, then removed those which are control-flow-exceptions and must have missed this one.

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -153,6 +157,7 @@ def adminbackup
          Crowbar::Error::StartStepOrderError,
          Crowbar::Error::EndStepRunningError,
          Crowbar::Error::SaveUpgradeStatusError => e
+    log_exception(e)
```
jsuchome: I think it does not belong here

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -248,6 +253,7 @@ def mode
   rescue ::Crowbar::Error::SaveUpgradeModeError,
          ::Crowbar::Error::SaveUpgradeStatusError,
          ::Crowbar::Error::UpgradeError => e
+    log_exception(e)
```
jsuchome: ... and here

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -43,6 +43,7 @@ def prepare
   rescue ::Crowbar::Error::StartStepRunningError,
          ::Crowbar::Error::StartStepOrderError,
          ::Crowbar::Error::SaveUpgradeStatusError => e
+    log_exception(e)
```
jsuchome: I think it does not belong here

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -116,6 +119,7 @@ def adminrepocheck
     end
   rescue Crowbar::Error::UpgradeError,
          StandardError => e
+    log_exception(e)
```
jsuchome: it seems better to split the handling of `StandardError` and `Crowbar::Error::UpgradeError`

skazi0:
[bsc#1032377 - admin repocheck already running](https://trello.com/c/824vXKAc/403-bsc-1032377-admin-repocheck-already-running)

PR #1207, opened 2017-04-06
Revert wicked ifreload related changes (bsc#1032751)

rhafer:
Calling ifreload here seems to have been a bad idea as it cause various
issue in HA setups:
- pacemaker mananged VIPs assigned to any of the reloaded interfaces are
  temporary lost, Causing resource failures in the cluster.
- reloading an interface causes a small network outage that can
  de-stabilize the cluster, leading to fenced nodes and other bad
  things.

https://bugzilla.suse.com/show_bug.cgi?id=1032751

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -569,7 +569,6 @@ when "suse"
         nic: nic,
         pre_up_script: pre_up_script
       })
-      notifies :create, "ruby_block[wicked-ifreload-required]", :immediately
     end
     if ifs[nic.name]["gateway"]
       template "/etc/sysconfig/network/ifroute-#{nic.name}" do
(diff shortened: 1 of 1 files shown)


rhafer:
@gosipyan, @vmoravec  you might want to test this (or the Cloud 6 variant) to see if that fixes the HA deployment issues.

rhafer:
> But I'm not sure if this PR makes sense without an alternative solution to the problem for which these wicked calls were originally added.

The problem that the original change tried to solve might be a little less severe. Hard to tell.

With this patch applied the old problem is back. That is, if you install a wicked maintenance update after deployment but before rebooting the node at least once. It will loose network connectivity (and get fenced if it was part of a cluster). I don't have a new solution for that issue yet, need to have another chat with the wicked maintainers.

PR #1213, opened 2017-04-11
upgrade: Download the latest crowbar_join from the server

jsuchome:
Download and execute latest crowbar_join from the provisioner server
so it is not rewritten during chef-client run.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-chef-upgraded.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-chef-upgraded.sh.erb
@@ -26,8 +26,12 @@ if [[ -f $UPGRADEDIR/crowbar-chef-upgraded-ok ]] ; then
     exit 0
 fi
 
+# Download the latest crowbar-join from the server
+curl -s -o /usr/sbin/crowbar_join_tmp <%= @crowbar_join %>
+
 # Move node to ready state and execute chef-client for the first time
-crowbar_join --start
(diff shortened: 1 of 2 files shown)


chef/cookbooks/crowbar/templates/default/crowbar-chef-upgraded.sh.erb
```diff
@@ -26,8 +26,11 @@ if [[ -f $UPGRADEDIR/crowbar-chef-upgraded-ok ]] ; then
     exit 0
 fi
 
+# Download the latest crowbar-join from the server
+curl -s -o /usr/sbin/crowbar_join_tmp <%= @crowbar_join %>
```
vuntz: You also need to chmod a+x the file.

jsuchome:
[Fixing upgrade breakage caused by rewritten crowbar_join](https://trello.com/c/MbTQpr56/227-fixing-upgrade-breakage-caused-by-rewritten-crowbar-join)

chef/cookbooks/crowbar/templates/default/crowbar-chef-upgraded.sh.erb
jsuchome: Added `+x` (should be enough)

rsalevsky:
Could you add the bsc number?

jsuchome:
Tested manually, seems to solve the problem

jsuchome:
Normal gating job is somehow stuck, but notice that the upgrade one has passed

PR #1214, opened 2017-04-11
upgrade: Do not stop services at remote node

jsuchome:
neutron-agents, nova-compute and libvirtd are managed by cluster
and should only be stopped from there


aspiers:
Oh sorry, I get it now. A slightly more verbose commit message wouldn't hurt :) e.g. an extra sentence explaining how the change works.

PR #1215, opened 2017-04-11
upgrade: Do not skip non-kvm nodes

jsuchome:
When non-kvm compute nodes do not have other role
(that would group them with 'controller' or 'other' nodes)
we need to remember to upgrade them during compute_nodes substep.

We didn't see this before probably because xen nodes in other tests had some additional role, so they were upgraded before the real compute_nodes substep (which is not a problem in "normal" mode).


jsuchome:
[P2: bsc#1033641 - [Upgrade] Non-kvm compute nodes may be ignored during the upgrade](https://trello.com/c/I22Hlo0U/235-p2-bsc-1033641-upgrade-non-kvm-compute-nodes-may-be-ignored-during-the-upgrade)

PR #1216, opened 2017-04-12
Fix for lbaasv2 migration with several clusters

abel-navarro:
When the controller nodes are not the same as the network nodes, we have
to retrieve the SQL connection needed for the lbaasv2 migration from the
neutron server.

Reference: https://bugzilla.suse.com/show_bug.cgi?id=1033632

### DIFF
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -188,6 +188,7 @@ template "/usr/sbin/crowbar-router-migration.sh" do
   action :create
 end
 
+neutron_server = search(:node, "run_list_map:neutron-server").first
 template "/etc/neutron/lbaas-connection.conf" do
   source "lbaas-connection.conf"
   mode "0640"
@@ -196,7 +197,7 @@ template "/etc/neutron/lbaas-connection.conf" do
   action :create
   variables(
     lazy do
-      { sql_connection: node[:neutron][:db][:sql_connection] }
+      { sql_connection: neutron_server[:neutron][:db][:sql_connection] }
     end
   )
   only_if { roles.include? "neutron-network" }


chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -188,16 +188,16 @@
...
   source "lbaas-connection.conf"
   mode "0640"
   owner "root"
   group "root"
   action :create
   variables(
-    lazy do
-      { sql_connection: node[:neutron][:db][:sql_connection] }
-    end
+     sql_connection: neutron-server[:neutron][:db][:sql_connection]
```
jsuchome: Why would you remove that lazy evaluation? It's there for possibility of node (or neutron-server) being nil

abel-navarro: right, i'll try to rewrite that with the node search function inside the lazy block.

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -188,16 +188,16 @@
   action :create
 end
 
+neutron-server = search(:node, "role:neutron-server").first
```
vmoravec: A variable in ruby can't have dash in its name

abel-navarro:
[bsc#1033632 - crowbar-lbaas-evacuation.sh failed: Could not parse rfc1738 URL from string](https://trello.com/c/wBa62c4k/520-bsc-1033632-crowbar-lbaas-evacuation-sh-failed-could-not-parse-rfc1738-url-from-string)

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -196,8 +196,7 @@
   action :create
   variables(
     lazy do
-      { sql_connection: node[:neutron][:db][:sql_connection] }
-    end
+      { sql_connection: search(:node, "role:neutron-server").first[:neutron][:db][:sql_connection] }
```
jsuchome: `end` is missing

jsuchome: And even with that end it fails

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -196,8 +196,7 @@
...
   )
   only_if { roles.include? "neutron-network" }
```
jsuchome: Don't we need to extend this as well?

jsuchome:
Please add the bug number and "upgrade: " prefix to the commit message

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -188,16 +188,17 @@
...
   action :create
   variables(
-    lazy do
-      { sql_connection: node[:neutron][:db][:sql_connection] }
-    end
+    lazy {
+      { sql_connection: neutron_server[:neutron][:db][:sql_connection] }
+    }
   )
   only_if { roles.include? "neutron-network" }
```
jsuchome: I think you need to make sure that `neutron_server` exists (`only_if { roles.include?("neutron-network") && !neutron_server.nil? }`

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -188,16 +188,17 @@
...
   source "lbaas-connection.conf"
   mode "0640"
   owner "root"
   group "root"
   action :create
   variables(
-    lazy do
-      { sql_connection: node[:neutron][:db][:sql_connection] }
-    end
+    lazy {
```
rsalevsky:
```
lazy do
...
end
````

abel-navarro:
@rsalevsky fixing that

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
abel-navarro: i thought about that. As it is right now, the file will be created without the sql_connection information if that neutron-server is missing. Otherwise the file won't be created at all.

The first case helped me identify that the script was actually running but with wrong arguments so it was fast to find the cause. The latter could be a bit misleading.

In either case, both will fail miserably because we don't handle this situation nicely. I'd prefer to leave as it is now, which is mirroring the nova-controller code section.

jsuchome: > As it is right now, the file will be created without the sql_connection information if that neutron-
> server is missing.

I don't think so. If `neutron-server` is nil, chef will fail when trying to evaluate `neutron_server[:neutron]...`

PR #1217, opened 2017-04-18
Fix node bulk edit to save only the changed nodes (bsc#1008838)

Itxaka:
We were not excluding the admin node from the target_platform and license_key checks as
it doesnt have an allocated value sent on POST.
Also, we didnt check for a blank value (nil or empty or "") in the public_name values.
Both of this issues triggered a save of all nodes, even if the values were the same.

### DIFF
--- a/crowbar_framework/app/controllers/nodes_controller.rb
+++ b/crowbar_framework/app/controllers/nodes_controller.rb
@@ -139,7 +139,7 @@ class NodesController < ApplicationController
               dirty = true
             end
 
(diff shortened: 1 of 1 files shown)


Itxaka:
[P4: bsc#1008838 - Bulk edit success message mentions all nodes when only one was updated](https://trello.com/c/FB7jvGJw/183-p4-bsc-1008838-bulk-edit-success-message-mentions-all-nodes-when-only-one-was-updated)

PR #1218, opened 2017-04-20
bsc#1033537: Fix domain caching

cmurphy:
The Node model goes through Crowbar::Settings to get the domain name of
a node. Without this patch, Crowbar::Settings will use the domain
attribute of the dns barclamp and, to avoid repeated database queries,
store it permanently as a class variable. This is a problem because if
the domain attribute is changed in the dns barclamp, the change will not
be reflected in the Node model, which means that if a node is removed
and re-added, its new domain name will be properly reflected in Chef but
crowbar will respond with an HTTP 404 Not Found when a node is queried.

This "caching" strategy has been part of the code apparently since the
beginning of time[1] and it's unclear what the performance gain was
thought to be at the time, so one might think that a reasonable option
to fix this would be to undo the "caching" strategy and simply query the
Proposal relation when the domain needs to be fetched. However, this is
not reasonable because the crowbar dashboard UI queries the node list
every few seconds and this causes a lookup of the domain for each node,
so with a reasonably large number of nodes this would easily overload
the database server. Rails' own SQL caching mechanism is not sufficient
here because query caches are destroyed at the end of an action[2].

This patch solves the problem by writing the proposal revision number
to Rails' cache when the dns barclamp is updated, and storing both the
revision number and the domain attribute as a class variable in the
Crowbar::Settings class. The Crowbar::Settings class then queries the
Rails cache to see if the dns proposal has been updated, and if so
updates its copy of the domain attribute. This way we avoid repeated
trips to the database but changes to the proposal will be propogated.

[1] https://github.com/crowbar/crowbar/commit/f95325fc7ec27a2c8ed4a184eca95af8f428ac3d
[2] http://guides.rubyonrails.org/caching_with_rails.html#sql-caching

### DIFF
--- a/crowbar_framework/app/models/proposal.rb
+++ b/crowbar_framework/app/models/proposal.rb
@@ -68,6 +68,13 @@ class Proposal < ActiveRecord::Base
     else
       item["deployment"][self.barclamp]["crowbar-revision"] += 1
     end
+    # The domain attribute of the dns barclamp is needed when looking up a Node.
+    # We don't want to go to the database every time we need it since that will
+    # be many times per second, but nor can we statically record it since it may
+    # change. Instead we use Rails' caching mechanism to store the new revision,
+    # and when we look it up elsewhere we first check the revision to see if the
+    # barclamp has been updated and we require a trip to the database.
+    cache_crowbar_revision if barclamp == 'dns'
   end
 
   def raw_attributes
@@ -257,5 +264,10 @@ class Proposal < ActiveRecord::Base
 
   def update_proposal_id
     properties["id"] = "#{barclamp}-#{name}"
+  end
+
+  def cache_crowbar_revision
+    cache_key = "deployment_#{barclamp}_crowbar_revision"
+    Rails.cache.write(cache_key, item["deployment"][barclamp]["crowbar-revision"])
   end
 end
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/proposal.rb
```diff
@@ -68,6 +68,13 @@ def increment_crowbar_revision!
     else
       item["deployment"][self.barclamp]["crowbar-revision"] += 1
     end
+    # The domain attribute of the dns barclamp is needed when looking up a Node.
+    # We don't want to go to the database every time we need it since that will
+    # be many times per second, but nor can we statically record it since it may
+    # change. Instead we use Rails' caching mechanism to store the new revision,
+    # and when we look it up elsewhere we first check the revision to see if the
+    # barclamp has been updated and we require a trip to the database.
+    cache_crowbar_revision if barclamp == 'dns'
```
jsuchome: Wouldn't it be possible to cache the revision from `dns_service.rb` to avoid this barclamp specific action?

aplanas: or cache the value for every barclamp? `cache_crowbar_revision` looks ready for that

cmurphy: @jsuchome I think yes, that's probably better
@aplanas I did it that way at first but decided there was more likely no reason other barclamps would want to cache this value

cmurphy:
This is actually a lot more problematic than I originally thought. I wrote up some thoughts on the trello card https://trello.com/c/J8sKouiL and I'm closing this change for now.

PR #1219, opened 2017-04-20
puma: Disable puma_worker_killer reaper logs

rsalevsky:
This change disables the puma reaper logs which look like
`PumaWorkerKiller: Consuming 100 mb with master and 2 workers.`.


jsuchome:
I'm fine with disabling those log messages.
But are you sure we're not losing any other kind of logs?

rsalevsky:
I'm trusting the upstream docu in this case, so if this is not totally wrong then yes I'm sure.

PR #1220, opened 2017-04-21
network: Add support for xmit_hash_policy (bsc#1033917)

s-t-e-v-e-n-k:
Add support for the xmit_hash_policy property of NIC bonding to be queried
and set, as well as adding the existing miimon property and new
xmit_hash_policy property to the network schema.

Drive-by correctly setting miimon rather than hardcoding 100 for the
interfaces(5) template.

### DIFF
--- a/chef/cookbooks/barclamp/libraries/nic.rb
+++ b/chef/cookbooks/barclamp/libraries/nic.rb
@@ -605,6 +605,15 @@ class ::Nic
       self
     end
 
+    def xmit_hash_policy
+      sysfs("bonding/xmit_hash_policy").split[0]
+    end
+
+    def xmit_hash_policy=(xmit_hash_policy)
+      sysfs_put("bonding/xmit_hash_policy", xmit_hash_policy)
+      self
+    end
+
     def down
       slaves.each{ |s|s.down }
       super
@@ -625,7 +634,7 @@ class ::Nic
       nil
     end
 
-    def self.create(nic,mode=6,miimon=100)
+    def self.create(nic, mode=6, miimon=100, xmit_hash_policy="layer2")
       Chef::Log.info("Creating new bond #{nic}")
       if self.exists?(nic)
         raise ::ArgumentError.new("#{nic} already exists.")
@@ -642,6 +651,7 @@ class ::Nic
       iface = ::Nic.new(nic)
       iface.mode = mode
       iface.miimon = miimon
+      iface.xmit_hash_policy = xmit_hash_policy
       iface.up
       iface
     end
--- a/chef/cookbooks/network/templates/default/suse-cfg.erb
+++ b/chef/cookbooks/network/templates/default/suse-cfg.erb
@@ -59,7 +59,7 @@ ETHERDEVICE=<%=quote(iface["parent"])%>
    # Settings specific to bond devices (ifcfg-bonding(5))
    when @nic.kind_of?(Nic::Bond) -%>
 BONDING_MASTER=yes
-BONDING_MODULE_OPTS=<%=quote("mode=#{@nic.mode} miimon=#{@nic.miimon}")%>
+BONDING_MODULE_OPTS=<%=quote("mode=#{@nic.mode} miimon=#{@nic.miimon} xmit_hash_policy=#{@nic.xmit_hash_policy}")%>
 <% iface["slaves"].each_with_index do |slave,i| -%>
 BONDING_SLAVE_<%=i%>=<%=quote(slave)%>
 <%   end -%>
--- a/chef/data_bags/crowbar/migrate/network/101_add_bonding_miimon_xmit_hash_policy.rb
+++ b/chef/data_bags/crowbar/migrate/network/101_add_bonding_miimon_xmit_hash_policy.rb
@@ -1,0 +1,19 @@
+def upgrade(ta, td, a, d)
+  unless a["teaming"].key? "miimon"
+    a["teaming"]["miimon"] = ta["teaming"]["miimon"]
+  end
+  unless a["teaming"].key? "xmit_hash_policy"
+    a["teaming"]["xmit_hash_policy"] = ta["teaming"]["xmit_hash_policy"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  unless ta["teaming"].key? "miimon"
+    a["teaming"].delete "miimon"
+  end
+  unless ta["teaming"].key? "xmit_hash_policy"
+    a["teaming"].delete "xmit_hash_policy"
+  end
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-network.json
+++ b/chef/data_bags/crowbar/template-network.json
@@ -8,7 +8,9 @@
       "enable_tx_offloading": true,
       "mode": "single",
       "teaming": {
-        "mode": 1
+        "mode": 1,
+        "miimon": 100,
+        "xmit_hash_policy": "layer2"
       },
       "interface_map": [
         {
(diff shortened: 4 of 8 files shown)


chef/data_bags/crowbar/template-network.schema
```diff
@@ -20,7 +20,9 @@
               "type": "map",
               "required": true,
               "mapping": {
-                "mode": { "type": "int", "required": true }
+                "mode": { "type": "int", "required": true },
+                "miimon": { "type": "int", "required": true },
+                "xmit_hash_policy": { "type": "str", "required": true, "pattern": "/^layer2$|^layer2+3$|^layer3+4$|^encap2+3$|^encap3+4$/" }
```
aplanas: Do we need a migration?

s-t-e-v-e-n-k: I actually wrote one, I completely forgot to git add it. :-(

chef/data_bags/crowbar/template-network.json
```diff
@@ -8,7 +8,9 @@
       "enable_tx_offloading": true,
       "mode": "single",
       "teaming": {
-        "mode": 1
+        "mode": 1,
+        "miimon": 100,
+        "xmit_hash_policy": "layer2"
```
vuntz: I'm 80% sure we want to allow changing these settings on a per-network basis. That's why we have the `team_mode` attribute for networks, that can override the `mode` attribute here.

s-t-e-v-e-n-k: Do we want both, IE, my current change for global settings plus per-network, or only per-network?

vuntz: I'd say both.

s-t-e-v-e-n-k:
If anyone is curious enough to test this https://gist.github.com/s-t-e-v-e-n-k/7986ad03beadb5855ba5452d10b344b9 is a copy of my notes and the two network.json.patch files I used.

chef/cookbooks/barclamp/libraries/nic.rb
```diff
@@ -605,6 +605,15 @@ def miimon=(millisecs)
       self
     end
 
+    def xmit_hash_policy
+      sysfs("bonding/xmit_hash_policy")
```
rhafer: Are we sure that this is always present? It seems setting ```xmit_hash_policy``` does only make sense for certain bonding modes (2, 4 and 5 according to https://www.kernel.org/doc/Documentation/networking/bonding.txt). Is the file also present when the bond is currently using any of the other modes?

s-t-e-v-e-n-k: At least in my testing, it's always present, and you can in fact always set it -- I suspect the kernel ignores the setting if the mode is not 2, 4 or 5.

rhafer: Ok. Thanks for clarifying.

bhdn:
A backport is needed as a PTF is being provided.

Itxaka:
provided backport at https://github.com/crowbar/crowbar-core/pull/1231

chef/cookbooks/barclamp/libraries/nic.rb
nicolasbock: @s-t-e-v-e-n-k are you sure this path always exists?

And did I read this correctly that the path resolves to something like `/sys/class/net/enp0s20f0u2/bonding`? Because if that's the case I don't see a `bonding` directory on my laptop for this NIC.

s-t-e-v-e-n-k: /sys/class/net/<foo>/bonding exists for *bonds*, not all NICs. modprobe bonding on your laptop will create a bond0 with no slaves, and that will have a bonding directory. Furthermore, that function is in the Nic::Bond class, not the generic Nic class.

dirkmueller:
waiting for testing feedback

nvin-krja:
I tested it and there seems to be a problem. We don't apply these settings in template file and it doesn't go in sysconfig files. I changed the template file to get values from iface[] and then it worked.

nvin-krja:
```
BONDING_MODULE_OPTS=<%=quote("mode=#{@nic.mode} miimon=#{iface["miimon"]}  xmit_hash_policy=#{iface["xmit_hash_policy"]}")%>
```

chef/data_bags/crowbar/migrate/network/101_add_bonding_miimon_xmit_hash_policy.rb
```diff
@@ -0,0 +1,19 @@
+def upgrade(ta, td, a, d)
+  unless a["network"]["teaming"].key? "miimon"
```
bhdn: `["network"]` is not needed for `a` and `ta` (ie it breaks the migration.)

chef/cookbooks/network/templates/default/suse-cfg.erb
```diff
@@ -59,7 +59,7 @@ ETHERDEVICE=<%=quote(iface["parent"])%>
    # Settings specific to bond devices (ifcfg-bonding(5))
    when @nic.kind_of?(Nic::Bond) -%>
 BONDING_MASTER=yes
-BONDING_MODULE_OPTS=<%=quote("mode=#{@nic.mode} miimon=#{@nic.miimon}")%>
+BONDING_MODULE_OPTS=<%=quote("mode=#{@nic.mode} miimon=#{@nic.miimon} xmit_hash_policy=#{@nic.xmit_hash_policy}")%>
```
bhdn: We had to change it to use `iface` instead of `@nic`, otherwise it wouldn't update in an already deployed cloud.

s-t-e-v-e-n-k: I have modified the earlier code to also set miimon and xmit_hash_policy on the nic object, if @rhafer thinks that is fine, then it also solves this problem as well (which I've tested and confirmed)

rhafer: I think that is ok.

chef/data_bags/crowbar/template-network.json
rhafer: @vuntz: Was that referring to ```xmit_hash_policy``` only? Or would you also like to see ```miimon``` changable per-network? AFAICS the current code only allows ```xmit_hash_policy```.

```miimon``` is only looked up in the "global" setting if I read the code correctly.

Comment (vuntz) [inline chef/data_bags/crowbar/template-network.json]: @rhafer doing `miimon` too would make sensen, indeed.

Comment (rhafer) [inline chef/data_bags/crowbar/template-network.json]: Ok, @s-t-e-v-e-n-k could you please also add ```miimon``` do the networks as well?

File: chef/data_bags/crowbar/migrate/network/101_add_bonding_miimon_xmit_hash_policy.rb [hunk 6]
```diff
@@ -0,0 +1,19 @@
...
+end
+
+def downgrade(ta, td, a, d)
+  unless ta["teaming"].key? "miimon"
+    a["teaming"].delete "miimon"
+  end
+  unless ta["teaming"].key? "xmit_hash_policy"
+    a["teaming"].delete "xmit_hash_policy"
+  end
+  return a, d
```
stefannica: Probably not important, but shouldn't the 'miimon' and 'xmit_hash_policy' attrs be deleted from the conduits as well, if present ?

bhdn: I think that's important. I tripped over it when testing downgrade some days ago.

chef/cookbooks/network/templates/default/suse-cfg.erb
stefannica: So does this mean that miimon/xmit_hash_policy can be updated post-deployment now ?

s-t-e-v-e-n-k: Yes, it can -- in the raw view for the network barclamp.

PR #1222, opened 2017-04-21
[3.0] crowbar: Ensure run_list is how we want it to be

sayalilunkad:
When we add a role to a node, we need the role in the run_list_map and
in the run_list, but if it was in one or the other, we were not doing
anything. This resulted in semi-broken setups (which could happen due to
crashes, or manual editing) not getting fixed when re-applying a
barclamp.

Now, when we change the run_list, we really ensure that it's how we want
it to be.

(cherry picked from commit b341f0b9b7ec70e6b2131f94ef3f39e6da60a89b)

### DIFF
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -709,21 +709,26 @@ class NodeObject < ChefObject
   end
 
   def add_to_run_list(rolename, priority, states = nil)
+    Rails.logger.debug("Ensuring #{name} has role #{rolename} with priority #{priority}")
+    save_it = false
     states = ["all"] unless states
-    crowbar["run_list_map"] = {} if crowbar["run_list_map"].nil?
+    crowbar["run_list_map"] ||= {}
     val = { "states" => states, "priority" => priority }
-    crowbar["run_list_map"][rolename] = val
-    Rails.logger.debug("crowbar[run_list_map][#{rolename}] = #{val.inspect}")
-    Rails.logger.debug("current state is #{self.crowbar['state']}")
-
-    # only rebuild the run_list if it effects the current state.
-    self.rebuild_run_list if states.include?("all") or states.include?(self.crowbar["state"])
+    if crowbar["run_list_map"][rolename] != val
+      crowbar["run_list_map"][rolename] = val
+      save_it = true
+    end
+    rebuild_run_list || save_it
   end
 
   def delete_from_run_list(rolename)
-    crowbar["run_list_map"] = {} if crowbar["run_list_map"].nil?
-    crowbar["run_list_map"][rolename] = { "states" => ["all"], "priority" => -1001 } unless crowbar["run_list_map"].nil?
-    crowbar_run_list.run_list_items.delete "role[#{rolename}]"
+    Rails.logger.debug("Ensuring #{name} doesn't have role #{rolename}")
+    crowbar["run_list_map"] ||= {}
+    if crowbar["run_list_map"].key?(rolename)
+      crowbar["run_list_map"].delete(rolename)
+      save_it = true
+    end
+    rebuild_run_list || save_it
   end
 
   def rebuild_run_list
@@ -743,12 +748,16 @@ class NodeObject < ChefObject
     end
     Rails.logger.debug("rebuilt run_list will be #{vals.inspect}")
 
+    old_run_list = crowbar_run_list.run_list_items.dup
+
     # Rebuild list
     crowbar_run_list.run_list_items.clear
     vals.each do |item|
       next if item[1]["priority"] == -1001 # Skip deleted items
       crowbar_run_list.run_list_items << "role[#{item[0]}]"
     end
+
+    old_run_list != crowbar_run_list.run_list_items
   end
 
   def run_list_to_roles
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1099,9 +1099,7 @@ class ServiceObject
 
             # An old node that is not in the new deployment, drop it
             unless new_nodes.include?(node_name)
-              @logger.debug "remove node #{node_name}"
-              pending_node_actions[node_name] = { remove: [], add: [] } if pending_node_actions[node_name].nil?
-
+              pending_node_actions[node_name] ||= { remove: [], add: [] }
               pending_node_actions[node_name][:remove] << role_name
 
               # Remove roles are  a way to "de-configure" things on the node
@@ -1145,12 +1143,9 @@ class ServiceObject
               next
             end
 
-            # A new node that we did not know before
-            unless old_nodes.include?(node_name)
-              @logger.debug "add node #{node_name}"
-              pending_node_actions[node_name] = { remove: [], add: [] } if pending_node_actions[node_name].nil?
-              pending_node_actions[node_name][:add] << role_name
-            end
+            pending_node_actions[node_name] ||= { remove: [], add: [] }
+            pending_node_actions[node_name][:add] << role_name
+
             nodes_in_batch << node_name unless nodes_in_batch.include?(node_name)
           end
         end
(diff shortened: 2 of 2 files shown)


sayalilunkad:
[P2: Bug 1030881 - #L3 Crowbar not updating new node run_list](https://trello.com/c/vP4LXrnN/247-p2-bug-1030881-l3-crowbar-not-updating-new-node-run-list)

crowbar_framework/app/models/node_object.rb
```diff
@@ -709,21 +709,24 @@ def crowbar_run_list(*args)
...
-    Rails.logger.debug("crowbar[run_list_map][#{rolename}] = #{val.inspect}")
-    Rails.logger.debug("current state is #{self.crowbar['state']}")
-
-    # only rebuild the run_list if it effects the current state.
-    self.rebuild_run_list if states.include?("all") or states.include?(self.crowbar["state"])
+    Rails.logger.debug("Ensuring #{name} has role #{rolename} with priority #{priority}")
+    save_it = false
+    crowbar["run_list_map"] ||= {}
+    if crowbar["run_list_map"][rolename] != { "priority" => priority }
+      crowbar["run_list_map"][rolename] = { "priority" => priority }
```
vuntz: In Cloud 6, you also need to have `"states" => states` in the hash.

sayalilunkad: Done, thanks!

crowbar_framework/app/models/node_object.rb
```diff
@@ -709,21 +709,25 @@ def crowbar_run_list(*args)
...
+      crowbar["run_list_map"][rolename] = { "states" => states }
```
vuntz: That doesn't look correct. I'd do something like:

```ruby
entry = { "states" => states, "priority" => priority }
if crowbar["run_list_map"][rolename] != entry
  crowbar["run_list_map"][rolename] = entry
  save_it = true
end
```

crowbar_framework/app/models/node_object.rb
```diff
@@ -709,21 +709,26 @@ def crowbar_run_list(*args)
...
   end
 
   def delete_from_run_list(rolename)
-    crowbar["run_list_map"] = {} if crowbar["run_list_map"].nil?
-    crowbar["run_list_map"][rolename] = { "states" => ["all"], "priority" => -1001 } unless crowbar["run_list_map"].nil?
-    crowbar_run_list.run_list_items.delete "role[#{rolename}]"
+    Rails.logger.debug("Ensuring #{name} doesn't have role #{rolename}")
+    crowbar["run_list_map"] ||= {}
+    if crowbar["run_list_map"].key?(rolename)
+      crowbar["run_list_map"].delete(rolename)
```
jsuchome: Are we sure this change does not break the upgrade? We're relying on searches in `run_list_map` quite a lot...

sayalilunkad: I can't be sure but as I understand this should only add the same roles to run_list and run_list_map so both are consistent with each other. @vuntz what do you think?

jsuchome: Yes, I understand. It just occurred to me that we might be relaying on that inconsistency.
Anyway, I'm deploying upgrade setup to test it...

jsuchome: ok, so it was false alarm, I tested with upgrade manually

vuntz:
@sayalilunkad didn't have time to check the updated patch, but gating is red.

sayalilunkad:
@vuntz yes I am looking into the failure but seems like its not directly related to the run_list, so the run_list is building but something caused by this change.

jsuchome:
I don't know how is it possible, but this PR seems to remove the "admin" network data from crowbar node ... effectively killing the network by chef-client run...

crowbar_framework/app/models/service_object.rb
```diff
@@ -1513,15 +1489,8 @@ def add_role_to_instance_and_node(barclamp, instance, name, prop, role, newrole)
...
-    unless node.role?(newrole)
-      node.add_to_run_list(newrole, local_chef_order, role_map[newrole])
-      save_it = true
-    end
-
-    unless node.role?("#{barclamp}-config-#{instance}")
-      node.add_to_run_list("#{barclamp}-config-#{instance}", local_chef_order)
-      save_it = true
-    end
+    save_it = node.add_to_run_list(newrole, local_chef_order) || save_it
```
jsuchome: Here there's an argument missing compared to previous version of `add_to_run_list` call

jsuchome: Here there's an argument missing (`role_map[newrole]`) compared to previous version of `add_to_run_list` call. 

When I add it, my manual test seems to succeed and add the network data correctly.

jsuchome: I wonder if it is in original (master) PR as well, please recheck...

sayalilunkad: Have you replaces it for all the call of add_to_run_list? And it is similar in the original Pr so no role_mape[newrole]

jsuchome: I did not. This one here seems to 'fix' my problem with network not being saved. But it might not be enough...

crowbar_framework/app/models/service_object.rb
```diff
@@ -1513,15 +1489,8 @@ def add_role_to_instance_and_node(barclamp, instance, name, prop, role, newrole)
...
-    unless node.role?(newrole)
-      node.add_to_run_list(newrole, local_chef_order, role_map[newrole])
-      save_it = true
-    end
-
-    unless node.role?("#{barclamp}-config-#{instance}")
-      node.add_to_run_list("#{barclamp}-config-#{instance}", local_chef_order)
-      save_it = true
-    end
+    save_it = node.add_to_run_list(newrole, local_chef_order, role_map[newrole]) || save_it
```
jsuchome: That's just a detail, but we do not need `|| save_it` in this line

crowbar_framework/app/models/service_object.rb
```diff
@@ -1145,12 +1143,9 @@ def apply_role(role, inst, in_queue)
...
             end
 
-            # A new node that we did not know before
-            unless old_nodes.include?(node_name)
-              @logger.debug "add node #{node_name}"
-              pending_node_actions[node_name] = { remove: [], add: [] } if pending_node_actions[node_name].nil?
-              pending_node_actions[node_name][:add] << role_name
-            end
+            pending_node_actions[node_name] ||= { remove: [], add: [] }
+            pending_node_actions[node_name][:add] << role_name
```
jsuchome: This is the problematic part. @sayalilunkad , revert this block to the previous state and try with that one.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1148,9 +1146,10 @@ def apply_role(role, inst, in_queue)
             # A new node that we did not know before
             unless old_nodes.include?(node_name)
               @logger.debug "add node #{node_name}"
-              pending_node_actions[node_name] = { remove: [], add: [] } if pending_node_actions[node_name].nil?
+              pending_node_actions[node_name] ||= { remove: [], add: [] }
               pending_node_actions[node_name][:add] << role_name
             end
```
vuntz: This part is different from the original patch, and the missing bit is actually a key part in fixing the bug.

sayalilunkad: Do you mean not having the unless condition here?  I think that was needed to make gating pass..

vuntz: > Do you mean not having the unless condition here?

Yes.

> I think that was needed to make gating pass..

Do we have details on why it was breaking with that change?

jsuchome: I do not remember exactly, I think there were no network data saved in the initial chef-client run when executing instcrowbar step

sayalilunkad: Just removed the condition to so see what fails exactly.

sayalilunkad:
@vuntz @jsuchome can you please review this again. It works now since the other two PR's are merged.

PR #1223, opened 2017-04-21
upgrade: Make sure to upgrade all nodes in normal node

jsuchome:

We've met some situations where the node was not upgraded during the controller phase, but was also missing nova-compute role. The effect was that such node was not upgraded at all.

UPDATE: Bug 1035528 is dedicated to this situation I was referring to


jsuchome:
> Also it would by nice if you could add some more reasoning for the change (e.g. under what
> circumstances the compute role can be missing)

I do not really understand why it happened in the bug that is mentioned here... the fact that proposal lists node having a role, while the node itself does not know about it seems to point to some broken setup.

PR #1225, opened 2017-04-24
crowbar: Do not save applied proposal as role too early in apply_role

vuntz:
Until now, we saved the role for the applied proposal early in
apply_role. However, apply_role could fail after that but still before
apply_role_pre_chef_client, resulting in the role containing incomplete
data (since apply_role_pre_chef_client is used to process the role and
add some info, like "use HA").

In case of such a failure (and that can happen when we fail to get a
lock for a node), then a periodic chef-client run would use a role with
incomplete information, possibly leading to changing the config on the
node wrongly.

Closes https://github.com/sap-oc/crowbar-openstack/issues/34

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1005,9 +1005,6 @@ class ServiceObject
       new_deployment.delete("elements_expanded")
     end
 
-    # make sure the role is saved
-    role.save
-
     # Build a list of old elements.
     # elements_expanded on the old role is guaranteed to exists, as we already
     # ran through apply_role with the old_role.  Cache is used for the case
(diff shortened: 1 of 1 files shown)


vuntz:
@matelakat any +1 here, since you tested the patch? :-)

matelakat:
> @matelakat any +1 here, since you tested the patch? :-)

sorry, I'm being slow here.

aspiers:
[(5) bsc#857375: role assignment isn't sufficiently fine-grained](https://trello.com/c/pQ7zWeVk/375-5-bsc-857375-role-assignment-isn-t-sufficiently-fine-grained)

aspiers:
Hmm, IIUC actually this change doesn't make things more fine-grained at all, it just makes them happen later.

vuntz:
> But doesn't this partially fix bsc#857375 (role assignment isn't orchestrated correctly (incl. conflict between apply_role and periodic chef runs + other issues)) and fully fix the correspondingTrello card? So I think the commit message needs to reference at least the bug.

Right it partially addresses bsc#857375, but it's actually not related to the trello card you linked to, as the fine-grained part there was more about "node C should not get compute role until node A with controller role has been applied".

PR #1226, opened 2017-04-26
dns: Disable domain field in UI (bsc#1033537)

cmurphy:
The domain field should not be changed after deployment. Chef uses this
field to identify nodes, including the admin node, and changing it
causes crowbar and chef to be able to unable to fetch data about a node.
This changes makes the field in the DNS barclamp UI disabled and adds a
note about it to discourage trying to make this change.

### DIFF
--- a/crowbar_framework/app/views/barclamp/dns/_edit_attributes.html.haml
+++ b/crowbar_framework/app/views/barclamp/dns/_edit_attributes.html.haml
@@ -3,7 +3,14 @@
     = header show_raw_deployment?, true
 
   .panel-body
-    = string_field :domain
+
+    %fieldset
+    - if @proposal.active?
+      = string_field :domain, :disabled => "disabled"
+      .alert.alert-info
+        = t('.domain_hint')
+    - else
+      = string_field :domain
 
     %fieldset
       %legend
--- a/crowbar_framework/config/locales/dns/en.yml
+++ b/crowbar_framework/config/locales/dns/en.yml
@@ -20,6 +20,7 @@ en:
     dns:
       edit_attributes:
         domain: 'Domain'
+        domain_hint: 'Changing the domain after initial deployment is not supported.'
         ip_list_hint: 'A comma-separated list of IP Addresses'
         server_hint: 'At least one nameserver or one node with the dns-server role must be specified.'
 


crowbar_framework/app/views/barclamp/dns/_edit_attributes.html.haml
```diff
@@ -3,7 +3,12 @@
     = header show_raw_deployment?, true
 
   .panel-body
-    = string_field :domain
+
+    %fieldset
+      = string_field :domain, :disabled => "disabled"
```
Itxaka: wouldnt this make the domain field always disabled, even on the first dns barclamp proposal creation?

Could we play with the `@proposal.active?` var to disable it once it has been deployed?

```
      = string_field :domain, :disabled => @proposal.active?
```

cmurphy: I think this is set using yast when you first install crowbar, so crowbar comes up with this proposal already activated and the user would never see the barclamp UI before it's active. Maybe someone else can confirm how it works?

jsuchome: @cmurphy I think you are right that we never see this not being deployed, but maybe there could be ways (if not now, than in future) to deploy the initial crowbar stack differently. So I'd say @Itxaka's proposal makes sense

cmurphy: Okay that's fair

crowbar_framework/config/locales/dns/en.yml
```diff
@@ -20,6 +20,7 @@ en:
     dns:
       edit_attributes:
         domain: 'Domain'
+        domain_hint: 'Changing the domain after deployment is not supported.'
```
jsuchome: I would say ***after initial deployment*** (It's probably implied, but kind of looks better for me)

cmurphy: Done

crowbar_framework/app/views/barclamp/dns/_edit_attributes.html.haml
cmurphy: Done

rsalevsky:
Code looks good but could you please add a Screenshot as this is a UI change?

cmurphy:
@rsalevsky  screenshot_2017-04-27_10-46-12

PR #1227, opened 2017-04-27
dashboard: Show a warning when the product is EOL

rsalevsky:
When the product is EOL we should warn the customer that the product is
EOL and out of support.

### DIFF
--- a/crowbar_framework/app/controllers/application_controller.rb
+++ b/crowbar_framework/app/controllers/application_controller.rb
@@ -1,6 +1,6 @@
 #
 # Copyright 2011-2013, Dell
-# Copyright 2013-2014, SUSE LINUX Products GmbH
+# Copyright 2013-2017, SUSE Linux GmbH
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -27,6 +27,7 @@ class ApplicationController < ActionController::Base
 
   before_action do |c|
     Crowbar::Sanity.cache! unless Rails.cache.exist?(:sanity_check_errors)
+    check_eol if Rails.cache.fetch(:eol_checked).blank?
   end
 
   before_filter :enforce_installer, unless: proc {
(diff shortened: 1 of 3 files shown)


crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -215,4 +216,17 @@ def upgrade
...
   end
+
+  def check_eol
+    branding = YAML.load_file(Rails.root.join("config", "branding.yml"))
+    return if branding["eol_date"].blank?
+    return if Time.parse(branding["eol_date"]).future?
+
+    flash[:alert] = I18n.t(
+      "dashboard.eol",
+      product: branding["page_title"],
```
jsuchome: Are we really talking about ***Crowbar*** product?

rsalevsky: In the case of this PR yes, but the branding (page_title) will be replaced in our package. But I'm unsure if we should set the EOL date in here or in our package.

jsuchome: Well, it looks like it would make sense to set both values form the package

rsalevsky: Ok, removed the date from the branding. Now this PR is very generic and we can activate it from the package.

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -27,6 +27,7 @@ class ApplicationController < ActionController::Base
 
   before_action do |c|
     Crowbar::Sanity.cache! unless Rails.cache.exist?(:sanity_check_errors)
+    check_eol unless Rails.cache.fetch(:eol_checked).blank?
```
jsuchome: Doesn't the `unless ... blank?` make sure that the `check_eol` is actually never called?

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -27,6 +27,7 @@ class ApplicationController < ActionController::Base
 
   before_action do |c|
     Crowbar::Sanity.cache! unless Rails.cache.exist?(:sanity_check_errors)
+    check_eol if Rails.cache.fetch(:eol_checked).blank?
```
dirkmueller: so that would be shown on any crowbar page, not just the "/" home page?

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -215,4 +216,18 @@ def upgrade
       end
     end
   end
+
+  def check_eol
+    Rails.cache.write(:eol_checked, true, expires_in: 24.hours)
+
+    branding = YAML.load_file(Rails.root.join("config", "branding.yml"))
+    return if branding["eol_date"].blank?
+    return if Time.parse(branding["eol_date"]).future?
```
dirkmueller: I wonder if we can easily digest that information from the update channels, because then we'd only have one place for tracking EOL date and not two..

rsalevsky: I found nothing in the channels but we could use the `/etc/products.d/suse-openstack-cloud.prod` file.

PR #1228, opened 2017-04-27
network: set max mtu of a parent to the max of its children

Itxaka:
iterate over all the children to obtain the max mtu that we need to
set to a parent nic. Before this, only the first vlan children of a parent
was used to set an mtu value, failing afterwards if the next vlan had a bigger
value.

Introduced in https://github.com/crowbar/crowbar-core/pull/1084

cc @nvin-krja

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -404,7 +404,7 @@ Nic.nics.each do |nic|
       if ifs[nic.name].key?("mtu") &&
           ifs.key?(parent_nic.name) &&
           ifs[parent_nic.name].key?("mtu") &&
-          ifs[parent_nic.name]["mtu"] < ifs[nic.name]["mtu"]
+          ifs[parent_nic.name]["mtu"].to_i < ifs[nic.name]["mtu"].to_i
         # parent has custom mtu lower than the nic custom mtu
         msg = "Interface #{parent_nic.name} has a custom mtu of #{ifs[parent_nic.name]["mtu"]} "\
                "but children vlan #{nic.name} wants a mtu of #{ifs[nic.name]["mtu"]}."\
@@ -417,7 +417,7 @@ Nic.nics.each do |nic|
 
       unless all_children.empty?
         max_mtu_nic = all_children.max_by { |k, v| v["mtu"] }[0]
-        ifs[nic.parent]["mtu"] = ifs[max_mtu_nic]["mtu"]
+        ifs[nic.parent]["mtu"] = ifs[max_mtu_nic]["mtu"].to_i
         parent_nic.mtu = ifs[max_mtu_nic]["mtu"]
       end
     end


Itxaka:

nvin-krja:
We must check for default MTU 1500 as well in the first section of 1084 patch. Throw an error only if there is custom MTU on parent less than that of VLAN.

nvin-krja:
And the error message says - Network mtu configuration is invalid. But if there is no custom mtu in configuration for parent then it is not invalid. It is just default.

Itxaka:
@nvin-krja updated with your comment.

nvin-krja:
I am no ruby expert but this worked for me.

    # 2) set the mtu for the parent if needed
    if ifs[nic.name]["mtu"].to_i != 1500
      # we want the highest mtu to end up in the ifcfg-$parent config
      parent_nic = Nic.new(nic.parent)
      all_children = ifs.select{|k,v| v["parent"] == nic.parent}
      max_mtu_nic = all_children.max_by{|k,v| v["mtu"]}[0]
      ifs[nic.parent]["mtu"] = ifs[max_mtu_nic]["mtu"]
      parent_nic.mtu = ifs[max_mtu_nic]["mtu"]
      Chef::Log.info("vlan #{nic.name} wants mtu #{ifs[nic.name]["mtu"]} but " \
                     "parent #{parent_nic.name} wants no/lower mtu. Set mtu "\
                     "for #{parent_nic.name} to #{parent_nic.mtu} as this is the"\
                     "highest mtu found in #{parent_nic.name} children, from #{max_mtu_nic}")
    end


It still won't work for nova_fixed because parent for nova_fixed is ovs-system

Itxaka:
@nvin-krja The only change that I can see in your code different from what in there its the condition to apply it.

So instead of checking if the parent mtu is lower than the interface mtu, you propose to recalculate the parent mtu everytime an interface has a non-standard mtu set.

I cant see how would that fix anything tbh. Is this a race condition that we are hitting now and that change will fix it somehow that I cant see?

Itxaka:
@nvin-krja Updated with your comments.

The condition to apply is different from what you suggested, as we still have to respect if the network config has set a custom MTU for the parent right? In your example we are just setting the parent MTU no matter what the end user has set to the arent network MTU config

nvin-krja:
We should restructure the  code. I made a mistake in my code. I didn't check if parent is already present or not. If parent is already present then we don't have to create a new one. And yes, I take children from ifs because it is a new parent. 

We still have to add additional code for bond interface. Because for bonding, parent is always present. And we might have to check slaves for custom MTU. This is the pseudocode i commented on the bug. 

if interface is vlan and has custom MTU:
  if parent is not bond:
    if parent present in config and has custom MTU lower than vlan:
      throw error
    else:
      scan all children of this parent in config and use the highest MTU on parent
  if parent is bond:
    if parent bond interface has custom MTU lower than vlan:
      throw error
    check all slaves of parent bond
    if any slave interface in config has custom MTU lower than bond vlan:
      throw error
    else:
      scan all children of this bond parent in config and use the highest MTU on parent bond interface

Itxaka:
@nvin-krja I was just working ont he code you added to the bug into this PR, sorry for not seeing it before!

Itxaka:
ignore hound complaints and commit message, will redo it later when its final

chef/cookbooks/network/recipes/default.rb
```diff
@@ -397,34 +397,49 @@ def get_datapath_id_for_ovsbridge(bridge)
...
-    networks_using_parent.each do |net_name|
-      net = Barclamp::Inventory.get_network_by_type(node, net_name)
-      unless net.use_vlan
-        nic_parent = Nic.new nic.parent
-        if nic_parent.mtu.to_i < ifs[nic.name]["mtu"].to_i
-          msg = "#{nic.name} wants mtu #{ifs[nic.name]["mtu"]} but network #{net_name} " \
-                "using the parent nic #{nic.parent} wants a lower mtu #{net.mtu}. " \
-                "This network mtu configuration is invalid."
+  if nic.is_a?(Nic::Vlan) and nic.key?("mtu")
+    parent_nic = Nic.new(nic.parent)
```
nvin-krja: we should first check if parent is present

Itxaka: ah yes, you mentioned that a bond is also reported as a vlan right?

chef/cookbooks/network/recipes/default.rb
```diff
@@ -398,32 +398,49 @@ def get_datapath_id_for_ovsbridge(bridge)
...
-        nic_parent = Nic.new nic.parent
-        if nic_parent.mtu.to_i < ifs[nic.name]["mtu"].to_i
-          msg = "#{nic.name} wants mtu #{ifs[nic.name]["mtu"]} but network #{net_name} " \
-                "using the parent nic #{nic.parent} wants a lower mtu #{net.mtu}. " \
-                "This network mtu configuration is invalid."
+    if nic.parent
+      parent_nic = Nic.new(nic.parent)
+      if parent_nic.bond?
+        # parent is a bond
+        if parent_nic.key?("mtu") && parent_nic["mtu"] < ifs[nic.name]["mtu"].to_i
```
rhafer: Hm, I might be lacking some ruby knowledge here, but are you sure this works as expected? ```parent_nic``` is object of ```class Nic``` (or a subclass). Why are you treating it as a Hash?

Additionally, if I am not mistaken, the "mtu" of parent_nic will always be set to something (even when it's using the default values) so I believe this check (and the error message) are wrong here.

chef/cookbooks/network/recipes/default.rb
```diff
@@ -398,32 +398,51 @@ def get_datapath_id_for_ovsbridge(bridge)
...
+          # parent has custom mtu lower than the nic custom mtu
+          msg = "Parent bond #{parent_nic.name} has a custom mtu of #{parent_nic["mtu"]} "\
+                "but net #{nic.name} wants a mtu of #{nic["mtu"]}."\
+                "This configuration is invalid."
           Chef::Log.fatal(msg)
           raise msg
         end
+        # check bond slaves interfaces for mtu values
```
rhafer: Is this really needed? Under what circumstance can the bond slaves have a diffferent MTU than the bond itself?

chef/cookbooks/network/recipes/default.rb
Itxaka: Indeed on the first part, it will have a mtu function, not a key, I'll fix this.

And indeed it seems that on the second part, the mtu function will read from /sys/class/net/DEVICE/mtu which will always be present, even with no mtu set on our side, so this would always be raised.

Will change!

chef/cookbooks/network/recipes/default.rb
Itxaka: I think @nvin-krja found it on his tests

Itxaka:
@nvin-krja just pushed the latest version, can you test it on your test env?

nvin-krja:
There are lot of changes so I'll put the entire code here and point out what I changed.
```
   # Do some MTU checks/configuration for the parent of the vlan nic
   if nic.is_a?(Nic::Vlan)
     if nic.parent
       parent_nic = Nic.new(nic.parent)
       if parent_nic.is_a?(Nic::Bond)
         # parent is a bond
         # check bond slaves interfaces for mtu values
         parent_nic.slaves.each do |bond_iface| 
           if ifs[nic.name]["mtu"] &&
               ifs[bond_iface.name] &&
               ifs[bond_iface.name].key?("mtu") &&
               ifs[bond_iface.name]["mtu"] < ifs[nic.name]["mtu"]
             msg = "Interface #{bond_iface.name}, part of bond #{parent_nic.name} has a "\
                 "custom mtu of #{ifs[bond_iface.name]["mtu"]} but children vlan #{nic.name} wants a mtu "\
                 "of #{ifs[nic.name]["mtu"]}. This configuration is invalid."
             Chef::Log.fatal(msg)
             raise msg
           end
         end
         # no problems, set the highest mtu of the bond's children
         all_children = ifs.select { |k, v| v["parent"] == nic.parent &&  v.key?("mtu") }
         max_mtu_nic = all_children.max_by{|k,v| v["mtu"]}[0]
       else  
         # parent is not a bond
         if ifs[nic.name].key?("mtu") &&
             ifs.key?(parent_nic.name) &&
             ifs[parent_nic.name].key?("mtu") &&
             ifs[parent_nic.name]["mtu"] < ifs[nic.name]["mtu"]
           # parent has custom mtu lower than the nic custom mtu
           msg = "Interface #{parent_nic.name} has a custom mtu of #{ifs[parent_nic.name]["mtu"]} "\
                 "but children vlan #{nic.name} wants a mtu of #{ifs[nic.name]["mtu"]}."\
                 "This configuration is invalid."
           Chef::Log.fatal(msg)
           raise msg
         else
           # no problems, set the highest mtu of all children in the same interface
           all_children = ifs.select { |k, v| v["parent"] == nic.parent && v.key?("mtu") }
           max_mtu_nic = all_children.max_by{|k,v| v["mtu"]}[0]
         end
       end
       ifs[nic.parent]["mtu"] = ifs[max_mtu_nic]["mtu"]
       parent_nic.mtu = ifs[max_mtu_nic]["mtu"]
     end
   end
```

- The loop for bond slaves with 'next if <condition>' is incorrect. You have code after the condition which gets executed when there is no mtu defined on slave eth interfaces. 
- Changed ifs[bond_iface] to ifs[bond_iface.name]
- Get mtu value from ifs[] everwhere.
- Added additional check for ifs[nic.name].key?("mtu") in both if statements.
- Added condition in select child to select only if they have mtu key.
- Changed max_by(&:mtu) to max_by{|k,v| v["mtu"]}[0] because it didn't work.

I tested it for bond setup with custom MTU on vlan bond interface and normal vlan setup with custom MTUs on two different vlan interface to see if the code picks the maximum one. Tried a negative test case as well with parent set to default and custom MTU on vlan inteface. Didn't try bonded slave eth interface with custom mtu. I don't think we will ever hit that case because our our conduit_map setting in network.json file.

Itxaka:
Code seems ok, I pushed as a new commit to be able to see the differences better and also added an extra logging as I have the suspicion that we are doing some checks wrong thus getting failures on mkcloud

nvin-krja:
I tested it on default mkcloud setup without any custom MTU settings just to see if my changes break anything. It worked fine.

There might still be some corner cases but as this point I can't think of anything. If I do, I will test it more.

Itxaka:
Fixed hounds issues and rebased both commits into one.

Also changed the `if` on https://github.com/crowbar/crowbar-core/pull/1228/files#diff-43125a344f7a7312bf484096c5e371ecR408 for a `next unless`

chef/cookbooks/network/recipes/default.rb
```diff
@@ -398,32 +398,46 @@ def get_datapath_id_for_ovsbridge(bridge)
...
-                "using the parent nic #{nic.parent} wants a lower mtu #{net.mtu}. " \
-                "This network mtu configuration is invalid."
+    if nic.parent
+      parent_nic = Nic.new(nic.parent)
+      Chef::Log.info("Checking MTU for nic #{nic.name} with parent #{parent_nic.name}")
+      if parent_nic.is_a?(Nic::Bond)
+        # parent is a bond
+        # check bond slaves interfaces for mtu values
+        parent_nic.slaves.each do |bond_iface|
+          next unless ifs[nic.name].key?("mtu") &&
```
Itxaka: hound go home, you are drunk. I can clearly see 2 spaces in there

chef/cookbooks/network/recipes/default.rb
```diff
@@ -398,33 +398,47 @@ def get_datapath_id_for_ovsbridge(bridge)
...
+              ifs[bond_iface.name] &&
+              ifs[bond_iface.name].key?("mtu") &&
+              ifs[bond_iface.name]["mtu"] < ifs[nic.name]["mtu"]
+          msg = "Interface #{bond_iface.name}, part of bond #{parent_nic.name} has a "\
+               "custom mtu of #{ifs[bond_iface.name]["mtu"]} but children vlan #{nic.name} "\
+               "wants a mtu of #{ifs[nic.name]["mtu"]}. This configuration is invalid."
```
rhafer: (Basically just repeating my earlier comment, as it got hidden by the latest changes)
@nvin-krja Can you explain when exactly this can happen? I'd be a little surprised if crowbar sets up a bond on interfaces that has a different mtu than the bond itself. So I'd like to find out if we really need a special case.

nvin-krja: I didn't know how our conduit_map works in setting up a bond interface and I was thinking of all possible scenarios where we must check child configuration for custom MTU and that is why I put this in pseudocode. Now that I understand it, I realized that this will never happen. We cannot have an interface with custom MTU inside bond. 

We can remove from the code.

Itxaka: So we actually dont need a check for bond interface here then. Will push.

vuntz:
@nvin-krja are you still okay with the patch? Bogdano mentioned you found an issue just before we deployed it with the PTF (and that's why the PTF doesn't use that patch at the moment).

Itxaka:
@vuntz seems that @nvin-krja is out this week so this needs to wait at least until next week for a +1 from him :D

Itxaka:
Updated with an extra guard to protect against all_children being empty, as found by @nvin-krja during his tests.

I kept the commit separated to make it easier to review, will squash on merge.

nvin-krja:
+1

chef/cookbooks/network/recipes/default.rb
```diff
@@ -398,32 +398,28 @@ def get_datapath_id_for_ovsbridge(bridge)
...
+          ifs[parent_nic.name]["mtu"] < ifs[nic.name]["mtu"]
+        # parent has custom mtu lower than the nic custom mtu
+        msg = "Interface #{parent_nic.name} has a custom mtu of #{ifs[parent_nic.name]["mtu"]} "\
+               "but children vlan #{nic.name} wants a mtu of #{ifs[nic.name]["mtu"]}."\
+               "This configuration is invalid."
+        Chef::Log.fatal(msg)
+        raise msg
+      end
+      # set the highest mtu of all children in the same interface
+      all_children = ifs.select { |k, v| v["parent"] == nic.parent && v.key?("mtu") }
```
vuntz: Am I right that we're going to run this code for each children of a parent? Shouldn't we do the opposite and run it once for each parent?

chef/cookbooks/network/recipes/default.rb
```diff
@@ -398,32 +398,28 @@ def get_datapath_id_for_ovsbridge(bridge)
...
+
+      unless all_children.empty?
+        max_mtu_nic = all_children.max_by { |k, v| v["mtu"] }[0]
+        ifs[nic.parent]["mtu"] = ifs[max_mtu_nic]["mtu"]
+        parent_nic.mtu = ifs[max_mtu_nic]["mtu"]
```
vuntz: Shouldn't we *not* do this if the parent MTU is already higher?

chef/cookbooks/network/recipes/default.rb
```diff
@@ -398,32 +398,28 @@ def get_datapath_id_for_ovsbridge(bridge)
...
       parent_nic = Nic.new(nic.parent)
-      Chef::Log.info("vlan #{nic.name} wants mtu #{ifs[nic.name]["mtu"]} but " \
-                     "parent #{parent_nic.name} wants no/lower mtu. Set mtu "\
-                     "for #{parent_nic.name} to #{ifs[nic.parent]['mtu']}")
-      parent_nic.mtu = ifs[nic.parent]["mtu"]
+      Chef::Log.info("Checking MTU for nic #{nic.name} with parent #{parent_nic.name}")
+      if ifs[nic.name].key?("mtu") &&
+          ifs.key?(parent_nic.name) &&
+          ifs[parent_nic.name].key?("mtu") &&
+          ifs[parent_nic.name]["mtu"] < ifs[nic.name]["mtu"]
```
dirkmueller: why are we no longer doing a ".to_i" comparison like the code did before? don't we want to compare numerically here?

chef/cookbooks/network/recipes/default.rb
```diff
@@ -398,32 +398,28 @@ def get_datapath_id_for_ovsbridge(bridge)
...
+               "This configuration is invalid."
+        Chef::Log.fatal(msg)
+        raise msg
+      end
+      # set the highest mtu of all children in the same interface
+      all_children = ifs.select { |k, v| v["parent"] == nic.parent && v.key?("mtu") }
+
+      unless all_children.empty?
+        max_mtu_nic = all_children.max_by { |k, v| v["mtu"] }[0]
+        ifs[nic.parent]["mtu"] = ifs[max_mtu_nic]["mtu"]
```
dirkmueller: same here, why the removal of .to_i ?

chef/cookbooks/network/recipes/default.rb
Itxaka: Indeed, I thought this was already the case, so that check might have been overwritten in the multiple changes for the patch :(

will (re?)add it

chef/cookbooks/network/recipes/default.rb
Itxaka: Im guessing that the implicit conversions when comparing strings composed of only numbers was the reasoning, but indeed, explicit over implicit, will change.

chef/cookbooks/network/recipes/default.rb
Itxaka: Umm, while this is correct, the patch was done to fix the existing code issues more than to refactor for optimization.

Would this be a blocker for you @vuntz ? Maybe that would work better as a follow up card to improve this code?

chef/cookbooks/network/recipes/default.rb
Itxaka: ahh, actually that check its on line 407!

chef/cookbooks/network/recipes/default.rb
dirkmueller: @Itxaka I'm also struggling to find the signal in the noise.. could you perhaps undo the nic_parent -> parent_nic rename or split that into a separate PR so that the difference gets smaller? this is tricky code and we don't have good test coverage for it so I would like to see the guts of the change easily.

Itxaka: @dirkmueller I dont see whre I changed nic_parent to parent_nic in the code?

old line 426 is now at 421, the Chef log statement is just extended to include more information but using parent_nic otherwise, and the rest of the code is newly introduced (The check for all_children)

cmurphy:
@toabctl done, but in https://github.com/crowbar/crowbar-core/pull/1371

The source in github says "unknown repository" so maybe @Itxaka's branch was deleted? Not otherwise sure how to push back to this PR.

PR #1229, opened 2017-05-02
provisioner: Fix UEFI on non-x86 hosts again

dirkmueller:
The previous change accidentally always only handled "x86_64" because
it was never looking up the machine's architecture but only the fallback
case.


dirkmueller:
aarch64 gating fails because the updated code isn't actually used - it is built as x86_64.rpm :(

PR #1230, opened 2017-05-05
[3.0] network: set max mtu of a parent to the max of its children

Itxaka:
iterate over all the children to obtain the max mtu that we need to
set to a parent nic. Before this, only the first vlan children of a parent
was used to set an mtu value, failing afterwards if the next vlan had a bigger
value.

(cherry picked from commit https://github.com/crowbar/crowbar-core/commit/9c7a1b384b220613e32d90102e19c8b56a5496fc)

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -394,32 +394,25 @@ Nic.nics.each do |nic|
 
   # Do some MTU checks/configuration for the parent of the vlan nic
   if nic.is_a?(Nic::Vlan)
-    # 1) validate that the parent of vlan nic is not used by a network directly
-    # and the requested mtu for the parent is lower than the vlan nic mtu.
-    # That would not work and setting the mtu on the vlan would fail.
-    networks_using_parent = if_mapping.select { |net, ifaces| ifaces.include? nic.parent }.keys
-    networks_using_parent.each do |net_name|
-      net = Barclamp::Inventory.get_network_by_type(node, net_name)
-      unless net.use_vlan
-        nic_parent = Nic.new nic.parent
-        if nic_parent.mtu.to_i < ifs[nic.name]["mtu"].to_i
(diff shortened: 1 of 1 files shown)


Itxaka:
Updated with latest changes from https://github.com/crowbar/crowbar-core/commit/9c7a1b384b220613e32d90102e19c8b56a5496fc

Itxaka:
updated with changes from https://github.com/crowbar/crowbar-core/commit/c108dba2b9556840785411b2c4ab7d15c8bdda86

Itxaka:

Itxaka:
mkcloud HA failure seems related to manila no finding the flavor, doesnt seem to relate to this PR

vuntz:
@Itxaka I'd feel more comfortable cherry-picking https://github.com/crowbar/crowbar-core/pull/1084 and then https://github.com/crowbar/crowbar-core/pull/1228, instead of having one commit with both changes

Itxaka:
@vuntz doesnt #1228 overrides #1084 anyway? or do you want to preserve the history of both patches together?

Itxaka:
oh wait, there is a line missing from #1084 which is not on #1228 :( :(

vuntz:
> @vuntz doesnt #1228 overrides #1084 anyway? or do you want to preserve the history of both patches together?

When we backport, we tend to cherry-pick to preserve history and to link to how the thing was fixed in the main branch so we have the full context.

Itxaka:
Blocked until https://github.com/crowbar/crowbar-core/pull/1241/ is merged first, to preserve proper history of the changes

vuntz:
For the record, this should also be blocked because https://github.com/crowbar/crowbar-core/pull/1228 is still not merged...

Itxaka:
| For the record, this should also be blocked because #1228 is still not merged...

I **hope** that was understood, even if not said :P

Itxaka:
Needs update once https://github.com/crowbar/crowbar-core/pull/1228 is squashed and merged!

Itxaka:
recovered at https://github.com/crowbar/crowbar-core/pull/1371

PR #1087, opened 2017-02-09
network: Allow custom MTU for all networks (bsc#1024277)

toabctl:
Previously we allowed custom MTU values only for the "admin", "storage"
and "os_sdn" networks. This prevents users to create custom networks
with MTU!=1500 so allow custom MTUs for all networks.

For example, a network.json containing a network like:

"storageclient": {
  "add_bridge": false,
  "broadcast": "10.1.5.255",
  "conduit": "intf4",
  "mtu": 9000,
  "netmask": "255.255.255.0",
  "ranges": {
  "host": {
    "end": "10.1.5.239",
    "start": "10.1.5.10"
  }
}

works now.

PR #1232, opened 2017-05-05
[3.0] network: Allow custom MTU for all networks (bsc#1024277)

Itxaka:
Previously we allowed custom MTU values only for the "admin", "storage"
and "os_sdn" networks. This prevents users to create custom networks
with MTU!=1500 so allow custom MTUs for all networks.

For example, a network.json containing a network like:

"storageclient": {
  "add_bridge": false,
  "broadcast": "10.1.5.255",
  "conduit": "intf4",
  "mtu": 9000,
  "netmask": "255.255.255.0",
  "ranges": {
  "host": {
    "end": "10.1.5.239",
    "start": "10.1.5.10"
  }
}

works now.

(cherry picked from commit 9d242e6bab5d79dfaa04a04f803b8efc0fa2e022)

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -305,12 +305,8 @@ node["crowbar"]["network"].keys.sort{|a,b|
     net_ifs << our_iface.name
   end
   if network["mtu"]
-    if ["admin", "storage", "os_sdn"].include? name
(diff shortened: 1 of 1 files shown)


rhafer:
Just for completeness: this is backport of: https://github.com/crowbar/crowbar-core/pull/1087

Itxaka:

PR #1233, opened 2017-05-05
chef: Fix chef-solr downgrade migration (bsc#1037784)

rsalevsky:
We can delete the chef element directly as we add it this migration,
there is no need to delete the other elements first. Also the syntax was
wrong.

### DIFF
--- a/chef/data_bags/crowbar/migrate/crowbar/102_chef-solr.rb
+++ b/chef/data_bags/crowbar/migrate/crowbar/102_chef-solr.rb
@@ -12,14 +12,6 @@ def upgrade(ta, td, a, d)
 end
 
 def downgrade(ta, td, a, d)
-  unless ta["chef"].key?("solr_heap")
-    a.delete["chef"]["solr_heap"]
-  end
-  unless ta["chef"].key?("solr_tmpfs")
-    a.delete["chef"]["solr_tmpfs"]
-  end
-  unless ta["chef"]
-    a.delete["chef"]
-  end
+  a.delete("chef") unless ta.key?("chef")
   return a, d
 end


chef/data_bags/crowbar/migrate/crowbar/102_chef-solr.rb
```diff
@@ -12,14 +12,6 @@ def upgrade(ta, td, a, d)
 end
 
 def downgrade(ta, td, a, d)
-  unless ta["chef"].key?("solr_heap")
-    a.delete["chef"]["solr_heap"]
-  end
-  unless ta["chef"].key?("solr_tmpfs")
-    a.delete["chef"]["solr_tmpfs"]
-  end
-  unless ta["chef"]
```
dirkmueller: this one should have been ta.key("chef") ?

chef/data_bags/crowbar/migrate/crowbar/102_chef-solr.rb
```diff
@@ -12,14 +12,6 @@ def upgrade(ta, td, a, d)
...
-    a.delete["chef"]
-  end
+  a.delete("chef") unless ta.key?("chef")
```
dirkmueller: I'm not sure this is worth the discussion time - but I would feel better if we'd keep it symmetrical. what if some other migration adds chef keys? we'd still have to remove the "solr" keys then, but only those.

rsalevsky: In theory the other downgrade migrations have to run first so we would not have the other chef keys.

PR #1236, opened 2017-05-09
bind: set transfer-source to avoid xfer failures (bsc#1036601)

dirkmueller:

http://bugzilla.suse.com/show_bug.cgi?id=1036601

Tells bind to use the admin IP as source for zone transfers, even if VIPs from pacemaker are the primary addr at the moment.

requests?**

### DIFF
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -135,6 +135,8 @@ def make_zone(zone)
     master_ip = node[:dns][:master_ip]
   end
 
+  admin_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+
   Chef::Log.debug "Creating zone file for zones: #{zonefile_entries.inspect}"
   template "/etc/bind/zone.#{zone[:domain]}" do
     source "zone.erb"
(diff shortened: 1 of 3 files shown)


bmwiedemann:
mkcloud fails in dns with
```
undefined method `admin_addr' for Chef::Resource::Template
```

dirkmueller:
@bmwiedemann yeah.. does it fix the problem? Could you try it manually in an environment that had that issue?

bmwiedemann:
I could not get SOC7 deployed that far yet, so will test it later.
(I tried 3 times and got 3 different failures)

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -143,7 +146,8 @@ def make_zone(zone)
     group "root"
     notifies :reload, "service[bind9]"
     variables(zonefile_entries: zonefile_entries,
-              master_ip: master_ip)
+              master_ip: master_ip,
+              admin_addr: admin_addr)
```
bmwiedemann: admin_addr seems to be undefined here - likely because we are in a function here while the other use of admin_addr is not.

dirkmueller: should be fixed

bmwiedemann:
Finally was able to deploy a HA cloud again. Test shows that it mostly fixes the issue, but there is `/etc/bind/named.conf.default-zones` still containing slave entries without a transfer-source line. Better fix that, too.

dirkmueller:
@bmwiedemann updated

bmwiedemann:
I hit the bug again in new.cloud.suse.de and this patch fixed it.

PR #1237, opened 2017-05-19
Enhance network-json-resolve to also work with exported json for nodes

vuntz:
When debugging data from a deployment that we don't have access to, we
usually only get json describing the nodes and the network.json file.
Without direct access to the chef-server, we can't query chef to get the
data about the node.

So add a --node-json option to allow not using chef to get that
information. If used together with --network-json, it means we need no
access to chef-server at all.

### DIFF
--- a/crowbar_framework/lib/crowbar/conduit_resolver.rb
+++ b/crowbar_framework/lib/crowbar/conduit_resolver.rb
@@ -328,14 +328,14 @@ module Crowbar
 
     ## Return the DMI system attributes from the node
     def cr_dmi_system
-      return {} if @node[:dmi].nil? || @node[:dmi][:system].nil?
+      return {} if @node.automatic_attrs["dmi"].nil? || @node.automatic_attrs["dmi"]["system"].nil?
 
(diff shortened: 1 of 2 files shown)


crowbar_framework/lib/crowbar/conduit_resolver.rb
```diff
@@ -328,14 +328,14 @@ def cr_node_roles
 
     ## Return the DMI system attributes from the node
     def cr_dmi_system
-      return {} if @node[:dmi].nil? || @node[:dmi][:system].nil?
+      return {} if @node.automatic_attrs[:dmi].nil? || @node.automatic_attrs[:dmi][:system].nil?
```
rhafer: Using symbols here doesn't seem to work when loading a node from json. At least for me ```network-json-resolve --network-json network-config-default.json --node-json node.json bus_order <node>``` always return "No bus order defined for  ...." until I use strings instead of symbols here and in the following lines

vuntz: Thanks, fixed.

PR #1238, opened 2017-05-22
upgrade: Use more robust check to ensure DRBD replication started (bsc#1038826)

jsuchome:
We should try to do a better detection that the replication has started and is running.

So, the first output, after the drbd is started might look like

```
postgresql role:Secondary
  disk:Inconsistent
  d52-54-77-77-01-01 connection:Connecting
```

Then we have the replication phase, looking like
```
postgresql role:Secondary
  disk:Inconsistent
  d52-54-77-77-01-01 role:Primary
    replication:SyncTarget peer-disk:UpToDate done:0.08
```
After this is done, drbd reports something like this:
```
postgresql role:Secondary
  disk:UpToDate
  d52-54-77-77-01-01 role:Primary
    peer-disk:UpToDate
```

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
@@ -62,15 +62,35 @@ fi
 
 if [[ -f /sbin/drbdadm ]] ; then
     # Wait until DRBD is started
+    # Initially, 'drbdadm status' returns just empty string so let's get pass this stage.
     while [[ $(drbdadm status | wc -c) -eq 0 ]]; do
         log "Waiting for DRBD to bring the devices up..."
         sleep 1
     done
+    log "DRBD is up: `drbdadm status`"
+
+    # Wait for replication to start
+    # 'drbdadm status' could look like:
+    # postgresql role:Secondary
+    #  disk:Inconsistent
+    #    d52-54-77-77-01-01 connection:Connecting
+    while ! drbdadm status | grep -qi replication; do
+        log "Replication did not start yet: `drbdadm status`"
+        sleep 1
+    done
+    log "Replication started: `drbdadm status`"
+
     # Now wait after DRBD data are synced with the other node
-    while drbdadm status | grep -q Inconsistent; do
+    # While replicating, 'drbdadm status' looks like
+    # postgresql role:Secondary
+    #  disk:Inconsistent
+    #    d52-54-77-77-01-01 role:Primary
+    #      replication:SyncTarget peer-disk:UpToDate done:0.08
+    while drbdadm status | grep -qi -e inconsistent -e replication; do
         log "Replication status: `drbdadm status`"
         sleep 10
     done
+    log "Replication done: `drbdadm status`"
 fi
 
 <% end %>


nick-wang:
IMHO, 5 sec still can't guarantee everything. I may keep "sleep 1" for checking the device up. And add one more loop[1] to make sure the initial sync starting(unlikely the sync finished within 1 sec), then check the resync finished.

...
while drbdadm status | grep -q "done:"; do
    log "Resync starting: `drbdadm status`"
    sleep 1
done
...

chef/cookbooks/crowbar/templates/default/crowbar-post-upgrade.sh.erb
```diff
@@ -66,11 +66,21 @@ if [[ -f /sbin/drbdadm ]] ; then
...
+    # Wait for replication to start
+    while ! drbdadm status | grep -qi replication; do
+        log "Replication did not start yet: `drbdadm status`"
+        sleep 1
+    done
+    drbdadm status
+
     # Now wait after DRBD data are synced with the other node
-    while drbdadm status | grep -q Inconsistent; do
+    while drbdadm status | grep -qi -e inconsistent -e replication; do
```
rhafer: I wonder if we also need a check after the leaving this loop if the sync really was successful? I guess the fact that the ```drbdadm status``` output does not contain ```inconsistent``` or ```replication``` does not necessarily mean that the sync was successful, right? Doesn't it report something like "uptodate" when it is really done?

jsuchome: Good point. I have to deploy some drbd setup to check how the output should look like

jsuchome: OK, I've added the output from the real upgrade to the PR description. That also shows us that simple grepping for "uptodate" will not help us, as that string is already present during replication.

@nick-wang do you have an idea for a simple check after the replication was done, to make sure it ended correctly? Does it make sense to check the exit value of some command? Or could we expect some special error output?

rhafer: I don't know of sure but aren't we interesting in this:

```
postgresql role:Secondary
  disk:UpToDate
  ^^^^^^^^^^^^^
  d52-54-77-77-01-01 role:Primary
    peer-disk:UpToDate
```

AFAIK the first block of the ```drbdadm status``` output for a specific volume eflects the state of the local disk (i.e. the disk that the command is executed on).

ICBW though, and I guess it's better to clarify with @nick-wang .

jsuchome: Yes, I noticed that. But it's not simple grep for `disk:UpToDate`, see that the state during replicating contains `peer-disk:UpToDate` string.
Could be solved by finer grepping, but maybe we have some better option?

jsuchome:
@rhafer so after discussion with @nick-wang on irc, I think that this level of checking is good enough.

As I understand it, if there's really some problem during the replication, the output will continue to show `Inconsistent`, which keeps us in the last loop and the script eventually times out with the error.

rhafer:
> As I understand it, if there's really some problem during the replication, the output will continue to show Inconsistent, which keeps us in the last loop and the script eventually times out with the error.

Ok, thanks for clarifying

jsuchome:
@dirkmueller @rsalevsky it would be good if this fix could go out quickly, to prevent potential DRBD upgrade issues; is it possible?

PR #1239, opened 2017-05-22
network: Use wicked to control bond slaves (bsc#1035127)

s-t-e-v-e-n-k:
Currently, when adding slaves to a bond, the interface will be brought
down and then back up. If wicked is not told about this state of
affairs, wicked will add the configured IP address back onto the
interface when it is bought back up.
    
To counteract this, we run wicked ifdown for any slave interface, and
then run wicked ifup once the template files have been written.

requests?**

### DIFF
--- a/chef/cookbooks/barclamp/libraries/nic.rb
+++ b/chef/cookbooks/barclamp/libraries/nic.rb
@@ -566,7 +566,6 @@ class ::Nic
       usurp(slave)
       slave.down
       sysfs_put("bonding/slaves","+#{slave}")
-      slave.up
       slave
     end
 
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -209,6 +209,12 @@ sorted_networks.each do |network|
     ifs[bond.name]["addresses"] ||= Array.new
     ifs[bond.name]["slaves"] = Array.new
     base_ifs.each do |i|
+      # If the slave isn't already a member of this bond, it may be configured
+      # with an IP or DHCP, and we don't want wicked to re-apply it when the
+      # interface is brought back up.
+      unless bond.slaves.include? i
+        ::Kernel.system("wicked ifdown #{i.name}")
+      end
       bond.add_slave i
       ifs[bond.name]["slaves"] << i.name
       ifs[i.name]["slave"] = true
@@ -565,7 +571,9 @@ when "suse"
         nic: nic,
         pre_up_script: pre_up_script
       })
+      notifies :create, "ruby_block[wicked-ifup-required]", :immediately
     end
+
     if ifs[nic.name]["gateway"]
       template "/etc/sysconfig/network/ifroute-#{nic.name}" do
         source "suse-route.erb"
@@ -579,6 +587,30 @@ when "suse"
         action :delete
       end
     end
+  end
+
+  run_wicked_ifup = false
+
+  # This, when notified by the above "ifcfg" templates, sets run_wicked_ifup
+  # to true (which was initialized to false in the compile phase).
+  # run_wicked_ifup is later used as an "only_if" guard for the
+  # "wicked ifup all" call that is needs to happen when any of the config
+  # files got updated. The purpose of doing it this way (instead of notifying
+  # the "wicked-ifup-all" resource directly), is to make sure that the
+  # ifup is only run once after all ifcfg file have been updated and
+  # independent of how many of them were changed.
+  ruby_block "wicked-ifup-required" do
+    block do
+      run_wicked_ifup = true
+    end
+    action :nothing
+  end
+
+  # Mark all configured interfaces as up, so wicked will keep them that way.
+  bash "wicked-ifup-all" do
+    action :run
+    code "wicked ifup all"
+    only_if { run_wicked_ifup }
   end
 
   # Avoid running the wicked related thing on SLE11 nodes


chef/cookbooks/network/recipes/default.rb
```diff
@@ -209,6 +209,16 @@ def get_datapath_id_for_ovsbridge(bridge)
     ifs[bond.name]["addresses"] ||= Array.new
     ifs[bond.name]["slaves"] = Array.new
     base_ifs.each do |i|
+      # If the interface is configured, wicked will helpfully re-add the IP
+      # when the interface is downed and then brought back up -- which is
+      # required, since interfaces which are up can not be added as slaves.
+      # To stop this happening, remove the configuration file and reload the
+      # interface.
+      if ::File.exist?("/etc/sysconfig/network/ifcfg-#{i.name}")
```
rhafer: Unfortunately we can't do it this way. As it means that we will remove the config file and do a wicked ifreload with every single chef-client run. (As the config file for the slave is being recreated further down in the recipe).

Unfortunately I don't have a good idea for a solution to the problem myself yet. It's a bit of a mess. I guess we need to have a discussion with the wicked guys.

s-t-e-v-e-n-k: Hmmm, maybe we check if the configuration file contains an IP address, and then perform this block? That would solve the issue we're seeing, and mean it's a no-op for subsequent chef-client runs.

s-t-e-v-e-n-k: Thinking about it further, this code block is only executed when creating a bond ...

chef/cookbooks/network/recipes/default.rb
```diff
@@ -209,6 +209,20 @@ def get_datapath_id_for_ovsbridge(bridge)
...
+      # To stop this happening, remove the configuration file and reload the
+      # interface.
+      slave_config = "/etc/sysconfig/network/ifcfg-#{i.name}"
+      if ::File.exist?(slave_config)
+        if ::Kernel.system("grep -q IPADDR #{slave_config}")
+          Chef::Log.info("Slave #{i.name} has a configuration file, removing")
+          ::File.unlink(slave_config)
+          ::Kernel.system("wicked ifreload #{i.name}")
+        end
+      end
```
rhafer: I've had a longer discussion with the wicked developer today. And he suggested to try a slightly different approach. This is what we came up with:

1. Change Nic::Bond.add_slave() (https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/barclamp/libraries/nic.rb#L557) to use "wicked ifdown <slave>" instead of calling "ip link set down". This would (tempoary) remove the interface from wicked`s greedy fingers ;).
2. Remove the ```slave.up``` call from Nic::Bond.add_slave() (this is actually not needed as the interface gets up automatically if the bond is up)
3. After rewriting the config files (around here: https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/network/recipes/default.rb#L568) trigger a ```wicked ifup <interface>``` for the changed interfaces.

For completenss: He, also mentioned that it is a bad idea in general to reconfigure stuff manually (i.e. behind wicked's back) and that it would be better to either use wicked or setup things by hand. But for achieving that, we'd need to rewrite the complete networking bits in crowbar, which is probably a bit too much currently :(

chef/cookbooks/network/recipes/default.rb
```diff
@@ -580,6 +589,14 @@ def get_datapath_id_for_ovsbridge(bridge)
       end
     end
   end
+  # If we downed a slave with wicked, bring it back under wicked's control.
+  slaves_downed_with_wicked.each do |i|
+    bash "wicked ifup #{i}" do
```
Itxaka: should be `i.name` here and in all the below uses of `i`

Itxaka: oh, wait, you are pushing i.name into the array, disregard that :D

chef/cookbooks/network/recipes/default.rb
```diff
@@ -580,6 +589,14 @@ def get_datapath_id_for_ovsbridge(bridge)
       end
     end
   end
+  # If we downed a slave with wicked, bring it back under wicked's control.
+  slaves_downed_with_wicked.each do |i|
```
rhafer: I wonder if we really need ```slaves_downed_with_wicked``` here. It might just be ok to call ```wicked ifup <name>``` for all interfaces that got their configuration updated?

chef/cookbooks/network/recipes/default.rb
```diff
@@ -566,6 +572,13 @@ def get_datapath_id_for_ovsbridge(bridge)
         pre_up_script: pre_up_script
       })
     end
+    # Mark all changed interfaces as up, so wicked will keep them that way.
+    bash "wicked ifup #{nic.name}" do
+      code "wicked ifup #{nic.name}"
+      action :nothing
+      subscribes :run, "template[/etc/sysconfig/network/ifcfg-#{nic.name}]", :immediately
+      only_if { nic_ifcfg.updated_by_last_action? }
```
rhafer: Why is this needed? The ```subscribes``` already makes sure that the resource is only executed when the template is changed.

rhafer:
@s-t-e-v-e-n-k I've pushed an additional change to your branch. It changes the code to just call "ifup all" (like you suggested). I tried to make sure that it is only called once after all config files have been written. With that I was able to get through the neutron deployment in a bonding setup with HA enabled.

Feel free to squash that commit into yours.

rhafer:
The current gate failures look rather unrelated to the patch set. I restarted the jobs as I assume that that was just a temporary glitch.

chef/cookbooks/network/recipes/default.rb
```diff
@@ -581,6 +589,30 @@ def get_datapath_id_for_ovsbridge(bridge)
...
 
+  run_wicked_ifup = false
+
+  # This, when notified by the above "ifcfg" templates, sets run_wicked_ifup
+  # to true (which was initialized to false in the compile phase).
+  # run_wicked_ifup is later used as an "only_if" guard for the
+  # "wicked ifup all" call that is needs to happen when any of the config
+  # files got updated. The purpose of doing it this way (instead of notifying
+  # the "wicked-ifup-all" resource directly), is to make sure that the
+  # ifup is only run once after all ifcfg file have been update and
```
vuntz: s/update/updated/

chef/cookbooks/network/recipes/default.rb
```diff
@@ -581,6 +589,30 @@ def get_datapath_id_for_ovsbridge(bridge)
...
+  # independent of how many of them were changed.
+  ruby_block "wicked-ifup-required" do
+    block do
+      run_wicked_ifup = true
+    end
+    action :nothing
+  end
+
+  # Mark all changed interfaces as up, so wicked will keep them that way.
```
vuntz: Strictly speaking, I guess the comment is wrong: we simply ask wicked to mark all *configured* interfaces as up.

chef/cookbooks/barclamp/libraries/nic.rb
```diff
@@ -566,7 +566,6 @@ def add_slave(slave)
       usurp(slave)
```
vuntz: I'm slightly confused here: with the `ifdown` call that is being done, don't we kill the `usurp` call here?

vuntz: (Maybe @rhafer can help on this one ;-))

rhafer: > I'm slightly confused here:

Me too,

> with the ifdown call that is being done, don't we kill the usurp call here?

I guess you're right. But I need to take a deeper look :(

s-t-e-v-e-n-k: usurp() removes addresses, routes and flushes the device, which might be required if some interface was configured manually and there is no wicked configuration for it -- however, I can test this with the usurp() call removed, and see what happens.

rhafer: > usurp() removes addresses, routes and flushes the device

Right, but more importantly it also assigns those addresses and routes to the parent interface (the bond).  Which I think is what @vuntz worried about. Because "wicked ifdown" is removing the interface, there is nothing that "ursurp" can assing to the bond. However is might not really be an issue.

@vuntz: The bond will get its addresses and routes assigned further down in the recipe anyway (https://github.com/s-t-e-v-e-n-k/crowbar-core/blob/006506f42bf009f324ea25aa4538bbbcfca0a91b/chef/cookbooks/network/recipes/default.rb#L436). In the case where the bond is newly created that would also be the time when the interface is put "up" for the first time. If the slave interface had any kind of cluster VIP assigned (which would mean there was  a cluster setup without bonding. Somewhat unlikely) that would also be the place where the VIP would be taken away from the bond.
I might be missing something but I can't currently think of a case where the usurp would add any IP to the bond that the later code would not add. Or am I missing your point?

>  which might be required if some interface was configured manually and there is no wicked configuration for it -- however, I can test this with the usurp() call removed, and see what happens.

I don't think this is needed currently.

vuntz:
> Or am I missing your point?

You are addressing it. Just wanted to make sure that someone had thought about it, and you did. So I'm happy :-)

s-t-e-v-e-n-k:
I'll be pushing this up on Monday with the comment fixes so it can run the CI gauntlet again.

PR #1241, opened 2017-06-05
[3.0] network: Set MTU for VLAN parent interface (bsc#1024279)

Itxaka:
If a physical interface has VLAN(s) configured with a custom MTU set but
the underlying physical network is not used directly and has no MTU in
its config, use the MTU from the VLAN interface for the parent device.
If there are multiple VLANs, use the highest MTU for the parent.

Also, if the physical parent from a VLAN interface is configured but the
MTU is lower than one of its VLAN interfaces, abort the chef-client run
and raise a useful error message.

(cherry picked from commit fc4152dd2aa903959e9f978b1fb48f51f7146420)

Backported as requested on https://github.com/crowbar/crowbar-core/pull/1230 by @vuntz

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -379,6 +379,37 @@ Nic.nics.each do |nic|
     nic.tx_offloading = node["network"]["enable_tx_offloading"] || false
   end
 
+  # Do some MTU checks/configuration for the parent of the vlan nic
+  if nic.is_a?(Nic::Vlan)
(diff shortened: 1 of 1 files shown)


Itxaka:
[P2: Bug 1032914 - #L3 top and other commands output hang when run via ssh on control node via netns](https://trello.com/c/izTfP3mT/256-p2%3A-bug-1032914---%23l3-top-and-other-commands-output-hang-when-run-via-ssh-on-control-node-via-netns)

vuntz:
@Itxaka mkcloud failed here, I guess this should be triggered again.

PR #1242, opened 2017-06-07
crowbar: Fix re-running chef after failure in apply_role for admin nodes

vuntz:
When a chef call fails in apply_role, we try a second time to run chef.
However, for admin nodes, we were not remembering the pid/host pair
correctly, resulting in the second chef call to be invalid (for no
host).

This resulted in the creation of /var/log/crowbar/chef-client/.log,
which is a long-standing bug.


sayalilunkad:
[P2: bsc#1030881 - #L3 Crowbar not updating new node run_list](https://trello.com/c/vP4LXrnN/247-p2-bsc1030881---l3-crowbar-not-updating-new-node-runlist)

vuntz:
I rebased after merging a conflicting patch, so need some +1 again (github automatically dismissed the previous reviews, nice)

sayalilunkad:
Let's merge this @vuntz?

PR #1244, opened 2017-06-15
network: Also remove optional keys during downgrade

s-t-e-v-e-n-k:
When downgrading the schema after applying xmit_hash_policy, we should
iterate over the conduits defined and remove team_miimon and
team_xmit_hash_policy if they exist.

I'm not 100% certain it is, but it addresses a bug in the original migration as it landed.

requests?**

### DIFF
--- a/chef/data_bags/crowbar/migrate/network/101_add_bonding_miimon_xmit_hash_policy.rb
+++ b/chef/data_bags/crowbar/migrate/network/101_add_bonding_miimon_xmit_hash_policy.rb
@@ -15,5 +15,9 @@ def downgrade(ta, td, a, d)
   unless ta["teaming"].key? "xmit_hash_policy"
     a["teaming"].delete "xmit_hash_policy"
   end
+  a["conduit_map"].each do |conduit|
+    a["conduit_map"][conduit].delete "team_miimon"
+    a["conduit_map"][conduit].delete "team_xmit_hash_policy"
+  end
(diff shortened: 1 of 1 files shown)


jsuchome:
s/download/downgrade

s-t-e-v-e-n-k:
Let me just wipe that egg off my face ...

chef/data_bags/crowbar/migrate/network/101_add_bonding_miimon_xmit_hash_policy.rb
```diff
@@ -15,5 +15,13 @@ def downgrade(ta, td, a, d)
   unless ta["teaming"].key? "xmit_hash_policy"
     a["teaming"].delete "xmit_hash_policy"
   end
+  a["conduit_map"].each do |conduit|
+    if a["conduit_map"][conduit].key? "team_miimon"
```
rhafer: Technically this is not needed. Deleteing a  non existing key from a hash in ruby doesn't yield an error. OTOH, it doesn't really harm either.

vuntz: I would actually still fix it, to ensure people don't copy & paste it.

jsuchome:
The HA failure is not related, I think it's safe to merge this anyway.

PR #1245, opened 2017-06-15
Fix confusing name of the method that does the opposite than claims

jsuchome:
See the usage of the method in the code:
```
    valid_nodes.reject! do |node|
      node_is_invalid_for_role(node.name, role)
    end
```

Note that crowbar-ceph is using the old name, so it needs to be fixed together with this.

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -367,7 +367,7 @@ class ServiceObject
     # from the nodes list in the case the new proposal would not be valid, so
     # nodes that can't be added at all would not be returned.
     nodes.reject! do |node|
-      node_is_valid_for_role(node, role.to_s)
+      node_is_invalid_for_role(node, role.to_s)
     end
 
     [200, nodes]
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/service_object.rb
```diff
@@ -448,6 +448,12 @@ def node_is_valid_for_role(node, role)
       violates_remotes_constraint?(elements, role)
   end
 
+  # Keeping for backward compatibility only, node_is_valid_for_role
+  # was the original (and confusing) name
+  def node_is_valid_for_role(node, role)
```
rhafer: I am not sure I agree with this. Having both methods ```node_is_valid_for_role``` and ```node_is_invalid_for_role``` present and both of them doing the same thing seems highly confusing to me. And actually I am not able to find a spot out side service_object.rb where ```node_is_valid_for_role``` is currently used.

jsuchome: It's used in my current https://github.com/crowbar/crowbar-ceph/pull/97 :-)

jsuchome: But I agree leaving the old name is confusing again

rhafer:
> Note that crowbar-ceph is using the old name, so it needs to be fixed together with this.

@jsuchome It seems this is no longer true, right? https://github.com/crowbar/crowbar-ceph/pull/99 seems to have remove the last usage for that method.

jsuchome:
True! Did not notice that...

PR #1247, opened 2017-06-16
Add support for SLES12SP3 repositories

jsuchome:


jsuchome:
I'm actually not sure about the values in `repos-cloud.yml` - where do I find correct fingerprints?

crowbar_framework/lib/crowbar/product.rb
```diff
@@ -29,7 +29,7 @@ def self.is_ses?
 
     def self.ses_platform
       # Returns the platform used for SES
-      "suse-12.2"
+      "suse-12.3"
```
dirkmueller: do we want to switch to ses5 right away? I would expect this to have further fallout..

jsuchome: Does SES4 work with SP3?

jsuchome:
Maybe I need to remove all those `suse-12.2` repos as well..

dirkmueller:
@jsuchome no this is just because gating required the crowbar PR to be merged first. thats why I suggested the automation workaround which would have solved that as well ;)

jsuchome:
[#cloud8 Check there's no cookbook code that is SP2-specific #sp3](https://trello.com/c/TqH64Qo0/28-cloud8-check-theres-no-cookbook-code-that-is-sp2-specific-sp3)

crowbar_framework/lib/crowbar/product.rb
dirkmueller: no, its sp2, it means mkcloud needs to deploy ceph nodes with sp2 then (I think we used to have code for that)

crowbar_framework/config/repos-cloud.yml
```diff
@@ -276,3 +276,272 @@ suse-12.2:
...
+      repomd:
+        tag: ["obsproduct://build.suse.de/SUSE:SLE-12-SP3:Update:Products:Cloud8/suse-openstack-cloud/8/cd/x86_64",
+              "obsproduct://build.suse.de/Devel:Cloud:8:Staging/suse-openstack-cloud/8/cd/x86_64",
+              "obsproduct://build.suse.de/Devel:Cloud:8/suse-openstack-cloud/8/cd/x86_64"]
+        fingerprint: []
+      url:
+      ask_on_error: false
+    suse-openstack-cloud-8-pool:
+      name: "SUSE-OpenStack-Cloud-8-Pool"
+      required: "mandatory"
```
dirkmueller: I think we need to make it optional until the scc/ibs setup is there..

jsuchome: Done

crowbar_framework/lib/crowbar/product.rb
jsuchome: I could revert this to "suse-12.2" and we can create separate card for SES deployment investigation, what do you think?

jsuchome:
Travis failure is fixed with https://github.com/crowbar/crowbar-core/pull/1248

crowbar_framework/lib/crowbar/product.rb
dirkmueller: as said: lets try with sp3 and see how far we get

dirkmueller:
it went far enough..

PR #1249, opened 2017-06-19
Remove SES5 references; support SES4 which is based on SP2.

jsuchome:
SUSE Enterprise Storage 5 (SES5) is only built for SP3, while
SES4 is only built for SP2.
Only SES4 is deployable with crowbar.

(We tried to switch everything to SP3 in https://github.com/crowbar/crowbar-core/pull/1247)

~Also this (or some extension of it) will be needed https://github.com/SUSE-Cloud/automation/pull/2002~

### DIFF
--- a/crowbar_framework/config/repos-cloud.yml
+++ b/crowbar_framework/config/repos-cloud.yml
@@ -347,26 +347,6 @@ suse-12.3:
       url:
       smt_path: "SUSE/Updates/SLE-HA/12-SP3/aarch64/update"
       ask_on_error: false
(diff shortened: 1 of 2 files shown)


jsuchome:
It's possible we're gonna need to remove all SES references completely, but it would be good to have something to have our CI working meanwhile.

aplanas:
This is intended for master or for stable/4.0?

jsuchome:
@aplanas that's indeed for master (that is normally running SP3, but we do not have SP3-enabled ceph product)

PR #1253, opened 2017-06-26
[stable/3.0] network: Also remove optional keys during downgrade

s-t-e-v-e-n-k:
When downgrading the schema after applying xmit_hash_policy, we should
iterate over the conduits defined and remove team_miimon and
team_xmit_hash_policy.

(cherry picked from commit 02ec0507069c2e23fc2aabe540d5e545f71966b5)

requests?**


Itxaka:
travis failure is due to the rails-observer issue, so this https://github.com/crowbar/crowbar-core/pull/1251 should probably be backported

PR #1258, opened 2017-06-28
openvswitch-switch package was merged with main package in SLES12SP3

jsuchome:
Look for "Merge openvswitch and openvswitch-switch into a single package since there
was no compelling reason to keep the switch functionality in a separate subpackage" in https://build.suse.de/package/rdiff/SUSE:SLE-12-SP3:GA/openvswitch?linkrev=base&rev=2

### DIFF
--- a/chef/cookbooks/network/attributes/default.rb
+++ b/chef/cookbooks/network/attributes/default.rb
@@ -20,9 +20,9 @@ when "suse"
     "vlan"
   ]
   default[:network][:ovs_pkgs] = [
-    "openvswitch",
-    "openvswitch-switch"
+    "openvswitch"
   ]
+  default[:network][:ovs_service] = "openvswitch"
   # non-x86_64 use the upstream kernel modules
   if node[:kernel][:machine] == "x86_64"
     # openSUSE and SLES12SP2 use the module shipped with upstream kernel
@@ -30,11 +30,12 @@ when "suse"
       default[:network][:ovs_pkgs].push("openvswitch-kmp-default")
     end
   end
-  # SLES11 uses a different service name for openvswitch
-  if node[:platform] == "suse" && node[:platform_version].to_f < 12.0
-    default[:network][:ovs_service] = "openvswitch-switch"
-  else
-    default[:network][:ovs_service] = "openvswitch"
+  if node[:platform] == "suse" && node[:platform_version].to_f < 12.3
+    default[:network][:ovs_pkgs].push "openvswitch-switch"
+    # SLES11 uses a different service name for openvswitch
+    if node[:platform_version].to_f < 12.0
+      default[:network][:ovs_service] = "openvswitch-switch"
+    end
   end
 when "rhel"
   default[:network][:base_pkgs] = [


chef/cookbooks/provisioner/files/default/crowbar_join.init.suse
```diff
@@ -4,7 +4,7 @@
 # Required-Start:    $syslog $network $remote_fs sshd
 # Should-Start:      neutron-ovs-cleanup xend libvirtd
 # Required-Stop:     $syslog $network $remote_fs sshd
-# Should-Stop:       openvswitch-switch ovs-usurp-config-br-fixed ovs-usurp-config-br-public
+# Should-Stop:       ovs-usurp-config-br-fixed ovs-usurp-config-br-public
```
bmwiedemann: actually, the 'Should' lines are optional and should not hurt when they reference missing services.

Also, the package was merged, but does that mean the service went away?

jsuchome: I think the service is gone as well, at least it looks that way on my SP3 node. But I can remove this line change and keep just the package one

bmwiedemann: rpm -qpl shows
openvswitch.service
ovs-vswitchd.service
ovsdb-server.service

so maybe we should add ovs-vswitchd too?

chef/cookbooks/network/attributes/default.rb
```diff
@@ -35,6 +35,7 @@
     default[:network][:ovs_service] = "openvswitch-switch"
   else
     default[:network][:ovs_service] = "openvswitch"
+    default[:network][:ovs_pkgs] = ["openvswitch"]
```
dirkmueller: I personally would prefer to move the "exception" into the "legacy" code branch. e.g. changing line 22 above to only have "openvswitch" and add an extra if for suse < 12.3 pushing the additional package (or just not worry about that part at all and just change the default globally).

jsuchome: ok

bmwiedemann:
actually, Jiri's link also shows
```
+Provides:       %{name}-switch = %{version}
```
so `zypper in openvswitch-switch` should still work.

chef/cookbooks/provisioner/files/default/crowbar_join.init.suse
jsuchome: Where, to `# Should-Stop`? `openvswitch` isn't there as well...

rhafer: @jsuchome  "openvswitch-switch" was renamed to "openvswitch" already with SLES12-SP1. But anyway this change seems to be unneeded AFAICS "crowbar_join.init.suse " is only used on SLES11 anyway.

jsuchome:
[#cloud8 Make OpenStack barclamps work for SP3](https://trello.com/c/DxMDobCg/33-cloud8-make-openstack-barclamps-work-for-sp3)

PR #1259, opened 2017-06-29
crowbar: use unversioned package name for postgresql

jsuchome:


jsuchome:
[#cloud8 Make core barclamps work for SP3](https://trello.com/c/6V3mm8jm/32-cloud8-make-core-barclamps-work-for-sp3)

PR #1260, opened 2017-06-30
ipmi: Add support for bmc_interface (bsc#1046567)

s-t-e-v-e-n-k:
Add support for specifying the bmc_interface type in the IPMI barclamp.

Drive-by converting tabs to spaces in the JSON template file.

Co-Authored-By: Eugen Block <e.block@suse.com>

requests?**

### DIFF
--- a/chef/data_bags/crowbar/migrate/ipmi/101_bmc_interface.rb
+++ b/chef/data_bags/crowbar/migrate/ipmi/101_bmc_interface.rb
@@ -1,0 +1,11 @@
+def upgrade(ta, td, a, d)
+  unless a.key? "bmc_interface"
+    a["bmc_interface"] = ta["bmc_interface"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a.delete "bmc_interface"
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-ipmi.json
+++ b/chef/data_bags/crowbar/template-ipmi.json
@@ -3,19 +3,20 @@
   "description": "The default proposal for the ipmi barclamp",
   "attributes": {
     "ipmi": {
-	  "bmc_enable" : false,
-	  "bmc_user": "root",
-	  "bmc_password": "cr0wBar!",
+      "bmc_enable" : false,
+      "bmc_user": "root",
+      "bmc_password": "cr0wBar!",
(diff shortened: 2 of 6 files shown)


chef/data_bags/crowbar/migrate/ipmi/101_bmc_interface.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(ta, td, a, d)
+  unless a["ipmi"].key? "bmc_interface"
+    a["ipmi"]["bmc_interface"] = ta["ipmi"]["bmc_interface"]
```
cmurphy: Are you sure this is right? I'm getting a migration error, it looks like it might be just `a.key? "bmc_interface"` and `a["bmc_interface"] = ta["bmc_interface"]`

chef/data_bags/crowbar/template-ipmi.json
```diff
@@ -3,19 +3,20 @@
   "description": "The default proposal for the ipmi barclamp",
   "attributes": {
     "ipmi": {
-	  "bmc_enable" : false,
-	  "bmc_user": "root",
-	  "bmc_password": "cr0wBar!",
+      "bmc_enable" : false,
+      "bmc_user": "root",
+      "bmc_password": "cr0wBar!",
+      "bmc_interface": "lan",
```
toabctl: The parameter should be used https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/node.rb#L982 here then. And "lanplus" as default would then be the way to go.

crowbar_framework/app/models/node.rb
```diff
@@ -979,7 +983,7 @@ def net_rpc_cmd(cmd)
 
   def bmc_cmd(cmd)
     if bmc_address.nil? || get_bmc_user.nil? || get_bmc_password.nil? ||
-        !system("ipmitool", "-I", "lanplus", "-H", bmc_address, "-U", get_bmc_user, "-P", get_bmc_password, cmd)
+        !system("ipmitool", "-I", bmc_interface, "-H", bmc_address, "-U", get_bmc_user, "-P", get_bmc_password, cmd)
```
toabctl: It must be ```get_bmc_interface```

PR #1263, opened 2017-07-07
[4.0] Unclaimed disks: improve handling of multipath devices (bsc#1031065)

Itxaka:
dm_multipath devices were being skipped while its paths were being included
in the array of unclaimed devices, allowing the DRBD recipe to claim them
and later fail while running pvcreate.

Introduce two new methods in Disk for checking whether the device (or its
holder) is named as "mpath-*".

This will make sure that when having a multipath device with several holders
we add to the unclaimed list only the multipath device and we ignore
the holders.

(cherry picked from commit dd2cde1b8a232e73c13035c994b6bdca7389034a)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,26 @@
 language: ruby
 sudo: false
 
-rvm:
-  - 2.1.0
+rvm: 2.1.0
 
-env:
-  - SKIP_CHECKS=yes
+env: SKIP_CHECKS=yes
+
+matrix:
+  include:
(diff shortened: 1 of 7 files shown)


Itxaka:
[P2: bsc#1031065 - L3-Question: crowbar - pacemaker deployment fails with drbd and multipath](https://trello.com/c/8N5tUpaF/219-p2-bsc1031065-l3-question-crowbar-pacemaker-deployment-fails-with-drbd-and-multipath)

PR #1266, opened 2017-07-13
Use threads to acquire/release locks in apply_role

Itxaka:
Stolen from https://github.com/crowbar/crowbar-core/pull/896/ to rebase and improve (if needed)

crowbar: Use a thread pool when acquiring/release chef-client locks.

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1203,31 +1203,15 @@ class ServiceObject
         node_attr_cache[node_name]["windows"] || node_attr_cache[node_name]["admin"]
       end
 
-      begin
-        apply_locks = []
-        # do not use a map to set apply_locks: if there's a failure, we need the
-        # variable to contain the locks that were acquired; with a map, the
-        # variable would be empty.
-        nodes_to_lock.each do |node|
-          apply_lock = Crowbar::Lock::SharedNonBlocking.new(
-            logger: Rails.logger,
-            path: "/var/chef/cache/pause-file.lock",
-            node: node,
-            owner: "apply_role-#{role.name}-#{inst}-#{Process.pid}",
-            reason: "apply_role(#{role.name}, #{inst}, #{in_queue}) pid #{Process.pid}"
-          ).acquire
-          apply_locks.push(apply_lock)
-        end
-      rescue Crowbar::Error::LockingFailure => e
-        Rails.logger.progress("Failed to apply role #{role.name}")
-        message = "Failed to apply the proposal: #{e.message}"
-        update_proposal_status(inst, "failed", message)
-        return [409, message] # 409 is 'Conflict'
-      end
+      owner = "apply_role-#{role.name}-#{inst}-#{Process.pid}"
+      reason = "apply_role(#{role.name}, #{inst}, #{in_queue}) pid #{Process.pid}"
+      apply_locks, errors = lock_and_wait_for_chef_clients(nodes_to_lock, owner, reason)
 
-      # Now that we've ensured no new intervallic runs can be started,
-      # wait for any which started before we paused the daemons.
-      wait_for_chef_daemons(applying_nodes)
+      unless errors.empty?
+        message = "Failed to apply the proposal:\n#{errors.values.join("\n")}"
+        update_proposal_status(inst, "failed", message)
+        return [409, message] # 409 is 'Conflict', which makes sense for locks
+      end
     end
 
     # By this point, no intervallic runs should be running, and no
@@ -1399,7 +1383,7 @@ class ServiceObject
     update_proposal_status(inst, "failed", message)
     [405, message]
   ensure
-    apply_locks.each(&:release) if apply_locks.any?
+    release_chef_locks(apply_locks)
     restore_to_ready(applying_nodes) if applying_nodes.any?
     process_queue unless in_queue
   end
@@ -1539,6 +1523,58 @@ class ServiceObject
   end
 
   private
+
+  def release_chef_locks(locks)
+    return if locks.empty?
+
+    locks.each_slice(20) do |lock_batch|
+      threads = []
+      lock_batch.each do |lock|
+        threads.push(Thread.new { lock.release })
+      end
+      logger.debug "Waiting for #{threads.length} unlock threads to finish..."
+      ThreadsWait.all_waits(threads)
+    end
+  end
+
+  def lock_and_wait_for_chef_clients(nodes, lock_owner, lock_reason)
+    locks = []
+    errors = {}
+    locks_mutex = Mutex.new
+    errors_mutex = Mutex.new
+
+    nodes.each_slice(20) do |node_batch|
+      threads = {}
+      node_batch.each do |node|
+        thread = Thread.new do
+          begin
+            lock = Crowbar::Lock::SharedNonBlocking.new(
+              logger: @logger,
+              path: "/var/chef/cache/pause-file.lock",
+              node: node,
+              owner: lock_owner,
+              reason: lock_reason
+            ).acquire
+            locks_mutex.synchronize { locks.push(lock) }
+            # Now that we've ensured no new intervallic runs can be started,
+            # wait for any which started before we paused the daemons.
+            wait_for_chef_clients(node, logger: true)
+          rescue Crowbar::Error::LockingFailure => e
+            errors_mutex.synchronize { errors[node] = e.message }
+          end
+        end
+        threads[thread] = node
+      end
+
+      # wait for all running threads and collect the ones with a non-zero return value
+      logger.debug "Waiting for #{threads.keys.length} lock threads to finish..."
+      ThreadsWait.all_waits(threads.keys)
+
+      break unless errors.empty?
+    end
+
+    [locks, errors]
+  end
 
   def wait_for_chef_clients(node_name, options = {})
     options = options.include?(:logger) ? { logger: Rails.logger } : {}
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1207,15 +1207,15 @@ class ServiceObject
       reason = "apply_role(#{role.name}, #{inst}, #{in_queue}) pid #{Process.pid}"
       apply_locks, errors = lock_nodes(nodes_to_lock, owner, reason)
 
-      # Now that we've ensured no new intervallic runs can be started,
-      # wait for any which started before we paused the daemons.
-      wait_for_chef_daemons(nodes_to_lock)
-
       unless errors.empty?
         message = "Failed to apply the proposal:\n#{errors.values.join("\n")}"
         update_proposal_status(inst, "failed", message)
         return [409, message] # 409 is 'Conflict', which makes sense for locks
       end
+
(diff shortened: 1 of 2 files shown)


Itxaka:
> Isn't it a problem? Looking at the thread, it might mean in an unfortunate case that we lock a node, then wait for it, and so on... And in the meanwhile another periodic run kicked off on the next node, so we need to wait for that as well.

I would say that this is not a problem with this approach, before this we would wait in one thread for all of them to finish any existing chef runs, and now we wait for each node on its own thread. So if one of the nodes chef run is stuck, we will get the same failure on both approaches, at the same time.

Just take into account that the waiting is done on the own thread for the node, not in the main thread as before, so the only thing that this can delay it's its own thread.

Plus, the locks will be acquired first, so it wont be possible for another chef run to be queued on that node.

Itxaka:
In flowchart form:

```
+-----------------+                                           +-----------------+
|                 |                                           |                 |
|     Before      |                                           |      After      |
|                 |                                           |                 |
+--------+--------+                                           +--------+--------+
         |                                                             |
         |                                                             |
         |                                                             |
         |                     Next node                               |
         v               +--------------------+                        v
     Main thread         |                    |                    Main thread
         +               |                    |                        +
         |               |                    |                        |
         |               v                    +                        |
         +------>  Foreach node +------->  Get lock                    +------>  Foreach node
                        +                                                           +
                        |                                                           |
                        |                                                           |
                        |                                                           |
                        |                                                           +------->  Thread +------>  Get lock  +------->  Wait for chef run to finish
                        |                                                           |
                        |                                                           |
                        v                                                           +------->  Thread +------>  Get lock  +------->  Wait for chef run to finish
        Wait for all nodes to finish chef run                                       |
                        +                                                           |
                        |                                                           +------->  Thread +------>  Get lock  +------->  Wait for chef run to finish
                        |                                                           |
                        v                                                           |
                Main thread waiting                                                 v
                                                                            Main thread waiting

```

So in any case, the main thread in both cases its gonna need to wait for **any** node to finish its chef-client run before continuing the execution of apply_role.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1203,31 +1203,15 @@ def apply_role(role, inst, in_queue, bootstrap = false)
...
-            reason: "apply_role(#{role.name}, #{inst}, #{in_queue}) pid #{Process.pid}"
-          ).acquire
-          apply_locks.push(apply_lock)
-        end
-      rescue Crowbar::Error::LockingFailure => e
-        Rails.logger.progress("Failed to apply role #{role.name}")
-        message = "Failed to apply the proposal: #{e.message}"
+      owner = "apply_role-#{role.name}-#{inst}-#{Process.pid}"
+      reason = "apply_role(#{role.name}, #{inst}, #{in_queue}) pid #{Process.pid}"
+      apply_locks, errors = lock_and_wait_for_chef_clients(nodes_to_lock, owner, reason)
```
abel-navarro: minor nitpick, I'd rather rename the variable `apply_locks` to `applied_locks`.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1540,6 +1524,79 @@ def wait_for_chef_daemons(node_list)
...
 
+  THREAD_POOL_SIZE = 20
+
+  def release_chef_locks(locks)
+    return if locks.empty?
+
+    queue = Queue.new
+    locks.each { |l| queue.push l }
+
+    workers = (1...THREAD_POOL_SIZE).map do
```
aplanas: I am sure that I am misunderstanding something. IIUC the original code crate batchs of 20 threads, but I can see that this iterate over THREAD_POOL_SIZE, without empty the queue (that can be bigger than 20). I do not see this also in the `lock_and_wait_for_chef_clients` method

crowbar_framework/app/models/service_object.rb
```diff
@@ -1540,6 +1524,79 @@ def wait_for_chef_daemons(node_list)
...
+    locks = []
+    errors = {}
+
+    return [locks, errors] if nodes.empty?
+
+    locks_mutex = Mutex.new
+    errors_mutex = Mutex.new
+
+    queue = Queue.new
+    nodes.each { |n| queue.push n }
```
abel-navarro: can you confirm this queue is thread safe? otherwise some conflicts can arise.

Itxaka: It better be, as it seems that its only purpouse its to sync between threads :)

http://ruby-doc.org/stdlib-2.0.0/libdoc/thread/rdoc/Queue.html

crowbar_framework/app/models/service_object.rb
Itxaka: @aplanas can you explain this a bit further? 

I kind of understand what you mean here but not sure, so i'll be better to clarify, but Im guessing its that the queue could be >20 but we only create 20 threads max, so in a case where we deploy to 25 nodes, 5 nodes will not get a lock and be stuck in the queue, which could be fixed by working in 20 threads batches but doing it `unless queue.empty?`

Am i right here?

crowbar_framework/app/models/service_object.rb
Itxaka: I think that the var name was kept for easy diff between the old changes and new, I dont mind changing it :)

crowbar_framework/app/models/service_object.rb
aplanas: Yes. My Ruby-fu is low but if I do not misinterpret the code, you are only attending the first batch of threads. You need to iterate until the queue (that if from the same size that the `locks`) is indeed empty.

Or you are iterating in some other place?

matelakat: Gents, my ruby knowledge is also not the best, I guess the statement:

```ruby
      loop do
        begin
          node = queue.pop(true)
        rescue ThreadError
          puts "ERROR"
          break
        end
```
Does the trick, so each thread tries to pop, and if the queue is empty, it raises an error. 
that's how the queue is emptied @aplanas Having said that it does not communicate the intent too well, and it's quite obscure. See this "magic":

```ruby
2.1.2 :001 > q = Queue.new
 => #<Thread::Queue:0x00000001194730>
2.1.2 :002 > q.push 1
 => #<Thread::Queue:0x00000001194730>
2.1.2 :003 > q.pop
 => 1
2.1.2 :004 > q.pop
fatal: No live threads left. Deadlock?
        from (irb):4:in `pop'
        from (irb):4
        from /home/matelakat/.rvm/rubies/ruby-2.1.2/bin/irb:11:in `<main>'
```

Now the interesting part comes when we try to do another pop after that:
```ruby
2.1.2 :005 > q.pop
```
It blocks the calling thread (as it would be expected)

I might not be well educated enough to find an explanation for this behaviour.

matelakat: We would need to use `(0..THREAD_POOL_SIZE)` otherwise we will only have a pool size of `THREAD_POOL_SIZE-1`

matelakat:
> So in any case, the main thread in both cases its gonna need to wait for any node to finish its chef-client run before continuing the execution of apply_role.

Nice diagram, but not correct

crowbar_framework/app/models/service_object.rb
nicolasbock: I think it's either:

~~~
irb(main):011:0> (0...3).map { |i| "#{i}" }
=> ["0", "1", "2"]
~~~

or

~~~
irb(main):012:0> (1..3).map { |i| "#{i}" }
=> ["1", "2", "3"]
~~~

where in this examples `THREAD_POOL_SIZE = 3`.

aplanas: Oh I am ruby-blind. I see now the inner loop.

Still I do not get why to do that. The locks are release into a thread, but the main thread is still blocked in the `.join()`, so there is not performance improvement.

Why not to collect that in the main thread then?

matelakat: Yeah, I missed the true parameter. On the other side the behaviour is still interesting on the blocking case...

matelakat: sorry, yes

Itxaka: > Still I do not get why to do that. The locks are release into a thread, but the main thread is still blocked in the .join(), so there is not performance improvement.
> 
> Why not to collect that in the main thread then?

@vuntz may now the intent of this, tbh in my tests there doesnt seem to be a performance improvement from this patch, Im guessing it may be helpful in certain conditions or its only a middle step to introduce further improvements down the line?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1540,6 +1524,79 @@ def wait_for_chef_daemons(node_list)
...
+    return if locks.empty?
+
+    queue = Queue.new
+    locks.each { |l| queue.push l }
+
+    workers = (1...THREAD_POOL_SIZE).map do
+      Thread.new do
+        loop do
+          begin
+            lock = queue.pop(true)
```
nicolasbock: Ah, sorry, I missed that @Itxaka already used the `true` here :smile:

Itxaka:
@aplanas @matelakat @nicolasbock After a quick chat with Mate it seems that he's rigth and this could lead to a decrease in the thread pool if we find a lot of chef clients running thus slowing down the whole thing.
I moved the wait for the chef clients outside where it used to be, so we can lock the nodes ASAP adn then we can wait for them to finish.

Tested locally, seems to work as expected.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1203,31 +1203,19 @@ def apply_role(role, inst, in_queue, bootstrap = false)
...
+      owner = "apply_role-#{role.name}-#{inst}-#{Process.pid}"
+      reason = "apply_role(#{role.name}, #{inst}, #{in_queue}) pid #{Process.pid}"
+      apply_locks, errors = lock_nodes(nodes_to_lock, owner, reason)
 
       # Now that we've ensured no new intervallic runs can be started,
       # wait for any which started before we paused the daemons.
-      wait_for_chef_daemons(applying_nodes)
+      wait_for_chef_daemons(nodes_to_lock)
+
+      unless errors.empty?
```
matelakat: Don't we need to fail before we do `wait_for_chef_daemons` ?

Itxaka: ooops!

crowbar_framework/app/models/service_object.rb
```diff
@@ -1203,31 +1203,19 @@ def apply_role(role, inst, in_queue, bootstrap = false)
...
+        message = "Failed to apply the proposal:\n#{errors.values.join("\n")}"
         update_proposal_status(inst, "failed", message)
-        return [409, message] # 409 is 'Conflict'
+        return [409, message] # 409 is 'Conflict', which makes sense for locks
       end
 
       # Now that we've ensured no new intervallic runs can be started,
       # wait for any which started before we paused the daemons.
-      wait_for_chef_daemons(applying_nodes)
+      wait_for_chef_daemons(nodes_to_lock)
```
matelakat: maybe this was a bug in the original implementation to use `applying_nodes`?

Itxaka: Seems like it indeed!

PR #910, opened 2016-12-19
Decrease amount of node saves

vuntz:
Saving is not cheap, and most of the time, it's not needed (as no attribute changed, or because we will already save later).

PR #1267, opened 2017-07-14
Decrease amount of node saves

Itxaka:
from https://github.com/crowbar/crowbar-core/pull/910 to rebase , test and improve

Saving is not cheap, and most of the time, it's not needed (as no attribute changed, or because we will already save later).


Itxaka:
I deleted the original branch for some reason :/ closing and opening a new one

PR #1269, opened 2017-07-18
crowbar: explicitly require the modules to avoid circular dependencies

Itxaka:
In order to avoid circular dependencies when using threading
we should explicitly require the appropriate modules that we
are gonna use.

### DIFF
--- a/crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
+++ b/crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
@@ -15,6 +15,8 @@
 # limitations under the License.
 #
 
+require "remote_node"
+
 module Crowbar
   class Lock::SharedNonBlocking < Lock
     # This class implements a shared (i.e. non-exclusive) lock, by


jsuchome:
Is it just `RemoteNode` than needs loading at this time?

toabctl:
I guess this fixes ```Failed to apply the proposal: uncaught exception (Circular dependency detected while autoloading constant RemoteNode)``` ?

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
```diff
@@ -16,6 +16,7 @@
 #
 
 module Crowbar
+  include RemoteNode
```
matelakat: I would add a comment to explain why this is needed, so people don't remove it blindly.

crowbar_framework/config/application.rb
```diff
@@ -19,8 +19,7 @@
 
 module Crowbar
   class Application < Rails::Application
-    # Explicitely eager load /lib/crowbar/lock so we can use SharedNonBlocking
-    # with threading without hitting circular dependencies
+
     config.eager_load_paths += Dir["#{config.root}/lib/crowbar/lock"]
```
dirkmueller: did you want to remove the actual eager loading as well instead of just the comment that explains why the line is there?

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
rsalevsky: Are you sure you mean `include` and not `require`? These are two really different things, include takes all methods from the `RemoteNode` and includes them into `Crowbar`. The `require` just uses them, which makes more sense to me.

crowbar_framework/config/application.rb
Itxaka: I did delete the comment on mistake, let me recover it!

crowbar_framework/lib/crowbar/lock/shared_non_blocking.rb
Itxaka: umm, good point, I think you are right here.

Itxaka:
`12:44:57 +(mkcloud-common.sh:102) complain(): printf 'Error (13): %s\n' 'Multi DNS server test failed!'`

:(

PR #1271, opened 2017-07-18
crowbar: do a graceful shutdown/reboot when invoked via crowbar

dirkmueller:
Currently there is a hard reboot executed which is interrupting
a currently pending chef-client run. Normally that shouldn't be
a problem but in the HA cluster case it can cause significant harm.


dirkmueller:
(untested)

crowbar_framework/app/models/node.rb
```diff
@@ -1058,7 +1058,7 @@ def set_state(state)
       if @node[:platform_family] == "windows"
         net_rpc_cmd(:power_cycle)
       else
-        ssh_cmd("/sbin/reboot")
+        ssh_cmd("/usr/sbin/crowbar_join --stop || /sbin/reboot")
```
nicolasbock: Why not use `systemd` here?

dirkmueller: how does systemd help with reboot? is there a reboot command with systemd?

PR #1272, opened 2017-07-20
Decrease the amount of node saves

Itxaka:
Saving is not cheap, and most of the time, it's not needed (as no attribute changed, or because we will already save later).

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -16,6 +16,8 @@
 
 return if node[:platform_family] == "windows"
 
+dirty = false
+
 # Make sure packages we need will be present
 node[:network][:base_pkgs].each do |pkg|
   p = package pkg do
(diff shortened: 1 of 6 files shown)


chef/cookbooks/provisioner/recipes/setup_base_images.rb
```diff
@@ -441,24 +446,24 @@
...
-    node[:provisioner][:repositories][os] ||= Mash.new
-    node[:provisioner][:repositories][os][arch] ||= Mash.new
+    repositories[os] ||= Mash.new
+    repositories[os][arch] = Mash.new
 
     if File.exist?("#{os_dir}/crowbar-extra") && File.directory?("#{os_dir}/crowbar-extra")
       Dir.foreach("#{os_dir}/crowbar-extra") do |f|
         next unless File.symlink? "#{os_dir}/crowbar-extra/#{f}"
-        node[:provisioner][:repositories][os][arch][f] ||= Hash.new
+        repositories[os][arch][f] = Hash.new
```
Itxaka: I rather ignore this as all of the other vars are initialized the same way. I prefer to do a cleanup PR afterwards that changes them all instead.

chef/cookbooks/network/recipes/default.rb
```diff
@@ -16,6 +16,8 @@
 
 return if node[:platform_family] == "windows"
 
+dirty = false
```
matelakat: I will leave the comments here, as github doesn't let me comment on areas that are not in the hunks for some reason:

These links should refer to lines in the code that do `node.set` and I think we are missing `dirty` there.
https://github.com/Itxaka/crowbar-core/blob/2dabd459013cb90ea3571637cb0538ae146d44b4/chef/cookbooks/network/recipes/default.rb#L153

Itxaka: that key will only be set if there is nothing in that key:
```
unless node["network"]["ovs_datapath_ids"][bridge]
```

Itxaka: In any case, we dont do a node.save in there so it wont be needed. The point of the PR is to avoid doing node.save, not to avoid saving attributes. When we do a node.save, supposedly its because we need that value to be in the node instantly, so other nodes may access it (or searches), so in this case there doesnt seem to be a need of making that value available anywhere else, so no node.save in here :)

PR #1274, opened 2017-07-21
use public-foo hostnames (bsc#1110080)

bmwiedemann:
to be covered by CA certificate wildcards

e.g. *.p3.cloud.suse.de
does not cover public.cluster-services.p3.cloud.suse.de.
but it does public-cluster-services.p3.cloud.suse.de.

To make it easier to use SSL with wildcard CA certificates

by not introducing a subdomain level, which is not covered by SSL cert wildcards

requests?**
with public- in names it is possible to collide with node aliases that also get written to DNS

### DIFF
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -200,8 +200,8 @@ nodes.each do |n|
       alias_name = alias_name_no_net
     else
       net_name = network.name.gsub("_","-")
-      base_name = "#{net_name}.#{base_name_no_net}"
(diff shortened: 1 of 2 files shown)


vmoravec:
The proposed change conflicts with the proposal for cluster domain naming scheme in our docs, see chapter 11.1 for SOC7

bmwiedemann:
So we would have to update the documentation accordingly when this gets merged.

bmwiedemann:
[doc bug](https://bugzilla.suse.com/show_bug.cgi?id=1110080) filed

bmwiedemann:
https://github.com/bmwiedemann/crowbar-core/tree/publicname also has it added to the commit message, but I cannot edit the PR to take it as source.

PR #1275, opened 2017-07-21
crowbar: require crowbar instead of eager load

Itxaka:
There seems to be some issue with using eager load
where some required parameters that should be filled when
deploying a barclamp are empty.

This could be a side issue of using eager load to load some
classes on application start. Lets use the simpler require
instead.


Itxaka:
This may not fix it, but it could have something to do with it.

stefannica:
the mkcoud job is still failing with the same error

PR #1276, opened 2017-07-21
[WIP] dns: add more logging to the barclamp

Itxaka:
Lets see where the slave ips is run and what it contains when we hit the multidns issue


Itxaka:
@nicolasbock yep! Useless PR but I cannot get it to reproduce it locally properly or under supervised runs so I hope it fails (Watch it passing all jenkins runs with solid green just to mock me)

PR #1281, opened 2017-07-25
crowbar: New step in apply_role to save the config to a databag

Itxaka:
Generic block to save the insecure config of a barclamp to the
databag after the chef run. Its expected to be override by the
different services if they need to store more items.

requires: https://github.com/crowbar/crowbar-openstack/pull/1087


Itxaka:
Actually, let me do this in a different way from the crowbar-openstack repo instead to avoid dependencies and possible failures due to it being in different places

PR #1282, opened 2017-07-25
crowbar_join: Add a report timing handler

Itxaka:
Adds a timing handler that runs during the report phase of
the chef-client to show the 10 slowest resources during the chef run

Ideally the handlers should be moved out of the crowbar_join.sh
file and turned into a chef resource for a decend deploy
mechanism and the ability to configure them, but as a first step
I think its ok to include them like this (for the moment, maybe a PR will come next to fix that)

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
+++ b/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
@@ -238,6 +238,44 @@ EOF
         echo "$reboot_handler_reset_content" > /var/chef/handlers/reboot_handler_reset.rb
     fi
 
+    # create resource timing handler
+    read -d '' resource_timing_handler_content <<"EOF"
+class TimingHandler < Chef::Handler
+  def initialize(items=10)
+    super()
+    @items = items
+  end
+  def report
+    Chef::Log.info "Showing the #{@items} slowest resources"
+    all_resources.sort_by! {|r| r.elapsed_time }.reverse!
+    all_resources[0...@items].each do |r|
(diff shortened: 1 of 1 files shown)


Itxaka:
@matelakat we were talking about something like this yesterday I believe!

Itxaka:
@matelakat I think it only runs on install time.

The part of where that is deployed and it gets run its still a bit fuzzy to me, so instead of taking longer to do the investigation and extract the handlers and deploy them more properly, I just did a quick hack job to add this, mostly as a POC for what we talked about in a meeting on how to track what resources were taking time.

nicolasbock:
`crowbar_join` is run at boot.

matelakat:
It's fine as a POC, but I rather not merge it, unless we feature-swich it with some flag. I don't want to see experimental code changing any production environment.

nicolasbock:
I like @Itxaka suggestion to properly move this feature into a resource. Then we should also investigate whether the restart handlers that are currently patched into crowbar_join can't be integrated into a resource as well so that we can remove all of this crazy patching entirely.

Itxaka:
agree with @nicolasbock here, we should dedicate some time to this to move it to a proper resource (if we can), that way we can also choose to deploy it or not via a flag in the provisioner(?) barclamp.

nicolasbock:
Yes, without any serious kind of code reading, the provisioner barclamp sounds like the right place :)

Itxaka:
Not needed for the moment so closing

jsuchome:
Do you want this just as a help for some current debugging? Or should it land in the production code? What about that Mate's original idea, of including it only when some flag is present?

Itxaka:
served its purpose to test timings

PR #1284, opened 2017-07-31
service_object: Fix error logging (bsc#1051436)

rsalevsky:
The syntax was just wrong, we have to join the backtrace frist and then
print it.


rsalevsky:
Fixed hound.

PR #1285, opened 2017-08-01
Allow setting NVMe drive in ceph (bsc #1051298)

sayalilunkad:
NVMe device does not match any of the conditions to be a claimed disk
hence always unclaimed. Adding the pattern to the regex makes it claimable.


sayalilunkad:
[bsc#1051298 - Ceph barclamp cannot set NVMe drive as journal](https://trello.com/c/9tCCrC22/355-bsc1051298-ceph-barclamp-cannot-set-nvme-drive-as-journal)

PR #1287, opened 2017-08-01
testing: use knife to check syntax of cookbooks

Itxaka:
Use the installed knife command to check for cookbook
syntax. This will catch ruby issues on the cookbooks and also
template issues that we are not checking right now.

### DIFF
--- a/.knife-test.rb
+++ b/.knife-test.rb
@@ -1,0 +1,5 @@
+# knife config file for travis-ci jobs that require knife
+# set a dir for the cache in the HOME dir or it will try to write to /var/chef
+cache_type "BasicFile"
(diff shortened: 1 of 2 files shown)


Itxaka:
This is expected to fail the travis run until https://github.com/crowbar/crowbar-core/pull/1286 is merged and this rebased.

Good test for the validity of this change :D

Itxaka:
mmmh, travis got confused and did not test the proper sha for the change... github issues like yesterday??

Itxaka:
Unfortunately it seems that knife requires some dirs to be writable for the cache, so Im gonna try a couple more things to see if we can go over that properly :(

Itxaka:
Now it should work

PR #1292, opened 2017-08-03
Fix network_validator (bsc#1025642)

aplanas:
The previous network validator take the first private IP from the
system and expect that this is the admin IP.  This work OK in case
that eth0 is listed before eth1, but this is something that do not
happends always.

This change use the FQND information to deduce the admin IP, as
install-suse-cloud is doing.

### DIFF
--- a/crowbar_framework/lib/crowbar/validator/network_validator.rb
+++ b/crowbar_framework/lib/crowbar/validator/network_validator.rb
@@ -4,7 +4,7 @@ module Crowbar
       def validate_network
         `#{Rails.root.join("..", "bin/network-json-validator").to_s} \
           --admin-ip \
-          #{Socket.ip_address_list.detect(&:ipv4_private?).ip_address} \
+          #{IPSocket.getaddress(Socket.gethostname)} \
           #{network_json.to_s}`
       end
 


aplanas:
[P2: Bug 1025642 - #L3 network.json is invalid - Admin node ip validation fails during installation](https://trello.com/c/DCEFM2F7/389-p2-bug-1025642-l3-networkjson-is-invalid-admin-node-ip-validation-fails-during-installation)

crowbar_framework/lib/crowbar/validator/network_validator.rb
```diff
@@ -4,7 +4,7 @@ class NetworkValidator
       def validate_network
         `#{Rails.root.join("..", "bin/network-json-validator").to_s} \
           --admin-ip \
-          #{Socket.ip_address_list.detect(&:ipv4_private?).ip_address} \
+          #{IPSocket.getaddress(Socket.gethostname)} \
```
nicolasbock: How does it find the IP of the hostname? If the host has several IP addresses which one is found?

aplanas: `Socket.gethostname` returns the FQDN of the, that is in `/etc/hostname`, and the `IPSocket.getaddress` will return the IP defined for this host in `/etc/hosts`. And AFAIK in this file you cannot specify multiple IP address for the same host, so there is only one.

The goal here is to find the IP of the admin machine, and this is where this code is running

Itxaka: Testing this locally, if the hostname is missing, how it will try to resolve it? Because I dont have my hostname on /etc/hosts and it still resolves it to my VPN IP :D

aplanas: @Itxaka true. It can get it from DNS and/or NIS too.

nicolasbock: That makes sense @aplanas, thanks!

aplanas:
@rhafer fixed

@rhafer @nicolasbock @Itxaka @sayalilunkad can you re-review? The commit message change dismissed the previous reviews

crowbar_framework/lib/crowbar/validator/network_validator.rb
rhafer: > Socket.gethostname returns the FQDN from /etc/hostname

Sorry for nitpicking here, but technically gethostname contains what ever was configured as the hostname (using hostnamectl or hostname <hostname>). It doesn't read /etc/hostname. And in many cases this is just the unqualified hostname  (without the domain).
IPSocket.getaddress will then use the resolver (DNS and/or /etc/hosts or whatever is configured in nsswitch) to figure out the FQDN and the corresponding IP address.</nitpick>

aplanas: @rhafer I see. Thanks for fixing this : ))

nicolasbock: Ah, thanks @rhafer for the explanation! That makes a lot more sense :smile:

PR #1294, opened 2017-08-15
Fix network_validator (bsc#1025642)

aplanas:
The previous network validator take the first private IP from the
system and expect that this is the admin IP.  This work OK in case
that eth0 is listed before eth1, but this is something that do not
happends always.

This change use the FQDN information to deduce the admin IP, as
install-suse-cloud is doing.

(cherry picked from commit c907a905767638282d7375ede271d9b5dc689b6b)

### DIFF
--- a/crowbar_framework/lib/crowbar/validator/network_validator.rb
+++ b/crowbar_framework/lib/crowbar/validator/network_validator.rb
@@ -4,7 +4,7 @@ module Crowbar
       def validate_network
         `#{Rails.root.join("..", "bin/network-json-validator").to_s} \
           --admin-ip \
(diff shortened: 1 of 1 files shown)


aplanas:
[P2: Bug 1025642 - #L3 network.json is invalid - Admin node ip validation fails during installation](https://trello.com/c/DCEFM2F7/389-p2-bug-1025642-l3-networkjson-is-invalid-admin-node-ip-validation-fails-during-installation)

Itxaka:
@aplanas you migth need to rebase this once https://github.com/crowbar/crowbar-core/pull/1310 is merged

aplanas:
Rebased

PR #1295, opened 2017-08-16
Fix chef node attribute lookup

dirkmueller:
it seems .fetch on symbols doesn't work for Chef attributes, and the
behavior is subtle different for "def []" instead. so with other words:

  node.fetch(:crowbar_wall) -> IndexError: Key crowbar_wall does not
  exist

  node[:crowbar_wall] -> works
  node["crowbar_wall] -> works
  node.fetch("crowbar_wall") -> works

This seemingly harmless HoundCI fix entirely broke the boot device
and dns export logic of Cloud 7+.

requests?**

### DIFF
--- a/chef/cookbooks/ipmi/recipes/ipmi-discover.rb
+++ b/chef/cookbooks/ipmi/recipes/ipmi-discover.rb
@@ -35,10 +35,10 @@ unsupported = [
   "Unknown (Not Detected)"
 ]
 
-node.set["crowbar_wall"] = {} unless node["crowbar_wall"]
-node.set["crowbar_wall"]["ipmi"] = {} unless node["crowbar_wall"]["ipmi"]
-node.set["crowbar_wall"]["status"] = {} unless node["crowbar_wall"]["status"]
-node.set["crowbar_wall"]["status"]["ipmi"] = {} unless node["crowbar_wall"]["status"]["ipmi"]
+node.set["crowbar_wall"] ||= Mash.new
+node.set["crowbar_wall"]["ipmi"] ||= Mash.new
+node.set["crowbar_wall"]["status"] ||= Mash.new
+node.set["crowbar_wall"]["status"]["ipmi"] ||= Mash.new
 node.save
 
 if node[:platform_family] == "windows"
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -262,7 +262,7 @@ filename = \"discovery/x86_64/bios/pxelinux.0\";
       os = node[:provisioner][:default_os]
     end
 
-    boot_device = mnode.fetch(:crowbar_wall, {})[:boot_device]
+    boot_device = mnode.fetch("crowbar_wall", {})[:boot_device]
 
     append << node[:provisioner][:available_oses][os][arch][:append_line]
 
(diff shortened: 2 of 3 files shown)


toabctl:
Would be nice to set crowbar_wall always to a Mash. I think this is not enough.

Itxaka:
@toabctl I dont think it matters here. The problem is that we are using `fetch` on a `NodeObject` which doesn't seem to work with symbols no matter what:

```
chef > node[:test] = Mash.new
 => {} 
chef > node.save
 => <Chef::Node:0x1dffbdc @name="crowbar.vo9.cloud.suse.de"> 
chef > node.fetch(:test)
IndexError: Key test does not exist
	from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/node/attribute.rb:199:in `fetch'
	from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/node.rb:346:in `method_missing'
	from (irb):8
	from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/shef.rb:73:in `block in start'
	from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/shef.rb:72:in `catch'
	from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/shef.rb:72:in `start'
	from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/bin/shef:34:in `<top (required)>'
	from /usr/bin/shef:23:in `load'
	from /usr/bin/shef:23:in `<main>'
chef > node.fetch("test")
 => {} 
```

toabctl:
@Itxaka the problem is that we use in some places symbols and in others strings. If we make sure crowbar_wall is always a Mash, than that does not matter anymore

Itxaka:
@toabctl indeed, fetch does not seem decent enough to be used on our chef version.

We could always modify our chef fork to convert the keys like they do on newer chef versions, in which they force all getters and setters to run over a `convert_key` method which turns them into strings if they are symbols, with a simple patch:

```
diff --git a/chef/lib/chef/node/attribute.rb b/chef/lib/chef/node/attribute.rb
index 558163b90..7d7abcee1 100644
--- a/chef/lib/chef/node/attribute.rb
+++ b/chef/lib/chef/node/attribute.rb
@@ -184,6 +184,7 @@ class Chef
       end
 
       def fetch(key, default_value=nil, &block)
+	key = key.kind_of?(Symbol) ? key.to_s : key
         if get_keys.include? key
           determine_value(
             get_value(automatic, key),
```

This makes the fetch work as expected:

```
chef > node.fetch(:crowbar_wall)
 => {"network"=>{"interfaces"=>{"eth0"=>{"addresses"=>["192.168.237.10/24"], "type"=>"physical", "mtu"=>1500, "gateway"=>"192.168.237.1"}}, "nets"=>{"admin"=>["eth0"]}, "addrs"=>{"admin"=>["192.168.237.10/24"]}}} 

```

toabctl:
@Itxaka I like that idea! And did it at https://build.suse.de/request/show/138074 for the SOC7 chef code.

toabctl:
@dirkmueller beside the rubygem-chef fix, it would be nice to also fix https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/node.rb#L60 (and all others in that file) and also the other files in crowbar_framework/app/models/

dirkmueller:
@toabctl crowbar_framework is rails not chef, Mash isn't available there..

toabctl:
Backport is https://github.com/crowbar/crowbar-core/pull/1299

PR #1296, opened 2017-08-16
Avoid crashing chef on listing "installing" nodes (bsc#1050278)

dirkmueller:
Apparently there is a way to get nodes in a state without the [:network]
attribute being set. it shouldn't harm to just continue as if
nothing would have happened.

requests?**

### DIFF
--- a/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+++ b/chef/cookbooks/barclamp/libraries/barclamp_library.rb
@@ -35,7 +35,8 @@ module BarclampLibrary
       # returns a full network definition, including ranges; this doesn't
       # depend on the node being enabled for this network
       def self.get_network_definition(node, type)
-        if node[:network][:networks].nil? || !node[:network][:networks].key?(type)
+        if node[:network].nil? || node[:network][:networks].nil? ||
+            !node[:network][:networks].key?(type)
           nil
         else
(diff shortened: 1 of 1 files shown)


cmurphy:
Is this meant to close #1050278 entirely? Seems like there would be a deeper problem that caused the node object to get into a state like that.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -45,7 +45,7 @@ def self.get_network_definition(node, type)
       def self.list_networks(node)
         answer = []
         unless node[:crowbar].nil? || node[:crowbar][:network].nil? ||
-            node[:network][:networks].nil?
+            node[:network].nil? || node[:network][:networks].nil?
```
toabctl: Same needs to be done for the function ```def self.get_network_by_type(node, type)``` (line 61)

dirkmueller: done

dirkmueller:
@cmurphy I believe because of repeated forget/reallocate cycles the node ended up being removed from network barclamp but not added back. don't know exactly why yet.

toabctl:
Backport is https://github.com/crowbar/crowbar-core/pull/1297

PR #1300, opened 2017-08-17
Bootdisk detection: Skip cciss, prefer wwn's

dirkmueller:
On Systems with HP Smart Array Controllers, the discovery image
detects cciss udev links but those don't actually exist during
installation because for unknown reasons the installation system
doesn't actually load the cciss module. So skip them and prefer
wwn's, those are hopefully unique enough.

Funny side story: In at least one case, the cciss- link was based on the
controller's serial number (which is the same for both devices that
were attached). you just had one job..

requests?**


toabctl:
Backport is https://github.com/crowbar/crowbar-core/pull/1304

PR #1302, opened 2017-08-17
Set up Danger for commit linting

cmurphy:
Currently, we have no automated way of enforcing or encouraging good patch proposal practices, and it is therefore left up to the reviewer to notice and request changes for improper commit formatting and pull request structure. This is a problem because we have not actually formally agreed to rules around pull request structure, and the importance of enforcing it is subjective to each reviewer. This patch sets up a basic commit linter using Danger[1]. For now, the checks it performs are limited to the default commit formatting checks from the commit-lint plugin[2] as well as a check for the existence of merge commits in a pull request. For now, the checks are set to "warn" and will post comments but will not fail a build.

The bot will use the github account susecloudcommitpolice. Using this gem in TravisCI requires the bot to have an API token set as an environment variable in the TravisCI settings for the repository. The token (which in most workflows is a secret akin to a password) will be exposed in the TravisCI logs, which is required to enable the bot to post comments on pull requests. The bot will not be part of any SUSE-related organizations and does not have permission to do anything beyond posting comments on public repositories, so the only venue for abuse is by spamming, and if that occurs we can disable the bot.

[1] http://danger.systems/ruby/
[2] http://danger.systems/plugins/commit_lint.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ env: SKIP_CHECKS=yes
 
 matrix:
   include:
+    - gemfile: Gemfile
+      script:
+      - bundle exec danger
     - gemfile: crowbar_framework/Gemfile
       script:
        - cd crowbar_framework
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,0 +1,7 @@
+# Check basic commit message formatting
+commit_lint.check warn: :all
+
+# Ensure a clean commit history
+if git.commits.any? { |c| c.message =~ /^Merge branch/ }
+  warn('Please rebase to get rid of the merge commits in this PR')
+end
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,7 @@ group :development do
   gem "sprockets-standalone", "~> 1.2.1"
   gem "sprockets", "~> 2.11.0"
   gem "rspec", "~> 3.1.0"
+  gem "danger-commit_lint"
 end
 
 unless ENV["PACKAGING"] && ENV["PACKAGING"] == "yes"


cmurphy:
With permission I can add the required environment variable to the travis settings.

rsalevsky:
I like the idea but can we please use the existing susecloudapi user? Adding more and more bot users is not very helpful and we will lose track who owns them.

aspiers:
> Adding more and more bot users is not very helpful

Actually there is some benefit - it means that each bot codebase only gets the absolute minimum GitHub permissions it needs.

> and we will lose track who owns them.

Isn't that what https://externaltools.suse.de/ is for?

cmurphy:
I'm happy either way. We can create a new API token for the susecloudapi user so the token itself doesn't have more permissions than it needs. Would need the owner of that user to set that up in travis (http://danger.systems/guides/getting_started.html#setting-up-danger-to-run-on-your-ci).

Itxaka:
Im trying to test this and danger doesn't seem to run locally, insisting on getting `DANGER_GITHUB_API_TOKEN` even when run with `local`

Is it possible to run locally? I mean, that way we can use git hooks so its easier to fix before pushing

cmurphy:
@Itxaka it is possible to run locally (http://danger.systems/guides/troubleshooting.html#i-want-to-work-locally-on-my-dangerfile) but it's not useful in the way you're hoping, it's really meant to be run against pull requests.

Itxaka:
ahh, that sucks. Isnt there any other danger-like service that also allows local running? I dont want to run on the same issue as with hound, where is really difficult to check changes locally so I have to keep pushing commits to check their status :(

cmurphy:
> Isnt there any other danger-like service that also allows local running?

We could look into something like https://jorisroovers.github.io/gitlint/ or https://github.com/m1foley/fit-commit which seem better suited for local runs. The advantage I see with danger is it has TravisCI integration built into it so we don't have to think about it too hard. With the others we'd have to put in more effort to building a bot that makes comments, or we'd have to make it fail the travis build to force people to look at the CI results.

Glancing at the code it sort of looks like running on local branches should be possible, so maybe it's just not well documented. I can have a deeper look.

> I dont want to run on the same issue as with hound, where is really difficult to check changes locally so I have to keep pushing commits to check their status :(

Personally I never set up hound/rubocop locally because hound is so fast I don't have to wait for its feedback, and I don't really mind that my mistakes are public ;) but I get that not everyone has the same preferences.

aspiers:
> please send a mail to cloud-devel explaining how this is useful and why it is good and have a discusison there outside patch review.
>
> In general I think its a great idea, not trying to dismiss it , but lets widen the audience first.

+1 for widening the audience. I think there is still maybe some value in doing some preliminary research first, including a PoC (to avoid wasting everyone's time proposing something you don't properly understand, or which later proves to be unworkable), but I'm also fine with telling everyone about it now.

aspiers:
> Personally I never set up hound/rubocop locally because hound is so fast I don't have to wait for its feedback, and I don't really mind that my mistakes are public ;) but I get that not everyone has the same preferences.

Yeah, I actually dislike the public noise from Hound which could be avoided by running rubocop locally - I think it wastes people's time to a degree, *especially* the time of the PR submitter, due to the increased latency of pushing to GitHub and then waiting for Hound to report errors, vs. just immediately running it locally.

Itxaka:
> Yeah, I actually dislike the public noise from Hound which could be avoided by running rubocop locally - I think it wastes people's time to a degree, especially the time of the PR submitter, due to the increased latency of pushing to GitHub and then waiting for Hound to report errors, vs. just immediately running it locally.

Plus dont get me started when hound goes crazy and starts reviewing code that is not there anymore or refuses to clean up comments that are already addresses ¬_¬

rsalevsky:
@aspiers externaltools.suse.de only covers the SUSE org in the crowbar org we can do what we want. At the moment we have three api accounts and I only know the owner of one of them. 

As far as I know you can specify the access rights per token.

aspiers:
> As far as I know you can specify the access rights per token.

@rsalevsky OK thanks, if that's true then I agree we can consolidate onto one bot.

cmurphy:
We didn't reach a super solid "YES" on the mailing list thread but I think we can agree that it doesn't hurt to try it out. This is ready to go as-is, but we'll need the owner of the susecloudapi user to create a new API token and add it to the travis settings, and I don't know who that owner is.

aspiers:
@dirkmueller Presumably you are OK to dismiss your review now?

cmurphy:
@rsalevsky: @jdsn advised that it would be better to use a separate user since we've started hitting the rate limits with the susecloudapi user, so would it be okay to go ahead with susecloudcommitpolice user?

rsalevsky:
@cmurphy sure, but please document who owns this user and who as access

cmurphy:
@rsalevsky do you have a recommendation on where that should be documented?

rsalevsky:
@cmurphy I think so far we don't have any documentation regarding this topic; I would suggest to check with @jdsn.

aspiers:
@cmurphy I have identified some further checks which we could benefit from implementing in Danger:

- [ ] prohibit multiple commits within a PR with identical commit messages
- [ ] once [openSUSE/obs-service-tar_scm#207 (auto-append bug references from commit message bodies into generated changes)](https://github.com/openSUSE/obs-service-tar_scm/issues/207) is implemented, prohibit suffixes like "` (bsc#1234567)`" on the first line of commit messages
- [ ] ensure that any references to external bug/feature trackers include a full URL
- [ ] [require all commit messages to begin with a "`component: `" prefix in repos where this is appropriate](http://mailman.suse.de/mlarch/SuSE/cloud-devel/2017/cloud-devel.2017.09/msg00035.html)
- [ ] any PR with labels like `do not merge` / `WIP` / `needs work` extra should fail the Danger check and hence not be mergeable.

Of course these can and probably should be done as follow-up PRs.  I don't want to delay the merging of this any longer than absolutely necessary, otherwise this experiment will drag on too long.

aspiers:
> @rsalevsky do you have a recommendation on where that should be documented?

@cmurphy Under [the usual place for developer docs](https://github.com/SUSE/cloud/tree/master/docs).

cmurphy:
Documentation proposed here: https://github.com/SUSE/cloud/pull/182

I'm going to go ahead and merge this and then we can start thinking about adding new checks and adding this to other repositories

PR #1303, opened 2017-08-18
provisioner: Fix wwn usage for persistent disk names

toabctl:
Commit 5697817479c0a switched to use wwn if possible but forgot to change
one place which is now fixed with this commit.

Co-Authored-By: Dirk Mueller <dmueller@suse.com>

requests?**


sayalilunkad:
@toabctl dirk has already made this change here https://github.com/crowbar/crowbar-core/pull/1300

toabctl:
Change already proposed by Dirk: https://github.com/crowbar/crowbar-core/pull/1300

PR #1305, opened 2017-08-22
provisioner: Install crowbarctl

aplanas:
Update autoyast.xml.erb to install crowbarctl in each node.  This
is the first step to replace `curl` raw calls to the crowbar API,
and use `crowbarctl` instead.

This replacement will fix some issues in relation with `curl`,
https, redirections and BasicAuth passwords.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/base.rb
+++ b/chef/cookbooks/provisioner/recipes/base.rb
@@ -413,3 +413,21 @@ template "/etc/sh.shrc.local" do
   group "root"
   mode "0644"
 end
+
+crowbar_node = node_search_with_cache("roles:crowbar").first
+address = crowbar_node["crowbar"]["network"]["admin"]["address"]
+protocol = crowbar_node["crowbar"]["apache"]["ssl"] ? "https" : "http"
+server = "#{protocol}://#{address}"
+password = crowbar_node["crowbar"]["users"]["crowbar"]["password"]
+verify_ssl = !crowbar_node["crowbar"]["apache"]["insecure"]
+template "/etc/crowbarrc" do
+  source "crowbarrc.erb"
+  variables(
+    server: server,
+    password: password,
+    verify_ssl: verify_ssl
+  )
+  owner "root"
+  group "root"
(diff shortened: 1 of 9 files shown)


aplanas:
This needs the configuration file

cmurphy:
Added the config file

cmurphy:
@nicolasbock it works because it uses a default server value of 'http://127.0.0.1:80'.

nicolasbock:
Thanks @cmurphy, that makes sense.

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -409,3 +409,21 @@
...
+template "/etc/crowbarrc" do
+  source "crowbarrc.erb"
+  variables(
+    server: server,
+    password: password,
+    verify_ssl: verify_ssl
+  )
+  owner "root"
+  group "root"
+  mode "0o644"
```
vuntz: Please don't make it world-readable as it contains a password.

cmurphy: Wow I can't believe I did that

cmurphy: fixed

cmurphy:
@vuntz I fixed the permissions error, could you rereview?

cmurphy:
re-pushed to retrigger the ha job that had been deleted since I last looked at this

chef/cookbooks/provisioner/templates/default/crowbarrc.erb
```diff
@@ -0,0 +1,7 @@
+[default]
+server = <%= @server %>
+username = crowbar
```
jsuchome: Does this **have to be** `crowbar`?

cmurphy: I guess not, that could be changed here https://github.com/crowbar/crowbar-core/blob/master/chef/data_bags/crowbar/template-crowbar.json#L21

So this should probably be smarter.

PR #1306, opened 2017-08-22
Add Cloud8/Pike repository tags

stefannica:
This change fixes the current pikecloud8 crowbar allocation failures
Adds missing pike repository tags.
requests?**


stefannica:
@toabctl I updated the commit message, thanks

PR #1308, opened 2017-08-23
Create event hook infrastructure

scottwulf:
Rebase of PR #952 but only including the new infrastructure class

For usage examples, see:
https://github.com/crowbar/crowbar-core/pull/952
https://github.com/crowbar/crowbar-ha/pull/171
https://github.com/crowbar/crowbar-openstack/pull/717

There are several cases where we have events that should trigger some
activity in some other part of Crowbar. For instance:

 - when the public name of a node is saved, it may impact the endpoint
   of an OpenStack service
 - when the public name of the VIP of haproxy is changed, it impacts the
   endpoint of OpenStack services
 - when the keystone proposal is applied, we may want to reapply all
   proposals that depend on keystone

What we need for this is the ability to notify about the events in the
rails application and then dispatch the notifications to hooks which
listen for them and decide if some action should be triggered.

The main reason we didn't have this in the past is that we likely don't
want to do that in the foreground of the rails application. But now that
we have delayed_job, we can send the notifications and run the hooks in
the background.

In this commit, we add the simple infrastructure about notifications and
hooks:

 - the events are defined with a name and a hash that contains the
   details of the event. The structure of the hash depends on the event.

 - a simple dispatcher exists that simply connects the hooks to the
   events.

 - the hooks only exist for service objects for the time being; a
   service object simply needs to have a event_hook method to register
   the hook, and will need to filter for the events it cares about. The
   signature of event_hook is as follows:
     def event_hook(role, event, details)

It could be argued that the hooks should be registered for some specific
events (hence moving the filter to the event dispatcher), but it's not
worth the complexity for now.

### DIFF
--- a/crowbar_framework/lib/crowbar/event_dispatcher.rb
+++ b/crowbar_framework/lib/crowbar/event_dispatcher.rb
@@ -1,0 +1,51 @@
+#
+# Copyright 2017, SUSE
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+module Crowbar
+  class EventDispatcher
+    class << self
+      def trigger_hooks(event, details)
+        BarclampCatalog.barclamps.keys.each do |barclamp|
+          begin
+            cls = ServiceObject.get_service(barclamp)
+          rescue NameError
+            # catalog may contain barclamps which don't have services
+            next
+          end
+
+          next unless cls.method_defined?(:event_hook)
+
+          service = cls.new(Rails.logger)
+
+          proposals = Proposal.where(barclamp: barclamp)
+          proposals.each do |proposal|
+            next if proposal.role.nil?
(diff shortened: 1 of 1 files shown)


crowbar_framework/lib/crowbar/event_dispatcher.rb
```diff
@@ -0,0 +1,48 @@
...
+          service = cls.new(Rails.logger)
+
+          proposals = Proposal.where(barclamp: barclamp)
+          proposals.each do |proposal|
+            role = proposal.role
+            next if role.nil?
+            begin
+              service.event_hook(role, event, details)
+            rescue StandardError => e
+              raise "Error while executing event hook of #{barclamp} for #{event}: #{e.message}"
```
matelakat: Maybe we want to include that it is for a specific proposal - as if the barclamp has multiple proposals, it might fail in one of them?

crowbar_framework/lib/crowbar/event_dispatcher.rb
```diff
@@ -0,0 +1,48 @@
...
+          next unless cls.method_defined?(:event_hook)
+
+          service = cls.new(Rails.logger)
+
+          proposals = Proposal.where(barclamp: barclamp)
+          proposals.each do |proposal|
+            role = proposal.role
+            next if role.nil?
+            begin
+              service.event_hook(role, event, details)
```
matelakat: What happens if `event_hook` is implemented badly and raises an error? I guess in that case we raise and do not execute the hook for the rest of the barclamps?

matelakat:
[Notification Infrastructure](https://trello.com/c/In9nTfTH/34-notification-infrastructure)

toabctl:
@scottwulf we usually wait for 2 approvals before merging.

PR #1309, opened 2017-08-24
network: Ensure default route exists

cmurphy:
Sometimes when bringing up a new interface, wicked deletes the default
interface. We've seen this before[1] but we don't have a root cause.
This workaround ensures that after wicked ifup is called, the default
route is reapplied.

[1] https://bugzilla.suse.com/show_bug.cgi?id=1031077

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -638,6 +638,14 @@ when "suse"
     code "wicked ifup all"
     only_if { run_wicked_ifup }
   end
+  # Sometimes wicked deletes the default route upon bringing up a new interface,
+  # make sure it gets put back
+  unless ::Kernel.system("ip route |grep -q default")
+    nic = default_route[:nic]
+    gw = default_route[:gateway]
+    Chef::Log.info("Adding default route via #{gw} to #{nic}")
+    ::Kernel.system("ip route add default via #{gw} dev #{nic}")
+  end
 
   # Avoid running the wicked related thing on SLE11 nodes
   unless node[:platform] == "suse" && node[:platform_version].to_f < 12.0


chef/cookbooks/network/recipes/default.rb
```diff
@@ -638,6 +638,14 @@ def get_datapath_id_for_ovsbridge(bridge)
     code "wicked ifup all"
     only_if { run_wicked_ifup }
   end
+  # Sometimes wicked deletes the default route upon bringing up a new interface,
+  # make sure it gets put back
```
nicolasbock: Add the bug reference bsc#1031077?

chef/cookbooks/network/recipes/default.rb
```diff
@@ -638,6 +638,14 @@ def get_datapath_id_for_ovsbridge(bridge)
...
+  unless ::Kernel.system("ip route |grep -q default")
+    nic = default_route[:nic]
+    gw = default_route[:gateway]
+    Chef::Log.info("Adding default route via #{gw} to #{nic}")
+    ::Kernel.system("ip route add default via #{gw} dev #{nic}")
+  end
```
nicolasbock: Do I understand this correctly that this is a workaround while we see this issue with the default route being dropped? I can't quite tell from the bug report, do you have an inkling as to what is ultimately the cause?

nicolasbock: Ok, sorry, I just re-read your PR description and there you say it: no root cause known right now :frowning_face:

cmurphy:
I still think this is worth having but it turned out not to be the cause of the linked bug so I removed the tag

nicolasbock:
But it fixed the problem, right?

cmurphy:
@nicolasbock https://github.com/crowbar/crowbar-openstack/pull/1195 and https://github.com/crowbar/crowbar-openstack/pull/1197 fix the problem in the bug report, this fixes an unrelated problem.

nicolasbock:
I don't follow @cmurphy (and maybe this just means that it's time for another coffee :smile:). The bug report refers to the default route being lost. How are the other two PRs addressing the route? And this PR sets the default route. Very confused...

cmurphy:
@nicolasbock aha - sorry for my confusing wording, the bug I was referring to was https://bugzilla.suse.com/show_bug.cgi?id=1055188 which I had referenced in the commit subject, so I shouldn't have said "linked" because I had also added a link to https://bugzilla.suse.com/show_bug.cgi?id=1031077.

I do think this commit would fix that older bug #1031077 but I'm not sure that bug is really valid any more because it came from before we reverted the "wicked ifreload" thing.

nicolasbock:
Ok @cmurphy thanks, that makes sense. Now I remember that revert,
 41e310cf21642cf0dc2f2fa0b7294c1db7821b59. If it's really the case though that the loss of the default route was fixed by 41e310cf21642cf0dc2f2fa0b7294c1db7821b59 then do we still need your PR? If we can't definitely say that we don't have routing issues anymore then I think it's still a good idea to merge this.

cmurphy:
It was fixed at the time but there's now a `wicked ifup all` that for some reason seems to cause the same problem when it brings up br-fixed :(

On the other hand we also have https://github.com/crowbar/crowbar-core/blob/d07a2da7ab0c84115bd03f0881bf01f197ef1d75/chef/cookbooks/network/recipes/default.rb#L478 which re-adds the route on the next chef run, which is why this problem hasn't been as noticeable as the earlier problem.

s-t-e-v-e-n-k:
The wicked ifup all call was added due to issues with bonding -- we drop the underlying slaves with wicked ifdown before downing them to add them to the bonding interface (because if we don't, wicked will notice when we up the interface again, and reapply the configuration, which was leading to duplicate IPs.) The call to ifup all at the end is to instruct wicked that all interfaces have been configured -- is this not the case for br-fixed?

cmurphy:
@s-t-e-v-e-n-k At the point that ifup is run, br-fixed is already created and up. It has an address and it has two routes through it, one for the public network and one is the default route. When ifup is run, the default route disappears.

PR #1314, opened 2017-08-30
Disable upgrade API in Cloud8

skazi0:
It fixes bug https://bugzilla.suse.com/show_bug.cgi?id=1036762

Filter is added to the backend to block access to the upgrade API. All `/upgrade` endpoints return fixed JSON error message which will be displayed directly in CLI and will pop-up when user tries to run prechecks in the UI.

requests?**
https://trello.com/c/m8ROv8Hg/437-bug-1036762-upgrade-api-accessible-in-cloud7

### DIFF
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -15,6 +15,20 @@
 #
 
 class Api::UpgradeController < ApiController
+  # disable upgrade API until upgrade to next version is implemented
+  before_action do
+    # skip filter if we're in the middle of upgrade from previos version
+    unless File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-running")
+      render json: {
+        errors: {
+          unexpected_error: {
+            data: "Upgrade not yet supported in this version"
+          }
+        }
+      }, status: :unprocessable_entity
+    end
+  end
+
   skip_before_filter :upgrade
 
   api :GET, "/api/upgrade", "Show the Upgrade progress"
--- a/crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
+++ b/crowbar_framework/spec/controllers/api/upgrade_controller_spec.rb
@@ -29,6 +29,14 @@ describe Api::UpgradeController, type: :request do
     end
     let(:tarball) { Rails.root.join("spec", "fixtures", "crowbar_backup.tar.gz") }
 
+    before do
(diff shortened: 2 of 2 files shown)


skazi0:
[Bug 1036762 - Upgrade API accessible in Cloud7](https://trello.com/c/m8ROv8Hg/437-bug-1036762-upgrade-api-accessible-in-cloud7)

jsuchome:
Hm. I get
```
Filter chain halted as #<Proc:0x000000096b4a78@/opt/dell/crowbar_framework/app/controllers/api/upgrade_controller.rb:19> rendered or redirected
```
in production.log, but the UI still moves to the upgrade landing page.

jsuchome:
Besides, the bug is about Cloud7, this code is Cloud8... ?

Itxaka:
Same, on cloud8 going to the upgrade url logs the filter halt but still redirects me. On the CLI it works as expected thougth

```
root@crowbar:/opt/dell/crowbar_framework # crowbarctl upgrade status
{"errors":{"unexpected_error":{"data":"Upgrade not supported in this version"}}}

```

skazi0:
@jsuchome, @Itxaka: the "Filter chain halted" is just INFO and I think this is expected behavior with filters.
The `http://<crowbar>/upgrade` path will not be blocked completely by this as this is static content served directly from apache and crowbar service doesn't have anything to do with that. The point of this change is to block users from calling some (potentially unsafe) APIs not from seeing the upgrade UI.
@jsuchome you're right about Cloud8/Cloud7 but I think this filter is still appropriate for Cloud8. I will create another PR for Cloud7 which will be more selective as we want the APIs to work on Cloud7 during upgrade from Cloud6 but we don't want users to start new upgrades from Cloud7.

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -15,6 +15,20 @@
...
 
 class Api::UpgradeController < ApiController
+  # disable upgrade API until upgrade to next version is implemented
+  before_action do
+    # skip filter if we're in the middle of upgrade from previos version
+    unless File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-running")
+      render json: {
+        errors: {
+          unexpected_error: {
+            data: "Upgrade not supported in this version"
```
jsuchome: Just a detail, but it might be nice to have the message better, at least by saying "it is not _yet_ supported".

PR #1316, opened 2017-08-31
Retry to resolve a hostname before continue

toabctl:
Before running commands we check if a node is ready or the chef-client
is still running. Before doing that, we try to resolve the hostname
and continue with the resolved IP.
It seems that in some cases, resolve_host() does not work and returns
an empty string. The crowbar production.log then tells us:

INFO -- Waiting for already running chef-client on
  d52-54-77-77-01-03.vk2.cloud.suse.de (); next attempt in 10 seconds
  until 2017-08-31 04:36:53 +0000

but inside of "()" should be the resolved IP address. This message is
repeated until we reach the timeout. This repetition is useless
because we retry with an empty IP.
To minimize the impact of this problem, retry to resolve the hostname
for some time.

### DIFF
--- a/crowbar_framework/lib/remote_node.rb
+++ b/crowbar_framework/lib/remote_node.rb
@@ -28,6 +28,8 @@ module RemoteNode
   def ready?(host, timeout = 60, sleep_time = 10, options = {})
     timeout_at = Time.now + timeout
     ip = resolve_host(host)
+    return false if ip.empty?
+
     msg = "Waiting for host #{host} (#{ip}); next attempt in #{sleep_time} seconds until #{timeout_at}"
 
     nobj = Node.find_by_name(host)
@@ -42,6 +44,7 @@ module RemoteNode
   def chef_ready?(host, timeout = 60, sleep_time = 10, options = {})
     timeout_at = Time.now + timeout
     ip = resolve_host(host)
+    return false if ip.empty?
     msg = "Waiting for already running chef-client on #{host} (#{ip}); next attempt in #{sleep_time} seconds until #{timeout_at}"
     wait_until(msg, timeout_at.to_i, sleep_time, options) do
       port_open?(ip, 22) && !chef_clients_running?(ip)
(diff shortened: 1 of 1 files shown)


toabctl:
The problem occured on https://ci.suse.de/job/openstack-mkcloud/81061/

crowbar_framework/lib/remote_node.rb
```diff
@@ -111,6 +111,12 @@ def chef_clients_running?(host_or_ip)
   # Workaround for similar issue
   # http://projects.puppetlabs.com/issues/2776
   def resolve_host(host)
+    sleep_time = 5
+    timeout_at = Time.now + 60
+    msg = "Waiting to resolve hostname '#{host}'; next attempt in #{sleep_time} seconds until #{timeout_at}"
+    wait_until(msg, timeout_at.to_i, sleep_time, {}) do
+      !`dig +short #{host} | head -n1`.rstrip.empty?
+    end
```
cmurphy: You could store `dig +short #{host} | head -n1`.rstrip in a variable here and avoid the extra dig call after the loop

aspiers:
@toabctl Please trim the boilerplate text if you are not using it!

crowbar_framework/lib/remote_node.rb
toabctl: > Please trim the boilerplate text if you are not using it!

@a

> Please trim the boilerplate text if you are not using it!

@aspiers done. But it's the commit msg that is important! :)

toabctl:
> Please trim the boilerplate text if you are not using it!

@aspiers done. But it's the commit msg that is important! :)

crowbar_framework/lib/remote_node.rb
toabctl: Done

toabctl:
> But if after the new retries the resolution still fails then you have the same problem as before. Please can you also fix the calling code to handle the case where it never gets an IP?

Done

toabctl:
> Do you know why that happens? Is bind/dnsmasq is being restarted on the admin node (as it happens when applying the dns barclamp)? Anyway we could avoid that?

@rhafer bind seems to be restarted. From the crowbar node admin log:
```
[2017-08-31T04:16:49+00:00] INFO: service[bind9] restarted 
```

and from the crowbar production.log the first occurence fot the loop is at 2017-08-31T04:16:53.334690

rhafer:
> bind seems to be restarted. From the crowbar node admin log:

[..]

> and from the crowbar production.log the first occurence fot the loop is at 2017-08-31T04:16:53.334690

According to /var/log/message bind is already back then. But the resolver is configured to use the local dnsmasq first (see etc/resolv.conf). And that one is configured to forward to one the the running bind instances (node1 and node2 also have bind now). It seem those instances are not yet done with the zone transfer, according to the logs on node1 in finishes the zone transfer at 2017-08-31T04:17:03.510496. It might be that dnsmasq was forwarding queries to them. Anyway, I digress.

toabctl:
@aspiers are you ok with merging this now?

toabctl:
Backport for stable/4.0 is https://github.com/crowbar/crowbar-core/pull/1323

PR #1318, opened 2017-08-31
dhcp fix for UEFI (bsc#961536)

matelakat:
Without the specific dhcp option:

always-reply-rfc1048 true;

Some machines fail to boot (DL160) and print something like this on the
console:

```
Loading Linux ...
error: time out opening `suse-...../loader/linux`
Loading initial ramdisk ...
error: you need to load the kernel first.

Press any key to...
```

### DIFF
--- a/chef/cookbooks/dhcp/templates/default/dhcpd.conf.erb
+++ b/chef/cookbooks/dhcp/templates/default/dhcpd.conf.erb
@@ -18,6 +18,9 @@ option path-prefix code 210 = text;
 # have to hack syslog.conf to complete the redirection).
 log-facility local7;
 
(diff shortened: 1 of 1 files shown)


matelakat:
[UEFI booting issue @ SAP](https://trello.com/c/JhP1Tits/87-uefi-booting-issue-sap)

Itxaka:
Is there any downside to having this set to true? After reading about the flag I cannot see any, but just in case...

nicolasbock:
@Itxaka I don't see any either :)

toabctl:
Backport for stable/4.0 is https://github.com/crowbar/crowbar-core/pull/1320

PR #1320, opened 2017-09-01
dhcp fix for UEFI (bsc#961536)

toabctl:
Without the specific dhcp option:

always-reply-rfc1048 true;

Some machines fail to boot (DL160) and print something like this on the
console:

```
Loading Linux ...
error: time out opening `suse-...../loader/linux`
Loading initial ramdisk ...
error: you need to load the kernel first.

Press any key to...
```

(cherry picked from commit 7652fd8c1edbb24c7137bdb2e42328701e9bc47e)

requests?**


matelakat:
[UEFI booting issue @ SAP](https://trello.com/c/JhP1Tits/87-uefi-booting-issue-sap)

PR #1321, opened 2017-09-01
Properly pass command args to bmc_cmd

toabctl:
Multiple string arguments mean that the system method runs the first
argument without a shell and passes subsequent arguments quoted to that
command [1].
This means that if the cmd string contains multiple substrings separated
by whitespace the whole string gets quoted into one argument and
passed to ipmitool which is incorrect.
By first splitting cmd on whitespace and then splatting the resulting
list into a list of arguments to system, all whitespace separated
substrings become separate arguments to ipmitool.

[1] https://ruby-doc.org/core-2.2.0/Kernel.html#method-i-system

(cherry picked from commit e339ca6de997df9828e05775e00777f9132da618)

### DIFF
--- a/crowbar_framework/app/models/node.rb
+++ b/crowbar_framework/app/models/node.rb
@@ -982,8 +982,9 @@ class Node < ChefObject
   end
 
   def bmc_cmd(cmd)
+    cmd_list = cmd.split
     if bmc_address.nil? || get_bmc_user.nil? || get_bmc_password.nil? ||
-        !system("ipmitool", "-I", get_bmc_interface, "-H", bmc_address, "-U", get_bmc_user, "-P", get_bmc_password, cmd)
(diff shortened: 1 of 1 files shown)


toabctl:
> did anyone test the new code?

@bmwiedemann according to https://github.com/crowbar/crowbar-core/pull/1315 , @Itxaka did

nicolasbock:
@bmwiedemann, you can test this with
```
$ cat test.rb 
cmd="-h test.rb"
system("ls", "-l", cmd)
cmd_list = cmd.split
system("ls", "-l", *cmd_list)
```
which produces
```
$ ruby test.rb 
ls: invalid option -- ' '
Try 'ls --help' for more information.
-rw-r--r-- 1 nbock nbock 92 Sep  1 04:38 test.rb
```

PR #1322, opened 2017-09-01
Allow to discover IPMI settings for deployed nodes

bmwiedemann:
https://bugzilla.suse.com/show_bug.cgi?id=885890
When IPMI was enabled after nodes were already deployed, it could not be used for STONITH.

We also include the ipmi-discover recipe in 'ready' state

**Is there additional information worth sharing?**
The discover code will not run by default, only when a user does
`crowbarctl proposal edit ipmi default`
to add `ready` under
```
  "deployment": {
    "ipmi": {
      "element_states": {
        "ipmi":
```

### DIFF
--- a/chef/cookbooks/ipmi/recipes/role_ipmi.rb
+++ b/chef/cookbooks/ipmi/recipes/role_ipmi.rb
@@ -15,7 +15,7 @@
 #
 
 if CrowbarRoleRecipe.node_state_valid_for_role?(node, "ipmi", "ipmi")
-  if ["discovering", "hardware-installing", "readying"].include? node[:state]
+  if ["discovering", "hardware-installing", "readying", "ready"].include? node[:state]
     include_recipe "ipmi::ipmi-discover"
   end
 


bmwiedemann:
It was intended to be a partial fix (see PR description), because always running this would add extra time to chef-client runs

bmwiedemann:
I think there was some way to only run chef code, when a user does apply/commit a (certain?) proposal, but not in the regular runs

matelakat:
[IPMI barclamp improvements](https://trello.com/c/vYorAmTd/93-ipmi-barclamp-improvements)

matelakat:
We would like to re-work this as part of the attached trello card

bmwiedemann:
depends. in some regards, rebooting can be expensive, when you have guest VMs running on the node and would have to migrate them off.
But maybe a crowbar_join --start would also do the trick?

skazi0:
@bmwiedemann I haven't tested it but the part during reboot that does the transition through `readying` is exactly `crowbar_join` call so probably this will be enough.

bmwiedemann:
change is probably not needed then

PR #1326, opened 2017-09-06
Read only ipmi

skazi0:
**NOTE: https://github.com/crowbar/crowbar-ha/pull/232 is needed to complete this functionality wrt STONITH**

When IPMI boards are pre-configured with static IPs/users Crowbar should not try to allocate new IPs or change passwords. Only the discovery part should be done by Crowbar to detect the settings.

New option is added which disables all write operations on IPMI interface.

requests?**
https://trello.com/c/Q7laJy5q/91-readonly-ipmi-use-existing-ipmi-settings-no-bmc-nat

### DIFF
--- a/crowbar_framework/app/views/barclamp/ipmi/_edit_attributes.html.haml
+++ b/crowbar_framework/app/views/barclamp/ipmi/_edit_attributes.html.haml
@@ -7,7 +7,14 @@
     = string_field :bmc_user
     = password_field :bmc_password
     = string_field :bmc_interface
-    = boolean_field :use_dhcp
-    = boolean_field :ignore_address_suggestions
+    = boolean_field :bmc_reconfigure,
+     "data-showit" => "true",
+     "data-showit-target" => "#reconfigure_container",
+     "data-showit-direct" => "true"
+    %span.help-block
+      = t(".bmc_reconfigure_hint")
+    #reconfigure_container
+      = boolean_field :use_dhcp
+      = boolean_field :ignore_address_suggestions
     = boolean_field :bmc_nat_enable
     = boolean_field :debug
--- a/crowbar_framework/config/locales/ipmi/en.yml
+++ b/crowbar_framework/config/locales/ipmi/en.yml
@@ -23,6 +23,8 @@ en:
         bmc_user: 'BMC User'
         bmc_password: 'BMC Password'
         bmc_interface: 'BMC Interface'
+        bmc_reconfigure: 'Modify BMC Settings'
+        bmc_reconfigure_hint: 'When this option is disabled, Crowar will not reconfigure BMC. All BMC network settings need to be properly pre-configured on all nodes and users/passwords need to match the ones specified above.'
         use_dhcp: 'BMC Should Use DHCP (Not explicitly assigned)'
         ignore_address_suggestions: 'Ignore Already Configured Address'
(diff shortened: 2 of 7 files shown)


skazi0:
[readonly IPMI - use existing IPMI settings - no bmc-nat](https://trello.com/c/Q7laJy5q/91-readonly-ipmi-use-existing-ipmi-settings-no-bmc-nat)

crowbar_framework/app/views/barclamp/ipmi/_edit_attributes.html.haml
```diff
@@ -7,7 +7,12 @@
     = string_field :bmc_user
     = password_field :bmc_password
     = string_field :bmc_interface
-    = boolean_field :use_dhcp
-    = boolean_field :ignore_address_suggestions
+    = boolean_field :bmc_reconfigure,
+     "data-showit" => "true",
```
matelakat: What do these `data-showit` things do?

cmurphy:
Not opposed to adding this for safety, but in theory if the bmc is preconfigured and ignore_address_suggestions is set to false, crowbar won't reset it

https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/ipmi_service.rb#L83

Is there something going wrong where crowbar is resetting the IP address even when crowbar_wall is populated with the discovered info?

crowbar_framework/app/views/barclamp/ipmi/_edit_attributes.html.haml
skazi0: They hide the parameters in reconfigure_container if reconfigure is disabled. The `use_dhcp` and `ignore_address_suggestions` make no sense if reconfiguration is not done.

skazi0:
@cmurphy yeah, with proper configuration of `bmc` network and "suggestions" enabled and password matching the existing password, the whole thing should be a noop without this change, but there are so many things that can go wrong (e.g. losing access to IPMI) that we decided that it's better for the UX to have such option.

crowbar_framework/config/locales/ipmi/en.yml
```diff
@@ -23,6 +23,7 @@ en:
         bmc_user: 'BMC User'
         bmc_password: 'BMC Password'
         bmc_interface: 'BMC Interface'
+        bmc_reconfigure: 'Modify BMC Settings'
```
vuntz: This should be made more specific to mention that it's only about networking. AFAICT, we still modify the user/password.

skazi0: The plan is to make it really 'read-only': https://github.com/skazi0/crowbar-core/blob/6ceee2664d9d7a0777e31d5b90552194bc418c20/chef/cookbooks/ipmi/recipes/ipmi-configure.rb#L118

vuntz: Oh, missed that. Can you add a hint in the UI that it is required that all servers have the same user/password for the BMC?

skazi0: done (not tested yet)

skazi0:
@vuntz well... it behaved the same before this change. For example if you enable `use_dhcp` after you install some nodes this setting will have no effect on those, only on new ones.

PR #1327, opened 2017-09-08
Increase inotify max user instances

abel-navarro:
It has been found that during heavy network creation by Neutron (>1K
networks) dnsmasq is unable to complete due to lack of inotify user
instances.

The system default is 128 instances which doesn't cover this situation.
By (short) experimentation 1024 has been found to be a comprehensive
value.

Reference: https://bugs.launchpad.net/charms/+bug/1593041

Found while investigating: https://trello.com/c/M0umdQTM

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -92,6 +92,13 @@ execute "enable netfilter for bridges" do
   subscribes :run, resources(cookbook_file: "modprobe-bridge.conf"), :delayed
 end
 
+# Increase inotify max user instances
+# https://bugs.launchpad.net/charms/+bug/1593041
+execute "increasing inotify max user instances" do
+  command "echo 1024 > /proc/sys/fs/inotify/max_user_instances"
+  action :nothing
+end
+
 conduit_map = Barclamp::Inventory.build_node_map(node)
 Chef::Log.debug("Conduit mapping for this node:  #{conduit_map.inspect}")
 route_pref = 10000


abel-navarro:
[OVS issue with more than 1000 networks](https://trello.com/c/M0umdQTM/57-ovs-issue-with-more-than-1000-networks)

chef/cookbooks/network/recipes/default.rb
```diff
@@ -92,6 +92,13 @@
   subscribes :run, resources(cookbook_file: "modprobe-bridge.conf"), :delayed
 end
 
+# Increase inotify max user instances
+# https://bugs.launchpad.net/charms/+bug/1593041
+execute "increasing inotify max user instances" do
+  command "echo 1024 > /proc/sys/fs/inotify/max_user_instances"
+  action :nothing
```
Itxaka: This would not survive any reboots right? Plus rigth now its not doing anything :)

abel-navarro:
@nicolasbock dnsmasq does indeed consume [one inotify instance](http://thekelleys.org.uk/gitweb/?p=dnsmasq.git;a=blob;f=src/inotify.c;h=b5a17ddc682191cfa3360f1a6235cbf0920da2d6;hb=HEAD#l92). Since we are using one dnsmasq process per network, inotify will consume as many instances as networks.

In our tests we were creating 1k networks, so 1024 fits there. In my environment there are only 6 other processes consuming inotify instances: systemd, systemd-udevd, dbus-daemon, cron, nscd and an isolated dnsmasq instance. That means we roughly need one inotify instance per network we want to create. My choice of 1024 inotify instances may be limiting in case we want to create more networks. Probably a 8k limit would be enough for now. Suggestions are accepted.

abel-navarro:
@toabctl will fix that.

@Itxaka right, will make is persist during reboots.

nicolasbock:
@abel-navarro, thanks! If we are creating 1000 networks then the new limit of 1024 seems awfully close to what you need. I agree with your suggestion to raise this limit even further to maybe 8k.

chef/cookbooks/network/recipes/default.rb
nicolasbock: That's a good point @Itxaka. @abel-navarro we need to notify this resource somewhere or just run it.

abel-navarro:
Probably crowbar-core is not the right place to set the inotify values since that affects all nodes and this is just for network nodes. I created a new PR (https://github.com/crowbar/crowbar-openstack/pull/1268) to set it in the network nodes as well as making it persistent to reboots. So I'll be closing this PR and continue discussion there.

PR #1329, opened 2017-09-18
Add utils_systemd_service_restart LWRP

vuntz:
This resource adds (or removes) a systemd config snippet for a service,
to change the behavior to Restart=on-failure for that service. This
means systemd will automatically restart the service on failures, which
makes a relatively good self-healing solution without pacemaker.

See systemd.service(5) for details on Restart=

Note that the fetch_name_from_service_resource attribute of this
resource enables automatically fetching the systemd service name from
the service resource, which helps remove duplication of code in the
cookbooks when the service name is being overridden in the attributes of
the service resource. For instance, this will work as expected:
```
  service "foobar" do
    service_name "notfoobar"
  end

  utils_systemd_service_restart "foobar"
```

### DIFF
--- a/chef/cookbooks/apache2/metadata.rb
+++ b/chef/cookbooks/apache2/metadata.rb
@@ -4,6 +4,7 @@ license "Apache 2.0"
 description "Installs and configures all aspects of apache2 using Debian style symlinks with helper definitions"
 long_description IO.read(File.join(File.dirname(__FILE__), "README.rdoc"))
 version "0.99.3"
+depends "utils"
 recipe "apache2", "Main Apache configuration"
 recipe "apache2::mod_alias", "Apache module 'alias' with config file"
 recipe "apache2::mod_auth_basic", "Apache module 'auth_basic'"
--- a/chef/cookbooks/apache2/recipes/default.rb
+++ b/chef/cookbooks/apache2/recipes/default.rb
@@ -222,3 +222,5 @@ service "apache2" do
   action [:enable, :start]
   ignore_failure true
 end
+
+utils_systemd_service_restart "apache2"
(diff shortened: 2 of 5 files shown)


aplanas:
I would prefer to consider it a bug in the package. In this way will be beneficial for Ardana, raw openSUSE packages and any alternative method of deployment

vuntz:
> I would prefer to consider it a bug in the package. In this way will be beneficial for Ardana, raw openSUSE packages and any alternative method of deployment

Good point. I'll defer to @aspiers on how it should be handled; he may have given this some thoughts, and I don't have a strong feeling.

(note however that I'd need a backport to stable/4.0, where we can't change this in the service files)

aspiers:
If we put this in the packages then we'd have to disable it again for the (default) case where Pacemaker is managing the services.  Is that even doable?  If we need the packages to support both approaches, my hunch is that it's going to be easier if they are kept less opinionated, i.e. the restart is layered on top like it is here (and could easily be done in Ardana too in a similar way).

aplanas:
@aspiers in this case we need to coordinate with the packages:

```
rpm-packaging> grep -R 'Restart' * | wc -l
38
```

vuntz:
> If we put this in the packages then we'd have to disable it again for the (default) case where Pacemaker is managing the services. Is that even doable? If we need the packages to support both approaches, my hunch is that it's going to be easier if they are kept less opinionated, i.e. the restart is layered on top like it is here (and could easily be done in Ardana too in a similar way).

I updated the pull request to actually allow anything in the package: if we call `:disable` on the resource, then it will create a config snippet with `Restart=no` and if we call `:enable`, it will create a config snippet with `Restart=on-failure` (or some other value, depending on the attributes).

How does that sound?

aplanas:
@vuntz If this works with Pacemaker, I think IMHO that is a great solution.

PR #1331, opened 2017-09-20
all: Make systemd restart services on failures

vuntz:
This offers some automatic recovery, which is nice. It turns out it's
also recommended by systemd documentation, in case a service is
long-running, which is the case here.

It's worth mentioning that in the past, the periodic chef-client has
been providing this, to some extent. But this means the recovery could
take up to 15 minutes. It was actually not working in the case of the
chef server infrastructure (chef-server, chef-solr, couchdb, etc.), and
systemd can really help here.

This is of course disabled when a service is managed by Pacemaker, as we
do not want to have systemd interfere with Pacemaker in that case. Note
that this commit doesn't touch any service managed by Pacemaker, though.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/default.rb
+++ b/chef/cookbooks/crowbar/recipes/default.rb
@@ -301,6 +301,13 @@ template "/etc/chef/solr.rb" do
   notifies :restart, "service[chef-solr]", :delayed
 end
 
+# Make systemd restart chef services (and their deps) on failure
+utils_systemd_service_restart "rabbitmq-server"
+utils_systemd_service_restart "couchdb"
+utils_systemd_service_restart "chef-solr"
+utils_systemd_service_restart "chef-expander"
+utils_systemd_service_restart "chef-server"
+
 if node[:platform_family] == "suse"
   cookbook_file "/etc/tmpfiles.d/crowbar.conf" do
     owner "root"
@@ -319,6 +326,7 @@ if node[:platform_family] == "suse"
   service "crowbar" do
     action :enable
   end
+  # No need for utils_systemd_service_restart: Restart= is already in the .service file
 else
   %w(chef-server-api chef-server-webui chef-solr rabbitmq-server).each do |f|
     file "/etc/logrotate.d/#{f}" do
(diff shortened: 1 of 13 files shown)


chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -301,6 +301,13 @@
   notifies :restart, "service[chef-solr]", :delayed
 end
 
+# Make systemd restart chef services (and their deps) on failure
+utils_systemd_service_restart "rabbitmq-server"
+utils_systemd_service_restart "couchdb"
+utils_systemd_service_restart "chef-solr"
```
abel-navarro: I would put the service restart call right after the `service` block, as you did with the other services.

chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -301,6 +301,13 @@
   notifies :restart, "service[chef-solr]", :delayed
 end
 
+# Make systemd restart chef services (and their deps) on failure
+utils_systemd_service_restart "rabbitmq-server"
```
abel-navarro: Won't this crowbar dependency for rabbitmq overlap with the one with use for the cluster?

vuntz: No, that's the rabbitmq on the admin server, used by chef. It's completely independent of anything we do with rabbitmq for openstack.

chef/cookbooks/crowbar/recipes/default.rb
vuntz: I didn't do it in this case because I wanted to keep all these lines together (since we actually don't deal with these services in the recipe).

So let's wait for another reviewer and get a 3rd opinion :-)

chef/cookbooks/resolver/recipes/default.rb
```diff
@@ -51,6 +51,9 @@
     end
     not_if { node["crowbar"]["admin_node"] && File.exist?("/var/lib/crowbar/install/disable_dns") }
   end
+  utils_systemd_service_restart "dnsmasq" do
+    not_if { node["crowbar"]["admin_node"] && File.exist?("/var/lib/crowbar/install/disable_dns") }
```
aspiers: Why is this line necessary?  Wouldn't setting the Restart option be irrelevant if the service is disabled?  Makes me wonder if the `not_if` in the previous `service` block is actually the wrong approach because if someone had manually enabled it, it wouldn't disable it.

vuntz: You're correct, it's irrelevant, but on the other hand, it makes little sense to create a snippet if we don't do anything for the service. So somehow, it doesn't hurt either, and I view it more as consistency between the two resources.

PR #1333, opened 2017-09-22
bind: set transfer-source to avoid xfer failures (bsc#1036601)

bmwiedemann:
https://bugzilla.suse.com/show_bug.cgi?id=1036601

Currently bind isn't explicitely told which IP address to
request zone transfers from, while the server is configured to
specifically only accept zone transfer requests from the specific
list of whitelisted 2ndary slave ip addresses. Without that
change it happened that bind happened to choose another IP address
from the admin network (like an unrelated VIP allocated from admin)
for transfer, failing to sync up with master.

Cherry-picked from commit 6f450c3375b5437d443ec666560539

### DIFF
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -130,6 +130,8 @@ def make_zone(zone)
     master_ip = node[:dns][:master_ip]
   end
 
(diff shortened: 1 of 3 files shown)


dirkmueller:
needs to be reopened from a personal branch for the mkcloud whitelisting check to pass

PR #1334, opened 2017-09-22
provisioner: Make chef splay configurable

skazi0:
For large installations many concurrent chef-client runs can cause significant load on chef-server and crowbar.

Exposing `splay` parameter enables users to increase it to distribute the chef-server load caused by chef-clients.

requests?**
https://trello.com/c/dcIa46O8/118-change-chef-client-splay

### DIFF
--- a/chef/data_bags/crowbar/migrate/provisioner/201_chef_splay.rb
+++ b/chef/data_bags/crowbar/migrate/provisioner/201_chef_splay.rb
@@ -1,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["chef_splay"] = ta["chef_splay"] unless a.key? "chef_splay"
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a.delete("chef_splay")
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-provisioner.json
+++ b/chef/data_bags/crowbar/template-provisioner.json
@@ -215,6 +215,7 @@
           "debug": "debug"
         }
       },
+      "chef_splay": 20,
       "chef_client_runs": 900
     }
   },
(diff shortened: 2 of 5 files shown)


skazi0:
[User modifiable chef-client splay](https://trello.com/c/dcIa46O8/118-change-chef-client-splay)

chef/data_bags/crowbar/migrate/provisioner/201_chef_splay.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["chef_splay"] = ta["chef_splay"]
```
Itxaka: if we dont have a guard here and we backport this migration, it would erase the current configured value with the default for an environment which was updated from soc7 to soc8

skazi0:
@Itxaka something like this?

chef/data_bags/crowbar/template-provisioner.json
```diff
@@ -215,14 +215,15 @@
           "debug": "debug"
         }
       },
+      "chef_splay": 20,
```
vuntz: Can we bump the default to 300? 20s is really to slow.

skazi0: We agreed with @matelakat that defaults change can be done as separate PR. Let me know if this change is needed here.

matelakat: That should be a separate PR

mnmehta:
@skazi0 Please note that we need to update the "state" method in crowbar_framework/app/models/node.rb so that node state does not get marked as "unknown" prematurely. I see this a lot on the SAP scale cloud because the current logic makes it unknown if it has not seen a chef_client run within chef_client_runs + a 300s buffer ... let's add splay into this.

matelakat:
> @skazi0 Please note that we need to update the "state" method in crowbar_framework/app/models/node.rb so that node state does not get marked as "unknown" prematurely. I see "unknown" state often on the SAP scale cloud (on which I have set splay to 1800) because the current logic makes it "unknown" if it has not seen a chef_client run within chef_client_runs seconds + a 300s buffer ... let's add splay into this.

@mnmehta creating a trello card for that.

matelakat:
https://trello.com/c/WXL2ZVg6

PR #1335, opened 2017-09-22
dns: fail on empty server list

bmwiedemann:
Discovered when working on http://bugzilla.novell.com/show_bug.cgi?id=1058017
Nodes that are not 'ready' are not considered available to serve as DNS server, but this can lead to situations were nodes drop all entries from /etc/resolv.conf and dnsmasq.conf and cannot resolve any DNS anymore.

we detect that case and raise a (fatal) exception
to not screw up existing resolv.conf and dnsmasq.conf
which would make it hard to leave this state,
because rabbitmq (and possibly others (especially with SSL)) needs DNS to work.

requests?**
There is probably more to that bug, such as nodes losing roles, not just config flags under certain conditions.

### DIFF
--- a/chef/cookbooks/resolver/recipes/default.rb
+++ b/chef/cookbooks/resolver/recipes/default.rb
@@ -26,6 +26,11 @@ if dns_list.empty? && \
   dns_list = (node[:dns][:forwarders] + node[:dns][:nameservers]).flatten.compact
 end
 
+if dns_list.empty?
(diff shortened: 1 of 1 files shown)


bmwiedemann:
extended the commit message.
btw: the original change was on SOC6 where we use chef-solr's search, which has been refactored out meanwhile, but I guess that the other logic still applies.

dirkmueller:
don't think is is needed with the proper fix (#1340)

PR #1337, opened 2017-09-26
dns: Run dns-client role only in readying/ready/applying states

dirkmueller:
This is like nearly all other roles.

The change only really impacts the discovery image, but it turns out
that with the discovery image, we use DHCP, so we get the DNS info from
DHCP in that case.

(cherry picked from commit 8e980e6f5d02f78b54113b7d92a16009812c2ca7)

requests?**


rsalevsky:
Closing this PR as we won't pickup new changes for Cloud 6 maintenance updates. If this is still important for newer versions please feel free to reopen this PR against master.

PR #1339, opened 2017-09-26
[POC] skip unready nodes on barclamp deployment

Itxaka:
Background:

-  just ignore not ready/not accessible nodes (they will be automatically updated on their next local chef-client run)


Itxaka:
This is just one approach to the nodes not ready issue, but the problem with this is that we need to add a list of states in which we would allow the node to be skipped because if not, depending on the state, the deployment will fails (on shutdown for example)

Itxaka:

Itxaka:
#1341 sounds much better than this.

PR #1340, opened 2017-09-26
ohai: Skip saving etc/passwd entries (bsc#1058017)

dirkmueller:
These can be a really large subtree that prevents indexing in solr, so
don't save it.

Large LDAP directories were fetched and then the solr index was incomplete due to oversized data being skipped by solr.

We stop storing user information.

requests?**


chef/cookbooks/ohai/recipes/default.rb
```diff
@@ -112,6 +112,9 @@
 # drop relatively big attributes that we know we won't use
 node.automatic_attrs["kernel"].delete("modules")
 
+# drop the collected Etc.passwd entries, these can be huge
+node.automatic_attrs["etc"].delete("passwd")
```
Itxaka: why not disable the passwd plugin directly on the client.rb file?

`Ohai::Config[:disabled_plugins] << "passwd"`

dirkmueller:
[P0: Bug 1058017 - L3: rabbitmq failing due to chef / crowbar setting wrong dnsmasq.conf](https://trello.com/c/zr2OXXlF/459-p0-bug-1058017-l3-rabbitmq-failing-due-to-chef-crowbar-setting-wrong-dnsmasqconf)

chef/cookbooks/ohai/recipes/default.rb
dirkmueller: hmm, thats even nicer, but it also removes the groups. will check

PR #1341, opened 2017-09-27
Remove unready nodes from deployment

Itxaka:
If we set the experimental skip_unready_nodes config to true and provide
a list of roles, we try to skip nodes that are not ready from the
deployment of the barclamp so we dont need all nodes to be ready to
deploy a barclamp.
    
This is useful when dealing with a high number of nodes (i.e.
nova-compute) and we dont need all those other nodes to be ready to add
a new node to a role

Background: 

just ignore not ready/not accessible nodes (they will be automatically updated on their next local chef-client run)

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -248,9 +248,9 @@ class ServiceObject
 #   process_queue - see what we can execute
 #
 
-  def queue_proposal(inst, element_order, elements, deps, bc = @bc_name)
+  def queue_proposal(inst, element_order, elements, deps, bc = @bc_name, pre_cached_nodes = {})
     Crowbar::DeploymentQueue.new(logger: Rails.logger).queue_proposal(
-      bc, inst, elements, element_order, deps
+      bc, inst, elements, element_order, deps, pre_cached_nodes
     )
   end
 
@@ -945,6 +945,11 @@ class ServiceObject
     # Cache some node attributes to avoid useless node reloads
     node_attr_cache = {}
 
+    # experimental option
+    skip_unready_nodes_enabled = Rails.application.config.experimental.fetch(
+      "skip_unready_nodes", {}
+    ).fetch("enabled", false)
+
     # Part I: Looking up data & checks
     #
     # we look up the role in the database (if there is one), the new one is
@@ -970,50 +975,6 @@ class ServiceObject
     new_elements = new_deployment["elements"]
     element_order = new_deployment["element_order"]
 
-    # When bootstrapping, we don't run chef, so there's no need for queuing
-    if bootstrap
-      pre_cached_nodes = {}
-      # do not try to process the queue in any case
-      in_queue = true
-    else
-      # Attempt to queue the proposal.  If delay is empty, then run it.
-      deps = proposal_dependencies(role)
-      delay, pre_cached_nodes = queue_proposal(inst, element_order, new_elements, deps)
-      unless delay.empty?
-        # force not processing the queue further
-        in_queue = true
-        # FIXME: this breaks the convention that we return a string; but really,
-        # we should return a hash everywhere, to avoid this...
-        return [202, delay]
-      end
-
-      Rails.logger.debug "delay empty - running proposal"
-    end
-
-    # expand items in elements that are not nodes
-    expanded_new_elements = {}
-    new_deployment["elements"].each do |role_name, nodes|
-      expanded_new_elements[role_name], failures = expand_nodes_for_all(nodes)
-      unless failures.nil? || failures.empty?
-        Rails.logger.fatal(
-          "apply_role: Failed to expand items #{failures.inspect} for role \"#{role_name}\""
-        )
-        message = "Failed to apply the proposal: cannot expand list of nodes for role \"#{role_name}\", following items do not exist: #{failures.join(", ")}"
-        Rails.logger.progress("Failed to apply role #{role.name}")
-        update_proposal_status(inst, "failed", message)
-        return [405, message]
-      end
-    end
-    new_elements = expanded_new_elements
-
-    # save list of expanded elements, as this is needed when we look at the old
-    # role. See below the comments for old_elements.
-    if new_elements != new_deployment["elements"]
-      new_deployment["elements_expanded"] = new_elements
-    else
-      new_deployment.delete("elements_expanded")
-    end
-
     # Build a list of old elements.
     # elements_expanded on the old role is guaranteed to exists, as we already
     # ran through apply_role with the old_role.  Cache is used for the case
@@ -1027,6 +988,64 @@ class ServiceObject
         old_elements = old_deployment["elements"]
       end
     end
+
+    # When bootstrapping, we don't run chef, so there's no need for queuing
+    if bootstrap
+      pre_cached_nodes = {}
+      # do not try to process the queue in any case
+      in_queue = true
+    else
+      # Attempt to queue the proposal.  If delay is empty, then run it.
+      deps = proposal_dependencies(role)
+      if skip_unready_nodes_enabled
+        elements_without_unready, pre_cached_nodes = skip_unready_nodes(
+          @bc_name, inst, new_elements, old_elements
+        )
+        delay, pre_cached_nodes = queue_proposal(
+          inst, element_order, elements_without_unready, deps, @bc_name, pre_cached_nodes
+        )
+      else
+        delay, pre_cached_nodes = queue_proposal(inst, element_order, new_elements, deps)
+      end
+
+      unless delay.empty?
+        # force not processing the queue further
+        in_queue = true
+        # FIXME: this breaks the convention that we return a string; but really,
+        # we should return a hash everywhere, to avoid this...
+        return [202, delay]
+      end
+
+      Rails.logger.debug "delay empty - running proposal"
+    end
+
+    new_elements, failures, msg = expand_items_in_elements(new_deployment["elements"])
+    unless failures.nil?
+      Rails.logger.progress("apply_role: Failed to apply role #{role.name}")
+      update_proposal_status(inst, "failed", msg)
+      return [405, msg]
+    end
+
+    # save list of expanded elements, as this is needed when we look at the old
+    # role. See below the comments for old_elements.
+    if new_elements != new_deployment["elements"]
+      new_deployment["elements_expanded"] = new_elements
+    else
+      new_deployment.delete("elements_expanded")
+    end
+
+    if skip_unready_nodes_enabled
+      # if we have removed nodes from the list, make sure to expand them and overwrite the
+      # new_elements var so we dont try to run chef-client on those not-ready nodes
+      new_elements, failures, msg = expand_items_in_elements(elements_without_unready)
+      unless failures.nil?
+        Rails.logger.progress("apply_role: Failed to apply role #{role.name}")
+        update_proposal_status(inst, "failed", msg)
+        return [405, msg]
+      end
+    end
+
+    # use the same order as in the old deployment if the element order is not filled yet
     element_order = old_deployment["element_order"] if (!old_deployment.nil? and element_order.nil?)
 
     Rails.logger.debug "old_deployment #{old_deployment.pretty_inspect}"
@@ -1421,6 +1440,22 @@ class ServiceObject
     []
   end
 
+  def expand_items_in_elements(elements)
+    # expand items in elements that are not nodes
+    expanded_new_elements = {}
+    elements.each do |role_name, nodes|
+      expanded_new_elements[role_name], failures = expand_nodes_for_all(nodes)
+      next if failures.nil? || failures.empty?
+      Rails.logger.fatal(
+        "apply_role: Failed to expand items #{failures.inspect} for role \"#{role_name}\""
+      )
+      msg = "Failed to apply the proposal: cannot expand list of nodes " \
+        "for role \"#{role_name}\", following items do not exist: #{failures.join(", ")}"
+      return [nil, failures, msg]
+    end
+    [expanded_new_elements, nil, nil]
+  end
+
   def add_role_to_instance_and_node(barclamp, instance, name, prop, role, newrole)
     node = Node.find_by_name(name)
     if node.nil?
@@ -1724,5 +1759,35 @@ class ServiceObject
       Rails.logger.error(([e2.message] + e2.backtrace).join("\n"))
       [400, e2.message]
     end
+  end
+
+  def skip_unready_nodes(bc, inst, new_elements, old_elements)
+    logger.debug("skip_unready_nodes: enter for #{bc}:#{inst}")
+    skip_unready_nodes_roles = Rails.application.config.experimental.fetch(
+      "skip_unready_nodes", {}
+    ).fetch("roles", [])
+    pre_cached_nodes = {}
+    cleaned_elements = new_elements.deep_dup
+    skip_unready_nodes_roles.each do |role|
+      # only do something if we have the same role on both old and new
+      next unless new_elements.key?(role) && old_elements.key?(role)
+      # we only can skip nodes that are on both old and new, as we know that those old nodes had
+      # the roles applied and will eventually become consistent with the deployment due to the
+      # periodic chef run
+      shared_elements = new_elements[role] & old_elements[role]
+      shared_elements.each do |n|
+        pre_cached_nodes[n] ||= Node.find_by_name(n)
+        node = pre_cached_nodes[n]
+        next if node.nil?
+        # skip if nodes are on ready or crowbar_upgrade state, we dont need to do anything
+        next if ["ready", "crowbar_upgrade"].include?(node.crowbar["state"])
+        logger.warn(
+          "Node #{n} is skipped until next chef run for #{bc}:#{inst} with role #{role}"
+        )
+        cleaned_elements[role].delete(n)
+      end
+    end
+    logger.debug("skip_unready_nodes: exit for #{bc}:#{inst}")
+    [cleaned_elements, pre_cached_nodes]
   end
 end
--- a/crowbar_framework/config/experimental.yml
+++ b/crowbar_framework/config/experimental.yml
@@ -1,6 +1,26 @@
 default: &default
   disallow_restart:
     enabled: false
+  skip_unready_nodes:
+    enabled: false
+    roles:
+     - bmc-nat-client
+     - ceilometer-agent
+     - deployer-client
+     - dns-client
+     - ipmi
+     - logging-client
+     - nova-compute-ironic
+     - nova-compute-kvm
+     - nova-compute-qemu
+     - nova-compute-vmware
+     - nova-compute-xen
+     - nova-compute-zvm
+     - ntp-client
+     - provisioner-base
+     - suse-manager-client
+     - swift-storage
+     - updater
 
 development:
   <<: *default
--- a/crowbar_framework/lib/crowbar/deployment_queue.rb
+++ b/crowbar_framework/lib/crowbar/deployment_queue.rb
@@ -26,10 +26,9 @@ module Crowbar
     # Receives proposal info (name, barclamp), list of nodes (elements), on which the proposal
     # should be applied, and list of dependencies - a list of {barclamp, name/inst} hashes.
     # It adds them to the queue, if possible.
-    def queue_proposal(bc, inst, elements, element_order, deps)
+    def queue_proposal(bc, inst, elements, element_order, deps, pre_cached_nodes)
       logger.debug("queue proposal: enter for #{bc}:#{inst}")
       delay = []
-      pre_cached_nodes = {}
       begin
         lock = acquire_lock("queue")
 
@@ -41,7 +40,9 @@ module Crowbar
 
         # Delay is a list of nodes that are not in ready state. pre_cached_nodes
         # is an uninteresting optimization.
-        delay, pre_cached_nodes = add_pending_elements(bc, inst, element_order, elements, queue_me)
+        delay, pre_cached_nodes = add_pending_elements(
+          bc, inst, element_order, elements, queue_me, pre_cached_nodes
+        )
 
         # We have all nodes ready.
         if delay.empty?
@@ -333,7 +334,6 @@ module Crowbar
 
       # Delay is the list of nodes that are not ready and are needed for this deploy to run
       delay = []
-      pre_cached_nodes = {}
       begin
         # Check for delays and build up cache
         # FIXME: why?
@@ -374,10 +374,9 @@ module Crowbar
       # Check to see if we should delay our commit until nodes are ready.
       delay = []
       nodes.each do |n|
-        node = Node.find_by_name(n)
+        pre_cached_nodes[n] ||= Node.find_by_name(n)
+        node = pre_cached_nodes[n]
         next if node.nil?
-
-        pre_cached_nodes[n] = node
         # allow commiting proposal for nodes in the crowbar_upgrade state
         state = node.crowbar["state"]
         delay << n if (state != "ready" && state != "crowbar_upgrade") && !delay.include?(n)
--- a/crowbar_framework/lib/crowbar/deployment_queue.rb
+++ b/crowbar_framework/lib/crowbar/deployment_queue.rb
@@ -147,11 +147,11 @@ module Crowbar
 
             next unless dependencies_satisfied?(item.properties["deps"])
 
-            nodes_map = elements_to_nodes_to_roles_map(
+            nodes_map, pre_cached_nodes = elements_to_nodes_to_roles_map(
               prop["deployment"][item.barclamp]["elements"],
               prop["deployment"][item.barclamp]["element_order"]
             )
-            delay, pre_cached_nodes = elements_not_ready(nodes_map.keys)
+            delay, = elements_not_ready(nodes_map.keys, pre_cached_nodes)
             proposal_to_commit = { barclamp: item.barclamp, inst: item.name } if delay.empty?
           end
 
@@ -278,11 +278,12 @@ module Crowbar
     # should be emptied.  FIXME: looks like bc-inst: value should be a list, not
     # a hash?
     def remove_pending_elements(bc, inst, elements)
-      nodes_map = elements_to_nodes_to_roles_map(elements)
+      nodes_map, = elements_to_nodes_to_roles_map(elements)
 
       # Remove the entries from the nodes.
       new_lock("BA-LOCK").with_lock do
         nodes_map.each do |node_name, data|
+          # TODO(itxaka): Maybe we can optimize this to use the node cache safely?
           node = Node.find_by_name(node_name)
           next if node.nil?
           unless node.crowbar["crowbar"]["pending"].nil? or node.crowbar["crowbar"]["pending"]["#{bc}-#{inst}"].nil?
@@ -296,7 +297,7 @@ module Crowbar
     # Create map with nodes and their element list
     # Transform ( {role => [nodes], role1 => [nodes]} hash to { node => [roles], node1 => [roles]},
     # accounting for clusters
-    def elements_to_nodes_to_roles_map(elements, element_order = [])
+    def elements_to_nodes_to_roles_map(elements, element_order = [], pre_cached_nodes = {})
       nodes_map = {}
       active_elements = element_order.flatten
 
@@ -311,7 +312,8 @@ module Crowbar
 
         # Add the role to node's list
         nodes.each do |node_name|
-          if Node.find_by_name(node_name).nil?
+          pre_cached_nodes[node_name] ||= Node.find_by_name(node_name)
+          if pre_cached_nodes[node_name].nil?
             logger.debug "elements_to_nodes_to_roles_map: skipping deleted node #{node_name}"
             next
           end
@@ -320,12 +322,14 @@ module Crowbar
         end
       end
 
-      nodes_map
+      [nodes_map, pre_cached_nodes]
     end
 
     # Get a hash of {node => [roles], node1 => [roles]}
     def add_pending_elements(bc, inst, element_order, elements, queue_me, pre_cached_nodes = {})
-      nodes_map = elements_to_nodes_to_roles_map(elements, element_order)
+      nodes_map, pre_cached_nodes = elements_to_nodes_to_roles_map(
+        elements, element_order, pre_cached_nodes
+      )
 
       # We need to be sure that we're the only ones modifying the node records at this point.
       # This will work for preventing changes from rails app, but not necessarily chef.


Itxaka:

nicolasbock:
This looks so familiar @Itxaka. Did you submit that PR before?

Itxaka:
@nicolasbock similar but not equal :) https://github.com/crowbar/crowbar-core/pull/1339

nicolasbock:
Ah, thanks @Itxaka, and I thought that I am starting to see things :smile:. So it's either this one or #1339 then, right?

nicolasbock:
Ah, thanks @Itxaka, and I thought that I am starting to see things :smile:. So it's either this one or #1339 then, right?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1713,4 +1714,34 @@ def active_update(proposal, inst, in_queue, bootstrap = false)
...
+        next if node.nil?
+        # skip if nodes are on ready or crowbar_upgrade state, we dont need to do anything
+        next if ["ready", "crowbar_upgrade"].include?(node.crowbar["state"])
+        logger.warn(
+          "Node #{n} is being removed from the deployment for #{bc}:#{inst} with role #{role}"
+        )
+        logger.warn(
+          "Node #{n} state is: #{node.crowbar["state"]}"
+        )
+        # TODO: raise an error if this is the only node on the role and its being skipped?
```
skazi0: Why? It will throw errors when you enable skipping for "role1" but you didn't want to deploy this role at all. Note that nova-compute-* roles are mostly exclusive so you won't be able to enable it for more than one without getting errors.

Itxaka: I was thinking that if we have a role with just one node in it, and there is some validation on crowbar-framework to make sure that there is at least 1 node for that role before deploying the barclamp, if we just remove that, then the deployment will fail, probably with an obscure error as it expects at least one node there, so in this case we may be should fail before continuing the execution?

I dont think this affects nova-compute-kvm but maybe it affects dns-client (the other role that this feature is targeted to)

Itxaka:
@nicolasbock indeed, althougth I think the other PR has some issues that are more complex to resolve and doesnt get us where we want, so we should ignore it for the moment.

crowbar_framework/app/models/service_object.rb
skazi0: Do you have a link to some code which does such validation?

nicolasbock:
Thanks @Itxaka

crowbar_framework/app/models/service_object.rb
Itxaka: https://github.com/crowbar/crowbar-openstack/search?utf8=%E2%9C%93&q=validate_at_least_n_for_role&type=

crowbar_framework/app/models/service_object.rb
```diff
@@ -1713,4 +1714,34 @@ def active_update(proposal, inst, in_queue, bootstrap = false)
...
+    skip_unready_nodes_config["roles"].each do |role|
+      next unless elements.key?(role)
+      elements[role].each do |n|
+        # TODO: deal with clusters...how?????
+        node = Node.find_by_name(n)
+        next if node.nil?
+        # skip if nodes are on ready or crowbar_upgrade state, we dont need to do anything
+        next if ["ready", "crowbar_upgrade"].include?(node.crowbar["state"])
+        logger.warn(
+          "Node #{n} is being removed from the deployment for #{bc}:#{inst} with role #{role}"
```
abel-navarro: I guess 'node is being skipped until next chef-client run' does give more information.

crowbar_framework/config/application.rb
```diff
@@ -63,5 +63,7 @@ class Application < Rails::Application
         Rails.logger.warn "Failed to load chef"
       end
     end
+    # experimental options
+    config.experimental = config_for(:experimental)
```
aspiers: How does the Crowbar admin enable this option?  I know this is only a POC currently but it would be nice if this was available via the UI / API.

aspiers: Discussed on IRC - initially it's a `config/experimental.yml` file with toggles for each type of experimental option, but later we can expose it via the UI / API.

Itxaka:
If https://github.com/crowbar/crowbar-core/pull/1307 gets merged then we need to rebase this to have the updated `experimental.yml`

crowbar_framework/app/models/service_object.rb
```diff
@@ -978,6 +978,7 @@ def apply_role(role, inst, in_queue, bootstrap = false)
     else
       # Attempt to queue the proposal.  If delay is empty, then run it.
       deps = proposal_dependencies(role)
+      new_elements = skip_unready_nodes(@bc_name, inst, new_elements)
```
vuntz: You need to pass `old_role` (or `old_elements` actually, but that means moving a block of code) here: if a node was not in the previous deployment, then we should not skip it when it's unready as it means the apply was partly about adding that node.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1713,4 +1714,34 @@ def active_update(proposal, inst, in_queue, bootstrap = false)
       [400, e2.message]
     end
   end
+
+  def skip_unready_nodes(bc, inst, elements)
+    skip_unready_nodes_config = Rails.application.config.experimental["skip_unready_nodes"]
+    # return if the feature is disabled
+    return elements unless skip_unready_nodes_config["enabled"]
+    logger.warn("Experimental skip_unready_nodes feature enabled")
```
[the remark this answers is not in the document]
Itxaka: sure!

crowbar_framework/app/models/service_object.rb
Itxaka: Thing is, if we skip below this lines then we still get the same issue, and that is that the proposal gets queued as there are nodes not ready, thus blocking the proposal deployment until those unready nodes are back online, which is the point of the card attached to this.

Now that I think about it, maybe a better solution would be just to skip showing the unready nodes in the ui list of the barclamp, but that would only work trougth UI, not trougth CLI.

Maybe a different approach is needed for this ¬_¬

crowbar_framework/app/models/service_object.rb
```diff
@@ -1713,4 +1714,34 @@ def active_update(proposal, inst, in_queue, bootstrap = false)
...
+    logger.debug("skip_unready_nodes: enter for #{bc}:#{inst}")
+    cleaned_elements = elements.deep_dup
+    skip_unready_nodes_config["roles"].each do |role|
+      next unless elements.key?(role)
+      elements[role].each do |n|
+        # TODO: deal with clusters...how?????
+        node = Node.find_by_name(n)
```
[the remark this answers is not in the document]
Itxaka: Same as above, if we move this below the line that would fill the `pre_cached_nodes` then the point of the patch disappears as the proposal gets queued.

crowbar_framework/app/models/service_object.rb
Itxaka: +1

crowbar_framework/app/models/service_object.rb
Itxaka: Actually I can move the old_elements block here instead :)

Itxaka: And actually looking at the code, and the logs we should not care if its `old_elements` or `new_elements` here.

If any `old_elements` are not in `new_elements`, they wont get checked, so no skip. And only elements in `new_elements` will be checked for ready-ness to queue the proposal so those are the ones that we want to skip.

vuntz: We should probably chat, because "then we still get the same issue, and that is that the proposal gets queued as there are nodes not ready," is actually what I think we should do if there's a node that was not in the proposal before.

vuntz: One other thing: we need to be super extra careful about one line that happens later on:

```
new_deployment["elements_expanded"] = new_elements
```

Right now, this would break because you would actually have only a subset of elements in there while we really really really want all elements, even the unready ones for that line. So you very likely need to use a second variable.

crowbar_framework/app/models/service_object.rb
vuntz: To be honest, I don't think it's an issue and that's related to my earlier comment: if we only skip unready nodes that were already in the proposal before, then there will still be a node with the role -- even if chef-client is not run with the apply.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1713,4 +1714,34 @@ def active_update(proposal, inst, in_queue, bootstrap = false)
       [400, e2.message]
     end
   end
+
+  def skip_unready_nodes(bc, inst, elements)
+    skip_unready_nodes_config = Rails.application.config.experimental["skip_unready_nodes"]
```
vuntz: Could we possibly re-read the yaml config file here? I'd hate to have people restart crowbar just to change this setting.

Itxaka: well, thats kind of the point of having this into a file instead of just a toggle. We could store that into the database as well so it would be easier, but I was thinking that when you need to go and restart crowbar, there is intention and plan it and follow up on the changes you did, so people dont just go, enable and forget. They understand that they are changing something and it requires a restart to enable, so they may be more careful this way.

All about that physiological barrier, maybe Im reading to much into this, but I know I think it twice and plan when having to restart something for a new option/config/feature VS just toggling something somewhere and forget.

Itxaka:
After a small chat with @vuntz this is more cleaqr to me <sup>it also helps that I read his comments</sup> So I will be changing this to only skip node that are not reqady **if** they were already on the old deployment+new deployment, as those are safe to skip and will catch up when they are back with the help of the periodic chef-client run.

crowbar_framework/config/experimental.yml
```diff
@@ -0,0 +1,14 @@
+default: &default
+  skip_unready_nodes:
+    enabled: false
+    roles:
+     - dns-client
```
vuntz: Btw, here's a list of roles I would add here:

- deployer-client
- provisioner-base
- ipmi
- bmc-nat-client
- dns-client
- logging-client
- ntp-client
- suse-manager-client
- updater
- nova-compute-ironic
- nova-compute-kvm
- nova-compute-qemu
- nova-compute-vmware
- nova-compute-xen
- nova-compute-zvm
- ceilometer-agent

Itxaka:
@vuntz updated with a different approach as we talked, here are the main changes for easy review:

 - `queue_proposal` now accepts a `pre_cached_nodes` argument, which default to empty to not break the previous workflow
 - moved the creation of the `old_elements` to the top of the `apply_role` function so we can use it for knowing which elements are in the old and new deployment
 - changed `skip_unready_nodes` to take into account that the nodes would be both in the old and new deployment
 - `skip_unready_node` now fills `pre_cached_nodes` so we dont incur in a performance loss when activated
 - calls `queue_proposal` with the `pre_cached_nodes` if the experimental feature is enabled
 - run the `queue_proposal` with the list of cleaned nodes if the experimental feature is enabled, otherwise do not touch anything
 - do not override the `elements` var anymore so we dont affect the rest of the code

crowbar_framework/app/models/service_object.rb
```diff
@@ -970,6 +975,20 @@ def apply_role(role, inst, in_queue, bootstrap = false)
     new_elements = new_deployment["elements"]
     element_order = new_deployment["element_order"]
 
+    # Build a list of old elements.
+    # elements_expanded on the old role is guaranteed to exists, as we already
+    # ran through apply_role with the old_role.  Cache is used for the case
+    # when pacemaker barclamp is deactivated.  elements_expanded gets updated
+    # by pacemaker barclamp.
+    old_elements = {}
```
Itxaka: Im not 100% sure that moving this up here does not have any repercussions. I checked the code and I dont think it has, especially because I left the `element_order = old_deployment["element_order"] if (!old_deployment.nil? and element_order.nil?)` line in place, which seems the one that needs to be there due to the comment found on line 1034-1035:

```
 # save list of expanded elements, as this is needed when we look at the old
     # role. See below the comments for old_elements.
# role. See below the comments for old_elements.
```


@vuntz is this ok-ish at 100% to be moved?

crowbar_framework/app/models/service_object.rb
```diff
@@ -970,6 +975,20 @@ def apply_role(role, inst, in_queue, bootstrap = false)
...
+    old_deployment = old_role.override_attributes[@bc_name] unless old_role.nil?
+    unless old_deployment.nil?
+      old_elements = old_deployment["elements_expanded"]
+      if old_elements.nil?
```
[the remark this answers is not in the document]
Itxaka: I may not change this so its easy to review as code moving places instead of code moving+refactoring

Itxaka:
crap, conflicting files found, time to rebase :(

crowbar_framework/lib/crowbar/deployment_queue.rb
```diff
@@ -26,10 +26,10 @@ def initialize(logger: Rails.logger)
     # Receives proposal info (name, barclamp), list of nodes (elements), on which the proposal
     # should be applied, and list of dependencies - a list of {barclamp, name/inst} hashes.
     # It adds them to the queue, if possible.
-    def queue_proposal(bc, inst, elements, element_order, deps)
+    def queue_proposal(bc, inst, elements, element_order, deps, pre_cached_nodes)
       logger.debug("queue proposal: enter for #{bc}:#{inst}")
       delay = []
-      pre_cached_nodes = {}
+      pre_cached_nodes = pre_cached_nodes
```
vuntz: Useless line now? :-)

crowbar_framework/app/models/service_object.rb
vuntz: Looks safe.

crowbar_framework/app/models/service_object.rb
```diff
@@ -978,7 +997,17 @@ def apply_role(role, inst, in_queue, bootstrap = false)
     else
       # Attempt to queue the proposal.  If delay is empty, then run it.
       deps = proposal_dependencies(role)
-      delay, pre_cached_nodes = queue_proposal(inst, element_order, new_elements, deps)
+      if skip_unready_nodes_enabled
+        elements_without_unready, pre_cached_nodes = skip_unready_nodes(
```
vuntz: We're still missing something here. We need to have a line that looks like `new_elements = elements_without_unready` at some point, but after `new_deployment["elements_expanded"] = new_elements` and also we need to make sure to expand all clusters in `elements_without_unready` when we do this.

Without this, we will still try to create the chef lock and run chef-client on unready nodes, which will fail if one of the unready nodes is just down.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1725,4 +1741,38 @@ def active_update(proposal, inst, in_queue, bootstrap = false)
       [400, e2.message]
     end
   end
+
+  def skip_unready_nodes(bc, inst, new_elements, old_elements)
+    skip_unready_nodes_config = Rails.application.config.experimental["skip_unready_nodes"]
```
vuntz: The check earlier on is much safer (with all the `fetch` magic). Also... it's questionable if we need the check in both places ;-)

crowbar_framework/app/models/service_object.rb
```diff
@@ -1725,4 +1741,38 @@ def active_update(proposal, inst, in_queue, bootstrap = false)
...
+    # return if the feature is disabled
+    return new_elements unless skip_unready_nodes_config["enabled"]
+    logger.warn("Experimental skip_unready_nodes feature enabled")
+    logger.debug("skip_unready_nodes: enter for #{bc}:#{inst}")
+    nodes_pre_cached = {}
+    cleaned_elements = new_elements.deep_dup
+    skip_unready_nodes_config["roles"].each do |role|
+      # only do something if we have the same role on both old and new
+      next unless new_elements.key?(role) || old_elements.key?(role)
```
vuntz: Don't you mean `&&`?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1725,4 +1741,38 @@ def active_update(proposal, inst, in_queue, bootstrap = false)
...
+      # we only can skip nodes that are on both old and new, as we know that those old nodes had
+      # the roles applied and will eventually become consistent with the deployment due to the
+      # periodic chef run
+      shared_elements = new_elements[role] & old_elements[role]
+      shared_elements.each do |n|
+        node = Node.find_by_name(n)
```
vuntz: Please do something like

```ruby
nodes_pre_cached[n] ||= Node.find_by_name(n)
node = nodes_pre_cached[n]
```
This matters for nodes that exist in multiple roles in the same proposal.

Btw, any reason to have used `nodes_pre_cached` instead of `pre_cached_nodes` (which is used everywhere else)?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1725,4 +1741,38 @@ def active_update(proposal, inst, in_queue, bootstrap = false)
...
+        next if node.nil?
+        nodes_pre_cached[n] = node
+        # skip if nodes are on ready or crowbar_upgrade state, we dont need to do anything
+        next if ["ready", "crowbar_upgrade"].include?(node.crowbar["state"])
+        logger.warn(
+          "Node #{n} is being removed from the deployment for #{bc}:#{inst} with role #{role}"
+        )
+        logger.warn(
+          "Node #{n} state is: #{node.crowbar["state"]}"
+        )
```
vuntz: I really dislike that there are two log lines and not just one, but if you care about it, then keep it that way.

crowbar_framework/app/models/service_object.rb
Itxaka: ah indeed, I should remove this now as its useless!

crowbar_framework/app/models/service_object.rb
Itxaka: woops

crowbar_framework/app/models/service_object.rb
Itxaka: ah yeah, good catch.

> Btw, any reason to have used nodes_pre_cached instead of pre_cached_nodes (which is used everywhere else)?

Im just that dumb and thougth `nodes_pre_cached` was the actual var, never got to check it back :)

crowbar_framework/app/models/service_object.rb
Itxaka: I think the second line can be removed, I just used as a debug flag during tests to make sure that the nodes being skipped are really in a non-ready status.

crowbar_framework/lib/crowbar/deployment_queue.rb
Itxaka: well duh ¬_¬

Itxaka:
How about now @vuntz ?

 - Fixed all the issues you mentioned
 - Fixed `elements_not_ready` in the `deployment_queue.rb` to use the `pre_cached_nodes` properly
 - Pass the `pre_cached_nodes` var down to `add_pending_elements` in order to take advantage of the cache properly through all the queue methods
 - moved the expansion of the items in the elements into its own method to reuse it `expand_items_in_elements`
 - overrides the `new_elements` with an expanded `elements_without_unready` after we store the `new_deployment` with the original `new_elements`

crowbar_framework/app/models/service_object.rb
```diff
@@ -1421,6 +1432,24 @@ def proposal_dependencies(role)
     []
   end
 
+  def expand_items_in_elements(elements)
+    # expand items in elements that are not nodes
+    expanded_new_elements = {}
+    failures = []
+    elements.each do |role_name, nodes|
+      expanded_new_elements[role_name], failures = expand_nodes_for_all(nodes)
```
vuntz: You're overriding `failures` here: if there's a failure for the first element, but not the second element, this will go unnoticed.

Itxaka: Good point, I should break here like in the original code when we find an error 👍

Itxaka: plus we have no message to return :/

Itxaka:
Updated: 
 - properly return from `expand_items_in_elements` if a failure is found
 - properly return a message from `expand_items_in_elements` if a failure is found so we can show it to the user

crowbar_framework/lib/crowbar/deployment_queue.rb
```diff
@@ -374,10 +374,9 @@ def elements_not_ready(nodes, pre_cached_nodes = {})
       # Check to see if we should delay our commit until nodes are ready.
       delay = []
       nodes.each do |n|
-        node = Node.find_by_name(n)
+        pre_cached_nodes[n] ||= Node.find_by_name(n)
+        node = pre_cached_nodes[n]
```
vuntz: Since you're fixing things in this file: you may want to do a followup PR to use the cache in `elements_to_nodes_to_roles_map` too...

crowbar_framework/app/models/service_object.rb
```diff
@@ -1421,6 +1432,25 @@ def proposal_dependencies(role)
...
+      unless failures.nil? || failures.empty?
+        Rails.logger.fatal(
+          "apply_role: Failed to expand items #{failures.inspect} for role \"#{role_name}\""
+        )
+        message = "Failed to apply the proposal: cannot expand list of nodes for role \"#{role_name}\", following items do not exist: #{failures.join(", ")}"
+        Rails.logger.progress("Failed to apply role #{role.name}")
+        update_proposal_status(inst, "failed", message)
+        return [expanded_new_elements, failures, message]
```
vuntz: To make it even more visible it's an error, I would return `[nil, failures, message]`

crowbar_framework/app/models/service_object.rb
```diff
@@ -1421,6 +1432,25 @@ def proposal_dependencies(role)
...
+    failures = []
+    elements.each do |role_name, nodes|
+      expanded_new_elements[role_name], failures = expand_nodes_for_all(nodes)
+      unless failures.nil? || failures.empty?
+        Rails.logger.fatal(
+          "apply_role: Failed to expand items #{failures.inspect} for role \"#{role_name}\""
+        )
+        message = "Failed to apply the proposal: cannot expand list of nodes for role \"#{role_name}\", following items do not exist: #{failures.join(", ")}"
+        Rails.logger.progress("Failed to apply role #{role.name}")
+        update_proposal_status(inst, "failed", message)
```
vuntz: You do not have access to `inst` here; and I probably wouldn't do it here anyway, but in the caller (that's debatable, of course, but it's too magic otherwise). This may also be true for the log bits -- they may be better in the caller.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1421,6 +1432,25 @@ def proposal_dependencies(role)
...
+        return [expanded_new_elements, failures, message]
+      end
+    end
+    [expanded_new_elements, failures, ""]
```
vuntz: Can you ensure that `failures` is always `nil` or `[]` here? This would simplify the check in the caller.

crowbar_framework/lib/crowbar/deployment_queue.rb
Itxaka: done!

crowbar_framework/app/models/service_object.rb
Itxaka: ah, good thinking!

crowbar_framework/app/models/service_object.rb
Itxaka: done!

crowbar_framework/app/models/service_object.rb
Itxaka: done! message removed altogether

crowbar_framework/app/models/service_object.rb
Itxaka: done!

crowbar_framework/app/models/service_object.rb
```diff
@@ -1014,19 +1039,23 @@ def apply_role(role, inst, in_queue, bootstrap = false)
...
-    old_deployment = old_role.override_attributes[@bc_name] unless old_role.nil?
-    unless old_deployment.nil?
-      old_elements = old_deployment["elements_expanded"]
-      if old_elements.nil?
-        old_elements = old_deployment["elements"]
+    if skip_unready_nodes_enabled
+      # if we have removed nodes from the list, make sure to expand them and overwrite the
+      # new_elements var so we dont try to run chef-client on those not-ready nodes
+      new_elements, failures = expand_items_in_elements(elements_without_unready)
+      unless failures.nil? || failures.empty?
```
vuntz: Remove `|| failures.empty?`

vuntz:
Btw, @Itxaka: any reason to not add all the roles from https://github.com/crowbar/crowbar-core/pull/1341#discussion_r143192986 ?

Itxaka:
> Btw, @Itxaka: any reason to not add all the roles from #1341 (comment) ?

Mainly because Im testing this locally and didnt want to introduce extra roles until Im sure it works as expected. I'll update them in the push Im doing rigth now.

crowbar_framework/app/models/service_object.rb
Itxaka: done

Itxaka:
> Strictly speaking, I think we may be too aggressive with the deployment queue change, as the cache does conflict with the locking mechanism that is used (we acquire a lock before loading nodes, to ensure we have the latest data). But I think the locking mechanism is actually completely broken anyhow, so practically speaking, this may be fine.

@vuntz if this is a concern I can back off the `pre_cached_nodes` use on `remove_pending_elements` which I thikn its the one that may affect the locking mechanism until we have more time to investigate it, or even drop the commit altogether and we can do the improvements later.

vuntz:
> > Strictly speaking, I think we may be too aggressive with the deployment queue change, as the cache does conflict with the locking mechanism that is used (we acquire a lock before loading nodes, to ensure we have the latest data). But I think the locking mechanism is actually completely broken anyhow, so practically speaking, this may be fine.
>
> @vuntz if this is a concern I can back off the pre_cached_nodes use on remove_pending_elements which I thikn its the one that may affect the locking mechanism until we have more time to investigate it, or even drop the commit altogether and we can do the improvements later.

I'll leave this up to you. I'm fine trying this. We just need to keep the change in mind, in case there's some weird bug coming in about nodes having outdated data.

Itxaka:
@vuntz ok, backip up that and leaving a comment so we can investigate a safe way of using the cache instead of blocking ro introducing issues with this

vuntz:
@Itxaka this obviously needs the rebase now

crowbar_framework/app/models/service_object.rb
```diff
@@ -990,21 +1019,17 @@ def apply_role(role, inst, in_queue, bootstrap = false)
...
-        message = "Failed to apply the proposal: cannot expand list of nodes for role \"#{role_name}\", following items do not exist: #{failures.join(", ")}"
-        Rails.logger.progress("Failed to apply role #{role.name}")
-        update_proposal_status(inst, "failed", message)
-        return [405, message]
-      end
+    new_elements, failures = expand_items_in_elements(new_deployment["elements"])
+    unless failures.nil?
+      Rails.logger.fatal(
+        "apply_role: Failed to expand items #{failures.inspect} for role \"#{role_name}\""
+      )
```
vuntz: Argh. When we moved the error message here, outside of the helper, we lost access to `role_name`. I'd suggest to simply remove the mention? Or to go back and build the error message in the helper.

You could actually have the "cannot expand list of nodes..." built in the helper, and then add the prefix here (since we have two slightly different messages).

Itxaka: we still need the message to set `update_proposal_status` with a proper error message. So Im guessing moving the logging into the helper, then returning the msg like before should work for this case

crowbar_framework/lib/crowbar/deployment_queue.rb
```diff
@@ -277,11 +278,12 @@ def dequeue_proposal_no_lock(bc, inst)
     # should be emptied.  FIXME: looks like bc-inst: value should be a list, not
     # a hash?
     def remove_pending_elements(bc, inst, elements)
-      nodes_map = elements_to_nodes_to_roles_map(elements)
+      nodes_map, = elements_to_nodes_to_roles_map(elements)
 
       # Remove the entries from the nodes.
       new_lock("BA-LOCK").with_lock do
         nodes_map.each do |node_name, data|
+          # TODO(itxaka): Maybe we can optimize this to use the node cache safely?
```
vuntz: I think you're correct. But can be in a followup pull request :-)

Itxaka:
as soon as I test the crowbar client changes for the  other PR I'll apply and test this

Itxaka:
Found some issues while testing, it seems that the queue_proposal does **not** like it one bit when you pass a list of cached nodes instead of just names :)

crowbar_framework/app/models/service_object.rb
```diff
@@ -978,7 +997,17 @@ def apply_role(role, inst, in_queue, bootstrap = false)
     else
       # Attempt to queue the proposal.  If delay is empty, then run it.
       deps = proposal_dependencies(role)
-      delay, pre_cached_nodes = queue_proposal(inst, element_order, new_elements, deps)
+      if skip_unready_nodes_enabled
+        elements_without_unready, pre_cached_nodes = skip_unready_nodes(
+          @bc_name, inst, new_elements, old_elements
+        )
+        delay, pre_cached_nodes = queue_proposal(
+          inst, element_order, elements_without_unready, deps, pre_cached_nodes
```
Itxaka: Missing `@bc_name` as parameter, we are now passing `pre_cached_nodes` as `@bc_name` so the queue function gets craaaazy

Itxaka:
After the small fix it seems to work as expected:

```
P, [2017-10-13T10:24:25.984122 #25270:0x0000000621c018] PROGRESS -- Start applying role ipmi-config-default
D, [2017-10-13T10:24:25.994817 #25270:0x0000000621c018] DEBUG -- skip_unready_nodes: enter for ipmi:default
W, [2017-10-13T10:24:26.111455 #25270:0x0000000621c018]  WARN -- Node d52-54-77-77-01-05.vo9.cloud.suse.de is skipped until next chef run for ipmi:default with role bmc-nat-client
W, [2017-10-13T10:24:26.172171 #25270:0x0000000621c018]  WARN -- Node d52-54-77-77-01-05.vo9.cloud.suse.de is skipped until next chef run for ipmi:default with role ipmi
D, [2017-10-13T10:24:26.172287 #25270:0x0000000621c018] DEBUG -- skip_unready_nodes: exit for ipmi:default
D, [2017-10-13T10:24:26.174600 #25270:0x0000000621c018] DEBUG -- delay empty - running proposal
```

Itxaka:
Adding a new node which is on problem status results into a new role, makes the proposal to be queued, even if the node was already in another role and was skipped for that role as expected:

```
D, [2017-10-13T10:32:17.470474 #25272:0x000000057a0260] DEBUG -- skip_unready_nodes: enter for ipmi:default
W, [2017-10-13T10:32:17.684578 #25272:0x000000057a0260]  WARN -- Node d52-54-77-77-01-05.vo9.cloud.suse.de is skipped until next chef run for ipmi:default with role ipmi
D, [2017-10-13T10:32:17.684694 #25272:0x000000057a0260] DEBUG -- skip_unready_nodes: exit for ipmi:default
D, [2017-10-13T10:32:18.661882 #25272:0x000000057a0260] DEBUG -- queue proposal: adding ipmi:default to the queue
```

Itxaka:
@vuntz I think that covers the testing of what we expected this to do:

 - If the node was on old deployment and is on new deployment remove it from the list so we don not queue the proposal
 - If the node was **not** on old deployment and is on new deployment, do nothing to it and let the proposal be queued
 - Separate those checks per role

PR #1342, opened 2017-09-28
Add chef_splay to allowed time without update

skazi0:
After chef_splay was added, node state can change to "unknown" because interval between chef-client runs can be longer.

This change adds chef_splay to calculation to get upper limit for chef-client run properly.

requests?**
https://trello.com/c/WXL2ZVg6/148-crowbar-consider-splay-when-calculating-node-status


skazi0:
[crowbar: Consider splay when calculating node status](https://trello.com/c/WXL2ZVg6/148-crowbar-consider-splay-when-calculating-node-status)

PR #1343, opened 2017-09-28
crowbar: Fill pre_cached_nodes with one search

Itxaka:
As we reset the pre_cached_nodes before calling
remote_chef_client_threads the caceh will be empty and the nodes will
have to be loaded one by one, which in a higher number of nodes
translates to a higher number of chef api calls.

Instead, pre-fill the pre_cached_nodes by doing just one search to
include all the nodes that we are gonna modify. This should help with
performance when we have a high number of nodes to run chef onto.

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1305,6 +1305,13 @@ class ServiceObject
         node_attr_cache[node_name]["windows"]
       end
 
+      # refill pre_cached_nodes in one call instead of allowing
+      # #remote_chef_client_threads to load node by node, thus speeding it up?
+      nodes_to_run.each_slice(50) do |partial_nodes|
+        search = (partial_nodes.map { |n| "name:#{n}" }).join(" OR ")
+        NodeObject.find(search).map { |n| pre_cached_nodes[n.name] = n }
+      end
+
       threads = remote_chef_client_threads(nodes_to_run, pre_cached_nodes,
                                            roles)
 


Itxaka:
tested, seems to work.

When benchmarking `find` vs `find_by_name` with just one node, there seems a minor performance loss of about 0.4s per 100 calls:

```
irb(main):096:0> Benchmark.bm do |x|; x.report { 100.times do ; NodeObject.find("name:d52-54-77-77-01-01.vo9.cloud.suse.de"); end }; end
       user     system      total        real
   1.720000   0.080000   1.800000 (  5.330846)

irb(main):105:0> Benchmark.bm do |x|; x.report { 100.times do ; NodeObject.find_by_name("d52-54-77-77-01-01.vo9.cloud.suse.de"); end }; end
       user     system      total        real
   1.780000   0.050000   1.830000 (  5.128966)
```

Of course this depends a lot on the node load so you can get differences as big as 0.8s or as small as 0.1s 

And when using 5 nodes the thing just skewers a bit for the `find` but not by that much:
```
irb(main):126:0> Benchmark.bm do |x|; x.report { 50.times do ; NodeObject.find("name:d52-54-77-77-01-01.vo9.cloud.suse.de OR name:d52-54-77-77-01-02.vo9.cloud.suse.de OR name:d52-54-77-77-01-03.vo9.cloud.suse.de OR name:d52-54-77-77-01-04.vo9.cloud.suse.de OR name:d52-54-77-77-01-05.vo9.cloud.suse.de"); end }; end
       user     system      total        real
   3.590000   0.070000   3.660000 ( 10.244180)
irb(main):128:0> Benchmark.bm do |x|; x.report { 50.times do ; NodeObject.find_by_name("d52-54-77-77-01-01.vo9.cloud.suse.de"); NodeObject.find_by_name("d52-54-77-77-01-02.vo9.cloud.suse.de");NodeObject.find_by_name("d52-54-77-77-01-03.vo9.cloud.suse.de");NodeObject.find_by_name("d52-54-77-77-01-04.vo9.cloud.suse.de");NodeObject.find_by_name("d52-54-77-77-01-05.vo9.cloud.suse.de"); end }; end
       user     system      total        real
   4.130000   0.110000   4.240000 ( 11.980574)

```

So it doesnt seem that big of a gain. It needs more benchmarking with a higher number of nodes

crowbar_framework/app/models/service_object.rb
```diff
@@ -1305,6 +1305,19 @@ def apply_role(role, inst, in_queue, bootstrap = false)
         node_attr_cache[node_name]["windows"]
       end
 
+      # refill pre_cached_nodes in one call instead of allowing
+      # #remote_chef_client_threads to load node by node, thus speeding it up
+      search = ""
+      nodes_to_run.each do |n|
+        search << "name:#{n}" if n.equal? nodes_to_run.first
```
skazi0: why not do something like `search = nodes_to_run.join(" OR ")` ?

nicolasbock:
@Itxaka When you run these benchmarks again, could you calculate the mean and std. deviation too so we can see what spread the timings have? Without those it's hard to tell how different the results are.

crowbar_framework/app/models/service_object.rb
Itxaka: we would still need to do something extra for the first as the format is "name:NODE1 OR name:NODE2 OR name:NODE3"

skazi0: like this `search - (nodes_to_run.map { |n| "name:"+n }).join(" OR ")` ?

Itxaka: nice! changing it!

Itxaka:
For this to work we would need 2 different patches to the thin web server that crowbar uses and chef-solr or to move the search into batches.

skazi0:
Have you checked what size of batch would work (considering different length of node FQDNs)? I think that batching would be needed anyway to make it safe for large scale clouds.

Itxaka:
> Have you checked what size of batch would work (considering different length of node FQDNs)? I think that batching would be needed anyway to make it safe for large scale clouds.

I just pushed a change to do it on slices of 50. I may have a look when I have time to see if it increases the time for the queries before going further, as Im not sure it will :/

Itxaka:
wrong button!

Itxaka:
not needed

PR #1345, opened 2017-10-02
Combine batches of successive roles for same nodes,

sjamgade:
we can speed-up the application of (n+1)th role if the n if both the
roles are being applied on the same node. This speedup of deployment of
ceilometer by atleast 1m20s (measured: 90sec) and swift by ~20s.

eg. In our 2node deployment if ceilometer{server,central} are always
applied on the same roles, given they have different priorities, they are
to be applied. one after the other.

This does not violate any order constraints as the application procedure
of (n+1)th role is transparent to the nth role

Speed up the chef-client execution


sjamgade:
[Combine batches of successive roles for same nodes](https://trello.com/c/RHz8v6MV/501-combine-batches-of-successive-roles-for-same-nodes)

aspiers:
Two more thoughts:

1. The answer given to "Why is this change necessary?" was "Speed up the chef-client execution", but IIUC that's not precisely what's happening; it's actually *reducing* the number of times `chef-client` is run rather than speeding up execution of any single run - right?  So if I'm right please change the wording on that to clarify.

2. Actually, this also assumes that the cookbooks for (say) ceilometer would be ordered so that in the chef-client runs for the combined batch, any actions for `ceilometer-agent` would be performed after `ceilometer-central`.  And that sounds like it could *potentially* be a fundamental flaw in this optimization.  I'll leave it to @vuntz to comment on whether or not this is likely to be an issue.  I suspect it's fine for most if not all cookbooks, but it would be introducing a new assumption into the design so we'd have to go through all of them to make sure.  In particular I wonder whether it would break compute HA deployment.

aspiers:
And another:

3. Orchestration is [hard](https://etherpad.nue.suse.com/p/cloud-orchestration) ;-/

crowbar_framework/app/models/service_object.rb
```diff
@@ -1169,7 +1171,14 @@ def apply_role(role, inst, in_queue, bootstrap = false)
         end
       end # roles.each
 
-      batches << [roles, nodes_in_batch] unless nodes_in_batch.empty?
+      if pnodes_in_batch == nodes_in_batch && !proles.nil?
```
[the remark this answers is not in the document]
sjamgade:
```element_order = [
[b1]
[c1,c2]
]
batch 1: node1=[b1] node2=[b1]
batch 2: node1=[c1]
```
It think : this could break, if `c1` needs `b1` to be entirely functional

crowbar_framework/app/models/service_object.rb
```diff
@@ -1169,7 +1171,14 @@ def apply_role(role, inst, in_queue, bootstrap = false)
...
+        # proles is same as batches[-1][0]
+        proles << roles
```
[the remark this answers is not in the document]
sjamgade: Exactly.
Will add it comments.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1169,7 +1171,14 @@ def apply_role(role, inst, in_queue, bootstrap = false)
...
+      else
+        proles = roles
```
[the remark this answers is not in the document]
sjamgade: yes

crowbar_framework/app/models/service_object.rb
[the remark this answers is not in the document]
aspiers: > this could break, if c1 needs b1 to be entirely functional

Not necessarily, but that demonstrates my point 2) below - it depends on whether all the recipes for b1 were ordered to happen before the recipes for c1.  For example:

```
element_order = [
[b1]
[c1,c2]
]
batch 1: node1=[b1] node2=[b1]
batch 2: node1=[c1] node2=[c1]
```

This PR would change that to:

```
batch 1: node1=[b1,c1] node2=[b1,c1]
```

If some of c1's recipes come before b1's in the runlist then there's a problem.  Normally runlist ordering is determined by `chef_order`, but this can be overridden per role by `element_run_list_order`.  The only place we actually use that is with pacemaker remotes, but even there I'm not sure we actually use it - maybe the code only has remnants of a previous usage which we reverted when @vuntz came up with a different approach?  I can't remember but @vuntz will know for sure.

sjamgade:
1. Done.

2. [ceilometer-{central,agent} have different priorities](https://github.com/sjamgade/crowbar-openstack/blob/7a1908b1e22f8cdae4491349f3570d8193b36421/chef/data_bags/crowbar/template-ceilometer.json#L60) which makes them **n** and **n+1**. 
Chef client respects these priorities.

3. Cannot agree more.

Itxaka:
Sorry to barge in here but:
>As you can see from my comments, there are some serious code legibility issues here. The vast majority of these issues existed with #apply_role for many years before you joined SUSE ;-) However it's super-important that any changes to it should remove technical debt, not add new technical debt, otherwise we just continue the trend of this method getting worse and worse over the years.

> In particular, #apply_role has quite clearly delimited blocks of logic, so it lends itself very well to extraction of a few new methods (one commit at a time!). This would make it far easier to see the inputs, outputs, and actions of each chunk, and thus the overall flow of logic. If you could do this before enhancing it then you'd probably get a +1000 from me :-)

Totally agree here after having to deal with apply_role, and exactly what I was thinking of doing shortly, extracting `apply_role` into a minimum of 4 methods (as those are the four big steps inside) in order to be able to refactor that monstrous 500 line method into smaller chunks for clarity and *maybe* being able to (kind of) easily try to reduce/refactor smaller pieces inside it.

Plus the test coverage for it is minimal.

@sjamgade I understand this falls outside of the scope of the card but let me know if you plan of doing anything to `apply_role` so we dont do the same work twice in case I got some free time to refactor it :D

sjamgade:
@aspiers gating passed. 
I am not sure if that covers the HA-Compute case you mentioned.

https://ci.suse.de/job/openstack-mkcloud/86719/

https://ci.suse.de/job/openstack-mkcloud/86733/

crowbar_framework/app/models/service_object.rb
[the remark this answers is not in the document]
sjamgade: > If some of c1's recipes come before b1's in the runlist then there's a problem. 
yes this is correct.

I did look at `element_run_list_order` in nova, ceilometer swift and other barclamps. 
Didnt find the issue you mentioned. As each one of them had explicit priorities.

vuntz:
I'm mixed here: this is probably fine at the moment, but this would conflict with a potential change where we restrict what chef is doing on apply (which is something we were slowly heading towards). So if anything, we should have a big comment explaining the shortcut and that this is not compatible with the minimal chef-client approach on apply.

crowbar_framework/app/models/service_object.rb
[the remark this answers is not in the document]
aspiers: Yeah, I think I must have done the grep wrong last time because I see them now.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1069,6 +1069,8 @@ def apply_role(role, inst, in_queue, bootstrap = false)
     proposal = Proposal.where(barclamp: @bc_name, name: inst).first
     save_proposal = false
 
+    nodes_previousbatch = nil
```
aspiers: Nitpick: please be consistent in your variable naming style.  "previous batch" is two words, not one.  Since the existing style is to put `_` between words, put it between all words ("nodes_previous_batch").  Also, since maximising legibility is an important goal, I prefer names which are as close as possible to natural English language, which is why I suggested `nodes_in_prev_batch` and `roles_for_prev_batch`.  Admittedly that's more of a personal preference, but still worth considering.

aspiers:
> Totally agree here after having to deal with apply_role, and exactly what I was thinking of doing shortly, extracting apply_role into a minimum of 4 methods (as those are the four big steps inside) in order to be able to refactor that monstrous 500 line method into smaller chunks for clarity and maybe being able to (kind of) easily try to reduce/refactor smaller pieces inside it.

YES that's *exactly* what I was thinking of!

> Plus the test coverage for it is minimal.
>
> @sjamgade I understand this falls outside of the scope of the card but let me know if you plan of doing anything to apply_role so we dont do the same work twice in case I got some free time to refactor it :D

It would make me sooo happy to see this refactoring done :-)

crowbar_framework/app/models/service_object.rb
```diff
@@ -919,6 +919,35 @@ def self.proposal_to_role(proposal, bc_name)
...
+  #  they are to be applied, one after the other.
+  #
+  #  In other words: it's actually reducing the number of times chef-client
+  #  is run rather than speeding up execution of any single run, by
+  #  merging batches together
+  #
+  #  a batch is [roles, nodes]
+  def mergebatches(batches)
+    merged_batches = []
+    unless bathces.empty?
```
aspiers: `batches` typo. If you are pushing code before testing it, please clearly label and mark the PR as WIP, otherwise there is a risk broken code will get merged.

crowbar_framework/app/models/service_object.rb
```diff
@@ -919,6 +919,35 @@ def self.proposal_to_role(proposal, bc_name)
...
+  #  eg. In our 2node deployment ceilometer{server,central} are always
+  #  applied on the same node, given that they have different priorities,
+  #  they are to be applied, one after the other.
+  #
+  #  In other words: it's actually reducing the number of times chef-client
+  #  is run rather than speeding up execution of any single run, by
+  #  merging batches together
+  #
+  #  a batch is [roles, nodes]
+  def mergebatches(batches)
```
aspiers: Yes, much better to have this as a separate method!

But as mentioned before, please stay consistent with the existing style, so this should be `merge_batches`.

crowbar_framework/app/models/service_object.rb
```diff
@@ -919,6 +919,35 @@ def self.proposal_to_role(proposal, bc_name)
     RoleObject.new role
   end
 
+  #  we can speed-up the application of (n+1)th role if both(n,n+1)
+  #  roles are being applied on the same node.
+  #
+  #  eg. In our 2node deployment ceilometer{server,central} are always
+  #  applied on the same node, given that they have different priorities,
+  #  they are to be applied, one after the other.
```
aspiers: It would be helpful to explicitly mention `element_run_list_order` here and explain its effect on the run-list, otherwise the reader will be wondering why/how they end up with different priorities.

crowbar_framework/app/models/service_object.rb
```diff
@@ -919,6 +919,35 @@ def self.proposal_to_role(proposal, bc_name)
     RoleObject.new role
   end
 
+  #  we can speed-up the application of (n+1)th role if both(n,n+1)
```
aspiers: Thanks for this comment block - definitely important to have it since this batch merging is not immediately obvious without explanation!

crowbar_framework/app/models/service_object.rb
```diff
@@ -919,6 +919,35 @@ def self.proposal_to_role(proposal, bc_name)
...
+  #
+  #  eg. In our 2node deployment ceilometer{server,central} are always
+  #  applied on the same node, given that they have different priorities,
+  #  they are to be applied, one after the other.
+  #
+  #  In other words: it's actually reducing the number of times chef-client
+  #  is run rather than speeding up execution of any single run, by
+  #  merging batches together
+  #
+  #  a batch is [roles, nodes]
```
aspiers: Thanks - I find comments like this really helpful for understanding the data structures being manipulated.

jloehel:
I tested these changes in a cloud environment with 4 nodes and develcloud7 as source. I compared the time with and without the patch. You can see the results here:
graph
It's definitely a performance gain for in example the Barclamp "ceilometer". The roles are merged correctly.
``` 
(Pdb) batches["A"]["ceilometer"]
'[[["ceilometer-server", ["ceilometer-central"], ["ceilometer-agent"]], ["d52-54-77-77-01-01.xxx"]]]'
(Pdb) batches["B"]["ceilometer"]
'[[["ceilometer-server"], ["d52-54-77-77-01-01.xxx"]], [["ceilometer-central"], ["d52-54-77-77-01-01.xxx"]], [["ceilometer-agent"], ["d52-54-77-77-01-01.xxx"]]]'
```
It looks good to me. 👍

crowbar_framework/app/models/service_object.rb
```diff
@@ -919,6 +919,36 @@ def self.proposal_to_role(proposal, bc_name)
     RoleObject.new role
   end
 
+  #  we can speed-up the application of (n+1)th role if both(n,n+1)
+  #  roles are being applied on the same node
+  #
+  #  eg. In our 2node deployment ceilometer{server,central} are always
+  #  applied on the same node, given that they have different priorities,
+  #  they are to be applied, one after the other, these priorities come
+  #  from element_run_list_order
```
aspiers: Two nitpicks here:

1. The priorities come from `chef_order` by default but can be overridden per role by `element_run_list_order`
2. `ceilometer-{server,central}` is just one example, but this code will merge any adjacent batches regardless of what run_list priority the roles in.  Is it possible to make the code automatically check that the run_list priority of batch `n` is lower than the priority of batch `n+1`, and only merge if that is true?  If that was added then I think this would feel safe enough to me for a +1 vote.

PR #1349, opened 2017-10-04
[4.0] Add chef_splay to allowed time without update

skazi0:
backport of #1342


skazi0:
[crowbar: Consider splay when calculating node status](https://trello.com/c/WXL2ZVg6/148-crowbar-consider-splay-when-calculating-node-status)

PR #1350, opened 2017-10-05
crowbar: Forward protocol to rails (bsc#1059733)

cmurphy:
Without this patch, when crowbar is deployed with the HTTP+HTTPS option
and the user is connected with HTTPS, any redirect (such as after a
proposal is applied or when a node is deleted) will force the user back
to an HTTP connection. This is obviously insecure. This patch adds a
setting to the crowbar apache vhost to forward the request protocol to
rails so that the ssl property of the rails request object will be set
and the redirect_to method in rails will do the right thing.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/apache.conf.erb
+++ b/chef/cookbooks/crowbar/templates/default/apache.conf.erb
@@ -45,6 +45,9 @@
     SSLCertificateChainFile <%= @ssl_crt_chain_file %>
     <% end %>
 
+    RequestHeader set X-Forwarded-Proto 'https'
+    RequestHeader set Forwarded 'proto=https'
+
     DocumentRoot /opt/dell/crowbar_framework/public
 
     ErrorLog /var/log/apache2/crowbar-error.log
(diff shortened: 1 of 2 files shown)


chef/cookbooks/crowbar/templates/default/apache.conf.erb
```diff
@@ -44,6 +44,7 @@
     <% unless @ssl_crt_chain_file.nil? or @ssl_crt_chain_file.empty? %>
     SSLCertificateChainFile <%= @ssl_crt_chain_file %>
     <% end %>
+    RequestHeader set X_FORWARDED_PROTO 'https'
```
nicolasbock: According to [1] there is a standardized version of this header called `forwarded`. I have no idea how standardized that really is in terms of browser support, but we could add the 'standard' header just to make doubly sure that the browser gets what we are asking it to do:
```
Forwarded: proto=https
```

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto

cmurphy: Good catch - it looks like the RFC for that is only three years old https://tools.ietf.org/html/rfc7239 so I'm guessing it's not universal yet but would be good for future-proofing

cmurphy: done

cmurphy:
@nicolasbock @rsalevsky I had to rebase this, could you re-review? It's finally passing jenkins

cmurphy:
Backport at https://github.com/crowbar/crowbar-core/pull/1372

PR #1351, opened 2017-10-05
s390x: recognize DASD drives

aplanas:
DASD drives in s390x use a different name schema than other devices, like sda or hda. Change the regular expression to recognize them.

Indirectly this fix a bug in crowbar, where the nodes appears without hard disk information. Also is required for the Cinder barclamp, in order to create the partition and assign it to the cinder-volume group.

Tested in the Dallas system


aplanas:
[P2: Update Crowbar to show, in the node view, the hard disk information](https://trello.com/c/QMP5dvPF/137-p2-update-crowbar-to-show-in-the-node-view-the-hard-disk-information)

aplanas:
@nicolasbock sure thing, done

nicolasbock:
Thanks @aplanas!

PR #1352, opened 2017-10-05
upgrade: Don't fail without openstack db (bsc#1061777)

toabctl:
If only the admin (crowbar) node is deployed, an upgrade from SOC6 to
SOC7 currently fails with:

Creating OpenStack database backup failed

There is no node with the role "database-config-default" in that
case. Do not fail but continue the normal upgrade process if the role
was not found.

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -511,6 +511,12 @@ module Api
         end
 
         psql = postgres_params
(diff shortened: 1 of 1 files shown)


toabctl:
> I assume it was tested in real env.

yes.

toabctl:
Backport for stable/4.0 is https://github.com/crowbar/crowbar-core/pull/1356

PR #1353, opened 2017-10-05
updater: deal with cluster nodes when updating (bsc#983617, bsc#1061834)

Itxaka:
Make the update barclamp aware of HA nodes and deal with them properly.
When HA packages need to be updated, stop the HA services before the update
and start them again after the package have been updated.
When normal packages are updated, set the node in maintenance mode as mentioned
on the HA guide.
Also do not stop or set maintenance mode if the node is a remote_node.

Updates on normal nodes should not be affected by this changes.


Itxaka:
doesnt seem to be needed feel free to restore if so

PR #1354, opened 2017-10-05
nfs-server: Revert systemd Restart= bits for nfs services

vuntz:
This breaks in some mkcloud scenarios:
  nfsserver.service: Service has Restart= setting other than no, which isn't allowed for Type=oneshot services. Refusing.

### DIFF
--- a/chef/cookbooks/nfs-server/metadata.json
+++ b/chef/cookbooks/nfs-server/metadata.json
@@ -6,7 +6,6 @@
     "replacing": {
     },
     "dependencies": {
-      "utils": []
     },
     "groupings": {
     },
--- a/chef/cookbooks/nfs-server/recipes/default.rb
+++ b/chef/cookbooks/nfs-server/recipes/default.rb
@@ -55,7 +55,6 @@ package rpc_service
 service rpc_service do
   action [:enable, :start]
 end
-utils_systemd_service_restart rpc_service
 
 ["/var/log/crowbar/sledgehammer", "/updates"].each do |nfs_dir|
   directory nfs_dir do
@@ -71,7 +70,6 @@ service "nfs-kernel-server" do
   supports restart: true, status: true, reload: true
   action [:enable, :start]
 end
-utils_systemd_service_restart "nfs-kernel-server"
 
 execute "nfs-export" do
   command "exportfs -a"


bmwiedemann:
Found that it broke for "all" but not "all_noreboot" because nothing was calling `systemctl daemon-reload` to have it see the added files

chef/cookbooks/nfs-server/recipes/default.rb
```diff
@@ -55,7 +55,6 @@
 service rpc_service do
   action [:enable, :start]
 end
-utils_systemd_service_restart rpc_service
```
nicolasbock: I am probably misunderstanding something, but on a fresh SP2 install the service script for `rpcbind` is `notify` and not `oneshot`. And for those it's ok to set `Restart` other than no.

chef/cookbooks/nfs-server/recipes/default.rb
```diff
@@ -71,7 +70,6 @@
   supports restart: true, status: true, reload: true
   action [:enable, :start]
 end
-utils_systemd_service_restart "nfs-kernel-server"
```
nicolasbock: The `nfsserver` service is indeed `oneshot`.

chef/cookbooks/nfs-server/metadata.json
```diff
@@ -6,7 +6,6 @@
     "replacing": {
     },
     "dependencies": {
-      "utils": []
```
nicolasbock: Could you quickly explain what we use the `dependencies` key for?

Itxaka: This is the same as the usual `depends` on the metadata.rb of a chef recipe[0]. Supposedly, when you upload a cookbook to the chef server, it would compile the metadata.rb and store a json. If you only have the .json then it will not compile it and store the json directly after some checks.

I believe having only a .json is frowned upon, and the preference is to have the .rb file instead, but this is on more modern chef versions, not sure about chef 10.

[0] https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/apache2/metadata.rb#L7

nicolasbock: Ah thanks @Itxaka! In usual Chef fashion, there is a multitude of ways of doing the same thing! :smile: Maybe this is due to Ruby's Perl heritage. :wink:

chef/cookbooks/nfs-server/recipes/default.rb
dirkmueller: rpc is iirc autostarted by nfs, so removing that should be fine.

PR #1355, opened 2017-10-05
schema_migration: Provide a hook into ServiceObject (bsc#1058876)

rhafer:
This allows to execute barclamp specific maintenance code after a
schema migration ran for a proposal and its role. Subclasses, e.g.
DatabaseService can overwrite the default implementation (which is
a no-op).

This is a PR is laying the groundwork in order to fix:
https://bugzilla.suse.com/show_bug.cgi?id=1058876

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1412,6 +1412,15 @@ class ServiceObject
     # noop by default.
   end
 
+  # This callback provides a hook into the schema migration procedure.
+  # It is called from lib/schema_migration.rb after the schema migrations
+  # for a proposal (and its role, if present) have been executed. Parameters
+  # are the migrated proposal an role objects. The default implementation is
+  # a noop. (To be overwritten in sub classes)
+  def post_schema_migration_callback(proposal, role)
+    # noop by default
+  end
+
   #
   # Inputs: role = RoleObject of proposal being applied/queued.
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/service_object.rb
```diff
@@ -1400,6 +1400,15 @@ def apply_role_post_chef_call(old_role, role, all_nodes)
     # noop by default.
   end
 
+  # This callback provides a hook into the schema migration procedure.
+  # It is called from lib/schema_migration.rb after the schema migrations
+  # for a proposal (and its role, if present) have been executed. Parameters
+  # are the migrated proposal an role objects. The default implementation is
+  # a noop. (To be overwritten in sub classes)
+  def post_schema_migration_callback(proposal, role)
+    # noop by default
```
rsalevsky: maybe just do a `return`?

rhafer: @rsalevsky Why? Does that change anything? (I'd prefer to keep it like that, because the two methods above are actually defined in  the very same way)

rsalevsky: @rhafer Technically it's the same but it would be the usual way how it is done in ruby.

rhafer:
Rebased to get test package rebuild. (CI was failing because of that)

Also changed  post_schema_migration_callback according to @rsalevsky's suggestion

PR #1358, opened 2017-10-09
network: Partly revert 3d24a0f4cb - do not add Restart= for ovs

toabctl:
the openvswitch .service file is of Type=oneshot which is not
compatible with Restart=on-failure .

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -43,7 +43,6 @@ if node[:network][:needs_openvswitch]
   end
   s.run_action :enable
   s.run_action :start
-  utils_systemd_service_restart node[:network][:ovs_service]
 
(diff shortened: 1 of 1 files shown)


toabctl:
The error happens eg on https://ci.suse.de/job/openstack-mkcloud/87146/console . ```/var/log/messages``` from the networking node has:
```
2017-10-05T14:22:41.785485+00:00 d52-54-77-77-01-01 systemd[1]: openvswitch.service: Service has Restart= setting other than no, which isn't allowed for Type=oneshot services. Refusing.
```

toabctl:
Related: https://github.com/crowbar/crowbar-openstack/pull/1357

toabctl:
Backport for stable/4.0 is https://github.com/crowbar/crowbar-core/pull/1360

PR #1359, opened 2017-10-09
nfs-server: Revert systemd Restart= bits for nfs services

vuntz:
This breaks in some mkcloud scenarios:
  nfsserver.service: Service has Restart= setting other than no, which isn't allowed for Type=oneshot services. Refusing.

This is https://github.com/crowbar/crowbar-core/pull/1354 for master. Not sure why I did it in the wrong order :/


rsalevsky:
Correct gating jobs is https://ci.suse.de/job/openstack-mkcloud/87708/console

PR #1361, opened 2017-10-09
[4.0] Backport skip unready nodes deployment

Itxaka:
If we set the experimental skip_unready_nodes config to true and provide
a list of roles, we try to skip nodes that are not ready from the
deployment of the barclamp so we dont need all nodes to be ready to
deploy a barclamp.

This is useful when dealing with a high number of nodes (i.e.
nova-compute) and we dont need all those other nodes to be ready to add
a new node to a role

Background:

just ignore not ready/not accessible nodes (they will be automatically updated on their next local chef-client run)

Backport of: https://github.com/crowbar/crowbar-core/pull/1341

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -248,9 +248,9 @@ class ServiceObject
 #   process_queue - see what we can execute
 #
 
-  def queue_proposal(inst, element_order, elements, deps, bc = @bc_name)
+  def queue_proposal(inst, element_order, elements, deps, bc = @bc_name, pre_cached_nodes = {})
     Crowbar::DeploymentQueue.new(logger: Rails.logger).queue_proposal(
-      bc, inst, elements, element_order, deps
+      bc, inst, elements, element_order, deps, pre_cached_nodes
     )
   end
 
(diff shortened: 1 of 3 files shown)


Itxaka:

crowbar_framework/app/models/service_object.rb
```diff
@@ -1014,19 +1034,21 @@ def apply_role(role, inst, in_queue, bootstrap = false)
...
-    # when pacemaker barclamp is deactivated.  elements_expanded gets updated
-    # by pacemaker barclamp.
-    old_elements = {}
-    old_deployment = old_role.override_attributes[@bc_name] unless old_role.nil?
-    unless old_deployment.nil?
-      old_elements = old_deployment["elements_expanded"]
-      if old_elements.nil?
-        old_elements = old_deployment["elements"]
+    # make sure the role is saved
+    role.save
```
vuntz: These two lines are wrong; I think that's due to a broken rebase.

Itxaka: Something weird happened here as it showed like you never merged the patch that affected this, even with a rebase. Im guessing that it was a github delay or something like that. Fixed it!

vuntz:
> (cherry picked from commit 493c26b)

This is not the right commit id anymore (in the last commit).

Itxaka:
> This is not the right commit id anymore (in the last commit).
indeed, fixed!

PR #1362, opened 2017-10-09
[4.0] Backport disallow chef restarts

Itxaka:
Allow end users to set a barclamp as "no service reboots allowed" to avoid automatic service restarts by chef .

There is several parts to this:

    A new experimental config file to enable/disable experimental features like this one.
        Found at crowbar_framework/config/experimental.yml
        Gets loaded into the rails config at crowbar start, available at Rails.application.config.experimental

    A chef dsl service resource monkey-patching
        allows us to skip the service restart in case that a flag is set on the crowbar-config -> disallow_restarts databag
        allows us to set the flag node[cookbook_name][:requires_restart] if a restart was skipped due to the disallow_restart flag with the barclamp, service name that requires restart, timestamp, and identification if this is a pacemaker service.

    Several new routes on chef framework:
        /restart_management/configuration
            GET: List the barclamps and if service reboots are disallowed for them
            POST(params: disallow_restarts, barclamp): Sets the disallow_restart flag for a barclamp
        /restart_management/restarts
            GET: Lists all of the services needing restarts
            POST(params: node, service): Cleans the restart flag for the given service in the given node

Backport of: https://github.com/crowbar/crowbar-core/pull/1307

### DIFF
--- a/chef/cookbooks/utils/libraries/restart_manager.rb
+++ b/chef/cookbooks/utils/libraries/restart_manager.rb
@@ -1,0 +1,61 @@
+module ServiceRestart
+  class RestartManager
+    def initialize(cookbook_name, node, new_resource, is_pacemaker_service)
+      @cookbook_name = cookbook_name
+      @node = node
+      @service_name = new_resource.name.to_sym
+      @is_pacemaker_service = is_pacemaker_service
+    end
+
+    def cookbook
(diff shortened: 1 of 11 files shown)


Itxaka:
[Avoid Chef re-starting services](https://trello.com/c/BwtWivrg/67-avoid-chef-re-starting-services)

PR #1363, opened 2017-10-11
utils: fix data bag loading on RestartManager

Itxaka:
This used to work as it was part of the chef DSL which provides the
handy helper data_bag_item but it stopped working as soon as we moved it
from there into an external file, as we no longer have access to that
helper method and need to use the full path to the class to load a
databag item


Itxaka:
[Avoid Chef re-starting services](https://trello.com/c/BwtWivrg/67-avoid-chef-re-starting-services)

PR #1365, opened 2017-10-13
utils: Add systemd override LWRP

Itxaka:
Some services require modifications to the system resource limits, such
as the maximum number of open files, that we cannot necessarily predict
a reasonable default for. Without this patch, there is no way to control
those limits from crowbar and they must be changed manually. This patch
adds a LWRP to allow setting resource limits.

Currently only the LimitNOFILE (systemd's name for the maximum
number of open files) limit is supported.

Traditionally this was managed by editing limits for service users
installed by packages in /etc/security/limits.conf. This is no longer
effective under systemd, though the documentation of this reality is
poor and the best reference I can find for this is a comment in a bug
report[1].

Since limits.conf is no longer effective, this patch manages limits by
creating a systemd unit file under /etc/systemd/system for the service
in question. Systemd appends these override unit files to the unit file
installed by the package plus the unit file created by pacemaker if
applicable. Open file limits aren't directly configurable in pacemaker,
as the contents of the unit file are hardcoded[2].

[1] https://bugzilla.redhat.com/show_bug.cgi?id=754285#c1
[2] https://github.com/ClusterLabs/pacemaker/blob/Pacemaker-1.1.17-rc1/lib/services/systemd.c#L603-L625

Co-authored-by: Colleen Murphy <colleen.murphy@suse.com>

This has been extracted from PR https://github.com/crowbar/crowbar-openstack/pull/974

### DIFF
--- a/chef/cookbooks/utils/providers/systemd_override_limits.rb
+++ b/chef/cookbooks/utils/providers/systemd_override_limits.rb
@@ -1,0 +1,99 @@
+#
+# Copyright 2017 SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Support whyrun
+def whyrun_supported?
+  true
+end
+
+action :create do
+  current_resource = @current_resource
+  converge_by("Create #{@new_resource}") do
+    execute "systemctl daemon-reload" do
+      action :nothing
+    end
+
+    directory _get_override_directory do
+      owner "root"
+      group "root"
+      mode "0755"
+    end
+
+    service_resource_name = _get_service_resource_name
+    template _get_override_file_path do
+      source "systemd_override_limits.conf.erb"
+      owner "root"
+      group "root"
+      mode "0644"
+      cookbook "utils"
+      variables(
+        limits: current_resource.limits
+      )
+      notifies :run, resources(execute: "systemctl daemon-reload"), :delayed
+      notifies :restart, resources(service: service_resource_name), :delayed
+    end
+    Chef::Log.info "#{@new_resource} created / updated"
+  end
+end
+
+action :delete do
+  if @current_resource.exists
+    converge_by("Delete #{@new_resource}") do
+      execute "systemctl daemon-reload" do
+        action :nothing
+      end
+
+      override_file_path = _get_override_file_path
+      service_resource_name = _get_service_resource_name
+      file override_file_path do
+        action :delete
+        only_if { ::File.exist?(override_file_path) }
+        notifies :run, resources(execute: "systemctl daemon-reload"), :delayed
+        notifies :restart, resources(service: service_resource_name), :delayed
+      end
+      Chef::Log.info "#{@new_resource} deleted"
+    end
+  else
+    Chef::Log.info "#{@current_resource} doesn't exist - can't delete."
+  end
+end
+
+def load_current_resource
+  @current_resource = Chef::Resource::UtilsSystemdOverrideLimits.new(@new_resource.name)
+  @current_resource.service_name(@new_resource.service_name)
+  @current_resource.limits(@new_resource.limits)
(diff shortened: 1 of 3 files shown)


Itxaka:
[increase max open file limits](https://trello.com/c/E0gqtDEF/151-increase-max-open-file-limits)

chef/cookbooks/utils/providers/systemd_override.rb
```diff
@@ -0,0 +1,104 @@
...
+def _get_unit_name
+  "#{@new_resource.service_name}.service"
+end
+
+def _get_override_directory
+  "/etc/systemd/system/#{_get_unit_name}.d"
+end
+
+def _get_override_file_path
+  "#{_get_override_directory}/60-limits.conf"
```
vuntz: Since this seems to be about limits only, should we call the LWRP `systemd_override_limits`?

Itxaka: +1

Itxaka:
> If you wouldn't mind adding a "Co-authored-by: Colleen Murphy colleen.murphy@suse.com" to the commit message I would appreciate it :)

Indeed I will! Never thought to do so, sorry for that 👍 

> But I thought the idea was to bring this into the existing systemd_service_restart resource, not to create a new one?

Well, Im guessing the idea was to move it into the utils cookbook so both systemd LWRPs are together in the same place. Having them into just one resource could work, but then we would need to do something like with keystone, with different actions for a single resource. As it seems to me right now that there is not much overlap between both actions, separating them seem clearer from a logic pov, but if people want it to have them in a single resource I could work that in.

IMO, its nicer this way so the resource naming makes it clear what is it doing.

chef/cookbooks/utils/templates/default/systemd_override_limits.conf.erb
```diff
@@ -0,0 +1,4 @@
+[Service]
+<% @limits.each do |name, value| -%>
+<%= name %>=<%= value %>
+<% end -%>
```
nicolasbock: missing newline.

Itxaka: done!

chef/cookbooks/utils/providers/systemd_override_limits.rb
```diff
@@ -0,0 +1,104 @@
...
+      end
+      Chef::Log.info "#{@new_resource} deleted"
+    end
+  else
+    Chef::Log.info "#{@current_resource} doesn't exist - can't delete."
+  end
+end
+
+def load_current_resource
+  @current_resource = Chef::Resource::CrowbarOpenstackSystemdOverride.new(@new_resource.name)
```
cmurphy: I think this needs to be renamed too

Itxaka: I was just wondering why the load_current_resource is needed here. The @current_resource.exists can be done so Im not sure what purpose this serves. Any hints?

cmurphy: It's just a requirement for building custom LWRPs https://docs-archive.chef.io/release/12-3/custom_resources.html#load-current-resource

chef/cookbooks/utils/providers/systemd_override_limits.rb
```diff
@@ -0,0 +1,104 @@
...
+      mode "0755"
+    end
+
+    service_resource_name = _get_service_resource_name
+    template _get_override_file_path do
+      source "systemd_override_limits.conf.erb"
+      owner "root"
+      group "root"
+      mode "0644"
+      cookbook "crowbar-openstack"
```
cmurphy: Woops, wrong cookbook now

Itxaka: Fixed!

chef/cookbooks/utils/providers/systemd_override_limits.rb
```diff
@@ -0,0 +1,104 @@
...
+        action :delete
+        only_if { ::File.exist?(override_file_path) }
+        notifies :run, resources(execute: "systemctl daemon-reload"), :delayed
+        notifies :restart, resources(service: service_resource_name), :delayed
+      end
+      override_directory = _get_override_directory
+      directory override_directory do
+        action :delete
+        only_if { ::File.exist?(override_directory) }
+      end
```
cmurphy: When I wrote this this was the only resource managing this directory. Now it's probably not appropriate to try to delete this directory.

Itxaka: ah indeed, good point indeed, we dont want to delete the directory now that the other `systemd_service_restart` resource also manages this dir

Itxaka: fixed!

PR #1366, opened 2017-10-18
network: keep ovs secure fail-mode (bsc#1063772)

stefannica:
Update the wicked pre-up scriptlets for OVS bridges to keep the "secure" ovs fail-mode on all neutron managed bridges (br-public, br-fixed), with the exception of those that are also used to implement the "admin" network, which needs to be enabled during boot.

Previously, all neutron managed OVS bridges were configured by wicked scriptlets to be in a "standalone" fail-mode during boot, which contributed to a condition that could enable outside traffic to be unconditionally forwarded through them if:
 - the node is hard-reset (which doesn't give wicked a chance to remove the OVS bridges as it would during a graceful shutdown)
 - the neutron openvswitch agent fails during startup in such a way that leaves the br-int OVS bridge configured to forward all traffic unconditionally

Implications:
 - after a hard-reset, the interfaces attached to networks implemented using neutron managed OVS bridges will be available until (and if) the neutron openvswitch agent starts up, with the exception of the admin network

Fixes https://bugzilla.suse.com/show_bug.cgi?id=1063772

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -576,6 +576,7 @@ when "suse"
 
       pre_up_script = "/etc/wicked/scripts/#{nic.name}-pre-up"
       datapath_id = get_datapath_id_for_ovsbridge nic.name
+      is_admin_nwk = if_mapping.key?("admin") && if_mapping["admin"].include?(nic.name)
 
       template pre_up_script do
         owner "root"
@@ -584,7 +585,8 @@ when "suse"
         source "ovs-pre-up.sh.erb"
         variables(
           bridgename: nic.name,
-          datapath_id: datapath_id
+          datapath_id: datapath_id,
+          is_admin_nwk: is_admin_nwk
         )
       end
     end
--- a/chef/cookbooks/network/templates/default/ovs-pre-up.sh.erb
+++ b/chef/cookbooks/network/templates/default/ovs-pre-up.sh.erb
@@ -2,4 +2,13 @@
 
 ovs-vsctl br-exists <%= @bridgename %> || exit 0
 ovs-vsctl set Bridge <%= @bridgename %> other-config:datapath-id=<%= @datapath_id %>
+<%
+  # remove the "secure" fail-mode for bridges that share an interface
+  # with the "admin" network, otherwise the admin network will be offline
+  # during boot and until the neutron OVS agent wakes up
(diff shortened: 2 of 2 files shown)


stefannica:
One thing that concerns me about this fix is having additional custom OVS networks (i.e. with `add_ovs_bridge = true`) defined by the user in network.json. These networks will not be enabled during boot.

rhafer:
> One thing that concerns me about this fix is having additional custom OVS networks (i.e. with add_ovs_bridge = true) defined by the user in network.json. These networks will not be enabled during boot.

I would not worry too much about that:

1. When a  new bridge is created is has no failmode set, which is equivalent to "standalone".
2. As part of a clean reboot wicked would during shutdown delete the bridge and recreate it when it's starting again (i.e. it should also have failmode standalone, with just the underlying eth as a port)
3. So it might really only cause problems if:
   - there is a hard reset (power-cycle or fencing)
   - some 3rd piece did set the bridge to secure before the reset
   - and that bridges is connected to a physical interface that is needed for boostrapping the machine

Or maybe I am misunderstanding your concern. If that is the case please clarify.

rhafer:
> Update the wicked pre-up scriptlets for OVS bridges to keep the "secure" ovs fail-mode on all neutron managed bridges (e.g. br-public, br-fixed, br-tunnel), with the exception of those that are also used to implement the "admin" network, which needs to be enabled during boot.

Actually this is only about br-public and br-fixed, As we don't touch br-tunnel in the network barclamp at all. It's purely managed by neutron-ovs-agent and no NIC should be plugged into it.

> Implications:
>     none of the interfaces attached to neutron managed OVS networks defined in network.json will be available until (and if) the neutron barclamp is successfully deployed and the neutron openvswitch agent starts up, with the exception of those attached to the admin network

I think this is not quite true. IIRC if you cleanly reboot a machine the br-fixed and br-public should come up in "standalone" mode regardless of the pre-up script. But as they are removed and recreated from scratch during a clean reboot they should not have any connection to br-int, so we shouldn't have the loop issue in that case. (I guess this is something that we need to verify)

stefannica:
Thanks @rhafer, I updated the implications section, as you suggested

stefannica:
@rhafer I think I should backport this to SOC7. Do you agree ?

rhafer:
> I think I should backport this to SOC7. Do you agree ?

Yes, agreed.

PR #1368, opened 2017-10-18
utils: Fix restart flag removal

Itxaka:
It looks like when you use fetch to fetch an attribute from the node
object we get the direct value, which if modified (deleted/updated)
doesnt get to be reflected into the node object, so when we deleted the
restart flag from a service and the node was saved, the modification we
did was not saved.

Instead, lets just ask for forgiveness when trying to delete the
attribute by using rescue to save from any issues of attributes not
existing when deleting the flags.


Itxaka:
[Chef restarting services - Issues](https://trello.com/c/3mpxiYym/175-chef-restarting-services-issues)

PR #1373, opened 2017-10-19
[4.0] Get 'all' nodes directly from CouchDB

skazi0:
backport of #1370


skazi0:

PR #1374, opened 2017-10-19
Skip chef run on nodes when applying

Itxaka:
Provides a method that barclamps can override to skip a node from the list of nodes in which run is gonna run.

Provides this with a default of allowing all nodes to avoid issues and only will trigger the check if the experimental option is enabled.

Kind of stolen from https://github.com/vuntz/crowbar-core/commit/321c736a17404413505649e286543e7d9a79cc83

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -950,6 +950,10 @@ class ServiceObject
       "skip_unready_nodes", {}
     ).fetch("enabled", false)
 
+    skip_unchanged_nodes_enabled = Rails.application.config.experimental.fetch(
+      "skip_unchanged_nodes", {}
+    ).fetch("enabled", false)
+
     # Part I: Looking up data & checks
     #
     # we look up the role in the database (if there is one), the new one is
@@ -989,24 +993,31 @@ class ServiceObject
       end
     end
 
+    pre_cached_nodes = {}
+    cleaned_elements = nil
+
     # When bootstrapping, we don't run chef, so there's no need for queuing
     if bootstrap
-      pre_cached_nodes = {}
       # do not try to process the queue in any case
       in_queue = true
     else
       # Attempt to queue the proposal.  If delay is empty, then run it.
       deps = proposal_dependencies(role)
-      if skip_unready_nodes_enabled
-        elements_without_unready, pre_cached_nodes = skip_unready_nodes(
-          @bc_name, inst, new_elements, old_elements
-        )
-        delay, pre_cached_nodes = queue_proposal(
-          inst, element_order, elements_without_unready, deps, @bc_name, pre_cached_nodes
-        )
-      else
-        delay, pre_cached_nodes = queue_proposal(inst, element_order, new_elements, deps)
+      cleaned_elements = new_elements.deep_dup
+
+      if skip_unchanged_nodes_enabled
+        cleaned_elements = skip_unchanged_nodes(cleaned_elements, old_role, role)
       end
+
+      if skip_unready_nodes_enabled
+        cleaned_elements, pre_cached_nodes = skip_unready_nodes(
+          @bc_name, inst, cleaned_elements, old_elements
+        )
+      end
+
+      delay, pre_cached_nodes = queue_proposal(
+        inst, element_order, cleaned_elements, deps, @bc_name, pre_cached_nodes
+      )
 
       unless delay.empty?
         # force not processing the queue further
@@ -1034,10 +1045,10 @@ class ServiceObject
       new_deployment.delete("elements_expanded")
     end
 
-    if skip_unready_nodes_enabled
+    unless cleaned_elements.nil?
       # if we have removed nodes from the list, make sure to expand them and overwrite the
       # new_elements var so we dont try to run chef-client on those not-ready nodes
-      new_elements, failures, msg = expand_items_in_elements(elements_without_unready)
+      new_elements, failures, msg = expand_items_in_elements(cleaned_elements)
       unless failures.nil?
         Rails.logger.progress("apply_role: Failed to apply role #{role.name}")
         update_proposal_status(inst, "failed", msg)
@@ -1449,6 +1460,11 @@ class ServiceObject
     []
   end
 
+  def skip_unchanged_node?(node_name, old_role, role)
+    # By default dont skip anything
+    false
+  end
+
   def expand_items_in_elements(elements)
     # expand items in elements that are not nodes
     expanded_new_elements = {}
@@ -1768,6 +1784,17 @@ class ServiceObject
       Rails.logger.error(([e2.message] + e2.backtrace).join("\n"))
       [400, e2.message]
     end
+  end
+
+  def skip_unchanged_nodes(elements, old_role, role)
+    cleaned_elements = {}
+    elements.each do |r|
+      cleaned_elements[r] ||= {}
+      elements[r].each do |node_name|
+        cleaned_elements[r] << node_name unless skip_unchanged_node?(node_name, old_role, role)
+      end
+    end
+    cleaned_elements
   end
 
   def skip_unready_nodes(bc, inst, new_elements, old_elements)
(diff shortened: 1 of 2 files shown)


Itxaka:

crowbar_framework/app/models/service_object.rb
```diff
@@ -1336,6 +1340,12 @@ def apply_role(role, inst, in_queue, bootstrap = false)
         node_attr_cache[node_name]["windows"]
       end
 
+      if skip_unchanged_nodes_enabled
+        nodes_to_run = nodes_to_run.reject do |node_name|
+          skip_batch_for_node?(batch, node_name, old_role, role)
+        end
+      end
```
vuntz: I really believe you want to remove the nodes in the same way you remove nodes from the elements in the other experimental feature. This will be much more beneficial, as it allows not creating the lock files.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1449,6 +1459,11 @@ def proposal_dependencies(role)
     []
   end
 
+  def skip_batch_for_node?(batch, node_name, old_role, role)
+    # By default dont skip anything
+    false
+  end
```
vuntz: I'd suggest adding some commit where we override this method for some of the core barclamps (ntp, dns & logging come to my mind since most of the time when reapplying, there's no change that impacts the client roles -- iirc, for dns there's only one attribute that has an impact)

crowbar_framework/app/models/service_object.rb
Itxaka: mmmh, that is true, that way we dont get delayed by trying to block nodes that we are not gonna run anyway

crowbar_framework/app/models/service_object.rb
Itxaka: mmh, will that make this PR slower to review+merge? I rather do it on a separate one, or use the nova one as example instead than adding it to this and making the whole PR being blocked.

vuntz: Ok.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1045,6 +1049,16 @@ def apply_role(role, inst, in_queue, bootstrap = false)
       end
     end
 
+    if skip_unchanged_nodes_enabled
```
vuntz: Hmm, I was thinking earlier than that ;-) When we check if the proposal has to be queued. Or is it not possible there? (Or maybe between what I suggest and what you did, it actually doesn't change things in reality?)

Itxaka: Any reason to move it into the queue part? I dont see any tbh and introducing this into the part where we check if we should delay the proposal or not seems a bit of mixing logic. The `skip_unready_nodes` was ok to introduce there because it affected the queuing part but this does not seem like a good fit for that, unless you are thinking of expanding this to also cover the unready nodes as well?

Itxaka: Or maybe you want to avoid loading the node objects for nodes we are not going to use anyway? In that case it would make sense to have it earlier that this, yeah...

Itxaka: Umm, Im finding it difficult to manage it that way properly as we could have both options for skipping the nodes active at the same time, so I would have to add more complexity into the mix to manage `new_elements` properly.

Do we really want to add more complexity to `apply_role`? Indeed we could gain some time by not loading the nodes that we are not gonna apply from the chef server but im not sure if in terms of maintainability its worth it :(

vuntz: > Umm, Im finding it difficult to manage it that way properly as we could have both options for skipping the nodes active at the same time, so I would have to add more complexity into the mix to manage new_elements properly.

Let's chat tomorrow; I'm clearly missing something as I don't see the additional complexity.

crowbar_framework/app/models/service_object.rb
```diff
@@ -989,20 +993,28 @@ def apply_role(role, inst, in_queue, bootstrap = false)
...
-      pre_cached_nodes = {}
       # do not try to process the queue in any case
       in_queue = true
     else
       # Attempt to queue the proposal.  If delay is empty, then run it.
       deps = proposal_dependencies(role)
+      filtered_elements = new_elements.deep_dup
+
+      if skip_unchanged_nodes_enabled
+        filtered_elements = skip_unchanged_nodes(new_elements, old_role, role)
```
vuntz: To be future-proof, s/new_elements/filtered_elements/ in case we add some filter before this one.

crowbar_framework/app/models/service_object.rb
```diff
@@ -989,20 +993,28 @@ def apply_role(role, inst, in_queue, bootstrap = false)
...
+
     # When bootstrapping, we don't run chef, so there's no need for queuing
     if bootstrap
-      pre_cached_nodes = {}
       # do not try to process the queue in any case
       in_queue = true
     else
       # Attempt to queue the proposal.  If delay is empty, then run it.
       deps = proposal_dependencies(role)
+      filtered_elements = new_elements.deep_dup
```
vuntz: I tend to understand "filtered elements" as "elements that didn't pass the filter", and you use it for the opposite thing. Is it just me? :-)

crowbar_framework/app/models/service_object.rb
```diff
@@ -989,20 +993,28 @@ def apply_role(role, inst, in_queue, bootstrap = false)
...
       if skip_unready_nodes_enabled
-        elements_without_unready, pre_cached_nodes = skip_unready_nodes(
-          @bc_name, inst, new_elements, old_elements
+        filtered_elements, pre_cached_nodes = skip_unready_nodes(
+          @bc_name, inst, filtered_elements, old_elements
         )
         delay, pre_cached_nodes = queue_proposal(
-          inst, element_order, elements_without_unready, deps, @bc_name, pre_cached_nodes
+          inst, element_order, filtered_elements, deps, @bc_name, pre_cached_nodes
         )
```
vuntz: This would need to be outside of this `if` statement, since right now if only the first filter is used, we don't enter this code path.

crowbar_framework/app/models/service_object.rb
Itxaka: ah thats a good one

crowbar_framework/app/models/service_object.rb
Itxaka: it may be, I changed it to cleaned elements

crowbar_framework/app/models/service_object.rb
Itxaka: yep, indeed, plus with the `cleaned_elements` we can put it outside in any case of the experimental options enabled

crowbar_framework/app/models/service_object.rb
```diff
@@ -989,25 +993,32 @@ def apply_role(role, inst, in_queue, bootstrap = false)
...
-        delay, pre_cached_nodes = queue_proposal(
-          inst, element_order, elements_without_unready, deps, @bc_name, pre_cached_nodes
-        )
-      else
-        delay, pre_cached_nodes = queue_proposal(inst, element_order, new_elements, deps)
       end
 
+      delay, pre_cached_nodes = queue_proposal(
+        inst, element_order, cleaned_elements, deps, pre_cached_nodes
```
vuntz: `@bc_name` is missing in the arguments.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1034,10 +1045,10 @@ def apply_role(role, inst, in_queue, bootstrap = false)
       new_deployment.delete("elements_expanded")
     end
 
-    if skip_unready_nodes_enabled
+    unless cleaned_elements.nil?
       # if we have removed nodes from the list, make sure to expand them and overwrite the
       # new_elements var so we dont try to run chef-client on those not-ready nodes
-      new_elements, failures, msg = expand_items_in_elements(elements_without_unready)
+      new_elements, failures, msg = expand_items_in_elements(filtered_elements)
```
vuntz: `filtered_elements` is the old name of the variable ;-)

Itxaka: fixed

crowbar_framework/app/models/service_object.rb
Itxaka: fixed

PR #1376, opened 2017-10-20
Get role data directly from CouchDB

skazi0:
The performance of `crowbarctl node list` and dashboard needs improvement.

This improves performance of node/role operations on large installations.
Chef API is skipped resulting in less JSON (de)serialization.

requests?**


skazi0:

crowbar_framework/app/models/role_object.rb
```diff
@@ -206,7 +206,8 @@ def self.find_roles_by_search(search)
 
   def self.find_role_by_name(name)
     begin
-      return RoleObject.new Chef::Role.load(name)
+      # get role directly from CouchDB, skip Chef API
+      return RoleObject.new Chef::Role.cdb_load(name)
     rescue Errno::ECONNREFUSED => e
       raise Crowbar::Error::ChefOffline.new
     rescue Net::HTTPServerException => e
```
vuntz: Actually: don't we need to check for some other exception (isn't there one if the role with that name doesn't exist)?

crowbar_framework/app/models/role_object.rb
```diff
@@ -206,7 +206,8 @@ def self.find_roles_by_search(search)
 
   def self.find_role_by_name(name)
     begin
-      return RoleObject.new Chef::Role.load(name)
+      # get role directly from CouchDB, skip Chef API
+      return RoleObject.new Chef::Role.cdb_load(name)
```
Itxaka: as we are hitting the database, we wont get those errors anymore no?

crowbar_framework/app/models/role_object.rb
skazi0: From the code it looks like we should get the same exceptions as both calls use `Chef::REST` behind the scenes just hitting different backends.
I'm wondering if it's worth introducing `Crowbar::Error::CouchDBOffline` exception for these new cases...

crowbar_framework/app/models/role_object.rb
skazi0: It's REST interface so I guess we should get the same kind of errors.

Itxaka: +1

crowbar_framework/app/models/role_object.rb
vuntz: @skazi0 I'd feel slightly more comfortable if we know for sure that a non-existing role raises the same 404 Net::HTTPServerException :-)

skazi0: @vuntz as always... you're right. CouchDB returns empty result-set if nothing is found. Chef API was transforming it to 404.

PR #1377, opened 2017-10-20
Prefetch all roles during node listing

skazi0:
We want to squeeze every second possible out of dashboard/`crowbarctl node list`.

Fetching of all roles and passing them to Node constructor gives some performance improvement over loading them one by one.

requests?**

### DIFF
--- a/crowbar_framework/app/models/node.rb
+++ b/crowbar_framework/app/models/node.rb
@@ -24,8 +24,12 @@ class Node < ChefObject
 
   self.chef_type = "node"
 
-  def initialize(node)
-    @role = RoleObject.find_role_by_name Node.make_role_name(node.name)
+  def initialize(node, role = nil)
+    @role = if role.nil?
+      RoleObject.find_role_by_name Node.make_role_name(node.name)
+    else
+      role
+    end
     if @role.nil?
       # An admin node can exist without a role - so create one
       if !node["crowbar"].nil? and node["crowbar"]["admin_node"]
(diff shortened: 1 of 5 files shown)


skazi0:

crowbar_framework/app/models/node.rb
```diff
@@ -24,8 +24,12 @@ class Node < ChefObject
 
   self.chef_type = "node"
 
-  def initialize(node)
-    @role = RoleObject.find_role_by_name Node.make_role_name(node.name)
+  def initialize(node, role = nil)
+    @role = if role.nil?
+      RoleObject.find_role_by_name Node.make_role_name(node.name)
+    else
+      role
```
vuntz: Maybe some paranoid check that the role name matches `Node.make_role_name(node.name)` would be good?

crowbar_framework/app/models/node.rb
```diff
@@ -1493,10 +1497,11 @@ def find(search)
       end
 
       if nodes.is_a?(Array) and nodes[2] != 0 and !nodes[0].nil?
+        roles = Hash[RoleObject.all.map.collect { |role| [role.name, role] }]
```
vuntz: Asking out loud: should we do this just if `search.nil?`?

crowbar_framework/app/models/node.rb
skazi0: This would be good if we have some kind of assertions as this would most likely be a coding error. I'd skip that for now.

crowbar_framework/app/models/node.rb
skazi0: Good point. Added.

PR #1378, opened 2017-10-20
[4.0] utils: Add systemd override LWRP

Itxaka:
Some services require modifications to the system resource limits, such
as the maximum number of open files, that we cannot necessarily predict
a reasonable default for. Without this patch, there is no way to control
those limits from crowbar and they must be changed manually. This patch
adds a LWRP to allow setting resource limits.

Currently only the LimitNOFILE (systemd's name for the maximum
number of open files) limit is supported.

Traditionally this was managed by editing limits for service users
installed by packages in /etc/security/limits.conf. This is no longer
effective under systemd, though the documentation of this reality is
poor and the best reference I can find for this is a comment in a bug
report[1].

Since limits.conf is no longer effective, this patch manages limits by
creating a systemd unit file under /etc/systemd/system for the service
in question. Systemd appends these override unit files to the unit file
installed by the package plus the unit file created by pacemaker if
applicable. Open file limits aren't directly configurable in pacemaker,
as the contents of the unit file are hardcoded[2].

[1] https://bugzilla.redhat.com/show_bug.cgi?id=754285#c1
[2] https://github.com/ClusterLabs/pacemaker/blob/Pacemaker-1.1.17-rc1/lib/services/systemd.c#L603-L625

Co-authored-by: Colleen Murphy <colleen.murphy@suse.com>
(cherry picked from commit 86e2b6c7cca6a8a92906e43ff9ec8e37eed0c541)

Backport-of: https://github.com/crowbar/crowbar-core/pull/1365

### DIFF
--- a/chef/cookbooks/utils/providers/systemd_override_limits.rb
+++ b/chef/cookbooks/utils/providers/systemd_override_limits.rb
@@ -1,0 +1,99 @@
+#
+# Copyright 2017 SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


Itxaka:
[increase max open file limits](https://trello.com/c/E0gqtDEF/151-increase-max-open-file-limits)

PR #1382, opened 2017-10-23
[4.0] Prefetch all roles during node listing

skazi0:
backport of #1377


skazi0:

PR #1383, opened 2017-10-23
Increase default chef_splay

skazi0:
In large clouds multipe chef-clients connecting to chef-server at the same time can increase load on the admin server.

Chef's SPLAY parameter controls random factor to chef-client interval.
This gives better load distribution for large clouds by avoiding multiple chef-clients connect to chef-server at the same time.

requests?**
https://trello.com/c/YyJWDjew/159-chef-splay-default-settings-documentation


skazi0:
[chef splay default settings/documentation](https://trello.com/c/YyJWDjew/159-chef-splay-default-settings-documentation)

PR #1385, opened 2017-10-23
ohai: don't up interfaces

dirkmueller:
The ohai plugin that forcefully up'ed all physical interfaces
can cause harm on a configuration where multiple physical interfaces
are connected to the same switch/native vlan. in that case
the ARP lookup seemingly randomly flips between the mac addresses
of both causing hard to debug hangs on the cloud due to ip/arp
mismatch.

In our case, eth1 was not configured to be used by crowbar and thus had a MTU of 1500, while traffic leaving eth0 was assuming that replies with 9000 bytes could be received.

in the error case /var/log/chef/client.log had
```
[2017-10-23T11:37:11+00:00] INFO: *** Chef 10.32.2 ***
[2017-10-23T11:47:13+00:00] ERROR: Timeout connecting to 192.168.80.10:4000 for 
/nodes/d8c-dc-d4-af-91-c0.ecp1.cloud.suse.de, retry 1/5
[2017-10-23T11:52:18+00:00] INFO: HTTP Request Returned 401 Unauthorized: Failed
 to authenticate. Please synchronize the clock on your client
[2017-10-23T11:52:18+00:00] ERROR: Net::HTTPServerException: 401 "Unauthorized"
[2017-10-23T11:52:18+00:00] FATAL: Stacktrace dumped to /var/chef/cache/chef-sta
cktrace.out
[2017-10-23T11:52:18+00:00] ERROR: Sleeping for 900 seconds before trying again
```

### DIFF
--- a/chef/cookbooks/ohai/files/default/plugins/crowbar.rb
+++ b/chef/cookbooks/ohai/files/default/plugins/crowbar.rb
@@ -198,8 +198,9 @@ Dir.foreach("/sys/class/net") do |entry|
   tcpdump_out = tcpdump_file(logical_name)
   Ohai::Log.debug("tcpdump to: #{tcpdump_out}")
 
-  if ! File.exists? tcpdump_out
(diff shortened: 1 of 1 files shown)


rhafer:
For some context: https://lwn.net/Articles/45373/

PR #1388, opened 2017-10-24
Added script for setting up chef cluster

skazi0:
There's currently no "automatic" or "documented" way of setting up chef-server cluster in SOC. Having multiple chef-server instances running at the same time to serve chef-client requests improves performance of large scale installations where single chef-server can be a bottle-neck and can be easily saturated by many concurrent requests from chef-clients.

The script can be used to enable/disable multi-worker chef cluster with apache load balancer.

requests?**

**Usage**
`/opt/dell/bin/chef-server-cluster enable` will configure and start chef cluster with default number of workers based on number of CPUs. If there was some `/etc/sysconfig/chef-server`, the old file will be backed up as `/etc/sysconfig/chef-server.old`. If there was some `/etc/apache2/vhosts.d/chef-server-cluster.conf` it will be used as-is. Modifications done by the script are scoped by marker comments.

`/opt/dell/bin/chef-server-cluster enable 8` will configure the cluster to use 8 workers.

`/opt/dell/bin/chef-server-cluster enable` can be called safely when cluster is already enabled and will re-configure the cluster appropriately depending on the arguments.

`/opt/dell/bin/chef-server-cluster disable` will disable the cluster and re-start single-worker chef-server. The config files are renamed with `.disabled` extension and services are restarted.


skazi0:

bin/chef-cluster
```diff
@@ -0,0 +1,115 @@
...
+
+case $cmd in
+  enable)
+    [ "$workercount" == "auto" ] && workercount=$(grep -c ^processor /proc/cpuinfo)
+    [ "$workercount" -ge 2 ] || { echo "ERROR: workercount must be >= 2"; exit 2; }
+
+    # apache configuration
+    mkdir -p "$webroot"
+    a2enmod -q lbmethod_byrequests || a2enmod lbmethod_byrequests
+    test -e "$vhostconfig" || cat <<EOF > $vhostconfig
```
vuntz: Don't we want to always write the file, in case we fix the template in the script?

bin/chef-cluster
```diff
@@ -0,0 +1,115 @@
...
+  ErrorLog /var/log/apache2/chef-cluster-error.log
+  CustomLog /var/log/apache2/chef-cluster-access.log combined
+
+  <Location />
+    Require all granted
+  </Location>
+
+  <Proxy balancer://chefworkers>
+    # $balancerstart
+    # $balancerend
```
vuntz: Should we set some timeout/retry/failontimeout settings? See https://serverfault.com/a/609557 for instance.

bin/chef-cluster
```diff
@@ -0,0 +1,115 @@
...
+  </Proxy>
+
+  <Location /balancer-manager>
+    SetHandler balancer-manager
```
vuntz: The doc says this requires mod_status to be enabled, and I think it's disabled by default. Did you hit that?

bin/chef-cluster
```diff
@@ -0,0 +1,115 @@
...
+}
+
+case $cmd in
+  enable)
+    [ "$workercount" == "auto" ] && workercount=$(grep -c ^processor /proc/cpuinfo)
+    [ "$workercount" -ge 2 ] || { echo "ERROR: workercount must be >= 2"; exit 2; }
+
+    # apache configuration
+    mkdir -p "$webroot"
+    a2enmod -q lbmethod_byrequests || a2enmod lbmethod_byrequests
```
vuntz: To be super-safe, shouldn't we also enable proxy_balancer? (In case the rest of the crowbar setup stops enabling it at some point)

bin/chef-cluster
vuntz: Ah, and also maybe explicitly set lbmethod here, in case several lbmethod modules are enabled?

bin/chef-cluster
```diff
@@ -0,0 +1,115 @@
...
+    echo "$vhostconfig created/updated"
+
+    # chef-server configuration
+    test -e "$serverconfig" && mv "$serverconfig" "$serverconfig.old"
+    cat <<EOF > $serverconfig
+# configuration generated by: chef-cluster script
+PORT=$baseport
+OPTIONS="-N -c $workercount"
+EOF
+    echo "$serverconfig created/updated "
```
vuntz: Trailing space in the string :-) But as said later, I'd drop it completely.

bin/chef-cluster
```diff
@@ -0,0 +1,115 @@
...
+    " "$vhostconfig"
+    echo "$vhostconfig created/updated"
+
+    # chef-server configuration
+    test -e "$serverconfig" && mv "$serverconfig" "$serverconfig.old"
+    cat <<EOF > $serverconfig
+# configuration generated by: chef-cluster script
+PORT=$baseport
+OPTIONS="-N -c $workercount"
+EOF
```
vuntz: I would actually do something like:

```shell
tmpserverconfig=`mktemp`
cat <<EOF > "$tmpserverconfig"
...
EOF
if cmp -s "$serverconfig" "$tmpserverconfig"; then
  rm -f "$tmpserverconfig"
else
  test -e "$serverconfig" && mv "$serverconfig" "$serverconfig.old"
  mv "$tmpserverconfig" "$serverconfig"
  systemctl restart chef-server
fi
```

This avoid backing up the same config file and restarting chef-server with no reason.

Btw, I've removed the `echo` because I would only keep one (see next comment).

bin/chef-cluster
```diff
@@ -0,0 +1,115 @@
...
+    echo "$serverconfig created/updated "
+
+    systemctl restart chef-server
+    systemctl reload apache2
+    echo "services restarted"
+
+    echo "chef cluster enabled with $workercount workers."
```
vuntz: I'd only keep that `echo` and remove all previous ones. Also, maybe "chef-server" instead of just "chef"?

bin/chef-cluster
```diff
@@ -0,0 +1,115 @@
...
+
+  disable)
+    mv "$vhostconfig" "$vhostconfig.disabled" 2> /dev/null || { echo "ERROR: disabling $vhostconfig failed. Is chef cluster enabled?"; exit 3; }
+    echo "$vhostconfig disabled"
+    mv "$serverconfig" "$serverconfig.disabled" 2> /dev/null || { echo "ERROR: disabling $serverconfig failed. Is chef cluster enabled?"; exit 4; }
+    echo "$serverconfig disabled"
+    systemctl reload apache2
+    systemctl restart chef-server
+    echo "services restarted"
+    echo "chef cluster disabled."
```
vuntz: I would accept calling this while it's already disabled, and only restart/reload when needed. I'm also not sure what's the benefit of moving the files with a `.disabled` prefix, since the script can recreate them.

```shell
if test -f "$vhostconfig"; then
  rm -f "$vhostconfig"
  systemctl reload apache2
fi
if test -f "$serverconfig"; then
  rm -f "$serverconfig"
  systemctl restart chef-server
fi
echo "chef-server cluster disabled."
```

bin/chef-cluster
```diff
@@ -0,0 +1,115 @@
...
+    awk -i inplace "
+      /$balancerstart/ { print; f=1} !f;
+      /$balancerend/ {
+        for (i=$baseport;i<$baseport+$workercount;i++) {
+          print \"    BalancerMember http://127.0.0.1:\"i
+        };
+        print;
+        f=0
+      }
+    " "$vhostconfig"
```
vuntz: I would use the same trick as below for the sysconfig file (with a tempfile and cmp) to create the vhost file, and only reload apache if needed.

bmwiedemann:
Can you extend the PR description about why someone would want a chef-server-cluster?

bin/chef-cluster
skazi0: done

bin/chef-cluster
skazi0: 1. if you disable the balancer and enable it again you will get a new file
2. I was thinking more about a case where someone would modify some of the file (e.g. log paths) and would like to change number of workers later. With current implementation the changes won't be lost.

bin/chef-cluster
skazi0: I was thinking about different methods but IMO in our case (i.e. when all workers run on the same host) there is very little difference between them.

skazi0: I will take a look at the timeouts.

bin/chef-cluster
skazi0: Nope. It worked fine without it.

skazi0: We can enable this module just in case if there are no drawbacks of doing so.

bin/chef-cluster
skazi0: Oh, great idea. I'll add this.

bin/chef-cluster
skazi0: The benefit is that these files can be modified by the user. They are not managed by chef and can exist before we enable the cluster and/or be modified while the cluster is running.

bin/chef-cluster
skazi0: done.

bin/chef-cluster
skazi0: removed.

bin/chef-cluster
skazi0: done

skazi0:
I added the logic to limit number of restarts (via temp config file). I kept the behavior of writing vhost file only if it exists and only modifying the worker list.
I wish there was some nice way of handling this in chef-server config. Currently user can have `/etc/sysconfig/chef-server` file with e.g. `CONFIG` or `LOGFILE` settings or even some `OPTIONS` and it's not easy to deal with those cases automatically. We could try to add a separate option (e.g. `WORKERCOUNT=<n>`) and handle this inside systemd service file but this would be more complex solution and might cause some other problems.

bin/chef-cluster
skazi0: OK, I did some more checking:
1. `lbmethod=byrequests` is the default so even if some other ones are enabled, this will be used if nothing is selected explicitly
2. if I kill one of the workers, it's re-spawned but with different port (bug?). in such state failover works correctly out of the box without any special parameters.
3. if I start some dummy server on one of the ports (e.g. `netcat -l 4002`) to simulate hung worker the client times out after standard timeout (default 60sec). I can't find any option to make it failover to some other worker in such cases.

bin/chef-cluster
skazi0: One nice usecase where user might want to modify the generated vhost file is opening access to `balancer-manager` status page from outside of localhost.

bin/chef-cluster
skazi0: RE 3. I added `failontimeout=On` which helps a bit by changing state of worker which timed out as 'error'. This state is kept for `retry` seconds (default 60).

skazi0: OK, I went through the code of `mod_proxy/mod_proxy_balancer` and it seems that failover after backend timeout is not implemented.

vuntz:
> do we support non-standard chef-server port?

We do not have to worry about it in the context of Crowbar, no.

bin/chef-cluster
```diff
@@ -0,0 +1,127 @@
...
+  enable)
+    [ "$workercount" == "auto" ] && workercount=$(grep -c ^processor /proc/cpuinfo)
+    [ "$workercount" -ge 2 ] || { echo "ERROR: workercount must be >= 2"; exit 2; }
+
+    # apache configuration
+    mkdir -p "$webroot"
+    a2enmod -q proxy_balancer || a2enmod proxy_balancer
+    a2enmod -q lbmethod_byrequests || a2enmod lbmethod_byrequests
+    tmpvhostconfig=$(mktemp)
+    test -e "$vhostconfig" && cat "$vhostconfig" > "$tmpvhostconfig" || cat <<EOF > $tmpvhostconfig
```
vuntz: `cat "$vhostconfig" > "$tmpvhostconfig"` is interesting, I would do `cp "$vhostconfig" "$tmpvhostconfig"` ;-)

I understand why you want to keep that; I kind of disagree, but I won't push too hard. My only request is that in the case of re-using the existing config file, we add some `echo` to explain we're reusing it.

Hmm, but actually: what happens if the user runs `chef-cluster enable 2` and then `chef-cluster enable 4`?

skazi0: `chef-cluster enable 4` will re-configure the whole thing to use 4 workers. That's the whole point for re-using existing vhost file.
Use case:
1. `chef-cluster enable 2` (or `auto` or whatever)
2. test the whole thing, modify vhost file a bit (e.g. expose `balancer-manager` or change some timeouts)
3. decide to add more workers
4. `chef-cluster enable 4` will do exactly that without losing any changed done to vhost file.
Same applies when you want to remove some workers.

skazi0:
Renamed `chef-cluster` to `chef-server-cluster`. Some comments were not updated.

skazi0:
@vuntz where in crowbar repo should it be? `/scripts/`?

skazi0:
closing in favor of https://github.com/crowbar/crowbar/pull/2337

PR #1389, opened 2017-10-24
ntp: Make listening net configurable (bsc#1047941)

cmurphy:
In cases where the admin network is firewalled from the outside world,
the cloud ntp servers will not be able to connect to external ntp
servers to synchronize. This patch adds the option to set additional
listen interfaces for nodes with the ntp-server role so those ntp
servers can sync with outside servers.

On the ntp-server host, only packets from configured external NTP servers
and from hosts in the admin network are allowed; everything else is
ignored. In addition to that, the rate-limiting feature has been
enabled for requests coming from the external NTP servers.

On the ntp-client hosts, the NTP configuration is not changed by this patch.

### DIFF
--- a/chef/cookbooks/ntp/recipes/default.rb
+++ b/chef/cookbooks/ntp/recipes/default.rb
@@ -15,7 +15,8 @@
 # limitations under the License.
 #
 
-local_admin_address = Barclamp::Inventory.get_network_by_type(node, "admin").address
+admin_net = Barclamp::Inventory.get_network_by_type(node, "admin")
+local_admin_address = admin_net.address
 
 unless Chef::Config[:solo]
   ntp_config = Barclamp::Config.load("core", "ntp")
@@ -29,6 +30,10 @@ ntp_servers.reject! { |n| n == local_admin_address }
 if node["roles"].include?("ntp-server")
   ntp_servers += node[:ntp][:external_servers]
   is_server = true
+  listen_network_addresses = (node[:ntp][:server_listen_on_networks] || []).map do |network|
+    Barclamp::Inventory.get_network_by_type(node, network).address
+  end
+  listen_network_addresses.reject! { |n| n == local_admin_address }
 end
 
 if node[:platform_family] == "windows"
@@ -73,7 +78,10 @@ else
     source "ntp.conf.erb"
     variables(ntp_servers: ntp_servers,
               admin_interface: local_admin_address,
+              admin_subnet: admin_net.subnet,
+              admin_netmask: admin_net.netmask,
               is_server: is_server,
+              listen_interfaces: listen_network_addresses || [],
               fudgevalue: 10,
               driftfile: driftfile)
     notifies :restart, "service[ntp]"
--- a/chef/cookbooks/ntp/templates/default/ntp.conf.erb
+++ b/chef/cookbooks/ntp/templates/default/ntp.conf.erb
@@ -4,6 +4,9 @@ driftfile <%= @driftfile %>
 
 interface ignore wildcard
 interface listen <%= @admin_interface %>
+<% @listen_interfaces.each do |interface| %>
+interface listen <%= interface %>
+<% end %>
 
 # Enable this if you want statistics to be logged.
 #statsdir /var/log/ntpstats/
@@ -31,9 +34,25 @@ server <%= ntp_server %> iburst minpoll 4
 # that might be intended to block requests from certain clients could also end
 # up blocking replies from your own upstream servers.
 
+<% if @is_server -%>
+# By default, ignore all inbound traffic from all hosts
+restrict -4 default ignore
+restrict -6 default ignore
+
+# Exchange time with all hosts in the admin subnet, but don't allow configuration.
+restrict <%= @admin_subnet %> mask <%= @admin_netmask %> notrap nomodify nopeer noquery
+
+# Exchange time with external NTP servers, but don't allow configuration
+# and use rate-limiting.
+<% @ntp_servers.each do |ntp_server| -%>
+restrict <%= ntp_server %> kod limited notrap nomodify nopeer noquery
+<% end -%>
+
+<% else -%>
 # By default, exchange time with everybody, but don't allow configuration.
-restrict -4 default kod notrap nomodify nopeer noquery
-restrict -6 default kod notrap nomodify nopeer noquery
+restrict -4 default notrap nomodify nopeer noquery
+restrict -6 default notrap nomodify nopeer noquery
+<% end -%>
 
 # Local users may interrogate the ntp server more closely.
 restrict 127.0.0.1
@@ -42,7 +61,6 @@ restrict ::1
 # Clients from this (example!) subnet have unlimited access, but only if
 # cryptographically authenticated.
 #restrict 192.168.123.0 mask 255.255.255.0 notrust
-
 
 # If you want to provide time to your local subnet, change the next line.
 # (Again, the address is an example only.)
--- a/chef/data_bags/crowbar/template-ntp.json
+++ b/chef/data_bags/crowbar/template-ntp.json
@@ -3,14 +3,15 @@
   "description": "Common NTP service for the cluster. An NTP server or servers can be specified and all other nodes will be clients of them.",
   "attributes": {
     "ntp": {
-      "external_servers": [ ]
+      "external_servers": [ ],
+      "server_listen_on_networks": [ ]
     }
   },
   "deployment": {
     "ntp": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 1,
+      "schema-revision": 100,
       "element_states": {
         "ntp-server": [ "readying", "ready", "applying" ],
         "ntp-client": [ "readying", "ready", "applying" ]
--- a/crowbar_framework/app/models/ntp_service.rb
+++ b/crowbar_framework/app/models/ntp_service.rb
@@ -74,6 +74,16 @@ class NtpService < ServiceObject
   def apply_role_pre_chef_call(old_role, role, all_nodes)
     Rails.logger.debug("NTP apply_role_pre_chef_call: entering #{all_nodes.inspect}")
 
+    net_svc = NetworkService.new @logger
+    listen_networks = role.default_attributes["ntp"]["server_listen_on_networks"] || []
+    server_nodes_names = role.override_attributes["ntp"]["elements"]["ntp-server"]
+    server_nodes = server_nodes_names.map { |n| Node.find_by_name(n) }
+    server_nodes.each do |node|
+      listen_networks.each do |network|
+        net_svc.allocate_ip "default", network, "host", node.name
+      end
+    end
+
     save_config_to_databag(old_role, role)
 
     Rails.logger.debug("NTP apply_role_pre_chef_call: leaving")
--- a/chef/data_bags/crowbar/migrate/ntp/100_add_listens.rb
+++ b/chef/data_bags/crowbar/migrate/ntp/100_add_listens.rb
@@ -1,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["server_listen_on_networks"] = ta["server_listen_on_networks"]
+  [a, d]
+end
+
+def downgrade(ta, td, a, d)
+  a.delete("server_listen_on_networks")
+  [a, d]
+end
--- a/chef/data_bags/crowbar/template-ntp.schema
+++ b/chef/data_bags/crowbar/template-ntp.schema
@@ -16,6 +16,11 @@
               "type": "seq",
               "required": true,
               "sequence": [ { "type": "str" } ]
+            },
+            "server_listen_on_networks": {
+              "type": "seq",
+              "required": false,
+              "sequence": [ { "type": "str" } ]
             }
           }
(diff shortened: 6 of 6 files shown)


chef/cookbooks/ntp/recipes/default.rb
```diff
@@ -29,6 +29,10 @@
 if node["roles"].include?("ntp-server")
   ntp_servers += node[:ntp][:external_servers]
   is_server = true
+  listen_network_addresses = (node[:ntp][:listen_networks] || []).map do |network|
+    Barclamp::Inventory.get_network_by_type(node, network).address
```
vuntz: You're assuming that the node has an address allocate for that network, but you're not ensuring this in `apply_role_pre_chef_call`.

chef/data_bags/crowbar/template-ntp.json
```diff
@@ -3,14 +3,15 @@
   "description": "Common NTP service for the cluster. An NTP server or servers can be specified and all other nodes will be clients of them.",
   "attributes": {
     "ntp": {
-      "external_servers": [ ]
+      "external_servers": [ ],
+      "listen_networks": [ ]
```
vuntz: Can we actually put `"admin"` as default value there? This would make this completely configurable, so the user could drop the admin network from there too.

This would require a validation check that `listen_networks` is not empty.

cmurphy: Right now this only applies to the ntp server node and the client nodes still listen only on the admin interface to synchronize with the admin node (or whatever nodes have the ntp-server role). You think that the clients should also be able to listen on a different network?

vuntz: Ah I missed that, sorry. Please discard that part of my comment :-)

vuntz: That being said, it may be good to rename the attribute to clarify that it's only for reaching the external servers?

cmurphy: ++

chef/cookbooks/ntp/recipes/default.rb
cmurphy: Do you think `apply_role_pre_chef_call` method should be doing the allocation or do you think it should be raising an error if an address isn't already allocated? This method is kind of mysterious to me.

cmurphy:
@vuntz thanks for dismissing your review but do you think you could also take another look at it? Especially the additions to `apply_role_pre_chef_call`

crowbar_framework/app/models/ntp_service.rb
```diff
@@ -74,6 +74,16 @@ def transition(inst, name, state)
   def apply_role_pre_chef_call(old_role, role, all_nodes)
     Rails.logger.debug("NTP apply_role_pre_chef_call: entering #{all_nodes.inspect}")
 
+    net_svc = NetworkService.new @logger
+    listen_networks = role.default_attributes["ntp"]["listen_networks"] || []
```
JanZerebecki: Also rename the attribute here to `server_listen_on_networks`.

chef/cookbooks/ntp/templates/default/ntp.conf.erb
```diff
@@ -4,6 +4,9 @@ driftfile <%= @driftfile %>
 
 interface ignore wildcard
 interface listen <%= @admin_interface %>
+<% @listen_interfaces.each do |interface| %>
+interface listen <%= interface %>
```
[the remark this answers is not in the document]
stefannica: Thanks for suggesting this @JanZerebecki . I updated the configuration file to use `restrict default ignore` as the default policy, while at the same time keeping the current configuration unchanged for the admin network (except I also removed the `kod` option, which has no effect unless `limited` is used too).

crowbar_framework/app/models/ntp_service.rb
stefannica: Good catch, thanks !

chef/cookbooks/ntp/templates/default/ntp.conf.erb
```diff
@@ -32,8 +35,11 @@ server <%= ntp_server %> iburst minpoll 4
 # up blocking replies from your own upstream servers.
 
 # By default, exchange time with everybody, but don't allow configuration.
-restrict -4 default kod notrap nomodify nopeer noquery
-restrict -6 default kod notrap nomodify nopeer noquery
+#restrict -4 default kod notrap nomodify nopeer noquery
+#restrict -6 default kod notrap nomodify nopeer noquery
+
+restrict -4 default ignore
+restrict -6 default ignore
```
cmurphy: Might be good to have a comment for these, it's a little confusing to see the "By default" comment above, then the defaults commented out, and then no explanation for this addition.

chef/cookbooks/ntp/templates/default/ntp.conf.erb
```diff
@@ -43,6 +49,7 @@ restrict ::1
 # cryptographically authenticated.
 #restrict 192.168.123.0 mask 255.255.255.0 notrust
 
+restrict <%= @admin_subnet %> mask <%= @admin_netmask %> notrap nomodify nopeer noquery
```
cmurphy: Comment for this would be nice too

cmurphy:
Oh, and thanks for pushing this forward!

chef/cookbooks/ntp/templates/default/ntp.conf.erb
stefannica: Yes I agree, a comment would be helpful here. Done

chef/cookbooks/ntp/templates/default/ntp.conf.erb
stefannica: Done, thanks.

stefannica:
> Oh, and thanks for pushing this forward!

Don't mention it :)

stefannica:
This is how I tested this change:
 - I used a preinstalled, 2 nodes SOC7 setup
 - I updated the crowbar-core package to include these changes
 - then, I configured a `bastion` network by adding it to the network barclamp:

```
    "bastion": {
      "conduit": "intf0",
      "vlan": 100,
      "use_vlan": true,
      "add_bridge": false,
      "mtu": 8900,
      "subnet": "192.168.111.0",
      "netmask": "255.255.255.0",
      "broadcast": "192.168.111.255",
      "ranges": {
        "host": {
          "start": "192.168.111.10",
          "end": "192.168.111.254"
        }
      }
    }
```
  Redeploying the network barclamp didn't configure anything new (i.e. eth0.100) on the admin node, but that was expected, because adding a bastion network is not supported post-installation (the bastion network [is configured on the admin node](https://github.com/crowbar/crowbar-core/blob/stable/4.0/crowbar_framework/app/models/network_service.rb#L365) only [when running in the hardware-installing state](https://github.com/crowbar/crowbar-core/blob/stable/4.0/crowbar_framework/app/models/network_service.rb#L322))
  - redeployed the NTP barclamp with `bastion` added to the `server_listen_on_networks` raw list attribute and `192.168.111.1` added to the list of external servers. This had the desired effect of adding the `eth0.100` - i.e. bastion interface - to the admin node, and the NTP configuration file was updated accordingly:
```
interface ignore wildcard
interface listen 192.168.249.10
interface listen 192.168.111.10
...
# By default, ignore all inbound traffic from all hosts
restrict -4 default ignore
restrict -6 default ignore

# Exchange time with all hosts in the admin subnet, but don't allow configuration.
restrict 192.168.249.0 mask 255.255.255.0 notrap nomodify nopeer noquery
restrict 192.168.111.1 kod limited notrap nomodify nopeer noquery

# Exchange time with external NTP servers, but don't allow configuration
# and use rate-limiting.
restrict 0.opensuse.pool.ntp.org kod limited notrap nomodify nopeer noquery
restrict 1.opensuse.pool.ntp.org kod limited notrap nomodify nopeer noquery
restrict 2.opensuse.pool.ntp.org kod limited notrap nomodify nopeer noquery
restrict 3.opensuse.pool.ntp.org kod limited notrap nomodify nopeer noquery

# Local users may interrogate the ntp server more closely.
restrict 127.0.0.1
restrict ::1
```
 - tested that NTP traffic between admin and the other cloud nodes is unaffected by these changes
 - then, I manually added the 192.168.111.1 IP on my host, where I had an NTP server running and reconfigured my NTP server to listen on that interface
 - I verified that the internal clients are unaffected and the configured external servers are reachable:
```
     remote           refid      st t when poll reach   delay   offset  jitter
==============================================================================
 LOCAL(0)        .LOCL.          10 l  229    8    0    0.000    0.000   0.000
-192.168.111.1   246.75.89.154    3 u   12   16  377    0.108    3.085   0.087
*athena.foxenhie 192.53.103.104   2 u   13   16  377   14.723    0.579   0.270
+moocat.me       134.71.66.21     2 u   12   16  337   15.010   -0.023   0.301
+greenore.zeip.e 140.203.204.77   2 u   13   16  377   14.760    0.095   0.275
-hadb1.smatwebde 216.218.254.202  2 u   13   16  377  121.046    2.516   0.160
```

cmurphy:
Was this verified with the customer? From https://bugzilla.suse.com/show_bug.cgi?id=1047941 it looks like the old version was satisfactory for them, so I'm a little hesitant that we're going in a different direction.

chef/cookbooks/ntp/templates/default/ntp.conf.erb
```diff
@@ -31,9 +34,25 @@ server <%= ntp_server %> iburst minpoll 4
...
+<% @ntp_servers.each do |ntp_server| -%>
+restrict <%= ntp_server %> kod limited notrap nomodify nopeer noquery
+<% end -%>
+
+<% else -%>
 # By default, exchange time with everybody, but don't allow configuration.
-restrict -4 default kod notrap nomodify nopeer noquery
-restrict -6 default kod notrap nomodify nopeer noquery
+restrict -4 default notrap nomodify nopeer noquery
+restrict -6 default notrap nomodify nopeer noquery
```
JanZerebecki: Delete the above 3 lines, as they override the default settings earlier in this change.

JanZerebecki:
> I'm a little hesitant that we're going in a different direction.

This is not a different direction. The changes that the customer verified are still there.

chef/cookbooks/ntp/templates/default/ntp.conf.erb
stefannica: @JanZerebecki , as I mentioned earlier, `kod` has no effect unless used in combination with `limited`, which is expressed as a warning in the logs:
```
KOD does nothing without LIMITED.
```
Also see [Reinhard's comment](https://bugzilla.suse.com/show_bug.cgi?id=1047941#c18) in the bug log.

stefannica:
> Was this verified with the customer? From https://bugzilla.suse.com/show_bug.cgi?id=1047941 it looks like the old version was satisfactory for them, so I'm a little hesitant that we're going in a different direction.

You're right in saying that the current version is quite different than what the customer tested as a PTF, but I also have to agree that we shouldn't make the NTP server public without configuring the minimum recommended [access restrictions](http://support.ntp.org/bin/view/Support/AccessRestrictions).

I'll try getting the customer to test this again as a PTF, but I have my doubts, seeing how the bug was already closed and then re-opened 3 months later because the patch wasn't merged.

Besides, I made sure to minimize the impact to the existing NTP external servers and internal clients/servers by keeping the same access policy for hosts in the admin network and for the configured external servers (I updated the commit message to state that), so this will only deny NTP service access to _unknown_ parties: hosts that are outside the admin network and not in the list of NTP external servers.

chef/cookbooks/ntp/templates/default/ntp.conf.erb
JanZerebecki: My suggestion has nothing to do with kod, but that these overwrite the ignore flag for default. From man ntp.conf: "last  match  found  defining  the  restriction  flags  associated with the entry."

stefannica: In that case, I think you're reading this wrong: there's an if/else block here that doesn't let that happen.

cmurphy:
[**@stefannica**](https://github.com/stefannica) commented on [May 17, 2018, 3:32 AM PDT](https://github.com/crowbar/crowbar-core/pull/1389#issuecomment-389821809 "2018-05-17T10:32:19Z - Replied by Github Reply Comments"):
> > Was this verified with the customer? From https://bugzilla.suse.com/show_bug.cgi?id=1047941 it looks like the old version was satisfactory for them, so I'm a little hesitant that we're going in a different direction.
> 
> You're right in saying that the current version is quite different than what the customer tested as a PTF, but I also have to agree that we shouldn't make the NTP server public without configuring the minimum recommended [access restrictions](http://support.ntp.org/bin/view/Support/AccessRestrictions).
> 
> I'll try getting the customer to test this again as a PTF, but I have my doubts, seeing how the bug was already closed and then re-opened 3 months later because the patch wasn't merged.
> 
> Besides, I made sure to minimize the impact to the existing NTP external servers and internal clients/servers by keeping the same access policy for hosts in the admin network and for the configured external servers (I updated the commit message to state that), so this will only deny NTP service access to _unknown_ parties: hosts that are outside the admin network and not in the list of NTP external servers.

Okay, +1 from me

PR #1391, opened 2017-10-25
[4.0]  Skip chef run on nodes when applying

Itxaka:
Provides a method that barclamps can override to skip a node from the list of nodes in which run is gonna run.

Provides this with a default of allowing all nodes to avoid issues and only will trigger the check if the experimental option is enabled.

Kind of stolen from vuntz/crowbar-core@321c736

Backport-of: https://github.com/crowbar/crowbar-core/pull/1374

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -950,6 +950,10 @@ class ServiceObject
       "skip_unready_nodes", {}
     ).fetch("enabled", false)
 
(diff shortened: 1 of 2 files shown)


Itxaka:

Itxaka:
[Backport "do not require all nodes to be ready" patch to SOC6CD](https://trello.com/c/3Ae6H2o4/177-backport-do-not-require-all-nodes-to-be-ready-patch-to-soc6cd)

PR #1394, opened 2017-10-25
Show IPMI links earlier

skazi0:
When IPMI interfaces are pre-configured and/or read-only mode is enabled discovered IPMI info was not used to display links in UI.

After this change, the links will be shown as soon as there is some IPMI address discovered from the BMC.

requests?**
https://trello.com/c/Gxhc3uD3/107-show-ipmi-link-in-ui-as-soon-as-the-addresses-are-available-in-crowbarwall

### DIFF
--- a/crowbar_framework/app/helpers/nodes_helper.rb
+++ b/crowbar_framework/app/helpers/nodes_helper.rb
@@ -541,7 +541,7 @@ module NodesHelper
 
   def node_link_list(node)
     link_list = [].tap do |result|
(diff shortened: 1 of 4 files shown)


skazi0:
[Show IPMI link in UI as soon as the addresses are available in crowbar_wall](https://trello.com/c/Gxhc3uD3/107-show-ipmi-link-in-ui-as-soon-as-the-addresses-are-available-in-crowbarwall)

PR #1396, opened 2017-10-25
[4.0] crowbar_framework: fix skip_unchanged_nodes

Itxaka:
There was 2 mistakes on the original method. The first, we were
iterating over the elements but only using one value instead of k,v so
it led to the value of r being the whole k+v which broke the hash
lookup. The second is that the initialization of the cleaned_elements
was a hash but a hash doesn't have the << method

For the first, the correct way of iterating over a hash and extracting
the k,v is now used

For the second, we initialize the attribute to an array as expected

(cherry picked from commit 70ab89e)
Backport-of: https://github.com/crowbar/crowbar-core/pull/1395


Itxaka:

PR #1400, opened 2017-10-27
Disable subject_cap danger check

cmurphy:
Since we usually prefix commit messages with their component, it doesn't
make sense to require it to be capitalized.


cmurphy:
@nicolasbock yes it would, this was just easier :) I can work on adding that back but I thought I'd get this out of people's way quickly for now.

nicolasbock:
:smile: We should make sure though that we don't forget about adding it at some point. I like these automatic checks and the more they can help us make commit messages consistent the better.

PR #1401, opened 2017-10-30
crowbar_framework: fix filtered nodes in apply_role

Itxaka:
As a side effect of introducing the experimental options for unchanged
and unready nodes into apply_role, there could be some cases in which
even if the nodes where removed from the list to be deployed we still
deployed them. For that issue, a check is done before adding the nodes
into the batches to check that indeed those nodes are going to be
deployed.

There was also a wrong comparision being done of the old_nodes vs
new_nodes in which we used the filtered list of nodes, but we want to
use the original node list without unfiltered to make sure that the
deployments didnt change. For this,the unfiltered list of nodes in the
new deployment is regenerated and used for the role comparision.

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1096,6 +1096,10 @@ class ServiceObject
     proposal = Proposal.where(barclamp: @bc_name, name: inst).first
     save_proposal = false
 
+    # recreate new_elements with all elements, in case some of the nodes
+    # were hit by the filtering of unready/unchanged nodes,
+    # as we need the full old/new deployment list to compare the role changes
+    new_elements_unfiltered, = expand_items_in_elements(new_deployment["elements"])
     # element_order is an Array where each item represents a batch of roles and
     # the batches must be applied sequentially in this order.
     element_order.each do |roles|
@@ -1114,7 +1118,7 @@ class ServiceObject
         next if role_name =~ /_remove$/
 
         old_nodes = old_elements[role_name] || []
-        new_nodes = new_elements[role_name] || []
+        new_nodes = new_elements_unfiltered[role_name] || []
 
         Rails.logger.debug "Preparing role #{role_name} for batch:"
         Rails.logger.debug "  Nodes in old applied proposal for role: #{old_nodes.inspect}"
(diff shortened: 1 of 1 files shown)


Itxaka:

susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
04686cd8a568124f6dd37f5b7bc44938c0783c31</td>
    </tr>
  </tbody>
</table>

</p>

crowbar_framework/app/models/service_object.rb
```diff
@@ -1188,6 +1192,9 @@ def apply_role(role, inst, in_queue, bootstrap = false)
               next
             end
 
+            # skip adding nodes to the batch unless they are really in the list to be deployed
+            next unless new_elements[role_name].include?(node_name)
```
vuntz: Can you move this to either:

 - a few lines earlier, before `pre_cached_nodes[node_name] ||= Node.find_by_name(node_name)` (to avoid loading data from chef without a good reason)
 - or after changing `pending_node_actions` (since at that point, we have the chef data and `pending_node_actions` is used to ensure consistency, and is worth doing if we have the chef data)

I would tend to go with the latter (the consistency is worth it, and I'm nearly sure we load the chef data in `apply_role_pre_chef_call` anyhow), but would be fine with the former.

vuntz: Btw, whatever option you pick, please add a comment to the code explaining the impact (cost of loading chef data vs consistency)

Itxaka: umm, Im nearly sure that we added this in this place to avoid having a wrong `pending_node_actions` with a node that then its not included on the `batch`. Although I dont remember if we tested setting it below the pending_node_actions, but I dont think it should matter much, as it would only affect nodes that have not changed attributes or roles we dont need to re-save the role list to them (thus speeding up time as well?)

we tested this extensively adding roles, removing roles and moving a node between roles and it look like it worked as expected.

@ilausuch do you remember our thinking process for setting this in here?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1170,6 +1174,10 @@ def apply_role(role, inst, in_queue, bootstrap = false)
         # If new_nodes is empty, we are just removing the proposal.
         unless new_nodes.empty?
           new_nodes.each do |node_name|
+            # skip adding nodes to the batch unless they are really in the list to be deployed
+            # do it before the Node load to avoid doing the call if the node is not there,
+            # as we dont want to spend cycles doing extra calls not needed
+            next unless new_elements[role_name].include?(node_name)
```
vuntz: I would love an empty line after that one :-)

crowbar_framework/app/models/service_object.rb
```diff
@@ -1170,6 +1174,10 @@ def apply_role(role, inst, in_queue, bootstrap = false)
         # If new_nodes is empty, we are just removing the proposal.
         unless new_nodes.empty?
           new_nodes.each do |node_name|
+            # skip adding nodes to the batch unless they are really in the list to be deployed
+            # do it before the Node load to avoid doing the call if the node is not there,
+            # as we dont want to spend cycles doing extra calls not needed
```
vuntz: Ah, actually, I'd request a comment like: "This also means we don't ensure that the node has all required roles (through the use of pending_node_actions); it's a reasonable trade-off in the context of this specific optimization, as the nodes should have all roles already (unless there's a bug or the customer broke things manually).".

Okay, it's too long, but you get the idea ;-)

PR #1402, opened 2017-10-30
[4.0] crowbar_framework:fix filtered nodes in apply_role

Itxaka:
As a side effect of introducing the experimental options for unchanged
and unready nodes into apply_role, there could be some cases in which
even if the nodes where removed from the list to be deployed we still
deployed them. For that issue, a check is done before adding the nodes
into the batches to check that indeed those nodes are going to be
deployed.

There was also a wrong comparision being done of the old_nodes vs
new_nodes in which we used the filtered list of nodes, but we want to
use the original node list without unfiltered to make sure that the
deployments didnt change. For this,the unfiltered list of nodes in the
new deployment is regenerated and used for the role comparision.

(cherry picked from commit adaa2b548543faa3c88c7262415bde369c1f7650)
Backport-of: https://github.com/crowbar/crowbar-core/pull/1401

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1096,6 +1096,10 @@ class ServiceObject
     proposal = Proposal.where(barclamp: @bc_name, name: inst).first
     save_proposal = false
 
+    # recreate new_elements with all elements, in case some of the nodes
+    # were hit by the filtering of unready/unchanged nodes,
+    # as we need the full old/new deployment list to compare the role changes
+    new_elements_unfiltered, = expand_items_in_elements(new_deployment["elements"])
     # element_order is an Array where each item represents a batch of roles and
     # the batches must be applied sequentially in this order.
     element_order.each do |roles|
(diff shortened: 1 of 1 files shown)


Itxaka:

Itxaka:
@vuntz cherry-picked latest commit with fixes from master

crowbar_framework/app/models/service_object.rb
```diff
@@ -1188,6 +1202,9 @@ def apply_role(role, inst, in_queue, bootstrap = false)
               next
             end
 
+            # skip adding nodes to the batch unless they are really in the list to be deployed
+            next unless new_elements[role_name].include?(node_name)
```
vuntz: Not a clean cherry-pick :-)

Itxaka: weeeeird, thats what I get for doing things quick without double checking :(

Itxaka: fixed!

PR #1404, opened 2017-10-31
dns: skip unchanged nodes from chef run

Itxaka:

time.

Requires: https://github.com/crowbar/crowbar-core/pull/1401

### DIFF
--- a/crowbar_framework/app/models/dns_service.rb
+++ b/crowbar_framework/app/models/dns_service.rb
@@ -188,4 +188,63 @@ class DnsService < ServiceObject
     instance = Crowbar::DataBagConfig.instance_from_role(old_role, role)
     Crowbar::DataBagConfig.save("core", instance, @bc_name, config)
   end
+
+  # try to know if we can skip a node from running chef-client
+  def skip_unchanged_node?(node, old_role, new_role)
+    @logger.debug("DNS skip_batch_for_node? entering: #{node}")
+
+    # if old_role is nil, then we are applying the barclamp for the first time
+    if old_role.nil?
+      @logger.debug("DNS skip_batch_for_node?: not skipping #{node} (new role)")
+      return false
+    end
+
+    # if attributes have changed, we need to run
+    return false if node_changed_attributes?(node, old_role, new_role)
+
+    # if the node changed roles, then we need to apply
+    return false if node_changed_roles?(node, old_role, new_role)
+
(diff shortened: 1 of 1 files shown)


Itxaka:
Tested manually and it works.

Itxaka:
[Skip batches when nothing has changed: DNS barclamp](https://trello.com/c/ObcI205O/189-skip-batches-when-nothing-has-changed-dns-barclamp)

ilausuch:
Hi, before merge this, #1405 #1406, I would like to test a generic alternative. I am working on that

crowbar_framework/app/models/dns_service.rb
```diff
@@ -188,4 +188,63 @@ def save_config_to_databag(old_role, role, server_nodes = nil)
...
+  end
+
+  def node_has_role?(node, role, role_name)
+    role.override_attributes[@bc_name]["elements"].each do |r_name, elements|
+      next if r_name != role_name
+      return true if elements.include?(node)
+    end
+
+    false
+  end
```
vuntz: I see that we're copying & pasting all these helpers everywhere. Can we put them directly in the crowbar framework instead of inside each barclamp?

crowbar_framework/app/models/dns_service.rb
```diff
@@ -188,4 +188,63 @@ def save_config_to_databag(old_role, role, server_nodes = nil)
...
+      @logger.debug("DNS skip_batch_for_node?: not skipping #{node} (new role)")
+      return false
+    end
+
+    # if attributes have changed, we need to run
+    return false if node_changed_attributes?(node, old_role, new_role)
+
+    # if the node changed roles, then we need to apply
+    return false if node_changed_roles?(node, old_role, new_role)
+
```
vuntz: In the case of dns, I would try to optimize things further and skip client-only nodes unless some attributes are changed (nameservers, additional_search_domains). I think all other attributes only matter for dns-server. This makes it possible to change the list of records without running chef of client nodes.

crowbar_framework/app/models/dns_service.rb
ilausuch: Yes, we are working on that

Itxaka:
@ilausuch is making a PR to make the checks more generic and then we can override on each barclamp. Will await for that to be merged, then rewrite this PR

vuntz:
Closing for now, this was part of https://github.com/crowbar/crowbar-core/pull/1408

PR #1406, opened 2017-10-31
NTP : skip unchanged nodes from chef run

ilausuch:

time.

Requires: #1401

### DIFF
--- a/crowbar_framework/app/models/ntp_service.rb
+++ b/crowbar_framework/app/models/ntp_service.rb
@@ -101,4 +101,63 @@ class NtpService < ServiceObject
     instance = Crowbar::DataBagConfig.instance_from_role(old_role, role)
     Crowbar::DataBagConfig.save("core", instance, @bc_name, config)
   end
+
+  # try to know if we can skip a node from running chef-client
+  def skip_unchanged_node?(node, old_role, new_role)
+    @logger.debug("NTP skip_batch_for_node? entering: #{node}")
+
+    # if old_role is nil, then we are applying the barclamp for the first time
+    if old_role.nil?
+      @logger.debug("NTP skip_batch_for_node?: not skipping #{node} (new role)")
+      return false
+    end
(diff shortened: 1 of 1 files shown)


ilausuch:
[Skip batches when nothing has changed: NTP barclamp](https://trello.com/c/jlgduYbh/190-skip-batches-when-nothing-has-changed-ntp-barclamp)

ilausuch:
Yes, It was tested

crowbar_framework/app/models/ntp_service.rb
```diff
@@ -101,4 +101,63 @@ def save_config_to_databag(old_role, role)
...
+  end
+
+  def node_has_role?(node, role, role_name)
+    role.override_attributes[@bc_name]["elements"].each do |r_name, elements|
+      next if r_name != role_name
+      return true if elements.include?(node)
+    end
+
+    false
+  end
```
vuntz: See https://github.com/crowbar/crowbar-core/pull/1404#discussion_r148483220

crowbar_framework/app/models/ntp_service.rb
```diff
@@ -101,4 +101,63 @@ def save_config_to_databag(old_role, role)
...
+      @logger.debug("NTP skip_batch_for_node?: not skipping #{node} (new role)")
+      return false
+    end
+
+    # if attributes have changed, we need to run
+    return false if node_changed_attributes?(node, old_role, new_role)
+
+    # if the node changed roles, then we need to apply
+    return false if node_changed_roles?(node, old_role, new_role)
+
```
vuntz: Note that we can skip running chef on client-only nodes if only the `external_servers` attribute is changed (and at the moment, it's the only attribute). We should take that into account here too.

vuntz:
Closing for now, this was part of https://github.com/crowbar/crowbar-core/pull/1408

PR #1408, opened 2017-11-02
skip unchanged nodes from chef run

ilausuch:
    
time.

Barclamps on witch this check will run are configured in experimental
configuration file

Requires: #1401

### DIFF
--- a/crowbar_framework/app/models/dns_service.rb
+++ b/crowbar_framework/app/models/dns_service.rb
@@ -188,4 +188,21 @@ class DnsService < ServiceObject
     instance = Crowbar::DataBagConfig.instance_from_role(old_role, role)
     Crowbar::DataBagConfig.save("core", instance, @bc_name, config)
   end
+
+  # try to know if we can skip a node from running chef-client
+  def skip_unchanged_node?(node_name, old_role, new_role)
+    # if old_role is nil, then we are applying the barclamp for the first time
+    return false if old_role.nil?
+
+    # if the node changed roles, then we need to apply
+    return false if node_changed_roles?(node_name, old_role, new_role)
+
+    # if attributes have changed, we need to run
+    return false if node_changed_attributes?(node_name, old_role, new_role)
+
+    # by this point its safe to assume that we can skip the node as nothing has changed on it
+    # same attributes, same roles so skip it
+    @logger.info("#{@bc_name} skip_batch_for_node? skipping: #{node_name}")
+    true
+  end
 end
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1826,4 +1826,16 @@ class ServiceObject
     logger.debug("skip_unready_nodes: exit for #{bc}:#{inst}")
     [cleaned_elements, pre_cached_nodes]
   end
+
+  # return true if the new attributes are different from the old ones
+  def node_changed_attributes?(node, old_role, new_role)
+    old_role.default_attributes[@bc_name] != new_role.default_attributes[@bc_name]
+  end
+
+  # return true if the node has changed roles
+  def node_changed_roles?(node, old_role, new_role)
+    roles_in_old = old_role.elements.keys.select { |r| old_role.elements[r].include?(node) }.sort
+    roles_in_new = new_role.elements.keys.select { |r| new_role.elements[r].include?(node) }.sort
+    roles_in_old != roles_in_new
+  end
 end
--- a/crowbar_framework/app/models/ntp_service.rb
+++ b/crowbar_framework/app/models/ntp_service.rb
@@ -101,4 +101,21 @@ class NtpService < ServiceObject
     instance = Crowbar::DataBagConfig.instance_from_role(old_role, role)
     Crowbar::DataBagConfig.save("core", instance, @bc_name, config)
   end
+
+  # try to know if we can skip a node from running chef-client
+  def skip_unchanged_node?(node_name, old_role, new_role)
+    # if old_role is nil, then we are applying the barclamp for the first time
+    return false if old_role.nil?
+
+    # if the node changed roles, then we need to apply
+    return false if node_changed_roles?(node_name, old_role, new_role)
+
+    # if attributes have changed, we need to run
+    return false if node_changed_attributes?(node_name, old_role, new_role)
+
+    # by this point its safe to assume that we can skip the node as nothing has changed on it
+    # same attributes, same roles so skip it
+    @logger.info("#{@bc_name} skip_batch_for_node? skipping: #{node_name}")
+    true
+  end
 end
--- a/crowbar_framework/app/models/provisioner_service.rb
+++ b/crowbar_framework/app/models/provisioner_service.rb
@@ -275,4 +275,21 @@ class ProvisionerService < ServiceObject
 
     [200, ""]
   end
+
+  # try to know if we can skip a node from running chef-client
+  def skip_unchanged_node?(node_name, old_role, new_role)
+    # if old_role is nil, then we are applying the barclamp for the first time
+    return false if old_role.nil?
+
+    # if the node changed roles, then we need to apply
+    return false if node_changed_roles?(node_name, old_role, new_role)
+
+    # if attributes have changed, we need to run
+    return false if node_changed_attributes?(node_name, old_role, new_role)
+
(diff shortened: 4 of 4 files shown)


ilausuch:
[Skip batches when nothing has changed: Generic](https://trello.com/c/yus5L4dW/192-skip-batches-when-nothing-has-changed-generic)

crowbar_framework/app/models/service_object.rb
```diff
@@ -1460,11 +1460,6 @@ def proposal_dependencies(role)
     []
   end
 
-  def skip_unchanged_node?(node_name, old_role, role)
-    # By default dont skip anything
-    false
-  end
```
vuntz: I would actually not change this method: we should keep it that way so that it gets overridden in each barclamp, because each barclamp may have some additional tests to decide on whether to skip or not. My request was really to put the helpers here (like you did) so they get shared.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1826,4 +1821,71 @@ def skip_unready_nodes(bc, inst, new_elements, old_elements)
...
+      return true
+    end
+
+    false
+  end
+
+  # return true if the node has changed roles
+  def node_changed_roles?(node, old_role, new_role, role_list)
+    # Check if some role of this barclamp has changed in the node
+    role_list.each do |role_name|
```
vuntz: You don't need to pass a `role_list`, you have this information in `{old,new}_role.override_attributes[@bc_name]["elements"]` already (the roles are the keys of these two hashes)

crowbar_framework/app/models/service_object.rb
```diff
@@ -1826,4 +1821,71 @@ def skip_unready_nodes(bc, inst, new_elements, old_elements)
...
+        return true
+      end
+    end
+
+    false
+  end
+
+  # check if this node has the role defined by role_name
+  def node_has_role?(node, role, role_name)
+    role.override_attributes[@bc_name]["elements"].each do |r_name, elements|
```
vuntz: Why not do something like:

```ruby
nodes_with_role = role.override_attributes[@bc_name]["elements"][role_name] || []
nodes_with_role.include?(node)
```

crowbar_framework/app/models/service_object.rb
```diff
@@ -1826,4 +1826,35 @@ def skip_unready_nodes(bc, inst, new_elements, old_elements)
...
+      logger.debug("#{@bc_name} skip_batch_for_node?: not skipping #{node} (attribute change)")
+      return true
+    end
+
+    false
+  end
+
+  # return true if the node has changed roles
+  def node_changed_roles?(node, old_role, new_role)
+    role_constraints.keys.each do |role_name|
```
vuntz: As said previously, you shouldn't use `role_constraints` but directly the content of `old_role` and `new_role`.

Something like:

```ruby
roles_in_old = old_role.elements.keys.select { |r| old_role.elements[r].include?(node) }.sort
roles_in_new = new_role.elements.keys.select { |r| new_role.elements[r].include?(node) }.sort
roles_in_old != roles_in_new
```

(completely untested)

ilausuch: I have tested it and works.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1826,4 +1826,35 @@ def skip_unready_nodes(bc, inst, new_elements, old_elements)
     logger.debug("skip_unready_nodes: exit for #{bc}:#{inst}")
     [cleaned_elements, pre_cached_nodes]
   end
+
+  # return true if the new attributes are different from the old ones
+  def node_changed_attributes?(node, old_role, new_role)
+    if old_role.default_attributes[@bc_name] != new_role.default_attributes[@bc_name]
+      logger.debug("#{@bc_name} skip_batch_for_node?: not skipping #{node} (attribute change)")
```
ilausuch: @vuntz I think it's better to remove any debug information from here and from the method node_changed_roles and move them to each implementation of skip_unchanged_node?
What do you think?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1826,4 +1826,22 @@ def skip_unready_nodes(bc, inst, new_elements, old_elements)
...
+    roles_in_old = old_role.elements.keys.select { |r| old_role.elements[r].include?(node) }.sort
+    roles_in_new = new_role.elements.keys.select { |r| new_role.elements[r].include?(node) }.sort
+    roles_in_old != roles_in_new
+  end
+
+  # check if this node has the role defined by role_name
+  def node_has_role?(node, role, role_name)
+    nodes_with_role = role.elements[role_name] || []
+    nodes_with_role.include?(node)
+  end
```
vuntz: Isn't that last method unused now?

crowbar_framework/app/models/dns_service.rb
```diff
@@ -188,4 +188,30 @@ def save_config_to_databag(old_role, role, server_nodes = nil)
...
+    # if attributes have changed, we need to run
+    if node_changed_attributes?(node_name, old_role, new_role)
+      logger.debug("#{@bc_name} skip_batch_for_node?: not skipping #{node} (attribute change)")
+      return false
+    end
+
+    # by this point its safe to assume that we can skip the node as nothing has changed on it
+    # same attributes, same roles so skip it
+    @logger.debug("#{@bc_name} skip_batch_for_node? skipping: #{node_name}")
+    @logger.debug("#{@bc_name} skip_batch_for_node? exiting: #{node_name}")
```
vuntz: I would remove the "exiting" log if it's always following a "skipping" log :-)

crowbar_framework/app/models/dns_service.rb
```diff
@@ -188,4 +188,30 @@ def save_config_to_databag(old_role, role, server_nodes = nil)
...
+  # try to know if we can skip a node from running chef-client
+  def skip_unchanged_node?(node_name, old_role, new_role)
+    @logger.debug("#{@bc_name} skip_batch_for_node? entering: #{node_name}")
+
+    # if old_role is nil, then we are applying the barclamp for the first time
+    return false if old_role.nil?
+
+    # if the node changed roles, then we need to apply
+    if node_changed_roles?(node_name, old_role, new_role)
+      logger.debug("#{@bc_name} skip_batch_for_node?: not skipping #{node} (role change)")
```
vuntz: I would actually not have any "not skipping" log; that's the default behavior and can be very spammy. I think we should only log when skipping.

crowbar_framework/app/models/dns_service.rb
```diff
@@ -188,4 +188,30 @@ def save_config_to_databag(old_role, role, server_nodes = nil)
     instance = Crowbar::DataBagConfig.instance_from_role(old_role, role)
     Crowbar::DataBagConfig.save("core", instance, @bc_name, config)
   end
+
+  # try to know if we can skip a node from running chef-client
+  def skip_unchanged_node?(node_name, old_role, new_role)
+    @logger.debug("#{@bc_name} skip_batch_for_node? entering: #{node_name}")
```
vuntz: Similarly (and it's not from you, I know), I would probably not log the "entering" bit. I don't see the benefit here.

crowbar_framework/app/models/dns_service.rb
```diff
@@ -188,4 +188,30 @@ def save_config_to_databag(old_role, role, server_nodes = nil)
...
+
+    # if attributes have changed, we need to run
+    if node_changed_attributes?(node_name, old_role, new_role)
+      logger.debug("#{@bc_name} skip_batch_for_node?: not skipping #{node} (attribute change)")
+      return false
+    end
+
+    # by this point its safe to assume that we can skip the node as nothing has changed on it
+    # same attributes, same roles so skip it
+    @logger.debug("#{@bc_name} skip_batch_for_node? skipping: #{node_name}")
```
vuntz: What about putting it at INFO level, not DEBUG? (`@logger.info`)

crowbar_framework/app/models/service_object.rb
ilausuch: Yes, I keep it because can be useful. It must be removed anyway?

Itxaka: I would say that we can remove it now and add it afterwards if needed?

PR #1410, opened 2017-11-03
crowbar: Fix skip_unready_nodes' for unready nodes

vuntz:
We also need to take into account the heartbeat as nodes that were not
probably shut down will still have the state "ready".Therefore we should
not look at the raw state attribute, but we should use the method that
interprets that raw state attribute.


susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
f4bc8ed5565f684d252e7e3da5cc43af9c4bbf46</td>
    </tr>
  </tbody>
</table>

</p>

PR #1412, opened 2017-11-03
crowbar: Avoid nil crash on applying errored nodes

dirkmueller:
When a node is in error state, it seems the chef apply fails
with

[2017-11-03T00:42:02-04:00] FATAL: NoMethodError: undefined method `key?' for nil:NilClass

This is due to "applying_for" not being set.

requests?**


aspiers:
CI failed in neutron proposal with:

    Can't connect to MySQL server on '192.168.251.86' (111 "Connection refused")

PR #1414, opened 2017-11-06
[POC] conduit_resolver: make the framework file a link

Itxaka:
NOT FOR MERGE. I jsut want to see the implications of doing this and if its viable or it breaks stuff. It probably does.

Instead of the chef file for conduit being a link to the
crowbar_framework file, reverse it so when dealing with chef tools that
cant deal with links, those tools work


Itxaka:
not moving forward

PR #1415, opened 2017-11-07
[4.0] Backport skip unchanged nodes from chef run

ilausuch:

time.

Barclamps on witch this check will run are configured in experimental
configuration file

Backport of #1408


ilausuch:
[Skip batches when nothing has changed: Generic](https://trello.com/c/yus5L4dW/192-skip-batches-when-nothing-has-changed-generic)

PR #1416, opened 2017-11-07
Deployer : Skip batches when nothing has changed

ilausuch:

time.

### DIFF
--- a/crowbar_framework/app/models/deployer_service.rb
+++ b/crowbar_framework/app/models/deployer_service.rb
@@ -138,4 +138,21 @@ class DeployerService < ServiceObject
     Rails.logger.debug("Deployer transition: exiting: #{name} for #{state}")
     return [200, { name: name }]
   end
+
+  # try to know if we can skip a node from running chef-client
+  def skip_unchanged_node?(node_name, old_role, new_role)
+    # if old_role is nil, then we are applying the barclamp for the first time
+    return false if old_role.nil?
+
+    # if the node changed roles, then we need to apply
+    return false if node_changed_roles?(node_name, old_role, new_role)
+
+    # no need to check if attributes changed because they are not used in
+    # cookbooks, just in the rails application
+
+    # by this point its safe to assume that we can skip the node as nothing has changed on it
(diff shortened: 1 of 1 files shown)


ilausuch:
Tested on 8.ha

crowbar_framework/app/models/deployer_service.rb
```diff
@@ -138,4 +138,21 @@ def transition(inst, name, state)
...
+  # try to know if we can skip a node from running chef-client
+  def skip_unchanged_node?(node_name, old_role, new_role)
+    # if old_role is nil, then we are applying the barclamp for the first time
+    return false if old_role.nil?
+
+    # if the node changed roles, then we need to apply
+    return false if node_changed_roles?(node_name, old_role, new_role)
+
+    # if attributes have changed, we need to run
+    return false if node_changed_attributes?(node_name, old_role, new_role)
```
vuntz: That's actually not true for the deployer. Can you remove the code and add an explicit comment? (the attributes are only used by the rails app)

crowbar_framework/app/models/deployer_service.rb
```diff
@@ -138,4 +138,20 @@ def transition(inst, name, state)
...
+
+  # try to know if we can skip a node from running chef-client
+  def skip_unchanged_node?(node_name, old_role, new_role)
+    # if old_role is nil, then we are applying the barclamp for the first time
+    return false if old_role.nil?
+
+    # if the node changed roles, then we need to apply
+    return false if node_changed_roles?(node_name, old_role, new_role)
+
+    # There is not need to check attributes because they aren't necesarry for the current barclamp
```
vuntz: Typo: "necessary"

Also, I would rephrase to "they are not used in cookbooks"; they are actually used in the barclamp (in the rails application).

vuntz:
Hrm, something is wrong, there are 4 commits which shouldn't be here.

vuntz:
> Hrm, something is wrong, there are 4 commits which shouldn't be here.

I fixed this.

PR #1417, opened 2017-11-07
Multi-threaded node saving

skazi0:
Saving node data takes a lot of time when applying proposal for barclamps like dns or network which affect most of the nodes.

With multiple chef-server workers running, saving multiple nodes in parallel gives significant performance improvement in large scale installations.

requests?**
https://trello.com/c/jGyRtgrr/195-save-node-data-in-parallel

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -152,6 +152,7 @@ class ServiceObject
 
   def set_to_applying(nodes, inst, pre_cached_nodes)
     with_lock "BA-LOCK" do
+      nodes_to_save = []
       nodes.each do |node_name|
         node = pre_cached_nodes[node_name]
         if node.nil?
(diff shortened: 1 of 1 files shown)


skazi0:
[Save node data in parallel](https://trello.com/c/jGyRtgrr/195-save-node-data-in-parallel)

skazi0:
@jsuchome idea came from @vuntz :)
@nicolasbock in my test runs with ~230 nodes time to apply dns proposal went down by ~24% (from ~10:45 to ~8:10)

Itxaka:
Im guessing that knowing that @nicolasbock is a scientists he wants raw data in order to reproduce the results independently. Science! :P

nicolasbock:
Yes, that's right @Itxaka, I always insist on evidence. :wink: You really can't just believe what you read on the Internet :smile:

nicolasbock:
And that's an impressive performance boost @skazi0 !

PR #1418, opened 2017-11-07
utils: extend systemd_service_restart

Itxaka:
Extend the systemd_service_restart helper to allow for just
writing+reloading the systemd configuration, allowing us to pass extra
parameters to add to a service in a generic way and reusing all the
existing code.

For example:

```
utils_systemd_service_restart "pacemaker" do
  extra_config {"TasksMax": 1024, "LimitNPROC": "800:1000"}
  action :override_config
end
```

would generate a
`/etc/systemd/system/pacemaker.service.d/crowbar-restart.conf` with the
following values:

```
[Service]
TasksMax=1024
LimitNPROC=800:1000
```

### DIFF
--- a/chef/cookbooks/utils/providers/systemd_service_restart.rb
+++ b/chef/cookbooks/utils/providers/systemd_service_restart.rb
@@ -79,3 +79,10 @@ action :disable do
 
   write_conf_snippet(new_resource, variables)
 end
+
+action :override_config do
+  variables = {
+    extra_config: new_resource.extra_config
+  }
+  write_conf_snippet(new_resource, variables)
+end
--- a/chef/cookbooks/utils/resources/systemd_service_restart.rb
+++ b/chef/cookbooks/utils/resources/systemd_service_restart.rb
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-actions :enable, :disable
+actions :enable, :disable, :override_config
 default_action :enable
 
 attribute :service, kind_of: String, name_attribute: true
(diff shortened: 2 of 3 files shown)


Itxaka:
[Increase TasksMax setting for pacemaker.service](https://trello.com/c/T1FGP9RR/194-increase-tasksmax-setting-for-pacemakerservice)

Itxaka:
@vuntz not sure about this. The naming does not help, as it has `restart` all over, but changing that now....doesnt seem like a good move. And creating another helper for this is just dumb as it would share everything from this one for no reason. Thoughts?

vuntz:
> @vuntz not sure about this. The naming does not help, as it has restart all over, but changing that now....doesnt seem like a good move. And creating another helper for this is just dumb as it would share everything from this one for no reason. Thoughts?

I'm mixed too.

I somehow like the idea to have different snippet file, which would likely mean creating a new LWRP. If we want to combine the things, then I would rename the current LWRP to something more generic, and add a compatibility LWRP calling the new one.

Itxaka:
Not needed anymore!

nicolasbock:
@vuntz What's your though on this PR in regards to https://github.com/crowbar/crowbar-core/pull/1461? Should we re-open this one here or do the work you suggest in #1461 ?

vuntz:
@nicolasbock no need to reopen this one, you can do the work in the other PR.

nicolasbock:
Thanks @vuntz

PR #1420, opened 2017-11-07
Improve skip_unchanged_node? for dns, logging, ntp, provisioner

ilausuch:
**crowbar: Add helper for skip_unchanged_node?**
    
The new "relevant_attributes_changed_if_roles?" helper checks if no
other attribute than the ignored has changed, and only considers this if
the node has no other role than the ones whitelisted.
    
This is useful to allow skipping "client" nodes when many attributes are
relevant only to "server" nodes.

**dns/logging/ntp/provisioner: Improve skip_unchanged_node?**

Use the new helper :-) Also always apply if the server changed.

### DIFF
--- a/crowbar_framework/app/models/ntp_service.rb
+++ b/crowbar_framework/app/models/ntp_service.rb
@@ -107,14 +107,27 @@ class NtpService < ServiceObject
     # if old_role is nil, then we are applying the barclamp for the first time
     return false if old_role.nil?
 
+    # if the servers have changed, we need to apply
+    old_servers = Set.new(old_role.elements["ntp-server"])
+    new_servers = Set.new(new_role.elements["ntp-server"])
+    return false if old_servers != new_servers
+
     # if the node changed roles, then we need to apply
     return false if node_changed_roles?(node_name, old_role, new_role)
 
-    # if attributes have changed, we need to run
-    return false if node_changed_attributes?(node_name, old_role, new_role)
+    # if we're a server, and any attribute has changed, then we need to apply
+    if new_role.elements["ntp-server"].include?(node_name) &&
+        node_changed_attributes?(node_name, old_role, new_role)
+      return false
+    end
 
-    # by this point its safe to assume that we can skip the node as nothing has changed on it
-    # same attributes, same roles so skip it
+    # if we're only a client, and relevant attributes have changed (we list
+    # here the ones to ignore), then we need to apply
+    return false if relevant_attributes_changed_if_roles?(node_name, old_role, new_role,
+      ["external_servers"], ["ntp-client"])
+
+    # by this point its safe to assume that we can skip the node as nothing has
+    # changed on it same attributes, same roles so skip it
     @logger.info("#{@bc_name} skip_batch_for_node? skipping: #{node_name}")
     true
   end
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1852,4 +1852,50 @@ class ServiceObject
     roles_in_new = new_role.elements.keys.select { |r| new_role.elements[r].include?(node) }.sort
     roles_in_old != roles_in_new
   end
+
+  # return true if no other attribute than the ignored has changed, but only if
+  # the node has no other role than the ones listed in only_for_roles
+  def relevant_attributes_changed_if_roles?(node, old_role, new_role, ignore_attr, only_for_roles)
+    # if only_for_roles has some sole
+    unless only_for_roles.nil? || only_for_roles.empty?
+      # get roles for this node
+      roles_in_new = new_role.elements.keys.select { |r| new_role.elements[r].include?(node) }
+
+      # return false if node has other roles that ones in the list
+      # (only_for_roles) -- this test is not for us
+      return false unless Set.new(roles_in_new).subset?(Set.new(only_for_roles))
+    end
+
+    # if the ingnore_attr has some element, apply filters
+    if ignore_attr.nil? || ignore_attr.empty?
+      old_role.default_attributes[@bc_name] != new_role.default_attributes[@bc_name]
+    else
+      # prepare a clone of default attributes of old and new roles
+      old_selected_attributes = old_role.default_attributes[@bc_name].deep_dup
+      new_selected_attributes = new_role.default_attributes[@bc_name].deep_dup
+
+      # function to remove all ignored attributes from a list
+      remove_ignored = lambda do |attributes, ignored|
+        ignored.each do |path|
+          iterator = attributes
+          path = path.split(".")
+
+          while path.length > 1
+            iterator = iterator[path[0]]
+            break if iterator.nil?
+            path.slice!(0)
+          end
+
+          iterator.delete(path[0]) unless iterator.nil?
+        end
+      end
+
+      # remove ignored attributes from old and new attributes
+      remove_ignored.call(old_selected_attributes, ignore_attr)
+      remove_ignored.call(new_selected_attributes, ignore_attr)
+
+      # return true if the attributes have changed, except for the ignored ones
+      old_selected_attributes != new_selected_attributes
+    end
+  end
 end
--- a/crowbar_framework/app/models/dns_service.rb
+++ b/crowbar_framework/app/models/dns_service.rb
@@ -194,14 +194,34 @@ class DnsService < ServiceObject
     # if old_role is nil, then we are applying the barclamp for the first time
     return false if old_role.nil?
 
+    # if the servers have changed, we need to apply
+    old_servers = Set.new(old_role.elements["dns-server"])
+    new_servers = Set.new(new_role.elements["dns-server"])
+    return false if old_servers != new_servers
+
     # if the node changed roles, then we need to apply
     return false if node_changed_roles?(node_name, old_role, new_role)
 
-    # if attributes have changed, we need to run
-    return false if node_changed_attributes?(node_name, old_role, new_role)
+    # if we're a server, and any attribute has changed, then we need to apply
+    if new_role.elements["dns-server"].include?(node_name) &&
+        node_changed_attributes?(node_name, old_role, new_role)
+      return false
+    end
 
-    # by this point its safe to assume that we can skip the node as nothing has changed on it
-    # same attributes, same roles so skip it
+    # if we're only a client, and relevant attributes have changed (we list
+    # here the ones to ignore), then we need to apply
+    return false if relevant_attributes_changed_if_roles?(node_name, old_role, new_role,
+      ["allow_transfer", "forwarders", "records"], ["dns-client"])
+
+    # note that the admin server, even if only dns-client, looks at forwarders...
+    # (cache admin node, to not look for it several times while applying)
+    @admin_node_name ||= Node.admin_node.name
+    return false if node_name == @admin_node_name &&
+        old_role.default_attributes[@bc_name]["forwarders"] != \
+            new_role.default_attributes[@bc_name]["forwarders"]
+
+    # by this point its safe to assume that we can skip the node as nothing has
+    # changed on it same attributes, same roles so skip it
     @logger.info("#{@bc_name} skip_batch_for_node? skipping: #{node_name}")
     true
   end
--- a/crowbar_framework/app/models/logging_service.rb
+++ b/crowbar_framework/app/models/logging_service.rb
@@ -101,4 +101,34 @@ class LoggingService < ServiceObject
     instance = Crowbar::DataBagConfig.instance_from_role(old_role, role)
     Crowbar::DataBagConfig.save("core", instance, @bc_name, config)
   end
+
+  # try to know if we can skip a node from running chef-client
+  def skip_unchanged_node?(node_name, old_role, new_role)
+    # if old_role is nil, then we are applying the barclamp for the first time
+    return false if old_role.nil?
+
+    # if the servers have changed, we need to apply
+    old_servers = Set.new(old_role.elements["logging-server"])
+    new_servers = Set.new(new_role.elements["logging-server"])
+    return false if old_servers != new_servers
+
+    # if the node changed roles, then we need to apply
+    return false if node_changed_roles?(node_name, old_role, new_role)
+
+    # if we're a server, and any attribute has changed, then we need to apply
+    if new_role.elements["logging-server"].include?(node_name) &&
+        node_changed_attributes?(node_name, old_role, new_role)
+      return false
+    end
+
+    # if we're only a client, and relevant attributes have changed (we list
+    # here the ones to ignore), then we need to apply
+    return false if relevant_attributes_changed_if_roles?(node_name, old_role, new_role,
+      ["external_servers"], ["logging-client"])
+
+    # by this point its safe to assume that we can skip the node as nothing has
+    # changed on it same attributes, same roles so skip it
+    @logger.info("#{@bc_name} skip_batch_for_node? skipping: #{node_name}")
+    true
+  end
 end
--- a/crowbar_framework/app/models/provisioner_service.rb
+++ b/crowbar_framework/app/models/provisioner_service.rb
@@ -281,14 +281,50 @@ class ProvisionerService < ServiceObject
     # if old_role is nil, then we are applying the barclamp for the first time
     return false if old_role.nil?
 
+    # if the servers have changed, we need to apply
+    old_servers = Set.new(old_role.elements["provisioner-server"])
+    new_servers = Set.new(new_role.elements["provisioner-server"])
+    return false if old_servers != new_servers
+
     # if the node changed roles, then we need to apply
     return false if node_changed_roles?(node_name, old_role, new_role)
 
-    # if attributes have changed, we need to run
-    return false if node_changed_attributes?(node_name, old_role, new_role)
+    # if we're a server, and any attribute has changed, then we need to apply
+    if new_role.elements["provisioner-server"].include?(node_name) &&
+        node_changed_attributes?(node_name, old_role, new_role)
+      return false
+    end
 
-    # by this point its safe to assume that we can skip the node as nothing has changed on it
-    # same attributes, same roles so skip it
+    # if we're only "base", and relevant attributes have changed (we list
+    # here the ones to ignore), then we need to apply
+    ignore_attributes = [
+      # only relevant for provisioner-server
+      "default_os",
+      "dhcp",
+      "enable_pxe",
+      "supported_oses",
+      # only impacts discovery, so only relevant for provisioner-server
+      "discovery",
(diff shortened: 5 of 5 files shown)


ilausuch:
This code is for your consideration. I have tested it in 8.ha.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1852,4 +1852,25 @@ def node_changed_roles?(node, old_role, new_role)
...
+  # return true if no other attribute than the ignored has changed. And that the node has no other
+  # role than the roles in only_for_roles
+  def relevant_attributes_changed_if_roles?(node, old_role, new_role, ignore_attributes, only_for_roles)
+    # sort roles of only_for_roles
+    only_for_roles = only_for_roles.sort
+
+    # return true if node has no other roles that ones in the list (only_for_roles)
+    roles_in_old = old_role.elements.keys.select { |r| old_role.elements[r].include?(node) }.sort
+    roles_in_new = new_role.elements.keys.select { |r| new_role.elements[r].include?(node) }.sort
+    return true if only_for_roles != roles_in_old || only_for_roles != roles_in_new
```
vuntz: First, you can ignore `old_role`: we only care about `new_role` here (since we only care about attributes that will impact the next chef run, not a previous run).

Then, I think what we want here is something closer to:

```ruby
require "set"
return true unless Set(roles_in_new).subset?(Set(only_for_roles))
```
That is: we don't need to make sure the two things are equal, but the roles the node have must be a subset of the "allowed" roles (`only_for_roles`).

And btw, if you move to sets, you don't need to sort things anymore ;-)

crowbar_framework/app/models/service_object.rb
```diff
@@ -1852,4 +1852,25 @@ def node_changed_roles?(node, old_role, new_role)
     roles_in_new = new_role.elements.keys.select { |r| new_role.elements[r].include?(node) }.sort
     roles_in_old != roles_in_new
   end
+
+  # return true if no other attribute than the ignored has changed. And that the node has no other
+  # role than the roles in only_for_roles
+  def relevant_attributes_changed_if_roles?(node, old_role, new_role, ignore_attributes, only_for_roles)
+    # sort roles of only_for_roles
+    only_for_roles = only_for_roles.sort
```
Itxaka: I think you can do `only_for_roles.sort!` and it will sort them in place, with no need to reassigning

crowbar_framework/app/models/service_object.rb
```diff
@@ -1852,4 +1852,25 @@ def node_changed_roles?(node, old_role, new_role)
     roles_in_new = new_role.elements.keys.select { |r| new_role.elements[r].include?(node) }.sort
     roles_in_old != roles_in_new
   end
+
+  # return true if no other attribute than the ignored has changed. And that the node has no other
+  # role than the roles in only_for_roles
+  def relevant_attributes_changed_if_roles?(node, old_role, new_role, ignore_attributes, only_for_roles)
```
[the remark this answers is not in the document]
Itxaka: can you fix this?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1852,4 +1852,25 @@ def node_changed_roles?(node, old_role, new_role)
...
+    roles_in_new = new_role.elements.keys.select { |r| new_role.elements[r].include?(node) }.sort
+    return true if only_for_roles != roles_in_old || only_for_roles != roles_in_new
+
+    #  return true if the attributes has changed except ignored
+    old_selected_attributes = old_role.default_attributes[@bc_name].reject do |k,v|
+      ignore_attributes.include?(k)
+    end
+    new_selected_attributes = new_role.default_attributes[@bc_name].reject do |k,v|
+      ignore_attributes.include?(k)
+    end
```
vuntz: This works fine until... we need to exclude attributes which are not toplevel. For example, in the provisioner barclamp, we'll want to exclude the `suse.autoyast` attribute (or `role.default_attributes[@bc_name]["suse"]["autoyast"]` if it's clearer).

The way I would deal with it is copy the old and new attributes like this:
```ruby
old_role_attributes = old_role.default_attributes[@bc_name].deep_dup
new_role_attributes = new_role.default_attributes[@bc_name].deep_dup
```

and then remove from these copies the ignored attributes if they exist. The reason I would do it this way is that this makes it much easier to deal with non-toplevel attributes. But I may be wrong :-)

Then we can compare things.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1852,4 +1852,25 @@ def node_changed_roles?(node, old_role, new_role)
...
   end
+
+  # return true if no other attribute than the ignored has changed. And that the node has no other
+  # role than the roles in only_for_roles
+  def relevant_attributes_changed_if_roles?(node, old_role, new_role, ignore_attributes, only_for_roles)
+    # sort roles of only_for_roles
+    only_for_roles = only_for_roles.sort
+
+    # return true if node has no other roles that ones in the list (only_for_roles)
+    roles_in_old = old_role.elements.keys.select { |r| old_role.elements[r].include?(node) }.sort
```
Itxaka: thinking out loud here, should we care if the node did not had the role before?

I mean, if it changed roles, it will trigger a run, so we may assume that it had the role before otherwise we would not be reaching this code, so maybe we can drop this?

In any case, it does not hurt

Itxaka: I was too late on this as I see that Vincent already mentions this :)

ilausuch: Yes, I didn't be sure about that. Thanks!
Sort was to optimize because other ways I had to sort twice the same list.

crowbar_framework/app/models/service_object.rb
ilausuch: I wasn't sure about old_role. Thank you for the new approach

vuntz:
I realized one more thing: we should never skip if there was any change for the ntp-server role (node that was added/removed).

Itxaka:
> I realized one more thing: we should never skip if there was any change for the ntp-server role (node that was added/removed).

Should that be catched by `return false if node_changed_roles?(node_name, old_role, new_role)` or do you mean to apply chef on all `ntp-client` nodes if there is a new `ntp-server`? probably the latter as the clients would need to update the server list if a new server appears

crowbar_framework/app/models/service_object.rb
```diff
@@ -1852,4 +1852,38 @@ def node_changed_roles?(node, old_role, new_role)
...
+  def relevant_attributes_changed_if_roles?(node, old_role, new_role, ignore_attr, only_for_roles)
+    # return true if node has no other roles that ones in the list (only_for_roles)
+    return true unless Set.new(roles_in_new).subset?(Set.new(only_for_roles))
+
+    # prepare a clone of default attributes of old and new roles
+    old_selected_attributes = old_role.default_attributes[@bc_name].clone
+    new_selected_attributes = new_role.default_attributes[@bc_name].clone
+
+    # function to remove all ignored attributes from a list
+    def remove_ignored(list, ignored)
```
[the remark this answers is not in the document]
ilausuch: Working on it

vuntz:
> > I realized one more thing: we should never skip if there was any change for the ntp-server role (node that was added/removed).
>
> Should that be catched by return false if node_changed_roles?(node_name, old_role, new_role) or do you mean to apply chef on all ntp-client nodes if there is a new ntp-server? probably the latter as the clients would need to update the server list if a new server appears

I meant the latter, indeed.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1852,4 +1852,38 @@ def node_changed_roles?(node, old_role, new_role)
     roles_in_new = new_role.elements.keys.select { |r| new_role.elements[r].include?(node) }.sort
     roles_in_old != roles_in_new
   end
+
+  # return true if no other attribute than the ignored has changed. And that the node has no other
+  # role than the roles in only_for_roles
+  def relevant_attributes_changed_if_roles?(node, old_role, new_role, ignore_attr, only_for_roles)
+    # return true if node has no other roles that ones in the list (only_for_roles)
+    return true unless Set.new(roles_in_new).subset?(Set.new(only_for_roles))
```
vuntz: `roles_in_new` is not defined.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1852,4 +1852,38 @@ def node_changed_roles?(node, old_role, new_role)
     roles_in_new = new_role.elements.keys.select { |r| new_role.elements[r].include?(node) }.sort
     roles_in_old != roles_in_new
   end
+
+  # return true if no other attribute than the ignored has changed. And that the node has no other
+  # role than the roles in only_for_roles
+  def relevant_attributes_changed_if_roles?(node, old_role, new_role, ignore_attr, only_for_roles)
+    # return true if node has no other roles that ones in the list (only_for_roles)
```
vuntz: s/no other roles/other roles/ I believe?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1852,4 +1852,41 @@ def node_changed_roles?(node, old_role, new_role)
     roles_in_new = new_role.elements.keys.select { |r| new_role.elements[r].include?(node) }.sort
     roles_in_old != roles_in_new
   end
+
+  # return true if no other attribute than the ignored has changed. And that the node has no other
+  # role than the roles in only_for_roles
+  def relevant_attributes_changed_if_roles?(node, old_role, new_role, ignore_attr, only_for_roles)
+    # get roles for this node
+    roles_in_new = new_role.elements.keys.select { |r| new_role.elements[r].include?(node) }.sort
```
vuntz: No need to call `sort` ;-)

crowbar_framework/app/models/service_object.rb
```diff
@@ -1852,4 +1852,41 @@ def node_changed_roles?(node, old_role, new_role)
...
+
+    # return true if node has other roles that ones in the list (only_for_roles)
+    return true unless Set.new(roles_in_new).subset?(Set.new(only_for_roles))
+
+    # prepare a clone of default attributes of old and new roles
+    old_selected_attributes = old_role.default_attributes[@bc_name].clone
+    new_selected_attributes = new_role.default_attributes[@bc_name].clone
```
vuntz: Asking @Itxaka: `clone` vs `deep_dup`: what's the right one?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1852,4 +1852,41 @@ def node_changed_roles?(node, old_role, new_role)
...
+
+        iterator.delete(path[0]) unless iterator.nil?
+      end
+    end
+
+    # remove ignored attributes from old and new attribures
+    remove_ignored.call(old_selected_attributes, ignore_attr)
+    remove_ignored.call(new_selected_attributes, ignore_attr)
+
+    #  return true if the attributes has changed except ignored
```
vuntz: `return true if the attributes have changed, except for the ignored ones`

crowbar_framework/app/models/service_object.rb
```diff
@@ -1852,4 +1852,41 @@ def node_changed_roles?(node, old_role, new_role)
...
+          iterator = iterator[path[0]]
+          break if iterator.nil?
+          path.slice(0)
+        end
+
+        iterator.delete(path[0]) unless iterator.nil?
+      end
+    end
+
+    # remove ignored attributes from old and new attribures
```
vuntz: s/attribures/attributes/

crowbar_framework/app/models/ntp_service.rb
```diff
@@ -110,8 +110,9 @@ def skip_unchanged_node?(node_name, old_role, new_role)
     # if the node changed roles, then we need to apply
     return false if node_changed_roles?(node_name, old_role, new_role)
```
vuntz: We'd also need a check as suggested in https://github.com/crowbar/crowbar-core/pull/1420#issuecomment-342800525

crowbar_framework/app/models/service_object.rb
```diff
@@ -1852,4 +1852,41 @@ def node_changed_roles?(node, old_role, new_role)
...
+
+    # return true if node has other roles that ones in the list (only_for_roles)
+    return true unless Set.new(roles_in_new).subset?(Set.new(only_for_roles))
+
+    # prepare a clone of default attributes of old and new roles
+    old_selected_attributes = old_role.default_attributes[@bc_name].clone
+    new_selected_attributes = new_role.default_attributes[@bc_name].clone
+
+    # function to remove all ignored attributes from a list
+    remove_ignored = lamdba do |list, ignored|
```
vuntz: s/lamdba/lambda/ ;-)

vuntz: Also I'd rename `list` to `attributes` to avoid confusion.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1852,4 +1852,41 @@ def node_changed_roles?(node, old_role, new_role)
...
+      ignored.each do |path|
+        iterator = list
+        path = path.split(".")
+
+        while path.length > 1
+          iterator = iterator[path[0]]
+          break if iterator.nil?
+          path.slice(0)
```
vuntz: and I think you want `slice!` not `slice`

crowbar_framework/app/models/service_object.rb
ilausuch: With clone we'll have a shallow copy where if there are a reference, it copies the reference to the same object. On the other hand, if we use deep_dup a new object will be created for this reference.
If will be references, and we don't want to change the information of any of them I guess must use deep_dup.
I am going to do tests

crowbar_framework/app/models/service_object.rb
ilausuch: Yes! thanks to this comment, I have realized that I had a mistake with my test. Fixed the test, I am working on the solution

crowbar_framework/app/models/ntp_service.rb
ilausuch: Working on that

crowbar_framework/app/models/ntp_service.rb
```diff
@@ -110,8 +110,13 @@ def skip_unchanged_node?(node_name, old_role, new_role)
...
 
-    # if attributes have changed, we need to run
-    return false if node_changed_attributes?(node_name, old_role, new_role)
+    # if attributes have changed except external_servers, we need to apply
+    return false if relevant_attributes_changed_if_roles?(node_name, old_role, new_role,
+      ["external_servers"], ["ntp-client"])
+
+    # if the node is a client and the servers has changed, we need to apply
+    return false if new_role.elements["ntp-client"].include?(node_name) &&
+    old_role.elements["ntp-server"].sort != new_role.elements["ntp-server"].sort
```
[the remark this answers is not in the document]
vuntz: Please fix this hound comment.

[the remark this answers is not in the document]
vuntz: Btw, no need for the `new_role.elements["ntp-client"].include?(node_name)` part of the test: if the ntp-server nodes change, we need to apply everywhere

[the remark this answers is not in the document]
ilausuch: Done

crowbar_framework/app/models/ntp_service.rb
```diff
@@ -110,8 +110,13 @@ def skip_unchanged_node?(node_name, old_role, new_role)
...
     return false if node_changed_roles?(node_name, old_role, new_role)
 
-    # if attributes have changed, we need to run
-    return false if node_changed_attributes?(node_name, old_role, new_role)
+    # if attributes have changed except external_servers, we need to apply
+    return false if relevant_attributes_changed_if_roles?(node_name, old_role, new_role,
+      ["external_servers"], ["ntp-client"])
+
+    # if the node is a client and the servers has changed, we need to apply
+    return false if new_role.elements["ntp-client"].include?(node_name) &&
```
vuntz: (readding my comment that is hidden now)

No need for the `new_role.elements["ntp-client"].include?(node_name)` part of the test: if there's a change in ntp-server nodes, then we need to apply everywhere.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1852,4 +1852,49 @@ def node_changed_roles?(node, old_role, new_role)
     roles_in_new = new_role.elements.keys.select { |r| new_role.elements[r].include?(node) }.sort
     roles_in_old != roles_in_new
   end
+
+  # return true if no other attribute than the ignored has changed. And that the node has no other
+  # role than the roles in only_for_roles
+  def relevant_attributes_changed_if_roles?(node, old_role, new_role, ignore_attr, only_for_roles)
```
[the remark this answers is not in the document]
ilausuch: @vuntz we can change the method name for relevant_attr_changed_if_roles? 
Do you agree?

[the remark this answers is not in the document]
vuntz: @ilausuch I don't mind, if you want to do so.

crowbar_framework/app/models/ntp_service.rb
```diff
@@ -110,8 +110,12 @@ def skip_unchanged_node?(node_name, old_role, new_role)
...
     return false if node_changed_roles?(node_name, old_role, new_role)
 
-    # if attributes have changed, we need to run
-    return false if node_changed_attributes?(node_name, old_role, new_role)
+    # if attributes have changed except external_servers, we need to apply
+    return false if relevant_attributes_changed_if_roles?(node_name, old_role, new_role,
+      ["external_servers"], ["ntp-client"])
+
+    # if the servers has changed, we need to apply
+    return false if old_role.elements["ntp-server"].sort != new_role.elements["ntp-server"].sort
```
vuntz: Last comment on this: I would actually make this the second test (before `node_changed_roles?`) as this one is very simple and generic, and doesn't depend on the node itself.

crowbar_framework/app/models/service_object.rb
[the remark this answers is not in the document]
ilausuch: In this case, I prefer to keep the current name is more clear.

ilausuch:
It was tested. However I prefer to run some more tests to validate last changes

ilausuch:
Some more tests done

susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
177c6bfdd310362e354f12622cdfe4df12b0147c</td>
    </tr>
  </tbody>
</table>

</p>

vuntz:
@ilausuch I fixed a couple of issues I found, and also did the change for other barclamps. Can you review? :-)

PR #1422, opened 2017-11-08
Fix thread pool creation and logging

skazi0:
On first sight the code looked like it was using `THREAD_POOL_SIZE` threads but it was one less. In addition the logging was misleading.

Fix creation of threads and logging.

requests?**


skazi0:
@jsuchome correct, this was just 19 instead of 20 threads

PR #1430, opened 2017-11-13
crowbar_join: Fix detection of reboot

vuntz:
We were using the runlevel command to find out if we're rebooting (to
set the appropriate state for the node), but that doesn't work anymore
with systemd. Instead, check if reboot.target is starting.


vuntz:
Given that https://github.com/crowbar/crowbar-core/pull/1431 includes this and has 2 +1, closing this one.

PR #1431, opened 2017-11-13
Fix issues when reinstalling a node

vuntz:
There were two issues when clicking "reinstall" for a node:

 - a race where we were rebooting the node too fast
 - the fact that the reboot of the node was actually triggering a state change, which could break the expected PXE config

(note that this contains https://github.com/crowbar/crowbar-core/pull/1430)

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
+++ b/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
@@ -63,6 +63,25 @@ log_to() {
     return $_ret
 }
 
+get_state() {
+  local output
+  local curlargs=(--connect-timeout 60 -s -L \
+      -H "Accept: application/json" -H "Content-Type: application/json" \
+      --max-time 240 --insecure --location)
+  [[ $CROWBAR_KEY ]] && curlargs+=(-u "$CROWBAR_KEY" --digest --anyauth)
+  # we expect something like {"value":"ready"}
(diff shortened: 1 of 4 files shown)


chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
```diff
@@ -63,6 +63,28 @@ log_to() {
...
 }
 
+get_state() {
+  local output
+  local curlargs=(--connect-timeout 60 -s -L \
+      -H "Accept: application/json" -H "Content-Type: application/json" \
+      --max-time 240 --insecure --location)
+  [[ $CROWBAR_KEY ]] && curlargs+=(-u "$CROWBAR_KEY" --digest --anyauth)
+  oldumask=`umask`
+  umask 077
```
Itxaka: so why do we set this? Its the first time I find this in a script. Is it necessary for the curl to work?

vuntz: TBH, I just copied what is done for `post_state` :-)

bmwiedemann: It only makes sense when you create (e.g. download) files, so that they will not be world-readable upon creation.
But the curl below only produces stdout, so the umask changes can be dropped here.

vuntz: There you go.

PR #1439, opened 2017-11-16
[4.0] ohai: Use uppercase for permanent MAC addresses

vuntz:
Everywhere else in ohai/chef, we use uppercase MAC addresses, so let's
ensure that what we get from ethtool is also uppercase.

(cherry picked from commit 415dcafd4469abc868a7b893452cf1bdb10d3ca1)

Backport of https://github.com/crowbar/crowbar-core/pull/1438


seife:
This is "not good enough".
We are seeing "E0:07:F6:..." vs "e0:7:f6..."

So the MAC address needs to be leading-zero-padded, too.

vuntz:
> So the MAC address needs to be leading-zero-padded, too.

Good catch, fixed.

PR #1441, opened 2017-11-17
network: run ifdown for vlan interface cleanup

stefannica:
When cleaning up vlan interfaces, simply running 'vconfig rem' doesn't have any effect for interfaces that are already managed by wicked. Only running 'wicked ifdown' or the more general 'ifdown' properly cleans up these interfaces.

Fixes [bsc#1063535](https://bugzilla.suse.com/show_bug.cgi?id=1063535)
Also see: https://trello.com/c/tkxpXh9w

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -151,6 +151,7 @@ def kill_nic(nic)
 
   Chef::Log.info("Interface #{nic.name} is no longer being used, deconfiguring it.")
   nic.destroy
+  ::Kernel.system("wicked ifdown #{nic.name}")
 
   kill_nic_files(nic)
 end


stefannica:
[Bug 1063535 - L3-Question: cloud installation fails: Creating new VLAN interface eth3.112 for bastion network](https://trello.com/c/tkxpXh9w/519-bug-1063535-l3-question-cloud-installation-fails-creating-new-vlan-interface-eth3112-for-bastion-network)

susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
b4e69410d1b65b7cbad6df5f6a5dfed4c02f277e</td>
    </tr>
  </tbody>
</table>

</p>

chef/cookbooks/barclamp/libraries/nic.rb
```diff
@@ -874,6 +874,7 @@ def parent
 
     def destroy
       super
+      ::Kernel.system("ifdown #{@nic}")
```
rhafer: Hm, wouldn't this belong into `kill_nic` in the network cookbook. I think the goal once was to keep nic.rb free of stuff related to the distro's network management framework (even though `ifdown` might be somewhat portable).

And even though the bugreport is about vlans, I guess the call makes sense for other types interfaces as well. But I might be overlooking something here.

stefannica:
@rhafer , I've done some additional testing and I think I found one problem with this approach: `ifdown` (which is using `wicked ifdown` under the hood) is actually removing the interface completely - be it vlan, bridge or OVS - which means that the nic.rb `destroy` calls need to be made __before__ the `ifdown` calls, otherwise they will fail.

Ideally, `wicked ifdown` should completely replace the `destroy` methods in nic.rb, but that's another story.

I'll update the fix to:
  - call `ifdown` after `destroy`
  - do it in the network barclamp instead of the nic.rb file

chef/cookbooks/network/recipes/default.rb
```diff
@@ -151,6 +151,7 @@ def kill_nic(nic)
 
   Chef::Log.info("Interface #{nic.name} is no longer being used, deconfiguring it.")
   nic.destroy
+  ::Kernel.system("wicked ifdown #{nic.name}")
```
dirkmueller: isn't the order the wrong way around? why not first tell wicked that we're going to kill it and then call nic.destroy, e.g. swap the two lines above?

dirkmueller: Ok, I see the comment above now.

bmwiedemann:
tested the patch with kbaikov and found that it avoided the network loop

PR #1445, opened 2017-11-24
Fix dependency status updates in deployment_queue

skazi0:
Proposal status for dependencies of queued proposal in deployment_queue was not updated correctly. It referred to 'proposal_status' action in DeployQueueController which was not implemented.

The correct place to query for proposal status is BarclampController.

requests?**
https://trello.com/c/SduJSFvx/216-error-when-fetching-proposal-status-on-deployment-queue-page


skazi0:
[Error when fetching proposal status on deployment queue page](https://trello.com/c/SduJSFvx/216-error-when-fetching-proposal-status-on-deployment-queue-page)

PR #1446, opened 2017-11-24
[4.0] Fix dependency status updates in deployment_queue

skazi0:
backport of #1445


skazi0:
[Error when fetching proposal status on deployment queue page](https://trello.com/c/SduJSFvx/216-error-when-fetching-proposal-status-on-deployment-queue-page)

PR #1447, opened 2017-11-28
crowbar : priorize provisioner barclamp execution

ilausuch:
After restart a node, Keystone requires the fernet keys and
they have to be imported from the founder node. To do that
is necessary to do a rsync, and to do that the ssh keys are
required. However in the current barclamp sort, the provisioner
is executed at the end of the list. So, In order to have the
ssh keys from the beginning this fix prioritize the provisioner
barclamp to be executed from the beginning.


ilausuch:
We realized that this is not a really solution to our problem. We can close it or keep this priority

PR #1450, opened 2017-12-04
upgrade: Services are no longer located under /etc/init.d

jsuchome:
These are the services we initentionally stop before the OpenStack upgrade.

During the disruptive upgrade, we're explicitly stopping OpenStack (and some related) services at before we go to next step of upgrading the packages. 

The locations have changed for SLES12SP3.

**Note**

It's possible we might rework the whole step or upgrade process. But for now, I'm only focusing on keeping the same functionality of the 6-7 upgrade

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
@@ -37,9 +37,9 @@ done
 <% end %>
 
 # stop cinder-volume at all nodes so the service could be removed from the database later
-if test -e /etc/init.d/openstack-cinder-volume; then
+if test -e /usr/sbin/rcopenstack-cinder-volume; then
     log "Stopping cinder-volume service"
-    /etc/init.d/openstack-cinder-volume stop
+    /usr/sbin/rcopenstack-cinder-volume stop
 fi
 
 <% else %>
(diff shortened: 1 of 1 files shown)


chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
```diff
@@ -48,11 +48,11 @@ fi
 # Note that for HA setup, they should be stopped by pacemaker.
 log "Stopping OpenStack services..."
 
-for i in /etc/init.d/openstack-* \
-         /etc/init.d/apache2 \
-         /etc/init.d/rabbitmq-server \
+for i in /usr/sbin/rcopenstack-* \
+         /usr/sbin/rcapache2 \
+         /usr/sbin/rcrabbitmq-server \
          /etc/init.d/ovs-usurp-config-* \
```
jsuchome: BTW, I do not know what this is and why we are stopping it here...

susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
dd33254ffa21fd72f2407f2705af6a1f90271c84</td>
    </tr>
  </tbody>
</table>

</p>

chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
Itxaka: this services should no longer exists I believe, reading the neutron recipe (both 4.0 and master) https://github.com/crowbar/crowbar-openstack/blob/stable/4.0/chef/cookbooks/neutron/recipes/common_agent.rb#L131

jsuchome: That makes sense, I could not find it anywhere

jsuchome:
Thanks, I've removed that one

jsuchome:
@skazi0 you're right, I was just aiming for the smallest replacement

PR #1451, opened 2017-12-04
provisioner: Extract ssh keys deployment

skazi0:
Provisioner cookbook which deploys authorized_keys is running last in the runlist but sometimes ssh keys can be needed earlier. Original usecase for this change is synchronization of fernet-keys from founder when new/reinstalled node joins the cluster.

This change extracts keys setup into a separate recipe for easier including in other cookbooks.

requests?**
https://trello.com/c/SBMOaFz1/187-reinstalling-node-in-existing-cluster-fails-because-fernet-keys-folder-does-not-exist

### DIFF
--- a/chef/cookbooks/provisioner/recipes/base.rb
+++ b/chef/cookbooks/provisioner/recipes/base.rb
@@ -26,13 +26,6 @@ package "ipmitool" do
   action :install
 end
 
-directory "/root/.ssh" do
-  owner "root"
-  group "root"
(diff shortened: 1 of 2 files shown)


skazi0:
[Reinstalling node in existing cluster fails, because fernet-keys folder does not exist](https://trello.com/c/SBMOaFz1/187-reinstalling-node-in-existing-cluster-fails-because-fernet-keys-folder-does-not-exist)

PR #1453, opened 2017-12-07
Filter out empty elements in crowbar lists

jhesketh:
When supplying a list (eg of hostnames: "host1, host2, , host3" etc)
an empty element "" would be set in the array. This change removes any
empty elements.

This change also stops empty strings from being converted to arrays
such as [""] (ie one empty element).

(bsc#1065903)


jhesketh:
Sorry, I had updated the commit message a few weeks ago but looks like I failed to force push to update this PR.

vuntz:
@jhesketh Will you create the backport for stable/4.0?

jhesketh:
Yep, created here: https://github.com/crowbar/crowbar-core/pull/1477

PR #1455, opened 2017-12-11
Revert "Disable upgrade API in Cloud8"

jsuchome:
This reverts commit 614c4e2f21a2a13ef9d74d09e8639cf7c6332e46.

Without reverting previous commit, we cannot test 7-8 upgrade


jsuchome:
Thanks

PR #1457, opened 2017-12-14
[4.0] provisioner: Extract ssh keys deployment

skazi0:
Provisioner cookbook which deploys authorized_keys is running last in
the runlist but sometimes ssh keys can be needed earlier. This change
extracts keys setup into a separate recipe for easier including in
other cookbooks.

(cherry picked from commit 1adf5c44571374f468155943845cf4d509ef76a0)

backport of #1451

### DIFF
--- a/chef/cookbooks/provisioner/recipes/base.rb
+++ b/chef/cookbooks/provisioner/recipes/base.rb
@@ -26,13 +26,6 @@ package "ipmitool" do
   action :install
 end
 
-directory "/root/.ssh" do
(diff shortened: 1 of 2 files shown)


skazi0:
[Reinstalling node in existing cluster fails, because fernet-keys folder does not exist](https://trello.com/c/SBMOaFz1/187-reinstalling-node-in-existing-cluster-fails-because-fernet-keys-folder-does-not-exist)

PR #1458, opened 2017-12-14
upgrade: fix pacemaker founder searchs

Itxaka:
When we introduced the new syncmarks for pacemaker
the founder was changed from a boolean indicating
that the node is a founder, to the fqdn of the founder

This results in the upgrade not working as expected because
it expects to find the boolean and the value to mark if
its a founder or not, which doesnt work with the new values

This patch tries to fix that by doing searches for the founder
value, extracting the fqdn of the founder and then loading
the founder node directly

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -335,7 +335,7 @@ module Api
         when "ceph"
           ::Node.find("roles:ceph-* AND ceph_config_environment:*").any?
         when "ha"
-          ::Node.find("pacemaker_founder:true AND pacemaker_config_environment:*").any?
+          ::Node.find("pacemaker_config_environment:*").any?
         end
       end
     end
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -444,7 +444,8 @@ module Api
         cinder_node = nil
         upgrade_nodes.each do |node|
           if node.roles.include?("cinder-controller") &&
-              (!node.roles.include?("pacemaker-cluster-member") || node["pacemaker"]["founder"])
+              (!node.roles.include?("pacemaker-cluster-member") ||
+                  node["pacemaker"]["founder"] == node[:fqdn])
             cinder_node = node
           end
           cmd = node.shutdown_services_before_upgrade
@@ -611,10 +612,10 @@ module Api
       end
 
       def upgrade_pacemaker_cluster(cluster_env)
-        founder = ::Node.find(
-          "pacemaker_founder:true AND " \
+        founder_name = ::Node.find(
           "pacemaker_config_environment:#{cluster_env}"
-        ).first
+        ).first[:pacemaker][:founder]
+        founder = ::Node.find_by_name(founder_name)
         upgrade_cluster founder, cluster_env
       end
 
@@ -838,16 +839,29 @@ module Api
 
         return upgrade_controllers_disruptive if upgrade_mode == :normal
 
-        network_node = ::Node.find(
-          "pacemaker_founder:true AND " \
+        network_cluster_members = ::Node.find(
+          "run_list_map:pacemaker-cluster-member AND " \
           "run_list_map:neutron-network AND NOT " \
           "run_list_map:neutron-server"
-        ).first
+        )
+
+        network_node = if network_cluster_members.empty?
+          nil
+        else
+          ::Node.find_by_name(network_cluster_members.first[:pacemaker][:founder])
+        end
+
         prepare_network_node(network_node) unless network_node.nil?
 
         # Now we must upgrade the clusters in the correct order:
         # 1. data, 2. API, 3. network
-        cluster_founders = ::Node.find("pacemaker_founder:true")
+
+        # search in all the cluster members to get a unique list of founders for all clusters
+        cluster_founders_names = ::Node.find("run_list_map:pacemaker-cluster-member").map! do |node|
+          node[:pacemaker][:founder]
+        end.uniq
+
+        cluster_founders = cluster_founders_names.map { |name| ::Node.find_by_name(name) }
 
         sorted_founders = cluster_founders.sort do |n1, n2|
           first_data = n1[:run_list_map].key? "database-server"
@@ -923,9 +937,9 @@ module Api
         Rails.logger.info("Upgrading controller nodes in cluster #{cluster}")
 
         non_founder_nodes = ::Node.find(
-          "pacemaker_founder:false AND " \
           "pacemaker_config_environment:#{cluster} AND " \
-          "run_list_map:pacemaker-cluster-member"
+          "run_list_map:pacemaker-cluster-member AND " \
+          "NOT fqdn:#{founder[:fqdn]}"
         )
         non_founder_nodes.select! { |n| !n.upgraded? }
 
@@ -1242,10 +1256,10 @@ module Api
           # find the controller in this cluster
           cluster_name = ServiceObject.cluster_name(cluster)
           cluster_env = "pacemaker-config-#{cluster_name}"
-          controller = ::Node.find(
-            "pacemaker_founder:true AND " \
+          founder_name = ::Node.find(
             "pacemaker_config_environment:#{cluster_env}"
-          ).first
+          ).first[:pacemaker][:founder]
+          controller = ::Node.find_by_name(founder_name)
 
           # restart remote resources for each node
           role.cluster_remote_nodes.each do |node|
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -79,7 +79,7 @@ end
 # Find out now if we have HA setup and pass that info to the script
 use_ha = roles.include? "pacemaker-cluster-member"
 remote_node = roles.include? "pacemaker-remote"
-cluster_founder = use_ha && node["pacemaker"]["founder"]
+is_cluster_founder = use_ha && node["pacemaker"]["founder"] == node[:fqdn]
 
 template "/usr/sbin/crowbar-shutdown-services-before-upgrade.sh" do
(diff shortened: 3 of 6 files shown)


Itxaka:
travis failure its expected as I have not fixed the tests, this is out for pre-reviews

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -335,7 +335,7 @@ def addon_deployed?(addon)
         when "ceph"
           ::Node.find("roles:ceph-* AND ceph_config_environment:*").any?
         when "ha"
-          ::Node.find("pacemaker_founder:true AND pacemaker_config_environment:*").any?
+          ::Node.find("pacemaker_founder:* AND pacemaker_config_environment:*").any?
```
jsuchome: Is this correct check? Isn't * true also for empty value?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -611,10 +611,11 @@ def cancel
       end
 
       def upgrade_pacemaker_cluster(cluster_env)
-        founder = ::Node.find(
-          "pacemaker_founder:true AND " \
+        founder_name = ::Node.find(
+          "pacemaker_founder:* AND " \
```
jsuchome: This clause is not necessary, as you're searching for cluster nodes only.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -838,16 +839,23 @@ def upgrade_controller_clusters
 
         return upgrade_controllers_disruptive if upgrade_mode == :normal
 
-        network_node = ::Node.find(
-          "pacemaker_founder:true AND " \
+        network_founders = ::Node.find(
+          "pacemaker_founder:* AND " \
```
jsuchome: Iit should be enough to exclude `pacemaker_founder:*` bit from here

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -838,16 +839,23 @@ def upgrade_controller_clusters
...
           "run_list_map:neutron-network AND NOT " \
           "run_list_map:neutron-server"
-        ).first
+        )
+        network_node = ::Node.load(network_founders.first[:pacemaker][:founder]) rescue nil
```
jsuchome: Is the `rescue nil` needed here? Won't the load return nil anyway, if the node does not exist?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -838,16 +839,23 @@ def upgrade_controller_clusters
...
         prepare_network_node(network_node) unless network_node.nil?
 
         # Now we must upgrade the clusters in the correct order:
         # 1. data, 2. API, 3. network
-        cluster_founders = ::Node.find("pacemaker_founder:true")
+        cluster_founders_names = ::Node.find("pacemaker_founder:*")
```
jsuchome: "cluster_founders_names" actually means all nodes in all clusters, right?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1242,10 +1250,10 @@ def prepare_remote_nodes
           # find the controller in this cluster
           cluster_name = ServiceObject.cluster_name(cluster)
           cluster_env = "pacemaker-config-#{cluster_name}"
-          controller = ::Node.find(
-            "pacemaker_founder:true AND " \
-            "pacemaker_config_environment:#{cluster_env}"
-          ).first
+          founder_name = ::Node.find(
+            "pacemaker_founder:* AND pacemaker_config_environment:#{cluster_env}"
```
jsuchome: Again, `pacemaker_founder:*` is not needed

crowbar_framework/app/models/api/crowbar.rb
Itxaka: there should be no empty values on `node[:pacemaker][:founder]` either its false/true or the fqdn.

Doing a search on a non-existent attribute will return always empty:
```
irb(main):014:0> NodeObject.find("keystone_not_exists:*")
=> []

```

crowbar_framework/app/models/api/upgrade.rb
Itxaka: dont we need to find all `neutron-network` nodes that are applied into clusters?

crowbar_framework/app/models/api/upgrade.rb
Itxaka: yes, but with the filtering below it turns into only founders names. I can use an extra var here or just chain all the methods below instead

crowbar_framework/app/models/api/upgrade.rb
Itxaka: +1

crowbar_framework/app/models/api/upgrade.rb
Itxaka: +1

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Or the search could be for something else, would `pacemaker:*` work as well?

crowbar_framework/app/models/api/crowbar.rb
jsuchome: Well, I think `::Node.find("pacemaker_config_environment:*")` gives the same result, right?

crowbar_framework/app/models/api/upgrade.rb
Itxaka: yes, but we want only the founders rigth? so we need to serach+filter here to extract the founders of all clusters

crowbar_framework/app/models/api/crowbar.rb
Itxaka: it does, let me change it

crowbar_framework/app/models/api/upgrade.rb
jsuchome: My point is, this first search does not get you just founders anyway (you need to refine it later), so you do not need to look for "pacemaker_founder:*". 
But I guess it works, so it's just nitpicking

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Well, it cannot hurt I guess

crowbar_framework/app/models/api/upgrade.rb
jsuchome: True. So this "pacemaker_founder:*" acts as an indication that it is a cluster member. Hm, that works, but it's not very readable.

Itxaka: yeah, might be easier to search for `run_list_map:pacemaker-cluster-member` so we know that its part of a cluster

Itxaka: updated to use that instead and changed var to make it clearer

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -611,10 +611,10 @@ def cancel
       end
 
       def upgrade_pacemaker_cluster(cluster_env)
-        founder = ::Node.find(
-          "pacemaker_founder:true AND " \
+        founder_name = ::Node.find(
           "pacemaker_config_environment:#{cluster_env}"
-        ).first
+        ).first[:pacemaker][:founder]
```
jsuchome: Doesn't this search fit at one line?

Itxaka: nope, 109 chars I counted!

crowbar_framework/app/models/api/upgrade.rb
Itxaka: no, but you need to search for all nodes in all clusters to get the founders names. We may be better with searching for `run_list_map:pacemaker-cluster-member` in here as well.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -838,16 +838,36 @@ def upgrade_controller_clusters
...
+        end.uniq
+
+        # prepare the search for the founders
+        if cluster_founders_names.empty?
+          cluster_founders = []
+        else
+          search = "fqdn:"
+          search << cluster_founders_names.join(" OR fqdn:")
+          cluster_founders = ::Node.find(search)
+        end
```
vuntz: That feels quite complicated, why not:

```ruby
cluster_founders = cluster_founders_names.map { |name| ::Node.load(name) }
```

vuntz:
Travis is red :/

crowbar_framework/app/models/api/upgrade.rb
Itxaka: becuase Im dumb probably, just guessing here :P

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -838,16 +838,34 @@ def upgrade_controller_clusters
...
+        cluster_founders_names = ::Node.find("run_list_map:pacemaker-cluster-member").map! do |node|
+          node[:pacemaker][:founder]
+        end.uniq
+
+        # prepare the search for the founders
+        cluster_founders = if cluster_founders_names.empty?
+          []
+        else
+          cluster_founders_names.map { |name| ::Node.find_by_name(name) }
+        end
```
vuntz: No need for the `if` thing ;-)

Itxaka: done!

Itxaka:
@jsuchome updated, I think its not missing anything now

jsuchome:
Something else is probably still missing - my nodes did not realize they need HA repositories during the upgrade

Itxaka:
> Something else is probably still missing - my nodes did not realize they need HA repositories during the upgrade

Actually I think you are rigth, but the problem is on the crowbar-ha patch as there is a couple of methods that still use the boolean value: ha_presence_check and health_report

Itxaka:
> Actually I think you are rigth, but the problem is on the crowbar-ha patch as there is a couple of methods that still use the boolean value: ha_presence_check and health_report

Ah, actually those 2 methods are called before admin upgrade so they are correct. Disregard this comment.

PR #1459, opened 2017-12-19
Upgrade services tweaks

jsuchome:
Do not stop galera and apache:
In the services step, we want to keep database and keystone running.

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -1147,7 +1147,6 @@ module Api
           )
         else
           node_api.upgrade
-          node_api.reboot_and_wait
           node_api.post_upgrade
           node_api.join_and_chef
         end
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1147,7 +1147,6 @@ def upgrade_one_node(name)
           )
         else
           node_api.upgrade
-          node_api.reboot_and_wait
```
skazi0: what happened to this?

jsuchome: The point is that `node_api.upgrade` already includes `reboot_and_wait` call.
Nothing that bad, it should not reboot twice (that's why we have that check for `rebooted_file`), but anyway it's not necessary.

skazi0: OK. Would be good to mention it in the PR description. I haven't noticed the second commit.

PR #1460, opened 2017-12-19
crowbar : accept alias in node transition call

ilausuch:
crowbarctl node transition <name> <state> didn't accept alias
as name of the node. The problem was in API controller. To fix it
first it tries to get the node information from name or alias and
then calls to transition function using the real name.

### DIFF
--- a/crowbar_framework/app/controllers/barclamp_controller.rb
+++ b/crowbar_framework/app/controllers/barclamp_controller.rb
@@ -114,11 +114,15 @@ class BarclampController < ApplicationController
   def transition
     id = params[:id]       # Provisioner id
     state = params[:state] # State of node transitioning
-    name = params[:name] # Name of node transitioning
+    name = params[:name] # Name or alias of node transitioning
 
-    unless valid_transition_states.include?(state)
-      render text: "State '#{state}' is not valid.", status: 400
-    else
+    unless state == "discovering"
+      node = Node.find_node_by_name_or_alias(name)
+      raise ActionController::RoutingError, "Not Found" if node.nil? # raise an error if not found
+      name = node.name # replace the alias with the real name
+    end
+
+    if valid_transition_states.include?(state)
       status, response = @service_object.transition(id, name, state)
       if status != 200
         render text: response, status: status
@@ -131,6 +135,8 @@ class BarclampController < ApplicationController
           render json: response
         end
       end
+    else
+      render text: "State '#{state}' is not valid.", status: 400
     end
   end
 


ilausuch:
[Check/fix `crowbarctl node *` alias support.](https://trello.com/c/T03pTNx0/222-check-fix-crowbarctl-node-alias-support)

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -116,20 +116,27 @@ def transition
     state = params[:state] # State of node transitioning
     name = params[:name] # Name of node transitioning
 
-    unless valid_transition_states.include?(state)
-      render text: "State '#{state}' is not valid.", status: 400
+    node = Node.find_node_by_name_or_alias(name)
+
+    if node.nil?
+      render text: "Node does not exist.", status: 404
     else
```
skazi0: maybe return here and save us one level of nesting for code below?

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -116,20 +116,27 @@ def transition
     state = params[:state] # State of node transitioning
     name = params[:name] # Name of node transitioning
```
skazi0: Update the comment to fit the "name or alias" behavior.

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -116,20 +116,27 @@ def transition
...
     else
-      status, response = @service_object.transition(id, name, state)
-      if status != 200
-        render text: response, status: status
-      else
-        # Be backward compatible with barclamps returning a node hash, passing
-        # them intact.
-        if response[:name]
-          render json: Node.find_by_name(response[:name]).to_hash
+      name = node.name
```
skazi0: Why not simply pass `node.name` to `transition()` below? The variable is not used anywhere else.

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -114,12 +114,17 @@ def versions
...
 
-    unless valid_transition_states.include?(state)
-      render text: "State '#{state}' is not valid.", status: 400
-    else
-      status, response = @service_object.transition(id, name, state)
+    node = Node.find_node_by_name_or_alias(name)
+
+    if node.nil?
+      render text: "Node does not exist.", status: 404
+      return
```
skazi0: I think you should put the `return` in front of `render` to get the same behavior as with ruby's "implicit" return values.

skazi0: Even better would be to raise an exception like in other controllers:
`raise ActionController::RoutingError.new("Not Found") if node.nil?`

The exception has common handling and will display proper error message.

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -114,12 +114,14 @@ def versions
...
-    name = params[:name] # Name of node transitioning
+    name = params[:name] # Name or alias of node transitioning
 
-    unless valid_transition_states.include?(state)
-      render text: "State '#{state}' is not valid.", status: 400
-    else
-      status, response = @service_object.transition(id, name, state)
+    node = Node.find_node_by_name_or_alias(name)
+
+    raise ActionController::RoutingError, "Not Found" if node.nil?
```
vuntz: I think CI doesn't pass because we actually need to not return an error if the state is `discovering`: that's how we initially create a node (so of course, `node.nil?` is true in that case). When this happens, we need to assume that `name` is indeed a name, and not an alias.

skazi0: @vuntz this case was handled previously by `valid_transition_states` ?

vuntz: Unless I misunderstand, no. Previously, we were not validating the name of the node, and therefore it was always accepted, regardless of the state we're transitioning too (which also means we could add a node directly in state ready, which is kind of wrong).

PR #1461, opened 2017-12-20
Systemd env

nicolasbock:

requests?**

### DIFF
--- a/chef/cookbooks/utils/providers/systemd_environment.rb
+++ b/chef/cookbooks/utils/providers/systemd_environment.rb
@@ -1,0 +1,108 @@
+#
+# Copyright 2018 SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Support whyrun
+def whyrun_supported?
+  true
+end
+
+action :create do
+  current_resource = @current_resource
+  converge_by("Create #{@new_resource}") do
+    execute "systemctl daemon-reload" do
+      action :nothing
+    end
+
+    directory _get_override_directory do
+      owner "root"
+      group "root"
+      mode "0755"
+    end
+
+    service_resource_name = _get_service_resource_name
+    template _get_override_file_path do
+      source "systemd_environment.conf.erb"
+      owner "root"
+      group "root"
+      mode "0644"
+      cookbook "utils"
+      variables(
+        environment: current_resource.environment
+      )
+      notifies :run, resources(execute: "systemctl daemon-reload"), :delayed
+      notifies :restart, resources(service: service_resource_name), :delayed
+    end
+    Chef::Log.info "#{@new_resource} created / updated"
+  end
+end
+
+action :delete do
+  if @current_resource.exists
+    converge_by("Delete #{@new_resource}") do
+      execute "systemctl daemon-reload" do
+        action :nothing
+      end
+
+      directory _get_override_directory do
+        action :delete
+        only_if { _dir_is_empty(_get_override_directory) }
+      end
+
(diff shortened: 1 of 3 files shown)


Itxaka:
I did something similar to allow doing override files for systemd services with extra configs in https://github.com/crowbar/crowbar-core/pull/1418

My issue was that I didnt want to duplicate the code already existing, because if Im not mistaken, most of this code is duplicated on systemd_service_restart.rb

So if you see on my PR @vuntz suggested expanding systemd_service_restart.rb with the new action (in this case `environment_file`?) and creating a middle layer to maintain compatibility with the cookbooks calling `systemd_service_restart`

Although Im not sure if there is time constraints for this feature?

nicolasbock:
Thanks @Itxaka ! I did basically copy the `restart` code :wink: The only time constraint in a way is that https://github.com/crowbar/crowbar-openstack/pull/614 depends on this one and is blocked that way until your PR gets merged. But on the other hand, your approach sounds much cleaner. Let's get https://github.com/crowbar/crowbar-core/pull/1418 to move forward then!

Itxaka:
@nicolasbock take into account that #1418 is closed because it was not needed anymore, so I did not develope further and it will need some work, while this is already done.... so if it needs to get in to unblock https://github.com/crowbar/crowbar-openstack/pull/614 I will say lets move with this and create a card/bug/whatever for converging both resources into a new one.

nicolasbock:
Ok, that sounds good. I have tested this resource with the midonet one on top and it works as advertised.

Itxaka:
@nicolasbock looks like github malfunctioned and no tests are being run :(

nicolasbock:
I have restarted the Jenkins build. Hopefully it will work now.

nicolasbock:
Rebased

nicolasbock:
There are some HA failures, but I don't see how those are related or caused by this PR>

nicolasbock:
Rebased.

vuntz:
If you feel brave, as suggested by @Itxaka, I think it makes sense to merge the LWRPs, and provide a compatibility layer for the old existing one.

nicolasbock:
While I am working on refactoring the `systemd` providers, could we merge this one to unblock https://github.com/crowbar/crowbar-openstack/pull/614 ?

nicolasbock:
@Itxaka and @AbelNavarro Could you re-review this PR? I would prefer if we merge it this way to unblock https://github.com/crowbar/crowbar-openstack/pull/614 and I will work on refactoring the `systemd` providers. Thanks!

chef/cookbooks/utils/providers/systemd_environment.rb
```diff
@@ -0,0 +1,99 @@
...
+end
+
+action :create do
+  current_resource = @current_resource
+  converge_by("Create #{@new_resource}") do
+    execute "systemctl daemon-reload" do
+      action :nothing
+    end
+
+    directory _get_override_directory do
```
AbelNavarro: we may need to delete this directory on `:delete`

chef/cookbooks/utils/providers/systemd_environment.rb
```diff
@@ -0,0 +1,99 @@
...
+def _get_unit_name
+  "#{@new_resource.service_name}.service"
+end
+
+def _get_override_directory
+  "/etc/systemd/system/#{_get_unit_name}.d"
+end
+
+def _get_override_file_path
+  "#{_get_override_directory}/60-environment.conf"
```
AbelNavarro: Is this file name the same for all services?

nicolasbock: No this is specific to this override here.

chef/cookbooks/utils/providers/systemd_environment.rb
nicolasbock: Good point

nicolasbock:
@AbelNavarro TBH I copy and pasted that from an existing `systemd` provider. I 'll check what the `converge_by` primitive does. Thanks!

nicolasbock:
@AbelNavarro Here is a nice explanation: https://stackoverflow.com/questions/35238385/chef-what-is-meant-by-converge-by

nicolasbock:
@AbelNavarro I added your suggested directory delete. Please have another look.

nicolasbock:
Thanks @AbelNavarro . It's not urgently needed at the moment, but could be useful I suppose.

JanZerebecki:
This branch is closed for new features. Please reopen if you feel otherwise.

PR #1464, opened 2018-01-02
add barclamp-salt

bmwiedemann:
This is useful to apply changes or run commands
on groups of cloud nodes in parallel using calls like

salt-ssh controller\* -r 'grep Hz /proc/cpuinfo'

or to quickly test which nodes are up using
salt-ssh \* test.ping

TODO: add repo management


chef/cookbooks/salt/recipes/role_salt_ssh.rb
```diff
@@ -0,0 +1,73 @@
...
+  # http://download.nue.suse.com/ibs/SUSE/Updates/SLE-Module-Adv-Systems-Management/12/x86_64/update/
+  package "salt-ssh" do
+    action :install
+  end
+
+  nodes = node_search_with_cache("roles:network")
+  roster = []
+  nodes.each do |n|
+    nalias = n["crowbar"]["display"]["alias"] rescue nil
+    base_name_no_net = n[:fqdn].chomp(".#{n[:dns][:domain]}")
```
dirkmueller: this fails with

```


================================================================================
Recipe Compile Error in /var/chef/cache/cookbooks/salt/recipes/role_salt_ssh.rb
================================================================================

NoMethodError
-------------

Cookbook Trace:
---------------
  /var/chef/cache/cookbooks/salt/recipes/role_salt_ssh.rb:29:in `block in from_file'
  /var/chef/cache/cookbooks/salt/recipes/role_salt_ssh.rb:27:in `each'
  /var/chef/cache/cookbooks/salt/recipes/role_salt_ssh.rb:27:in `from_file'

```

for unready nodes breaking node discovery

chef/cookbooks/salt/recipes/role_salt_ssh.rb
```diff
@@ -0,0 +1,74 @@
...
+# limitations under the License.
+#
+
+if CrowbarRoleRecipe.node_state_valid_for_role?(node, "salt", "salt-ssh")
+  # requires repos
+  # http://download.nue.suse.com/ibs/SUSE/Products/SLE-Module-Adv-Systems-Management/12/x86_64/product/
+  # http://download.nue.suse.com/ibs/SUSE/Updates/SLE-Module-Adv-Systems-Management/12/x86_64/update/
+  package "salt-ssh" do
+    action :install
+  end
```
vuntz: Please don't put the real content of recipe here, and create another recipe that gets included from here.

The idea is to keep these role recipes extremely simple, so that we can add more orchestration logic easily to each of them.

chef/cookbooks/salt/recipes/salt_ssh.rb
```diff
@@ -0,0 +1,55 @@
...
+# http://download.nue.suse.com/ibs/SUSE/Products/SLE-Module-Adv-Systems-Management/12/x86_64/product/
+# http://download.nue.suse.com/ibs/SUSE/Updates/SLE-Module-Adv-Systems-Management/12/x86_64/update/
+package "salt-ssh" do
+  action :install
+end
+
+nodes = node_search_with_cache("roles:network")
+roster = []
+nodes.each do |n|
+  next if n[:dns].nil?
```
dirkmueller: better search for ```roles:dns-client OR roles:dns-server``` then (these are attributes coming from dns barclamp). just in case there is an inconsistency between network and dns barclamp, that avoids this extra next then.

bmwiedemann:
I could squash all 4 of them, if you like. Makes it easier to cherry-pick.

nicolasbock:
It's just that they look like a bunch of fix up changes on top of first change. I have nothing against splitting the PR into several commits, but I don't think we should have fixup commits in there.

nicolasbock:
@bmwiedemann Are you planning on squashing the commits?

bmwiedemann:
@nicolasbock whichever way you like it. I dont care much. The unsquashed version will have a bit more explanation in the commit messages though..

salt.yml
```diff
@@ -0,0 +1,31 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+barclamp:
+  name: 'salt'
+  display: 'Salt'
+  description: 'Salt Management'
+  version: 0
+  user_managed: true
```
dirkmueller: can you set that to user_managed: false to hide it from default installs?

bmwiedemann:
please re-review

PR #1465, opened 2018-01-04
[provisioner] cleanup hyperv files

bmwiedemann:
saving ~63MB disk space and bandwidth

The total crowbar-core is 65MB atm, and it turned out, most of that is not needed anymore.

chef-client for Windows took up most of the space.


rsalevsky:
Commit message should start with `provisioner:` and some more explanation would be good.

bmwiedemann:
updated

dirkmueller:
needs to be force-merged as it touches binary blobs.

PR #1466, opened 2018-01-04
provisioner: cleanup hyperv files

bmwiedemann:
saving ~63MB disk space and bandwidth

The total crowbar-core is 65MB atm, and it turned out, most of that
is not needed anymore.


bmwiedemann:
The master version introduced some regression https://ci.suse.de/job/openstack-mkcloud/100076/parsed_console/ so we need to get that fixed first and then add the fix to this backport.

bmwiedemann:
I squashed the fixups from #1467 into this, but I could add the commit IDs to this commit message once the 1467 is good and merged.

nicolasbock:
Yes, I think that would be a good idea, if purely from a forensics point of view :)

bmwiedemann:
should be good to merge now.

dirkmueller:
does this work for the maintenance update case? I don't see a migration added and we are editing a proposal template..

dirkmueller:
not relevant for SOC7 anymore.

PR #1468, opened 2018-01-08
Add log info for crowbar SSL(bsc#1073998)

sayalilunkad:
This patch refactors the SSL implementation for crowbar and adds some logging info to make
debugging easier.


sayalilunkad:
[P3 Bug 1073998 - SOC 7 Crowbar can not be switched to SSL anymore](https://trello.com/c/REVtrBJU/650-p3-bug-1073998-soc-7-crowbar-can-not-be-switched-to-ssl-anymore)

PR #1469, opened 2018-01-08
crowbar: Avoid crash on precached deleted nodes

dirkmueller:
When applying a proposal with deleted nodes it crashes due to
pre_cached_nodes containing a node with nil value.

requests?**


vuntz:
@dirkmueller Will you create the backport for stable/4.0?

PR #1470, opened 2018-01-09
Cleanup the Hyper-V cleanup

vuntz:
I don't think it makes sense to remove half of the code and half of the files. So instead, restore small bits, add an exception to the code in case the Hyper-V setup is done and would fail due to the large files missing, and properly remove the OS from the list of supported OS, with a migration.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/setup_base_images.rb
+++ b/chef/cookbooks/provisioner/recipes/setup_base_images.rb
@@ -342,6 +342,88 @@ if node[:provisioner][:default_os].nil?
   dirty = true
 end
 
+unless node[:provisioner][:supported_oses].keys.select{ |os| /^(hyperv|windows)/ =~ os }.empty?
+  common_dir="#{tftproot}/windows-common"
+  extra_dir="#{common_dir}/extra"
+
+  directory "#{extra_dir}" do
+    recursive true
+    mode 0o755
+    owner "root"
+    group "root"
+    action :create
+  end
+
+  # Copy the crowbar_join script
+  cookbook_file "#{extra_dir}/crowbar_join.ps1" do
+    owner "root"
+    group "root"
+    mode "0644"
+    action :create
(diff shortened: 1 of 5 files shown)


susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
e4a0b59dcfffd306a021b663a5c4d156999235cf</td>
    </tr>
    <tr>
      <td>:warning:</td>
123401a54fc887ceff6ee34c2e84ea23865ecd29</td>
    </tr>
    <tr>
      <td>:warning:</td>
1c56599bfb77ba2878070d4835ef51606be615d0</td>
    </tr>
  </tbody>
</table>

</p>

bmwiedemann:
I see that the migration is useful, but why keep .ps1 files around? YAGNI says, If they are needed again later, they can be added back later.
And having some leftover 'if windows' codepathes does not hurt. Didnt want to spend more time on finding those.

vuntz:
> I see that the migration is useful, but why keep .ps1 files around? YAGNI says, If they are needed again later, they can be added back later.

For the same reason we have templates for RHEL nodes, and also because there's at least one other windows-only template that wasn't removed in the initial PR.

It doesn't hurt to keep them, and the argument in the initial PR was only about the size -- it doesn't apply for them.

chef/cookbooks/provisioner/recipes/setup_base_images.rb
```diff
@@ -342,6 +342,90 @@
   dirty = true
 end
 
+unless node[:provisioner][:supported_oses].keys.select{ |os| /^(hyperv|windows)/ =~ os }.empty?
+  raise "Binary files (chef-client, curl) need to be added back to the cookbook for Hyper-V support"
```
dirkmueller: add a link to the PR/git commit to revert?

PR #1471, opened 2018-01-10
[4.0] crowbar : accept alias in node transition call

ilausuch:
crowbarctl node transition <name> <state> didn't accept alias
as name of the node. The problem was in API controller. To fix it
first it tries to get the node information from name or alias and
then calls to transition function using the real name.

(cherry picked from commit 9273c1c2a29a1a8f28b0e45114fd657de70116a4)


ilausuch:
[Check/fix `crowbarctl node *` alias support.](https://trello.com/c/T03pTNx0/222-check-fix-crowbarctl-node-alias-support)

vuntz:
HA CI failed, but I'm going to merge because the deployment passed; things failed later on when testing openstack -- and that's clearly unrelated to this patch.

ilausuch:
+1

PR #1472, opened 2018-01-10
clean up UI display of recent Chef logs on failure

aspiers:
When `chef-client` fails, the most recent lines of the Chef logs are surfaced in the Crowbar proposal UI.  However they are not wrapped in anything, so the line breaks gets ignored, creating an unholy mess.  Not sure how this was tolerated for so long, but let's fix it.

Also a couple of refactorings to clean up the relevant method.

### DIFF
--- a/crowbar_framework/app/assets/stylesheets/content/alert.scss
+++ b/crowbar_framework/app/assets/stylesheets/content/alert.scss
@@ -7,4 +7,14 @@
   p:last-child {
     margin-bottom: 0
   }
+
+  pre {
+    background-color: $state-danger-pre-bg;
+    max-height: 20em;
+    overflow: scroll;
+
+    br {
+      display: none;
+    }
+  }
 }
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1816,21 +1816,21 @@ class ServiceObject
 
   def get_log_lines(node)
     begin
-      l_counter = 1
-      find_counter = 0
+      line_count = 0
+      last_delimiter_line = 0
       f = File.open("/var/log/crowbar/chef-client/#{node}.log")
       f.each do |line|
-        if line == "="*80
-           find_counter = l_counter
+        if line == "=" * 80
+          last_delimiter_line = line_count
         end
-        l_counter += 1
+        line_count += 1
       end
       f.seek(0, IO::SEEK_SET)
       logged_lines =
-        if (find_counter > 0) && (l_counter - find_counter) < 50
-          f.readlines[find_counter -3..l_counter]
+        if (last_delimiter_line > 0) && (line_count - last_delimiter_line) < 50
+          f.readlines[last_delimiter_line -3..line_count]
         else
-          f.readlines[l_counter-50..l_counter]
+          f.readlines[line_count-50..line_count]
         end
       "Most recent logged lines from the Chef run: \n\n<pre>" +
         logged_lines.join + "</pre>"
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1814,11 +1814,11 @@ class ServiceObject
     end
   end
 
-  def get_log_lines(pid)
+  def get_log_lines(node)
     begin
       l_counter = 1
       find_counter = 0
-      f = File.open("/var/log/crowbar/chef-client/#{pid}.log")
+      f = File.open("/var/log/crowbar/chef-client/#{node}.log")
       f.each do |line|
         if line == "="*80
            find_counter = l_counter
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1832,8 +1832,8 @@ class ServiceObject
         else
           f.readlines[l_counter-50..l_counter]
         end
-      "Most recent logged lines from the Chef run: \n\n" +
-        logged_lines.join(" ")
+      "Most recent logged lines from the Chef run: \n\n<pre>" +
+        logged_lines.join + "</pre>"
     rescue
       Rails.logger.error("Error reporting: Couldn't open /var/log/crowbar/chef-client/#{pid}.log ")
       raise "Error reporting: Couldn't open  /var/log/crowbar/chef-client/#{pid}.log"
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1826,12 +1826,15 @@ class ServiceObject
         line_count += 1
       end
       f.seek(0, IO::SEEK_SET)
-      logged_lines =
+      starting_line =
+        # If we found a delimiter in the last (say) 10 lines, we don't need
+        # to show all of the last 50.
         if (last_delimiter_line > 0) && (line_count - last_delimiter_line) < 50
-          f.readlines[last_delimiter_line -3..line_count]
+          last_delimiter_line - 3
         else
-          f.readlines[line_count-50..line_count]
+          line_count - 50
         end
+      logged_lines = f.readlines[starting_line..line_count]
       "Most recent logged lines from the Chef run: \n\n<pre>" +
         logged_lines.join + "</pre>"
     rescue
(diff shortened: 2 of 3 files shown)


susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
fda8ea9c187daaf3a3bbd1c8da4ffa3c17a06981</td>
    </tr>
  </tbody>
</table>

</p>

aspiers:
@nicolasbock Will do; I would have done that instantly but it took me a few moments to understand the cryptic variable names.

nicolasbock:
I am not surprised ;)

aspiers:
@nicolasbock Fixed. I don't agree with the remaining Hound warning.

aspiers:
Added a PoC attempt to get a nicer background colour.  Pretty sure this is wrong somehow; I've never made changes which could affect the branding CSS.

jsuchome:
@aspiers do you have some failure example, so we can see the differences? (not necessary screenshot, just the way to reproduce some error you're testing)

BTW, adding @skazi0 , he's far better than me in this areas

aspiers:
@jsuchome Any proposal failure - they are not hard to find, just check the Jenkins Green boards ;-) But yes I should include a screenshot here.

jsuchome:
So this is how it looks for me http://imagebin.suse.de/2107/img ... but I have not been able to use that `state-danger-pre-bg` variable, `rake assets:precompile` fails with Undefined variable: "$state-danger-pre-bg". ... which is probably caused my missing knowledge of how to correctly apply your patch...

aspiers:
@jsuchome Thanks for testing!  I did not test the SASS changes yet, so it's easily possible I screwed up somehow.  But interesting that you did not get the same white background for the Chef logs which I got (that's what the variable is supposed to fix).

Yes, you can see from the screenshot that there is an unwanted extra blank line in between every line of the Chef logs, but I have no clue where these extra `<br>` elements are coming from :-/

crowbar_framework/vendor/assets/stylesheets/variables.scss
```diff
@@ -340,6 +340,7 @@ $state-warning-border: darken(adjust-hue($state-warning-bg, -10), 5%);
 
 $state-danger-text: #a94442;
 $state-danger-bg: #f2dede;
+$state-danger-pre-bg: lighten($state-danger-bg, 5%);
```
skazi0: Move this to `bootstrap/_variables.scss`. `variables.scss` doesn't seem to be included anywhere.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1814,23 +1814,29 @@ def handle_validation_errors
...
+        # If we found a delimiter in the last (say) 10 lines, we don't need
+        # to show all of the last 50.
+        if (last_delimiter_line > 0) && (line_count - last_delimiter_line) < 50
+          last_delimiter_line - 3
+        else
+          line_count - 50
+        end
+      logged_lines = f.readlines[starting_line..line_count]
+      "Most recent logged lines from the Chef run: \n\n<pre>" +
+        logged_lines.join(" ") + "</pre>"
```
skazi0: Why not join with empty string? The space appears as an indentation starting from second line.

aspiers: Oops, good point :) Fixed.

crowbar_framework/vendor/assets/stylesheets/variables.scss
aspiers: Thanks, fixed.

aspiers:
> I think the problem with <BR> comes from: https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/views/shared/_flashes.html.haml#L6
>
> See: https://apidock.com/rails/ActionView/Helpers/TextHelper/simple_format

Good catch, that's it!  Any ideas how to cleanly fix it to stop doing that in this particular case?

skazi0:
@aspiers I don't have a ready solution. I will try to find something tomorrow.

aspiers:
@skazi0 Thanks, really appreciate your help :)

skazi0:
@aspiers the best solution I found is this: 

    = simple_format(message).tr("\n", '').html_safe

`simple_format` converts newlines to `<br>` or '<p>` but doesn't remove the newlines. In normal HTML these should be ignored and rendering is based on the above tags. We can safely remove the newlines and the rendering should be correct. `html_safe` flag set by `simple_format` is lost after `tr` so we need to mark the string as safe again.

Alternative is more hacky, css-only, not-sure-how-portable: `pre br { display: none; }` (you should probably add some class to this `<pre>` and use that in css to not break any other views).

aspiers:
Jacek Tomasiak <notifications@github.com> wrote:
>@aspiers the best solution I found is this:
>
>    = simple_format(message).tr("\n", '').html_safe
>
>`simple_format` converts newlines to `<br>` or '<p>` but doesn't remove the newlines. In normal HTML these should be ignored and rendering is based on the above tags. We can safely remove the newlines and the rendering should be correct. `html_safe` flag set by `simple_format` is lost after `tr` so we need to mark the string as safe again.

That would work, but since this is done in
app/views/shared/_flashes.html.haml, how would we actually implement
that for flashes which contain Chef logs or other <pre> blocks,
without breaking all the other flashes out there?

>Alternative is more hacky, css-only, not-sure-how-portable: `pre br { display: none; }` (you should probably add some class to this `<pre>` and use that in css to not break any other views).

Ooooh that's evil! ... I like it ;-)  By far the easiest solution.  If
you look at the PR you'll see that the CSS for this <pre> is already
wrapped inside an .alert { } block, so it wouldn't break anything
else.  I'll try it.

crowbar_framework/app/assets/stylesheets/content/alert.scss
```diff
@@ -7,4 +7,14 @@
   p:last-child {
     margin-bottom: 0
   }
+
+  pre {
+    background-color: $state-danger-pre-bg;
+    overflow: scroll;
+    max-height: 20em;
+
+    br {
```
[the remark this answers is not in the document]
aspiers: It looks like here Hound is being more aggressive than [the default behaviour of `scss-lint`](https://github.com/brigade/scss-lint/blob/3c1aeb4d7a5a5aba0a7aa1fc2e8fcd929026df36/lib/scss_lint/linter/README.md#selectordepth); I think this should be ignored.

aspiers:
Here's roughly what it looks like now, although I hacked the CSS into Chrome manually rather than going through the whole asset compilation dance, so the background colour for the `<pre>` might be wrong still.

chef-failures

aspiers:
Still not sure why the first line is outdented by 1 column.

nicolasbock:
But this does look a whole lot better than before!

skazi0:
@aspiers re outdent: have you re-applied the proposal after removing space from join? the rendered outputs are stored in proposal so only css stuff gets updated without re-applying.

skazi0:
@aspiers re my first solution and breaking other flashes: I don't think it changes anything besides fixing the empty lines in `<pre>`.

My thinking:

```
message = "first line\nsecond line"
simple_format(message) == "first line\n<br />second line"
```
When this is rendered as HTML the `\n` is ignored by the browser. The only exception is in `<pre>` or similar (e.g. css: `white-space: pre`). I don't see how removing the newlines from `simple_format` output could break anything anywhere.

aspiers:
> @aspiers re outdent: have you re-applied the proposal after removing space from join? the rendered outputs are stored in proposal so only css stuff gets updated without re-applying.

Yes this was a fresh test. The problem is somewhere else.

skazi0:
@aspiers re outdent again, take a closer look, it's not an outdent but a missing indent on the first line. Spaces are there for padding the line numbers.

aspiers:
@skazi0 Not sure I follow? But no outdent here, weirdly! - http://crowbar.vk3.cloud.suse.de/crowbar/neutron/1.0/proposals/default

aspiers:
Neutron failed in the gate (totally unrelated of course), giving us a nice opportunity to examine the results of this, and it looks good. So I think this is ready to go.

PR #1473, opened 2018-01-11
crowbar: fix call to gensslcert

aplanas:
Fix the call to `gensslcert`, to use the common name parameter, instead the country one


aplanas:
Tested in vk7

aplanas:
@nicolasbock  updated the github comment and the commit

aplanas:
@rsalevsky I do not think that this needs backport: https://github.com/crowbar/crowbar-core/blob/stable/4.0/chef/cookbooks/crowbar/recipes/default.rb#L390

PR #1475, opened 2018-01-12
[4.0] clean up UI display of recent Chef logs on failure

aspiers:
Backport of #1472.

When `chef-client` fails, the most recent lines of the Chef logs are surfaced in the Crowbar proposal UI.  However they are not wrapped in anything, so the line breaks gets ignored, creating an unholy mess.  Not sure how this was tolerated for so long, but let's fix it.

Also a couple of refactorings to clean up the relevant method.


vuntz:
HA CI is not relevant here, merging.

PR #1477, opened 2018-01-15
[4.0] Filter out empty elements in crowbar lists

jhesketh:
When supplying a list (eg of hostnames: "host1, host2, , host3" etc)
an empty element "" would be set in the array. This change removes any
empty elements.

This change also stops empty strings from being converted to arrays
such as [""] (ie one empty element).

(bsc#1065903)

(cherry picked from commit 247f621)

Backport of https://github.com/crowbar/crowbar-core/pull/1453


vuntz:
Merging as HA job is using ceph and not swift anyhow...

PR #1478, opened 2018-01-15
[4.0] Add all .log file to apache logrotate(bsc#1071317)

sayalilunkad:
Currently only the apache access and error logs are rotated by logrotate
but /var/log/apache also contains access and error logs from keystone,
ceilometer, barbican and other services which are not rotated.
This adds a file which rotates all the logs in the apache directory.

(cherry picked from commit 7df53dca5f144ac670037d1006d6977241f35cc3)

### DIFF
--- a/chef/cookbooks/apache2/recipes/default.rb
+++ b/chef/cookbooks/apache2/recipes/default.rb
@@ -153,6 +153,13 @@ unless node[:platform_family] == "suse"
   end
 end
 
+template "/etc/logrotate.d/apache2-all" do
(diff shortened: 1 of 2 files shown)


sayalilunkad:
[P2 Bug 1071317 - L3: log rotation broken for ceilometer-api, keystone-admin and keystone-public logs](https://trello.com/c/xO7kK6Dr/627-p2-bug-1071317-l3-log-rotation-broken-for-ceilometer-api-keystone-admin-and-keystone-public-logs)

PR #1479, opened 2018-01-15
crowbar: Attribute names in validation errors

skazi0:
When validation errors happened, only error message like "not a string" was displayed.

Attribute path was added to the message to better identify possible source of error.

requests?**


jsuchome:
Please, add an example of the new error

skazi0:
Example:
Before:
zrzut ekranu z 2018-01-16 11-50-05
After:
zrzut ekranu z 2018-01-16 11-48-59

PR #1480, opened 2018-01-16
provisioner: Update grub.efi location for aarch64

dirkmueller:
SP2+ are installing the grub2 efi compat symlink under /usr/lib/grub2
not /usr/lib64/grub2 like on x86_64. adjust paths accordingly


vuntz:
Since it's for SP2+, do we want to backport this to stable/4.0?

PR #1481, opened 2018-01-18
crowbar: Ensure epmd.socket is configured correctly

vuntz:
With new erlang and rabbitmq-server packages, rabbitmq now depends on
the system-wide epmd service, which will only listen on IP addresses
defined for epmd.socket.

We need epmd to listen on the admin IP address (the one that resolves
the FQDN), so configure things accordingly.

While install-suse-cloud will already do this setup correctly, we also
change the crowbar cookbook to ensure the config stays up-to-date.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/default.rb
+++ b/chef/cookbooks/crowbar/recipes/default.rb
@@ -309,6 +309,44 @@ utils_systemd_service_restart "chef-expander"
 utils_systemd_service_restart "chef-server"
 
 if node[:platform_family] == "suse"
+  # With new erlang packages, we move to a system-wide epmd service, with a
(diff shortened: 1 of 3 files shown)


chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -21,6 +21,8 @@
   include_recipe "bluepill"
 end
 
+include_recipe "barclamp::default"
```
Itxaka: Im guessing you do this to have the `Barclamp` namespace available, but as far as I understod, just depending on it should make it available to the recipe no? its loaded on the context of the chef run for sure so Im not sure if we need this.

vuntz: True, this required a small change, but done.

PR #1483, opened 2018-01-21
[4.0] crowbar: Attribute names in validation errors

vuntz:
Before this change, when validation errors happened, only error message
like "not a string" was displayed. Attribute path was added to the
message to better identify possible source of error.

(cherry picked from commit 656605aabc9575408151e15ca53c2b3cf43fd24b)

Backport of https://github.com/crowbar/crowbar-core/pull/1479


vuntz:
HA job has no impact here, merging.

PR #1485, opened 2018-01-23
provisioner: Fix UEFI based discovery for SP3+

dirkmueller:
The grub behavior in SP3 appears to have changed, it is looking
for grub.cfg in the same directory like the grub.efi resides, not
in the parent directory like before.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/setup_base_images.rb
+++ b/chef/cookbooks/provisioner/recipes/setup_base_images.rb
@@ -122,6 +122,7 @@ discovery_arches.each do |arch|
   # we use grub2; steps taken from
   # https://github.com/openSUSE/kiwi/wiki/Setup-PXE-boot-with-EFI-using-grub2
   grub2arch = arch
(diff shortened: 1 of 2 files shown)


chef/cookbooks/provisioner/recipes/setup_base_images.rb
```diff
@@ -158,11 +159,14 @@
...
-    code shim_code
-    action :nothing
-    subscribes :run, resources("template[#{uefi_dir}/default/grub.cfg]"), :immediately
+  # Secure Boot Shim
+  if arch == "x86_64"
+    bash "Copy UEFI shim loader with grub2" do
+      cwd "#{uefi_dir}/default/boot"
+      code shim_code
+      action :nothing
+      subscribes :run, resources("template[#{grubcfgfile}]"), :immediately
```
nicolasbock: Don't you have to update the template resource call as well to make it match the new path?

rsalevsky:
@dirkmueller This patch let's the crowbar installation fail in the CI.

PR #1486, opened 2018-01-24
crowbar-hacks: Skip activation of HPE UEFI entries

skazi0:
HPE servers have some standard UEFI boot entries which are disabled
by default. After crowbar activates them, they are re-created by the
firmware with disabled state. Crowbar activates these duplicates and
the whole cycle repeats after each reboot. This can lead to very long
lists of boot entries which can even cause the machine to fail booting.

The workaround for this problem is a list of entry names which are
skipped during the activation loop. Currently the list includes only
the HPE specific entries but can be extended in the future if needed.

requests?**
https://trello.com/c/Tx6hp73H/239-uefi-boot-entry-numbers-increase-with-every-hardware-installation

### DIFF
--- a/chef/cookbooks/crowbar-hacks/recipes/uefi.rb
+++ b/chef/cookbooks/crowbar-hacks/recipes/uefi.rb
@@ -14,12 +14,32 @@
 # limitations under the License.
 #
 
+# List of UEFI boot entries excluded from activation loop.
+# Some HP hardware re-creates these entries as disabled when they are activated
+# expanding the list of boot entries on every reboot.
+ignored_boot_entries = [
+  "Embedded UEFI Shell",
+  "Diagnose Error",
+  "System Utilities",
+  "Intelligent Provisioning",
+  "Boot Menu",
+  "Network Boot",
+  "Embedded Diagnostics",
+  "View Integrated Management Log"
+]
+
 ruby_block "uefi_boot_order_config" do
   block do
     if node["uefi"] && File.exist?("/sys/firmware/efi")
       node["uefi"]["boot"]["order"].each do |order|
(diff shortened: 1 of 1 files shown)


skazi0:
I tested this on two machines and it seems to work correctly. 
This is the output of `efibootmgr` after a couple of reboots:

```
BootCurrent: 0009
Timeout: 0 seconds
BootOrder: 000C,000B,000D,0002,0009,000A,0000,0001,0003,0004,0005,0006,0007,0008
Boot0000  Embedded UEFI Shell
Boot0001  Diagnose Error
Boot0002  System Utilities
Boot0003  Intelligent Provisioning
Boot0004  Boot Menu
Boot0005  Network Boot
Boot0006  Embedded Diagnostics
Boot0007  View Integrated Management Log
Boot0008* Generic USB Boot
Boot0009* Slot 2 : Smart HBA H240 Controller - 1117.76 GiB, RAID 0 Logical Drive(Target:0
Boot000A* Slot 2 : Smart HBA H240 Controller - 1117.76 GiB, RAID 0 Logical Drive(Target:0
Boot000B* Slot 1 Port 1 : HP Ethernet 10Gb 2-port 530SFP+ Adapter - CNA (PXE IPv6) 
Boot000C* Slot 1 Port 1 : HP Ethernet 10Gb 2-port 530SFP+ Adapter - CNA (PXE IPv4) 
Boot000D* sles-secureboot
```

The `Generic USB Boot` doesn't seem to be handled by the firmware as one of the default entries.

skazi0:
[UEFI boot entry numbers increase with every hardware installation](https://trello.com/c/Tx6hp73H/239-uefi-boot-entry-numbers-increase-with-every-hardware-installation)

PR #1497, opened 2018-01-31
provisioner: remove repos before adding ours

bmwiedemann:
to make it easier to re-run crowbar_register in case of errors

### DIFF
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -277,6 +277,7 @@ fi
 # Setup the repos
 # ---------------
 
+rm -f /etc/zypp/repos.d/*.repo
 <% @repos.keys.sort.each do |name| %>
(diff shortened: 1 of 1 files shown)


jsuchome:
can't users have some custom repos?

bmwiedemann:
They could have custom repos, but then those repo names could collide with crowbar's too ... and if they are not SUSE repos, packages could collide which we do not want to support.

nicolasbock:
But if the customer wants to add custom repositories, when would they need to add them? Is `crowbar_register` run only when you allocate a new node?

bmwiedemann:
it only runs once.

an alternative would be to do `zypper rr $repo` before adding it, so we only drop repos with our names.

jsuchome:
`zypper rr $repo` sounds indeed safer, however as pointed out in the case of upgrade, we actually do remove all present repos at some point ... so I think it does not hurt in this time, esp. when the user can add extra repos after runing crowbar_register

PR #1498, opened 2018-01-31
[salt] add barclamp-salt

bmwiedemann:
backport of #1464 because CI there is constantly failing [for unrelated issues](https://bugzilla.suse.com/show_bug.cgi?id=1075368)
and we want this for our SOC7 deployments anyway.


bmwiedemann:
should I squash commits together?

rsalevsky:
We can do this as separate internal package but that's nothing what should go to crowbar-core.

PR #1500, opened 2018-02-05
provisioner: Use BarclampLibrary to search for available boot disks

jsuchome:
Sometimes looking for a disk under /dev/disk/by-path does not give
correct results  - paths for some devices might yet exist.

So get the list of unclaimed devices from BarclampLibrary, which has it
from ohai. Exclude removable devices.

See https://bugzilla.suse.com/show_bug.cgi?id=1063523 for initial problem discussion

### DIFF
--- a/chef/cookbooks/provisioner/recipes/bootdisk.rb
+++ b/chef/cookbooks/provisioner/recipes/bootdisk.rb
@@ -23,6 +23,19 @@ if node["crowbar_wall"]["registering"]
  node.save
 end
 
+# Look at the disks known by BarclampLibrary first
+unless node[:crowbar_wall][:boot_device]
+  BarclampLibrary::Barclamp::Inventory::Disk.unclaimed(node, true).each do |unclaimed_disk|
+    next if unclaimed_disk.nil? || unclaimed_disk.removable?
+    dev = unclaimed_disk.name
+    unclaimed_disk.claim("Boot")
+    dev = dev.split("/")[-1]
+    node[:crowbar_wall][:boot_device] = dev
+    node.save
+    break
+  end
+end
+
 # Find the first thing that looks like a hard drive based on
 # PCI bus enumeration, and use it as the target disk.
 # Unless some other barclamp has set it, that is.


susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
801423054f6fa2db8539f8b839fbbf607301d51b</td>
    </tr>
  </tbody>
</table>

</p>

jsuchome:
@nicolasbock look at https://bugzilla.suse.com/show_bug.cgi?id=1063523#c7

jsuchome:
@vuntz what if I ignore `/dev/disk/by-path` checking completely and go directly looking under `/dev/disk/by-id`, is that also wrong?

vuntz:
> @vuntz what if I ignore /dev/disk/by-path checking completely and go directly looking under /dev/disk/by-id, is that also wrong?

Well, yes, it's the same thing, you lose the first pass that is used for ordering.

vuntz:
> > What about using this code, as a fallback solution in case the current code doesn't find what it wants?
> > 
> That's the problem - current code finds something, but we do not know that it is wrong. In the cases described in the bug, there was only one entry under /dev/disk/by-path, although there were 2 disks present (and both were found by ohai)

Ah. Interesting. And annoying. I don't have an immediate idea on how to fix this :/

KeithMnemonic:
This is blocking a bug from being closed. Can we get an update on this and determine whether we can do something to provide a workaround or if not close the bug as WONTFIX

jsuchome:
@JanZerebecki you were looking at the disk search code lately, do you have any idea for better alternative to this?

JanZerebecki:
Why does the claiming by mount point (line 11) fail? Fixing this would be better than a workaround for the fallback.

It seems `/dev/disk/by-path` often doesn't contain all things on the system that might be addressed as a disk. There may be bugs in the kernel and udev. The kernel doesn't declare the device (e.g. `/sys/block/sda/device`) for some block(device)s. 
Is that the case here? What does `ls -la /sys/block/REPLACE` show for the device in question? If that does not contain a device symlink, is there an entry for it somewhere under `/sys/devices/`?

E.g. all device mapper blocks, like created with lvm, even when they map to only one device, don't have a device declared in that way. Though they have entries like `/sys/devices/virtual/block/dm-0`.

Maybe there is an ordering for the existing `by-path` entries and this type of device from `by-id` that works. Maybe we need to detect Xen similar to the VirtualBox workaround.

jsuchome:
@JanZerebecki I have this in my setup:
```
# ls -al /dev/disk/by-path/
lrwxrwxrwx 1 root root   9 Apr 17 08:39 pci-0000:00:01.1-ata-1 -> ../../sdb
lrwxrwxrwx 1 root root   9 Apr 17 08:39 pci-0000:00:01.1-scsi-0:0:0:0 -> ../../sda
lrwxrwxrwx 1 root root   9 Apr 17 08:39 pci-0000:00:01.1-scsi-0:0:1:0 -> ../../sdb

# ls -la /sys/block/
lrwxrwxrwx  1 root root 0 Apr 17 08:39 sda -> ../devices/pci0000:00/0000:00:01.1/ata1/host0/target0:0:0/0:0:0:0/block/sda
lrwxrwxrwx  1 root root 0 Apr 17 08:39 sdb -> ../devices/pci0000:00/0000:00:01.1/ata1/host0/target0:0:1/0:0:1:0/block/sdb
lrwxrwxrwx  1 root root 0 Apr 17 08:41 vda -> ../devices/pci0000:00/0000:00:04.0/virtio0/block/vda
```

Links under `/dev/disk/by-path/` looks incorrect, that might be behind the buggy behavior

JanZerebecki:
> Links under /dev/disk/by-path/ look incorrect, that might be behind the buggy behavior
How are these incorrect?

Sorry, i forgot a slash. To check if the device entry is missing, please add the output of: `ls -la /sys/block/vda/`

jsuchome:
2 links under `/dev/disk/by-path/` point to the same `sdb`

jsuchome:
`/sys/block/[vs]d[ab]/` exists for all 3 disks

JanZerebecki:
> 2 links under /dev/disk/by-path/ point to the same sdb

That is by nature, the same path can be expressed in different ways. There is not only one path for each disk, even if there is only one physical path.

Also a disk path may not be unique among all disks. It always happens on different distributions. But even if that was introduced by accident, it now needs to stay for backwards compatibility. That means our code needs to cope with that. In this case pci-0000:00:01.1-ata-1 is a path for both sda and sdb, as ata1 has two disks behind it.

What I wanted to know is if /sys/block/vda/device exists? Because I suspect it missing causes udev to not add any `by-path` entries for it, but I haven't looked at udev to confirm.

JanZerebecki:
I agree. The current code already has code to find virtio block devices by iterating `/sys/block`, but outside the ordering via `path` and `dev`. So a solution could be to get the list of devices from somewhere that is not `by-path` and then look up (one or multiple?) path to get the same ordering. That means rewriting the rest of the file.

We should file a bug in the appropriate place, to get the `by-path` for this case fixed, even if we stop using it. Upstream has rules for virtio devices, I wonder why these don't work in this case:
https://github.com/systemd/systemd/blob/e2c2d70/rules/60-persistent-storage.rules#L34

JanZerebecki:
The commit seems to be the same as before?

jsuchome:
@JanZerebecki now, instead of taking first disk from BarclampLibrary, I take first that is not `removable?`

JanZerebecki:
Ah, ok. I think this either bypasses the ordering by skipping reading from `by-path` or won't work for this problem case as the code reading from `by-path` will override it, would need to look more closely which it is. But either one makes this not work.

That mkcloud adds the intended boot device under pci id 4 and adds another disk under pci id 1 makes this more complicated. Perhaps we can also/first fix that in mkcloud.

Why not get rid of the layering of half working/broken code and also rewrite the rest of the file?
To replace the reading from `by-path`: For each disk from BarclampLibrary get a path from `/sys` and sort by that.

jsuchome:
> I think this either bypasses the ordering by skipping reading from by-path ..

Yes, it skips the by-path block. It does not mean the ordering is bypassed, there's some ordering in the items generated by BarclampLibrary itself.

JanZerebecki:
The order from BarclampLibrary is derived from the order of `Dir["/sys/block/*"]` which should not be relied up on (as it is implementation defined). Also if it is sorted by name, that is the wrong order.

jsuchome:
Looks like not needed any more

PR #1509, opened 2018-03-06
crowbar: clean restart flags extended

ilausuch:
This commit extends the functionality of 690e9bb6cd71e93f04cdb4ddd1639a5a785894dc. This commit allows to the final user to disable the restart of services when the chef client runs. If it is active for any cookbook and one of its services need to clean the flag is raised. To clean this flag there is the command "clear_restart" that the user will run when after the manual restart. In this command the service name have to be specified. However is interesting to restart all services of the one cookbook or restart all services in a node.

This commit allows to specify the name of the cookbook, cleaning all the service of this cookbook, also allows to don't specify any cookbook that will clean all flags in all cookbooks.

### DIFF
--- a/crowbar_framework/app/controllers/api/restart_management_controller.rb
+++ b/crowbar_framework/app/controllers/api/restart_management_controller.rb
@@ -37,7 +37,8 @@ class Api::RestartManagementController < ApiController
   api :POST, "/api/restart_management/restarts", "Clean the service restart flags"
   header "Accept", "application/vnd.crowbar.v2.0+json", required: true
   param :node, String, desc: "Node name", required: true
-  param :service, String, desc: "Service to clean restart flag for", required: true
+  param :cookbook, String, desc: "Cookbook to clean restart flags for", required: false
+  param :service, String, desc: "Service to clean restart flag for", required: false
   api_version "2.0"
 
   api :GET, "/api/restart_management/restarts", "Get a list of services that need restart"
@@ -81,22 +82,39 @@ class Api::RestartManagementController < ApiController
 
   def restarts_post
     params.require(:node)
-    params.require(:service)
 
     node_name = params[:node]
+    cookbook = params[:cookbook]
     service = params[:service]
 
     node = get_node_or_raise(node_name)
+    dirty = false
 
-    managed_cookbooks.each do |cookbook|
-      next unless node.key? :crowbar_wall
-      next unless node[:crowbar_wall].key? :requires_restart
-      next unless node[:crowbar_wall][:requires_restart].key? cookbook
-      next unless node[:crowbar_wall][:requires_restart][cookbook].key? service
-      node[:crowbar_wall][:requires_restart][cookbook].delete(service)
+    if cookbook.nil?
+      node[:crowbar_wall].delete(:requires_restart)
+      Rails.logger.info("restart_management: node #{node_name} cleaned")
+      dirty = true
+    else
+      raise Crowbar::Error::NotFound unless managed_cookbooks.include? cookbook
+
+      if service.nil?
+        node[:crowbar_wall][:requires_restart].delete(cookbook)
+        Rails.logger.info("restart_management: cookbook #{cookbook} flags cleaned on node "\
+          "#{node_name}")
+        dirty = true
+      else
+        if node[:crowbar_wall][:requires_restart].key? cookbook
+          node[:crowbar_wall][:requires_restart][cookbook].delete(service)
+          Rails.logger.info("restart_management: service #{service} of cookbook #{cookbook} "\
+          "flags cleaned in node #{node_name}")
+          dirty = true
+        else
+          raise Crowbar::Error::NotFound("Cookbook #{cookbook} not found on node #{node_name}")
+        end
+      end
     end
 
-    node.save
+    node.save if dirty
 
     head :ok
   end
@@ -129,7 +147,7 @@ class Api::RestartManagementController < ApiController
   end
 
   def get_node_or_raise(node_name)
-    node = NodeObject.find("name:#{node_name}").first
+    node = NodeObject.find_node_by_name_or_alias(node_name)
     raise Crowbar::Error::NotFound if node.nil?
     node
   end
--- a/crowbar_framework/spec/controllers/api/restart_management_controller_spec.rb
+++ b/crowbar_framework/spec/controllers/api/restart_management_controller_spec.rb
@@ -17,7 +17,7 @@
 require "spec_helper"
 require "json"
 
-describe Api::RestartManagementController, type: :request, focus: true do
+describe Api::RestartManagementController, type: :request, restartmanagement: true do
   let(:headers) { { ACCEPT: "application/vnd.crowbar.v2.0+json" } }
 
   context "Feature enabled" do
@@ -52,6 +52,7 @@ describe Api::RestartManagementController, type: :request, focus: true do
       allow(NodeObject).to receive(:find_by_name).with("testing.crowbar.com").and_return(
         @testing_node
       )
+
       @databag_item = ::Chef::DataBagItem.load("crowbar-config", "disallow_restart")
 
       # mock experimental config to test the controller
@@ -164,14 +165,15 @@ describe Api::RestartManagementController, type: :request, focus: true do
       end
     end
 
-    context "POST restarts" do
-      it "cleans the restart flag for a service" do
+    context "POST restarts", restartmanagement_post_restarts: true do
+      it "cleans the restart flag for a service by service name",
+        restartmanagement_by_cookbook: true do
         @admin_node.set["crowbar_wall"]["requires_restart"]["nova"] = {
           "openstack-nova-api" => {}
         }
-        allow(NodeObject).to receive(:find).with(
-          "name:admin.crowbar.com"
-        ).and_return([@admin_node])
+        allow(NodeObject).to receive(:find_node_by_name_or_alias).with(
+          "admin.crowbar.com"
+        ).and_return(@admin_node)
 
         # check that the key is in ther first
         expect(
@@ -179,6 +181,7 @@ describe Api::RestartManagementController, type: :request, focus: true do
         ).to include("openstack-nova-api")
         post "/api/restart_management/restarts", {
           node: "admin.crowbar.com",
+          cookbook: "nova",
           service: "openstack-nova-api"
         }, headers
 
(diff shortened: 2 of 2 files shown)


ilausuch:
[disallow_restart feature: allow clearing whole node](https://trello.com/c/LX8YZEDF/270-disallowrestart-feature-allow-clearing-whole-node)

susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
52b23f08dc46b5e0be2e12d7714ab10a241ac8ef</td>
    </tr>
  </tbody>
</table>

</p>

crowbar_framework/app/controllers/api/restart_management_controller.rb
```diff
@@ -92,7 +91,7 @@ def restarts_post
       next unless node.key? :crowbar_wall
       next unless node[:crowbar_wall].key? :requires_restart
       next unless node[:crowbar_wall][:requires_restart].key? cookbook
-      next unless node[:crowbar_wall][:requires_restart][cookbook].key? service
+      next unless service.nil? || node[:crowbar_wall][:requires_restart][cookbook].key? service
```
Itxaka: `next unless service.nil?` should be on its own line, probably at the start of the loop to avoid doing any checks

Itxaka:
PR message and commit is confusing. What this endpoint does is to clear the flags on which services need a manual restart (i.e. they are already in the disallow_restart list and chef tried to restart them, but instead skipped the restart and set the flag that this needs a manual restart), not allowing restarts or not.

crowbar_framework/app/controllers/api/restart_management_controller.rb
ilausuch: Thank you

ilausuch: Thank you, but I think the reason to be there is to avoid the check of the service name to reset all of them. However I am testing.

ilausuch:
Thank you for your comments

ilausuch:
@nicolasbock yes, you are right the PR was closed. I was a reference for me. I have write the correct commit. Thanks @nicolasbock @skazi0

crowbar_framework/app/controllers/api/restart_management_controller.rb
```diff
@@ -130,6 +138,7 @@ def configuration_post
 
   def get_node_or_raise(node_name)
     node = NodeObject.find("name:#{node_name}").first
+    node = NodeObject.find_node_by_alias(node_name) if node.nil?
```
vuntz: You should use `NodeObject.find_node_by_name_or_alias`

crowbar_framework/app/controllers/api/restart_management_controller.rb
```diff
@@ -92,8 +91,17 @@ def restarts_post
...
       next unless node[:crowbar_wall].key? :requires_restart
       next unless node[:crowbar_wall][:requires_restart].key? cookbook
-      next unless node[:crowbar_wall][:requires_restart][cookbook].key? service
-      node[:crowbar_wall][:requires_restart][cookbook].delete(service)
+
+      # clear all services or, clear all services of a cookbook
+      if service.nil? || cookbook == service
+        node[:crowbar_wall][:requires_restart][cookbook].each_key do |the_service|
+          node[:crowbar_wall][:requires_restart][cookbook].delete(the_service)
+        end
```
vuntz: Can't we just do something like  `node[:crowbar_wall][:requires_restart].delete(cookbook)`?

Or possibly the following which rewrites the logic in the method (not necessarily better since it's harder to read, actually):

```ruby
if service.nil?
   node[:crowbar_wall].delete(:requires_restart)
else
  managed_cookbooks.each do |cookbook|
    [...]
    if cookbook == service
      node[:crowbar_wall][:requires_restart].delete(cookbook)
    elsif node[:crowbar_wall][:requires_restart][cookbook].key? service
      node[:crowbar_wall][:requires_restart][cookbook].delete(service)
    end
  [...]
`

File: crowbar_framework/app/controllers/api/restart_management_controller.rb (hunk 2 above)
Comment (ilausuch) [reply crowbar_framework/app/controllers/api/restart_management_controller.rb]: It was my first attempt, but It fails in some travis tests but not in my command line tests. So I change for find_node_by_alias and it works in travis tests, and in my command line tests. If I change it I should or modify find_node_by_name_or_alias to accept or this code to accept exceptions. However I've already tested it without good results. Let me know if knowing this information I have to change it anyway, plese

File: crowbar_framework/app/controllers/api/restart_management_controller.rb (hunk 3 above)
Comment (ilausuch) [reply crowbar_framework/app/controllers/api/restart_management_controller.rb]: I think it's a good optimization, let's change it

Comment (ilausuch) [reply crowbar_framework/app/controllers/api/restart_management_controller.rb]: This part could be outside of the loop
 
next unless node.key? :crowbar_wall
next unless node[:crowbar_wall].key? :requires_restart

File: crowbar_framework/app/controllers/api/restart_management_controller.rb (hunk 2 above)
Comment (vuntz) [reply crowbar_framework/app/controllers/api/restart_management_controller.rb]: I'd still do the change and try to figure out why travis fails. Else, it's a pattern that people will repeat and it's not what we want :-)

Comment (ilausuch) [reply crowbar_framework/app/controllers/api/restart_management_controller.rb]: Ok, thankyou, WIP

Comment (Itxaka) [reply crowbar_framework/app/controllers/api/restart_management_controller.rb]: @ilausuch the reason that this may fail to pass the tests may be because the `NodeObject.find_node_by_alias(node_name)` may not be mocked properly, so its expected to fail or not return the expected status.

Hit me up for a joint session on the testing stuff if you need some help to dig on that further :)

P.S.: Also there should be no need for 2 different node searches there, Im guessing you wanted to remove the first one.

**Comment by vuntz** [comment]
> From my point of view service should be left alone and moved to optional and a new param for cookbook should be introduced which would clean up the restart flags on a cookbook level. Optional as well, so if both of those are omitted, the whole node should be cleaned up of those flags.

Works for me. Not what is implemented yet, though ;-)

File: crowbar_framework/app/controllers/api/restart_management_controller.rb [hunk 4]
```diff
@@ -81,19 +81,29 @@ def restarts_get
...
 
     node = get_node_or_raise(node_name)
-
-    managed_cookbooks.each do |cookbook|
-      next unless node.key? :crowbar_wall
-      next unless node[:crowbar_wall].key? :requires_restart
-      next unless node[:crowbar_wall][:requires_restart].key? cookbook
-      next unless node[:crowbar_wall][:requires_restart][cookbook].key? service
-      node[:crowbar_wall][:requires_restart][cookbook].delete(service)
+    if node.key?(:crowbar_wall) && node[:crowbar_wall].key?(:requires_restart)
```
Itxaka: This `if` can be skipped if you use `next` as before
```
next unless node.key? :crowbar_wall
next unless node[:crowbar_wall].key? :requires_restart
```

Itxaka: was looking at the other code :dancer:

crowbar_framework/app/controllers/api/restart_management_controller.rb
```diff
@@ -37,7 +37,7 @@ def configuration
   api :POST, "/api/restart_management/restarts", "Clean the service restart flags"
   header "Accept", "application/vnd.crowbar.v2.0+json", required: true
   param :node, String, desc: "Node name", required: true
-  param :service, String, desc: "Service to clean restart flag for", required: true
+  param :service, String, desc: "Service to clean restart flag for", required: false
   api_version "2.0"
```
Itxaka: missing the new cookbook param

ilausuch:
@Itxaka fixed

crowbar_framework/app/controllers/api/restart_management_controller.rb
```diff
@@ -37,7 +37,8 @@ def configuration
   api :POST, "/api/restart_management/restarts", "Clean the service restart flags"
   header "Accept", "application/vnd.crowbar.v2.0+json", required: true
   param :node, String, desc: "Node name", required: true
-  param :service, String, desc: "Service to clean restart flag for", required: true
+  param :cookbook, String, desc: "Cookbook to clean restart flags for", required: false
```
AbelNavarro: This change may break backwards compatibility since you changed the API. Before it was iterating over all cookbooks, now you have to pass one. Are we ok to introduce this change now?

vuntz: I think it's fine as the feature is clearly marked as experimental

crowbar_framework/app/controllers/api/restart_management_controller.rb
```diff
@@ -81,19 +82,35 @@ def restarts_get
...
     node = get_node_or_raise(node_name)
 
-    managed_cookbooks.each do |cookbook|
-      next unless node.key? :crowbar_wall
-      next unless node[:crowbar_wall].key? :requires_restart
-      next unless node[:crowbar_wall][:requires_restart].key? cookbook
-      next unless node[:crowbar_wall][:requires_restart][cookbook].key? service
-      node[:crowbar_wall][:requires_restart][cookbook].delete(service)
+    raise Crowbar::Error::NotFound unless
+      node.key?(:crowbar_wall) && node[:crowbar_wall].key?(:requires_restart)
```
vuntz: Why do we raise an error here? I would just pretend it all went fine.

crowbar_framework/app/controllers/api/restart_management_controller.rb
```diff
@@ -81,19 +82,35 @@ def restarts_get
...
+
+    if cookbook.nil?
+      node[:crowbar_wall].delete(:requires_restart)
+      Rails.logger.info("restart_management : node #{node_name} cleaned")
```
vuntz: No space before ":" (same comment applies to the other log lines)

crowbar_framework/app/controllers/api/restart_management_controller.rb
```diff
@@ -81,19 +82,35 @@ def restarts_get
...
+    else
+      raise Crowbar::Error::NotFound unless managed_cookbooks.include? cookbook
+      raise Crowbar::Error::NotFound unless node[:crowbar_wall][:requires_restart].key? cookbook
```
vuntz: Same here: do we really want to raise an error (for the second line, the first one is fine).

crowbar_framework/app/controllers/api/restart_management_controller.rb
```diff
@@ -81,19 +82,35 @@ def restarts_get
...
+
+      if service.nil?
+        node[:crowbar_wall][:requires_restart].delete(cookbook)
+        Rails.logger.info("restart_management : cookbook #{cookbook} flags cleaned in node "\
+          "#{node_name}")
+      else
+        raise Crowbar::Error::NotFound unless
+          node[:crowbar_wall][:requires_restart][cookbook].include? service
```
vuntz: (and here again)

crowbar_framework/app/controllers/api/restart_management_controller.rb
ilausuch: I thought that if users make a mistake, if it doesn't return an error they could think that the operation has executed correctly.

However, WIP

crowbar_framework/app/controllers/api/restart_management_controller.rb
ilausuch: Ok, thanks for the advice

crowbar_framework/app/controllers/api/restart_management_controller.rb
```diff
@@ -81,19 +82,28 @@ def restarts_get
...
 
-    managed_cookbooks.each do |cookbook|
-      next unless node.key? :crowbar_wall
-      next unless node[:crowbar_wall].key? :requires_restart
-      next unless node[:crowbar_wall][:requires_restart].key? cookbook
-      next unless node[:crowbar_wall][:requires_restart][cookbook].key? service
-      node[:crowbar_wall][:requires_restart][cookbook].delete(service)
+    if cookbook.nil?
+      node[:crowbar_wall].delete(:requires_restart)
+      Rails.logger.info("restart_management : node #{node_name} cleaned")
```
vuntz: Still a space before `:` here :-)

crowbar_framework/app/controllers/api/restart_management_controller.rb
```diff
@@ -81,19 +82,28 @@ def restarts_get
...
+    else
+      raise Crowbar::Error::NotFound unless managed_cookbooks.include? cookbook
+
+      if service.nil?
+        node[:crowbar_wall][:requires_restart].delete(cookbook)
+        Rails.logger.info("restart_management: cookbook #{cookbook} flags cleaned in node "\
```
vuntz: s/in node/on node/

crowbar_framework/app/controllers/api/restart_management_controller.rb
```diff
@@ -81,19 +82,28 @@ def restarts_get
...
+          "#{node_name}")
+      else
+        node[:crowbar_wall][:requires_restart][cookbook].delete(service).nil?
+        Rails.logger.info("restart_management: service #{service} of cookbook #{cookbook} "\
+          "flags cleaned in node #{node_name}")
```
vuntz: s/in node/on node/

crowbar_framework/app/controllers/api/restart_management_controller.rb
```diff
@@ -81,19 +82,28 @@ def restarts_get
...
+      Rails.logger.info("restart_management : node #{node_name} cleaned")
+    else
+      raise Crowbar::Error::NotFound unless managed_cookbooks.include? cookbook
+
+      if service.nil?
+        node[:crowbar_wall][:requires_restart].delete(cookbook)
+        Rails.logger.info("restart_management: cookbook #{cookbook} flags cleaned in node "\
+          "#{node_name}")
+      else
+        node[:crowbar_wall][:requires_restart][cookbook].delete(service).nil?
```
vuntz: No need for `nil?`, I think.

vuntz: Also I believe you need to check if `node[:crowbar_wall][:requires_restart][cookbook]` exists, which may be why you had the `nil?` bit (for the test that got dropped)

crowbar_framework/app/controllers/api/restart_management_controller.rb
Itxaka: done

crowbar_framework/app/controllers/api/restart_management_controller.rb
Itxaka: done

crowbar_framework/app/controllers/api/restart_management_controller.rb
```diff
@@ -81,22 +82,39 @@ def restarts_get
...
+    else
+      raise Crowbar::Error::NotFound unless managed_cookbooks.include? cookbook
+
+      if service.nil?
+        node[:crowbar_wall][:requires_restart].delete(cookbook)
+        Rails.logger.info("restart_management: cookbook #{cookbook} flags cleaned on node "\
+          "#{node_name}")
+        dirty = true
+      else
+        if node[:crowbar_wall][:requires_restart].key? cookbook
```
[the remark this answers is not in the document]
Itxaka: ugh, no.

PR #164, opened 2015-12-10
provisioner: Do not require rsync of Leap install media

vuntz:
(ignore the first commit, it's from https://github.com/crowbar/crowbar-core/pull/159)

If /srv/tftpboot/opensuse-42.1/x86_64/install/ exists but does not
contain the required files, we will download the minimal bits that are
needed to allow Leap installation.

It's very debatable whether this should be done in the chef cookbook or
elsewhere, and the "security check" is quite weak (just a md5sum
check)...

PR #1511, opened 2018-03-08
[4.0] Crowbar services restart all services and restart by category backport

ilausuch:
This commit extends the functionality of #1307 and #164. These commits allow to the final user to disable the restart of services when the chef client runs. If it is active for any cookbook and one of its services need to clean the flag is raised. To clean this flag there is the command "clear_restart" that the user will run when after the manual restart. In this command the service name have to be specified. However is interesting to restart all services of the one cookbook or restart all services in a node.

This commit allows to specify the name of the cookbook, cleaning all the service of this cookbook, also allows to don't specify any cookbook that will clean all flags in all cookbooks.

Backport from #1509


ilausuch:
[disallow_restart feature: allow clearing whole node](https://trello.com/c/LX8YZEDF/270-disallowrestart-feature-allow-clearing-whole-node)

PR #1499, opened 2018-02-01
provisioner: replace curl calls to crowbar API with crowbarctl

jsuchome:
This replacement will fix some issues in relation with curl,
https, redirections and BasicAuth passwords."

**Why is this change necessary?**

This is part of ongoing work to use crowbarctl instead of direct curl calls to API. crowbarctl should know everything and no special options that we're building for curl command are necessary..

**Trello card**

https://trello.com/c/2SwQTBg1/399-p3-replace-curl-commands-in-crowbar-with-crowbarctl

**Requires**

sleshammer changes:
https://build.suse.de/package/show/home:jsuchome:branches:Devel:Cloud:8/sleshammer# -> https://build.suse.de/request/show/154768

PR #1513, opened 2018-03-09
[4.0] provisioner: replace curl calls to crowbar API with crowbarctl 

jsuchome:
Backport of https://github.com/crowbar/crowbar-core/pull/1499

This replacement will fix some issues in relation with curl,
https, redirections and BasicAuth passwords."

Why is this change necessary?

This is part of ongoing work to use crowbarctl instead of direct curl calls to API. crowbarctl should know everything and no special options that we're building for curl command are necessary..

Trello card

https://trello.com/c/2SwQTBg1/399-p3-replace-curl-commands-in-crowbar-with-crowbarctl

requires sleshammer update https://build.suse.de/request/show/157590

### DIFF
--- a/bin/gather_logs.sh
+++ b/bin/gather_logs.sh
@@ -18,6 +18,7 @@
 
 if [[ -f /etc/crowbar.install.key ]]; then
     export CROWBAR_KEY=$(cat /etc/crowbar.install.key)
+    export CROWBAR_PASS="$(sed -e 's/^machine-install://' <<< $CROWBAR_KEY)"
 fi
 mkdir -p /tmp/crowbar-logs
 tarname="${1-$(date '+%Y%m%d-%H%M%S')}"
@@ -47,11 +48,10 @@ sort_by_last() {
 	-o 'UserKnownHostsFile /dev/null')
     logs=(/var/log /etc)
     logs+=(/var/chef/cache /var/cache/chef /opt/dell/crowbar_framework/db)
-    curlargs=(-o /dev/null -D - --connect-timeout 30 --max-time 120 --insecure \
-    --location)
-    [[ $CROWBAR_KEY ]] && curlargs+=(--digest -u "$CROWBAR_KEY")
-    for to_get in nodes proposals roles; do
-	curl "${curlargs[@]}" "http://localhost/$to_get" || :
+    crowbarctl node list -U machine-install -P $CROWBAR_PASS
+
+    for to_get in proposals roles; do
+        crowbarctl $to_get proposal list crowbar
     done
     for node in $(sudo -H knife node list); do
 	tarfile="${node%%.*}-${tarname}.tar.gz"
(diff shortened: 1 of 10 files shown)


bin/gather_logs.sh
```diff
@@ -47,11 +48,10 @@ sort_by_last() {
...
     logs+=(/var/chef/cache /var/cache/chef /opt/dell/crowbar_framework/db)
-    curlargs=(-o /dev/null -D - --connect-timeout 30 --max-time 120 --insecure \
-    --location)
-    [[ $CROWBAR_KEY ]] && curlargs+=(--digest -u "$CROWBAR_KEY")
-    for to_get in nodes proposals roles; do
-	curl "${curlargs[@]}" "http://localhost/$to_get" || :
+    crowbarctl node list -U machine-install -P $CROWBAR_PASS
+
+    for to_get in proposals roles; do
+        crowbarctl $to_get proposal list crowbar
```
vuntz: Why do we need `-U machine-install -P $CROWBAR_PASS` a few lines earlier and not here? Isn't it simply missing?

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
```diff
@@ -64,25 +64,28 @@
           key_re='crowbar\.install\.key=([^ ]+)'
           if [[ $(cat /proc/cmdline) =~ $key_re ]]; then
               export CROWBAR_KEY="${BASH_REMATCH[1]}"
-              echo "$CROWBAR_KEY" >/mnt/etc/crowbar.install.key
-          elif [[ -f /mnt/etc/crowbar.install.key ]]; then
-              export CROWBAR_KEY="$(cat /mnt/etc/crowbar.install.key)"
+              echo "$CROWBAR_KEY" >/etc/crowbar.install.key
```
vuntz: Do we actually still need to create that file?

bin/gather_logs.sh
jsuchome: Ah, I guess I missed this

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
jsuchome: Good point, I was also wondering. But in the end I did not want to interfere in it

jsuchome:
> Generally speaking, there are various places where we add crowbarctl calls without --no-verify-ssl 
> while curl had --insecure, so this will definitely fail for self-signed certificates now.

True. It should work correctly for the cases where we write no-verify-ssl into crowbarrc file, but that's not always the case.

chef/cookbooks/provisioner/templates/default/autoyast-upgrade.xml.erb
vuntz: Agreed, this doesn't need to be handled here.

jsuchome:
[P3 Replace curl commands in crowbar with crowbarctl](https://trello.com/c/2SwQTBg1/399-p3-replace-curl-commands-in-crowbar-with-crowbarctl)

PR #1515, opened 2018-03-15
Allow removable block devices as physical drives

JanZerebecki:
This is determined from /sys/block/xxx/removable and my be set for SATA
disks.

(bsc#1085170)


chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -246,7 +246,7 @@ def held_by_multipath?
         def fixed
           # This needs to be kept in sync with the number_of_drives method in
           # node_object.rb in the Crowbar framework.
-          @device =~ /^([hsv]d|dasd|cciss|xvd|nvme)/ && !removable && !cinder_volume
+          @device =~ /^([hsv]d|dasd|cciss|xvd|nvme)/ && !cinder_volume
```
jsuchome: Do we really want this? Maybe it helps with the bug, but won't it bite us elsewhere?

JanZerebecki: I feel uneasy about it, too. Perhaps we can check for this driver and zero the removable flag for them.

How will it bite us?

jsuchome: I'm not sure it will, just saying that there was an intent in ignoring removable devices and we should investigate it.

JanZerebecki:
Similar problem: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=404927

JanZerebecki:
The code that is changed here was introduced in https://github.com/crowbar/crowbar-core/commit/ca76a723ffa2d617383ee05ab9c599dde1e1862e .

chef/cookbooks/barclamp/libraries/barclamp_library.rb
JanZerebecki: I think the main confusion would be with writeable media and USB sticks, though it is probably rare that those have unpartitioned space.

JanZerebecki:
This version only corrects the removable flag from controllers with the aacraid driver. Someone could still connect something removable to such an controller, however the chance is smaller.

chef/cookbooks/ohai/files/default/plugins/linux/block_device_extended.rb
```diff
@@ -25,6 +25,21 @@
         File.open("/sys/block/#{dir}/queue/#{check}") { |f| block[dir][check] = f.read_nonblock(1024).strip }
       end
     end
+
+    # correct that the aacraid driver always sets the removable flag
+    if block[dir]["removable"] == "1"
+      path = File.readlink("/sys/block/#{dir}").split("/")
```
jsuchome: Can't you use `block_device_dir` value here?

chef/cookbooks/ohai/files/default/plugins/linux/block_device_extended.rb
```diff
@@ -25,6 +25,21 @@
...
+      # example:
+      #   ../devices/pci0000:00/0000:00:02.0/0000:03:00.0/host0/target0:1:4/0:1:4:0/block/sdb
+      path[0] = "/sys"
```
jsuchome: I think with `File.realpath` you could avoid this trick of turning relative path to absolute

chef/cookbooks/ohai/files/default/plugins/linux/block_device_extended.rb
JanZerebecki: I could, but I wanted to keep within the style of the rest.

chef/cookbooks/ohai/files/default/plugins/linux/block_device_extended.rb
JanZerebecki: Done.

chef/cookbooks/ohai/files/default/plugins/linux/block_device_extended.rb
```diff
@@ -25,6 +25,22 @@
         File.open("/sys/block/#{dir}/queue/#{check}") { |f| block[dir][check] = f.read_nonblock(1024).strip }
       end
     end
+
+    # correct that the aacraid driver always sets the removable flag
+    if block[dir]["removable"] == "1"
+      path = File.readlink("/sys/block/#{dir}")
+      path = File.realpath(path, "/sys/block/").split("/")
```
jsuchome: Do you really need that `readlink` here? I think `path = File.realpath("/sys/block/#{dir}").split("/")` should work...

chef/cookbooks/ohai/files/default/plugins/linux/block_device_extended.rb
```diff
@@ -25,6 +25,22 @@
...
+      # example:
+      # /sys/devices/pci0000:00/0000:00:02.0/0000:03:00.0/host0/target0:1:4/0:1:4:0/block/sdb
+      path = path[0..4] + ["driver"]
+      path = path.join("/")
```
jsuchome: I do not really like you're using same variable for string and array...

JanZerebecki:
Yes, I incorporated both suggestions.

PR #1516, opened 2018-03-16
ohai: Do not fail when tcpdump output is missing

JanZerebecki:
ohai: Do not fail when tcpdump output is missing

As we might not have run tcpdump, we need to handle that its output file
might be missing.

ohai: show which interface an error happened on 

Follow up of: https://github.com/crowbar/crowbar-core/pull/1385

### DIFF
--- a/chef/cookbooks/ohai/files/default/plugins/crowbar.rb
+++ b/chef/cookbooks/ohai/files/default/plugins/crowbar.rb
@@ -242,7 +242,13 @@ networks.each do |network|
   sw_port = -1
   sw_port_name = nil
 
-  line = IO.readlines(tcpdump_out).grep(/Subtype Interface Name/).join ""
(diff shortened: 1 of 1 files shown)


jsuchome:
Why could tcpdump be missing? Can't we make sure it's there?

JanZerebecki:
The ouput would be missing when we failed to observe the interface having link and thus don't run tcpdump.

jsuchome:
@JanZerebecki I was asking why can't we make sure that ohai is available

JanZerebecki:
Related cases:
https://trello.com/c/6tq0Y7QD https://bugzilla.suse.com/show_bug.cgi?id=1084673
https://trello.com/c/mtY3MKOn https://bugzilla.suse.com/show_bug.cgi?id=1081187
https://trello.com/c/K0Pa9ZM8 https://bugzilla.suse.com/show_bug.cgi?id=1084319

PR #1520, opened 2018-03-27
network: Only set bond attributes when they change (bsc#1054268)

s-t-e-v-e-n-k:
One side effect of the change to support xmit_hash_policy means the
network recipe writes the miimon and xmit_hash_policy attributes to
sysfs every time chef runs, which causes the kernel to log that. Only
set the attributes if they are different from the current setting.

requests?**


susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
588585ceb306d997472ac1f867d47fd79bb2c582</td>
    </tr>
  </tbody>
</table>

</p>

PR #1524, opened 2018-03-29
Disable Danger subject_length check

cmurphy:
We typically include both the module component and a bug reference in
the subject of commit messages, which eats up space. Disable this check
even though 50 characters is standard.


cmurphy:
Closing in favor of https://github.com/crowbar/crowbar-core/pull/1525

PR #1528, opened 2018-04-04
[3.0] crowbar: use hostname file to validate host name

Itxaka:
Instead of using "hostname -f" to validate the host name,
just use the /etc/hostname to do so, as that is the source
for the hostname inside the backup, the comparison should
be done against the same file

(cherry picked from commit b9d8c2ea187b5ed1aa30ef2980277c3d1fa5d53f)

Backport-of: https://github.com/crowbar/crowbar-core/pull/1526


vuntz:
No need to block on ha gate for this one.

PR #1530, opened 2018-04-04
service: fix failure on transition to readying

dirkmueller:
When the salt barclamp is used, a transition to readying is failing
with this error:

  salt transition to default failed.
  wrong number of arguments (3 for 0)
  /opt/dell/crowbar_framework/app/models/service_object.rb:295:in `transition'
  /opt/dell/crowbar_framework/app/models/crowbar_service.rb:202:in `block (2 levels) in transition'
  /opt/dell/crowbar_framework/app/models/crowbar_service.rb:186:in `each'
  /opt/dell/crowbar_framework/app/models/crowbar_service.rb:186:in `block in transition'
  /opt/dell/crowbar_framework/app/models/crowbar_service.rb:185:in `each'
  /opt/dell/crowbar_framework/app/models/crowbar_service.rb:185:in `transition'
  /opt/dell/crowbar_framework/app/controllers/barclamp_controller.rb:126:in `transition'

This is due to incorrect number of arguments being specified on the
base service_object

requests?**

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -292,7 +292,7 @@ class ServiceObject
     [200, { versions: ["1.0"] }]
   end
 
-  def transition
+  def transition(inst, name, state)
     [200, {}]
   end
 
--- a/chef/data_bags/crowbar/template-salt.json
+++ b/chef/data_bags/crowbar/template-salt.json
@@ -23,10 +23,8 @@
       "config": {
         "environment": "salt-base-config",
         "mode": "full",
-        "transitions": true,
+        "transitions": false,
         "transition_list": [
-          "installed",
-          "readying"
         ]
       }
     }


crowbar_framework/app/models/service_object.rb
```diff
@@ -292,7 +292,8 @@ def versions
     [200, { versions: ["1.0"] }]
   end
 
-  def transition
+  def transition(inst, name, state)
+    Rails.logger.debug("transition: enter #{inst} #{name} #{state}")
```
vuntz: Please don't add just a "enter" log line, as we'll be looking for the "leave" log line. Can you either drop the verb, or make it "enter and leave"?

vuntz:
Btw, while this is indeed a bug, this is also a bug in the salt barclamp because I pretty much doubt the transition method should be called for it. That means it has some wrong metadata...

dirkmueller:
@nicolasbock the salt barclamp seems to be the first one not reimplementing this function

nicolasbock:
@dirkmueller That's sneaky :smile:

PR #1531, opened 2018-04-06
upgrade: use proper product name for cloud 8

Itxaka:
As the iso name has changed from cloud7[0] to cloud8[1] the product
name that zypper return has also changed, so we should make sure
to check the proper product name

[0] SUSE-OPENSTACK-CLOUD-7-x86_64-Media1.iso
[1] SUSE-OPENSTACK-CLOUD-CROWBAR-8-x86_64-Media1.iso

Should fix error `mkcloud error code 11: Some repository is missing on admin server. Cannot upgrade.` on the `cloud-mkcloud8-job-upgrade-disruptive-x86_64` job

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -281,7 +281,7 @@ module Api
             }
           end
 
-          cloud_available = repo_version_available?(products, "suse-openstack-cloud", "8")
+          cloud_available = repo_version_available?(products, "suse-openstack-cloud-crowbar", "8")
           ret[:openstack] = {
             available: cloud_available,
             repos: [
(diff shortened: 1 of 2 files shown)


jsuchome:
@Itxaka when did the change happen, do you have some link to history?

Itxaka:
@jsuchome this is probably due to the different ISOs now provided of SUSE Cloud as we now have:

* HPE-HELION-OPENSTACK-CLOUD-8-x86_64-Media1.iso -> Ardana
* SUSE-OPENSTACK-CLOUD-CROWBAR-8-x86_64-Media1.iso -> Crowbar
* SUSE-OPENSTACK-CLOUD-8-x86_64-Media1.iso -> Both?

So I guess that to avoid extra bandwidth when downloading the isos, the automation will only download the crowbar one, which makes the product name different.

Which makes me wonder if we should instead regex the product name as the end user could be using any 2 out of those 3 isos to upgrade, which would make this change useless.

Itxaka:
@jsuchome the change on the automation repo was done here: https://github.com/SUSE-Cloud/automation/pull/2383

PR #1534, opened 2018-04-10
upgrade: Remove pre-upgrade constraints from existing locations

jsuchome:
pre-upgrade constraints ensure that the service is not running at the
node with active pre-upgrade attribute. This is necessary during the upgrade
when we need to control which nodes can run services and which cannot.

So we must do a cleanup before we start the upgrade, otherwise all services would
stop at all nodes once we set pre-upgrade attribute to all nodes (which is the
initial step of the upgrade).

requests?**


jsuchome:
[Cleanup pacemaker location constraints related to upgrade](https://trello.com/c/fdPhVgG4/238-cleanup-pacemaker-location-constraints-related-to-upgrade)

jsuchome:
@vuntz It's true it would not hurt, but maybe it would be better to actually remove them _at the end of upgrade_ so next upgrade path (8-9) would not need to care. I'm not sure which way to go.

PR #1535, opened 2018-04-10
Allow sanity controller during upgrade

Itxaka:
As during the crowbar upgrade database step, a sanity check is
called, we need to skip the upgrade filter on the sanity
controller so it can be called even if we are in the middle of the
upgrade, otherwise we will be stuck in the database step

The sanity check was introduced in crowbar-init here: https://github.com/crowbar/crowbar-init/commit/29de8a8e2eeaf135ec028df05e6c285d3965be54


skazi0:
@Itxaka is this some new behavior for 7->8 upgrade? IIRC it was not required for 6->7 upgrade

skazi0:
@Itxaka ah, so this problem was introduced by my fix for database command failing because of failing sanity checks...

PR #1536, opened 2018-04-10
Add common cinder group for all nodes (bsc#1087896)

bhdn:
So that new nodes don't get different uids and clash when NFS is used.

Also related to bsc#968251.

https://bugzilla.suse.com/show_bug.cgi?id=1087896


bhdn:
[P5 Bug 1087896: inconsistent cinder uid/gid across nodes (unsa.org)](https://trello.com/c/o2T7Sh0o/920-p5-bug-1087896-inconsistent-cinder-uid-gid-across-nodes-unsaorg)

bhdn:
So I just realized that this patch alone will break existing deployments if released as a MU. As the newly chosen UID is in a different range than those currently use, breakage is guaranteed.

PR #1537, opened 2018-04-10
ohai: Correct aacraid drivers disk removable flag (bsc#1085170)

JanZerebecki:
This is determined from /sys/block/xxx/removable and is always set for
aacraid. It is useful to distinguish removable and hotswappable.
Removable should mean humans will remove this during normal operation
like an USB stick, where as hotswappable disks are only removed for servicing.

https://github.com/torvalds/linux/blob/7b1cd95d65eb3b1e13f8a90eb757e0ea232c7899/drivers/scsi/aacraid/aachba.c#L3117
gives its reason as: "Do not cache partition table for arrays". I
suspect that there is a better way for a kernel driver to behave, since
this is not the first controller with hotpluggable disks.

(cherry picked from commit a7312a8274312a0b1b0621e2578cf2290748b61c)
Backport of https://github.com/crowbar/crowbar-core/pull/1515


JanZerebecki:
Added a commit to make the Hound happy.

PR #1539, opened 2018-04-11
crowbar: test for find_node_by_name_or_alias

ilausuch:
Testing of find_node_by_name_or_alias method of node

### DIFF
--- a/crowbar_framework/spec/models/node_object_spec.rb
+++ b/crowbar_framework/spec/models/node_object_spec.rb
@@ -67,6 +67,21 @@ describe Node do
         expect(node.alias).to be == "testing"
       end
     end
+
+    describe "find_node_by_name_or_alias", find_node_by_name_or_alias: true do
+      it "returns nodes matching node (first part of method)" do
+        node = Node.find_node_by_name_or_alias("testing.crowbar.com")
+        expect(node).to_not be_nil
+        expect(node.alias).to be == "testing"
+      end
+
+      it "returns nodes matching alias (second part of method)" do
+        testing_node = Node.find_node_by_name("testing.crowbar.com")
+        allow(Node).to receive(:find_node_by_alias).with("testing_alias").and_return(testing_node)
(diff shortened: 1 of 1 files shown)


crowbar_framework/spec/models/node_object_spec.rb
```diff
@@ -67,6 +67,21 @@
...
+    describe "find_node_by_name_or_alias", find_node_by_name_or_alias: true do
+      it "returns nodes matching node (first part of method)" do
+        node = Node.find_node_by_name_or_alias("testing.crowbar.com")
+        expect(node).to_not be_nil
+        expect(node.alias).to be == "testing"
+      end
+
+      it "returns nodes matching alias (second part of method)" do
+        testing_node = Node.find_node_by_name_or_alias("testing")
+        allow(Node).to receive(:find_node_by_alias).with("testing2").and_return(testing_node)
```
jsuchome: Wouldn't it be better to add the alias to the fixture so we could actually test that the method works?

ilausuch: I tested it locally and didn't work but I'll test again directly on travis

ilausuch: Folowing your indications I added the alias directly over the fixture of testing inc crowbar/display section such as in node_ceph.crowbar.com.json but it didn't work. Alias always is "testing". In fact I have tested to change in ceph node, but It doesn't change. Then I checked that in the offline_couchdb/nodes_all.json was too. In case, I tested it too, but It didn't have any effect neither. I couldn't found the problem. So I keep the original version. Please share with me if you have a new clue to solve it. However, I think it won't be necessary because I am testing only the second part of the condition. This is the reason to force to first part to fail. Additionally find_node_by_alias is already tested.

toabctl:
@ilausuch can you stop merging into master please? we are next to the release and don't want to get $random changes into the GMC/GM...

PR #1542, opened 2018-04-12
[SCRD-3019] upgrade: re-enable openvswitch unit after package upgrade

Itxaka:
When the openvswitch package is upgrade the old systemd unit file is
deleted and a new one is installed, which makes the systemd service
lost the enable/disable value and use the default one (disabled)

So on 7->8 package upgrade we need to manually enable the sytemd
unit after the package upgrade.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.sh.erb
@@ -79,6 +79,13 @@ initiate_node_upgrade()
     # Upgrade the distribution non-interactively
     log "Executing zypper ref"
     zypper --no-color --releasever <%= @target_platform_version %> ref -f
+
+    # check to see if openvswitch is gonna be updated
+    is_openvswitch_updated=`zypper --no-color --non-interactive list-updates|grep -q openvswitch`
+    # check if openvswitch is enabled on boot
+    [ -f /etc/systemd/system/multi-user.target.wants/openvswitch.service ]
+    was_openvswitch_enabled=$?
+
     log "Executing zypper dist-upgrade"
     zypper --no-color --non-interactive dist-upgrade -l --recommends --replacefiles
     ret=$?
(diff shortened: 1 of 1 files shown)


jsuchome:
- you have a `sytemd` typo in commit message
- you might also mention explicitly that if we do not do it, we lose network connection to the nodes

Itxaka:
fixed!

chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.sh.erb
```diff
@@ -138,6 +138,11 @@ initiate_node_upgrade()
     rm -f /etc/nova/nova.conf.d/200-crowbar-upgrade.conf
     rm -f /etc/neutron/neutron-openvswitch-agent.conf.d/200-crowbar-upgrade.conf
 
+    # mark openvswitch as enabled so it comes correctly after a node reboot
+    # this is due to the package upgrade deleting the systemd unit and adding a new unit for the
+    # service, which causes this new service unit to not be enabled.
+    systemctl enable openvswitch
```
jsuchome: Could you check for openvswitch presence?

Also, is it possible that openvswitch was present before but disabled for some reason?

skazi0: yeah, what about systems with no openvswitch?

Itxaka: See new patch, will only do this if it finds the openvswitch in the list of packages to update

Itxaka:
RE: openvswitch installed but disabled

I think in that case it has been a manual intervention to disable it because if we are using openvswitch then your network depends on it right? 

I could also check if it was enable before and only enable it if so.

toabctl:
@Itxaka please stop merging $random changes into master. we try to prepare a release....

Itxaka:
@toabctl sorry about that, wont happen again!

chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.sh.erb
```diff
@@ -79,6 +79,13 @@ initiate_node_upgrade()
     # Upgrade the distribution non-interactively
     log "Executing zypper ref"
     zypper --no-color --releasever <%= @target_platform_version %> ref -f
+
+    # check to see if openvswitch is gonna be updated
+    is_openvswitch_updated=`zypper --no-color --non-interactive list-updates|grep -q openvswitch`
```
matthewoliver: The -q switch here only makes it quiet. So the return value would be 0 on success.  So this will always put nothing (as it returns nothing the stdout) into the is_openvswitch_updated wouldn't it? which would mean the renabling triggered by the 'if' on line 148 would never be be triggered. At least in my understanding.

jsuchome: @matthewoliver this was followed with https://github.com/crowbar/crowbar-core/pull/1558 (which we believe might not be needed after all because we might have package already...)

PR #1543, opened 2018-04-13
Add common cinder group for all nodes (bsc#1087896)

bhdn:
So that new nodes don't get different uids and clash when NFS is used.

Also related to bsc#968251.

This is a forward port of https://github.com/crowbar/crowbar-core/pull/1536

https://bugzilla.suse.com/show_bug.cgi?id=1087896


jsuchome:
I agree this should go to SOC8 for normal installations ... but we can't have it for maintenance update as it could break existing deployments when new node is being added

jsuchome:
@rsalevsky I believe this should go to the release

toabctl:
Looks like chaning uids during a MU is not allowed. So including this into GMC (also talked to @dirkmueller about it)

PR #1549, opened 2018-04-17
[4.0] upgrade: Fixes for prepare step

jsuchome:
The bash code will fail if we try to disable something that is not present.


jsuchome:
tested with HA and non-HA setup

PR #1550, opened 2018-04-18
rabbitmq: active the haproxy balancing option

ilausuch:
Added experimental option to active the haproxy balance for rabbitmq.

Related with crowbar/crowbar-openstack#1571. The precense of this option in the experimental config file is not required but show how to activate this feature.

### DIFF
--- a/crowbar_framework/config/experimental.yml
+++ b/crowbar_framework/config/experimental.yml
@@ -23,6 +23,8 @@ default: &default
      - updater
   skip_unchanged_nodes:
     enabled: false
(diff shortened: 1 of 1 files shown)


ilausuch:
[RabbitMQ connections rebalancing](https://trello.com/c/bMDQwg0D/246-rabbitmq-connections-rebalancing)

ilausuch:
yes! sure, I was testing, by default must be false

ilausuch:
Won't be necessary because now is automatic depending on if the proposal is already applied

PR #1552, opened 2018-04-19
upgrade: Change rabbitmq settings during the upgrade

jsuchome:
Default rabbitmq HA setup is using native clustering. We need
to allow switch to this default from DRBD/shared storage setup
that is used in SOC7.

Requires https://github.com/crowbar/crowbar-openstack/pull/1637
Alternative to https://github.com/crowbar/crowbar-core/pull/1557

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -699,12 +699,36 @@ module Api
         Rails.logger.info("Successfully finished disruptive upgrade of controller nodes")
       end
 
+      # Rabbitmq uses native clustering in SOC8, we need to adapt the configuration
+      def update_rabbitmq_setup
+        rabbit_cluster_members = ::Node.find(
+          "run_list_map:pacemaker-cluster-member AND " \
+          "run_list_map:rabbitmq-server"
+        )
+        return if rabbit_cluster_members.empty?
+
+        erlang_cookie = ServiceObject.new.random_password
+        prop = Proposal.find_by(barclamp: "rabbitmq", name: "default")
+        unless prop.raw_data["attributes"]["rabbitmq"]["cluster"]
+          prop.raw_data["attributes"]["rabbitmq"]["cluster"] = true
+          prop.raw_data["attributes"]["rabbitmq"]["erlang_cookie"] = erlang_cookie
+          prop.save
+        end
+
+        rabbit_cluster_members.each do |node|
+          node["rabbitmq"]["cluster"] = true
+          node["rabbitmq"]["erlang_cookie"] = erlang_cookie
+          node.save
+        end
+      end
+
       def do_controllers_substep(substep)
         if substep == :ceph_nodes
           join_ceph_nodes
           substep = :controller_nodes
         end
         if substep == :controller_nodes
+          update_rabbitmq_setup
           upgrade_controller_clusters
           upgrade_non_compute_nodes
           prepare_all_compute_nodes
--- a/crowbar_framework/spec/models/api/upgrade_spec.rb
+++ b/crowbar_framework/spec/models/api/upgrade_spec.rb
@@ -465,6 +465,7 @@ describe Api::Upgrade do
     it "successfully upgrades controller nodes" do
       allow(Api::Upgrade).to receive(:remaining_nodes).and_return(1)
       allow(Api::Upgrade).to receive(:join_ceph_nodes).and_return(true)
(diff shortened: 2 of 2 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -965,7 +965,16 @@ def upgrade_cluster(founder, cluster)
...
+
         node_api = Api::Node.new node.name
+
+        if node["run_list_map"].key? "rabbitmq-server"
+          prop = Proposal.find_by(barclamp: "rabbitmq", name: "default")
+          prop.raw_data["attributes"]["rabbitmq"]["cluster"] = true
+          prop.save
+          node_api.update_rabbitmq_config true
+        end
+
```
jsuchome: This rabbitmq related block should be in different method

crowbar_framework/app/models/api/node.rb
```diff
@@ -177,6 +177,14 @@ def disable_pre_upgrade_attribute_for(name)
       end
     end
 
+    # change the value of rabbitmq clustering
+    # (default for SOC8 is true, but false for SOC7)
+    def update_rabbitmq_config(new_value)
```
Itxaka: This needs to be executed in all the cluster nodes that have rabbitmq as currently only is executed on the first one

jsuchome: done

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -961,10 +962,30 @@ def upgrade_cluster(founder, cluster)
...
+      # Rabbitmq uses native clustering in SOC8, we need to adapt the configuration
+      def upgrade_rabbitmq_setup
+        rabbit_cluster_members = ::Node.find(
+          "run_list_map:pacemaker-cluster-member AND run_list_map:rabbitmq-server "
+        )
+        return if rabbit_cluster_members.empty?
+
+        prop = Proposal.find_by(barclamp: "rabbitmq", name: "default")
+        unless prop.raw_data["attributes"]["rabbitmq"]["cluster"]
+          prop.raw_data["attributes"]["rabbitmq"]["cluster"] = true
```
Itxaka: You also want to save the `["rabbitmq"]["erlang_cookie"]` here so its in sync with the node one.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -961,10 +962,30 @@ def upgrade_cluster(founder, cluster)
         Rails.logger.info("Nodes in cluster #{cluster} successfully upgraded")
       end
 
+      # Rabbitmq uses native clustering in SOC8, we need to adapt the configuration
+      def upgrade_rabbitmq_setup
+        rabbit_cluster_members = ::Node.find(
+          "run_list_map:pacemaker-cluster-member AND run_list_map:rabbitmq-server "
```
Itxaka: extra space at the end?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -961,10 +962,30 @@ def upgrade_cluster(founder, cluster)
...
+        return if rabbit_cluster_members.empty?
+
+        prop = Proposal.find_by(barclamp: "rabbitmq", name: "default")
+        unless prop.raw_data["attributes"]["rabbitmq"]["cluster"]
+          prop.raw_data["attributes"]["rabbitmq"]["cluster"] = true
+          prop.save
+        end
+
+        rabbit_cluster_members.each do |node|
+          node_api = Api::Node.new node.name
```
Itxaka: Arent they already `Node` objects here? Do we really need to transform them into an `Api::Node` object to save 2 variables?

Im just saying, having the node attribute modification here instead that on the `Api::Node` part, makes it clearer what this is doing and we can add comments to make clear that we need the same attrs on the Proposal and on the Node.

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Do you think I should pass this one to the nodes?

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Yeah. there's extra search, which seems unnecessary

So you think I should move `update_rabbitmq_config` to `Node`, or are you fine with saving it directly from here?

jsuchome:
We might want to do this only for DRBD case

jsuchome:
Maybe better place for these calls would be under `::Openstack::Upgrade`, something like `unset_db_synced`

PR #1553, opened 2018-04-19
[poc] rabbitmq: upgrade

Itxaka:
This goes with crowbar-openstack change:
```
diff --git a/chef/cookbooks/rabbitmq/recipes/role_rabbitmq_server.rb b/chef/cookbooks/rabbitmq/recipes/role_rabbitmq_server.rb
index 5a63c1d15..44c09c385 100644
--- a/chef/cookbooks/rabbitmq/recipes/role_rabbitmq_server.rb
+++ b/chef/cookbooks/rabbitmq/recipes/role_rabbitmq_server.rb
@@ -18,3 +18,9 @@ if CrowbarRoleRecipe.node_state_valid_for_role?(node, "rabbitmq", "rabbitmq-serv
   include_recipe "rabbitmq::rabbit"
   include_recipe "rabbitmq::monitor"
 end
+
+# when we are upgrading and on the proper step, run the rabbit recipe so we recreate any missing
+# resources after an upgrade
+if node[:state] == "crowbar_upgrade" && node["crowbar_upgrade_step"] == "rabbit_upgrade"
+  include_recipe "rabbitmq::rabbit"
+end
```

The idea here is to have an extra step when upgrading the controllers:

 * upgrade *only* rabbitmq packages and resources
 * reuse the chef rabbit recipe to ensure that the node state is in sync with what we want

As this will go on each node, the rabbitmq availability should be ok, I guess? clients will be shut down by that moment, and the neutron services should connect to the new server as they will listen on the same IPs, so there *will* be some downtime on rabbitmq, but as its in the middle of the upgrade and expected (API downtime is expected on non-disruptive upgrade) it should be ok (should, should, but its not tested to be true)

### DIFF
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -164,6 +164,59 @@ module Api
       reboot_and_wait
     end
 
+    # only rabbit upgrade
+    def upgrade_rabbit
+      prepare_repositories
+      delete_rabbit_resources
+      upgrade_rabbit_packages
+      create_rabbit_resources
+    end
+
+    # removes all rabbit resources from pacemaker
+    def delete_rabbit_resources
+      action = "delete_rabbit_resources"
+      save_node_action(action)
+      execute_and_wait_for_finish("/usr/sbin/#{action}.sh", 120)
+      Rails.logger.info("#{action} was successful.")
+    rescue StandardError => e
+      Api::Upgrade.raise_node_upgrade_error(
+        "Error while executing #{action} script. " + e.message
+      )
+    end
+
+    def upgrade_rabbit_packages
+      action = "upgrade_rabbit_packages"
+      save_node_action(action)
+      execute_and_wait_for_finish("/usr/sbin/#{action}.sh", 120)
+      Rails.logger.info("#{action} was successful.")
+    rescue StandardError => e
+      Api::Upgrade.raise_node_upgrade_error(
+        "Error while executing #{action} script. " + e.message
+      )
+    end
+
+    # creates pacemaker rabbit resources
+    def create_rabbit_resources
+      # save and change the status so we dont run the wrong recipe
+      backup_state = @node.crowbar["node_upgrade_state"]
+      @node.crowbar["node_upgrade_state"] = "rabbit_upgrade"
+      @node["rabbitmq"]["cluster"] = true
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -831,6 +835,14 @@ def prepare_network_node(network_node)
         delete_pacemaker_resources network_node
       end
 
+      def upgrade_rabbitmq_clusters
```
jsuchome: +1 for such method, yes

crowbar_framework/app/models/api/node.rb
```diff
@@ -164,6 +164,59 @@ def upgrade
...
+    def create_rabbit_resources
+      # save and change the status so we dont run the wrong recipe
+      backup_state = @node.crowbar["node_upgrade_state"]
+      @node.crowbar["node_upgrade_state"] = "rabbit_upgrade"
+      @node["rabbitmq"]["cluster"] = true
+      @node["rabbitmq"]["erlang_cookie"] = ServiceObject.new.random_password
+      @node.save
+      action = "create_rabbit_resources"
+      save_node_action(action)
+      execute_and_wait_for_finish("chef-client", 120)
```
jsuchome: Why extra chef-client run, isn't the one from `join_and_chef` enough?

crowbar_framework/app/models/api/node.rb
```diff
@@ -164,6 +164,59 @@ def upgrade
...
+    # only rabbit upgrade
+    def upgrade_rabbit
+      prepare_repositories
+      delete_rabbit_resources
+      upgrade_rabbit_packages
+      create_rabbit_resources
+    end
+
+    # removes all rabbit resources from pacemaker
+    def delete_rabbit_resources
```
jsuchome: No need, they are deleted by `delete_pacemaker_resources`

jsuchome:
An alternative https://github.com/crowbar/crowbar-core/pull/1552

crowbar_framework/app/models/api/node.rb
jsuchome: (but we need to adapt the delete script to take care of drbd resources)

Itxaka:
https://github.com/crowbar/crowbar-core/pull/1552 is nicer

PR #1557, opened 2018-04-23
upgrade: Change rabbitmq settings during the upgrade

jsuchome:
Default rabbitmq HA setup in SOC8 is using native clustering.
Let's switch to this default when upgrading from the setup that is using
DRBD storage.

Requires  https://github.com/crowbar/crowbar-openstack/pull/1637
Alternative to https://github.com/crowbar/crowbar-core/pull/1552

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -327,6 +327,21 @@ class CrowbarService < ServiceObject
       node.save
     end
 
+    # change rabbitmq/drbd setting to native queues setup (new default)
+    prop = Proposal.find_by(barclamp: "rabbitmq", name: "default")
+    role = prop.role
+    if role.default_attributes["rabbitmq"]["ha"]["storage"]["mode"] == "drbd"
+      role.default_attributes["rabbitmq"]["cluster"] = true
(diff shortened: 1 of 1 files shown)


jsuchome:
@rhafer This one moves the nodes modification to the one place where we are already doing it (keystone bootstrap, db_sync unset), so it looks more consistent to me

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -327,6 +327,15 @@ def prepare_nodes_for_os_upgrade
...
     end
 
+    # change rabbitmq/drbd setting to native queues setup (new default)
+    prop = Proposal.find_by(barclamp: "rabbitmq", name: "default")
+    role = prop.role
+    if role.default_attributes["rabbitmq"]["ha"]["storage"]["mode"] == "drbd"
+      role.default_attributes["rabbitmq"]["cluster"] = true
+      role.default_attributes["rabbitmq"]["erlang_cookie"] = random_password
+      role.save
+    end
```
rhafer: I think you need to update both, the proposal and the role. Otherwise the change will get lost if the user clicks apply on the proposal again.

PR #1558, opened 2018-04-24
upgrade: fix check for openvswitch package

Itxaka:
Instead of checking if the package is gonna be upgraded during
the upgrade, just check if the package is installed. This avoids
issues if the package has been already upgraded during a different
step or even before the upgrade.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.sh.erb
@@ -80,12 +80,6 @@ initiate_node_upgrade()
     log "Executing zypper ref"
     zypper --no-color --releasever <%= @target_platform_version %> ref -f
 
-    # check to see if openvswitch is gonna be updated
-    is_openvswitch_updated=`zypper --no-color --non-interactive list-updates|grep -q openvswitch`
-    # check if openvswitch is enabled on boot
-    [ -f /etc/systemd/system/multi-user.target.wants/openvswitch.service ]
-    was_openvswitch_enabled=$?
-
     log "Executing zypper dist-upgrade"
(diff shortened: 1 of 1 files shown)


chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.sh.erb
```diff
@@ -145,7 +146,7 @@ initiate_node_upgrade()
     rm -f /etc/nova/nova.conf.d/200-crowbar-upgrade.conf
     rm -f /etc/neutron/neutron-openvswitch-agent.conf.d/200-crowbar-upgrade.conf
 
-    if [ $is_openvswitch_updated == 0 ] && [ $was_openvswitch_enabled == 0 ]; then
+    if [ $is_openvswitch_installed == 0 ] && [ $was_openvswitch_enabled == 0 ]; then
```
jsuchome: I think it would work well enough to have `if rpm --quiet -q openvswitch` here ... well but anyway it's temporary solution

Itxaka: Indeed, because if the package is upgraded in iny other time, that check for `$was_openvswitch_enabled` will also fail

So it may be just easier to remove that part and just enable it if the package is installed

Itxaka:
Not needed, package has been fixed, revert on https://github.com/crowbar/crowbar-core/pull/1561

jsuchome:
Should not be needed if we manage to get https://github.com/SUSE-Cloud/automation/pull/2511 and https://github.com/SUSE-Cloud/automation/pull/2510 . (that would mean openvswitch could be installed from the test update channels)

jsuchome:
Finally not needed

PR #1559, opened 2018-04-24
ohai: Treat struct as fixed length instead of zero terminated string

JanZerebecki:
As the result gets written to the existing struct data, unpacking it can
be skipped. Otherwise unpack("P") with a specified length would be needed.
When unpack("p") is used on a struct it might read too many or not
enough bytes depending on when it finds a zero.


JanZerebecki:
@rhafer Thank you for spotting this.

JanZerebecki:
Decided to take a more defensive approach instead: 
While the result gets written to the existing memory of the ecmd
variable, unpack it again to avoid violating any assumptions that
existing memory doesn't change. (Testing didn't show anything making
this assumption, but just to be safe.)

Please review again.

PR #1565, opened 2018-04-26
crowbar: Self-heal if node is missing its node role in the run list

vuntz:
While this should never happen, it did happen for a user who was
manipulating chef data. And when this happens, this basically blocks all
of Crowbar.

It's trivial enough to self-heal to justify that it's worth it.


rsalevsky:
Was there a bugreport? This sound familiar. Also travis is failing.

vuntz:
> Was there a bugreport?

Nothing in bugzilla, something I got from a customer during a discussion.

PR #1570, opened 2018-05-02
[4.0] upgrade: Make sure crowbarctl is installed with correct setup

jsuchome:
crowbarctl is required by crowbar-join in SOC7, but proper configuration
was not created when upgrading from SOC6.
So let's do an explicit installation and configuration during the
upgrade.


jsuchome:
This actually seems to work.

PR #1571, opened 2018-05-04
control_lib: fix host allocation check

seife:
Before the "provisioner: replace curl calls to crowbar API with
crowbarctl" commit, __post_state did an implicit (at least partial)
get_state, updating ADMIN_ADDRESS. Without this, ADMIN_ADDRESS is
unset after allocation, causing the "PXE file contains .*install"
check to fail and the code having to time out after 300 seconds for
the host to be rebooted into installation.
Just add that __get_state in __post_state to fix this, also for other
unidentified cases where this implicit behaviour is expected.


AbelNavarro:
[Fix host allocation check in provisioner for faster host allocation / installation](https://trello.com/c/62b5QWFH/281-fix-host-allocation-check-in-provisioner-for-faster-host-allocation-installation)

PR #1573, opened 2018-05-09
upgrade: Do not upgrade compute-related packages to Pike

jsuchome:
It does not seem to be necessary to upgrade openstack packages
to from Newton to Pike version for them to work with Pike-based control plane.

Restart of current services (after control plane upgrade) is enough.

Note: these specific packages will be upgraded at some point,
but together with all packages on the node.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
@@ -77,29 +77,18 @@ log "No HA setup found..."
 
 <% if @compute_node %>
 
-log "Upgrading packages needed for nova-compute node"
-
-zypper --non-interactive up openstack-nova-compute
-
 log "Restarting services for Pike"
 
-zypper --non-interactive up <%= @neutron_agent %>
 <% unless @remote_node %>
 systemctl restart <%= @neutron_agent %>
 <% end %>
 
-<% if @l3_agent %>
-zypper --non-interactive up <%= @l3_agent %>
-<% unless @remote_node %>
+<% if @l3_agent && !@remote_node %>
 systemctl restart <%= @l3_agent %>
 <% end %>
-<% end %>
 
-<% if @metadata_agent %>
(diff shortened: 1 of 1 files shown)


jsuchome:
I'm actually bit surprised no change in rabbitmq settings is required ... maybe I did not really change it during the upgrade

jsuchome:
I agree - but still, we should test it specifically. 
(But it does not need to block this PR, we could create followup if we find anything)

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -77,30 +77,19 @@ log "No HA setup found..."
...
-zypper --non-interactive up <%= @neutron_agent %>
 <% unless @remote_node %>
 systemctl restart <%= @neutron_agent %>
 <% end %>
 
-<% if @l3_agent %>
-zypper --non-interactive up <%= @l3_agent %>
-<% unless @remote_node %>
+<% if @l3_agent && !@remote_node %>
 systemctl restart <%= @l3_agent %>
```
rhafer: Why is this still needed? You don't update the l3-, metadata- and cinder-agents so why would we need to restart?

jsuchome: Well I do not update _anything_, not even nova-compute. But those services were down when I got to that point. I guess you want to see more details why this happened.

rhafer: > Well I do not update anything, not even nova-compute.

I know, I just overlooked the fact that the script also still restarts nova-compute (it was hidden by the github UI) 

> But those services were down when I got to that point. I guess you want to see more details why this happened.

Well I guess it's the "crowbar-upgrade.rb" recipe doing it.

skazi0:
So the idea here is to remove those packages from "special handling" part but upgrade them together with the rest of OS?

jsuchome:
Right. The services work well together (Newton+Pike) even if not upgraded

PR #1574, opened 2018-05-11
Postpone the upgrade of compute nodes

jsuchome:
Once all other nodes have been upgraded, we want users to allow postponing the upgrade of compute nodes. 

**TODO**

- [x] unit tests in the library
- [x] API change (controller action for postpone/resume)
- [x] crowbar to read indicator file so it allows barclamps editing (3rd commit)
- [x] adapt crowbarctl See https://github.com/crowbar/crowbar-client/pull/174

### DIFF
--- a/crowbar_framework/app/controllers/application_controller.rb
+++ b/crowbar_framework/app/controllers/application_controller.rb
@@ -34,7 +34,8 @@ class ApplicationController < ActionController::Base
     Rails.env.test?
   }
   before_filter :upgrade, if: proc {
-    File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-running")
+    File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-running") &&
+      !File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-compute-nodes-postponed")
   }
   before_filter :sanity_checks, unless: proc {
     Rails.env.test? || \
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -149,10 +149,28 @@ class Api::UpgradeController < ApiController
   # The method runs asynchronously, so there's a need to poll for the status and possible errors
   def nodes
     if params[:component]
+      component = params[:component]
       upgrade_status = ::Crowbar::UpgradeStatus.new
       substep = upgrade_status.current_substep
       status = upgrade_status.current_substep_status
-      if ["all", "controllers"].include? params[:component]
+      if upgrade_status.compute_nodes_postponed? && component != "resume"
+        raise ::Crowbar::Error::UpgradeError,
+          "Upgrade is currently postponed. " \
+          "It has to be resumed before proceeding with any other action."
+      elsif ["resume", "postpone"].include? component
+        if status == :failed
+          raise ::Crowbar::Error::UpgradeError,
+            "Previous step ended with a failure. Not possible to postpone or resume."
+        end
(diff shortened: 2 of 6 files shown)


skazi0:
@jsuchome why do we need a separate file for that? isn't the field in status file enough?

jsuchome:
@skazi0 to combine it with this check https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/controllers/application_controller.rb#L37

I'm not saying we *need* extra file, but it seems like the straightforward update of current concept

susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

skazi0:
@jsuchome ah, so this is to make the before_filter as fast as possible, right? (i.e. no yaml parsing or locking just to check this flag)

jsuchome:
Right

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -34,7 +34,8 @@ class ApplicationController < ActionController::Base
     Rails.env.test?
   }
   before_filter :upgrade, if: proc {
-    File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-running")
+    File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-running") &&
+      !File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-compute-nodes-postponed")
```
rhafer: Does this imply that the crowbar's full functionality is available when the compute nodes upgrade is postponed? Sounds a little dangerous. E.g what happens if the nova barclamp is applied during this timeframe?

jsuchome: Yes, it is dangerous when the barclamp apply affects compute nodes. But that's how I understand the requirement.

There's actually no other effect of this postponing than allowing users to access the barclamps view. If we'd want to restrict it until the full upgrade is done, there's no need for postpone action. We could just enhance UI a little bit so it allows what current CLI already has (separating controllers and computes upgrade).

rhafer: Hm, there's doesn't seem to be  anything in the spec that requires the barclamps to be changeable while there are some notes not upgraded yet. But we might need some clarification on that. As far as I understood it just was just about giving the user some more control about when to upgrade which (compute) nodes. (I think we implemented most of that is already for the C6 to 7 upgrade)

If we really want the barclamp to be editable after the controllers are upgrade I guess we should restrict the change to barclamps that don't have any roles assigned to non-upgraded compute nodes (which might a little involved I guess).

jsuchome: > Hm, there's doesn't seem to be anything in the spec that requires the barclamps to be
> changeable while there are some notes not upgraded yet.

Maybe not in spec, but I remember this was @dirkmueller explanation

>  think we implemented most of that is already for the C6 to 7 upgrade

Indeed. Such limited functionality we already have.

> restrict the change to barclamps that don't have any roles assigned to non-upgraded compute > nodes 

Sounds reasonable, but might be tricky

jsuchome: But this approach is indeed up to discussion

jsuchome: Anyway, if this is the only problem, I'd leave it to followup PR

jsuchome:
@rhafer we might add subtask to https://jira.prv.suse.net/browse/SCRD-3428

jsuchome:
@skazi0 another review please?

PR #1575, opened 2018-05-11
[4.0] upgrade: Added pre-check for database type

jsuchome:
For now, we require users to have MariaDB deployed before the
upgrade can be started.

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -270,6 +270,10 @@ module Api
 
       def deployment_check
         ret = {}
+        # Only allow the upgrade for mariadb setup
+        # TODO: update the search if we allow multiple proposals
+        db_node = ::Node.find("roles:database-server").first
+        ret[:wrong_sql_engine] = true if db_node[:database][:sql_engine] != "mysql"
         # Make sure that node with nova-compute is not upgraded before nova-controller
         nova_order = BarclampCatalog.run_order("nova")
         ::Node.find("roles:nova-compute-*").each do |node|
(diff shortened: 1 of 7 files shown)


crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -270,6 +270,9 @@ def ha_config_check
 
       def deployment_check
         ret = {}
+        # Only allow the upgrade for mariadb setup
+        db_node = ::Node.find("roles:database-server").first
```
jhesketh: This is fine while we only allow one database proposal. Might be worth adding a TODO here to follow up after we decide on a database-migration strategy to check this is still adequate.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1556,14 +1556,22 @@ def sanity_check_errors(check)
       end
 
       def deployment_errors(check)
-        {
-          controller_roles: {
+        ret = {}
+        if check[:controller_roles]
+          ret[controller_roles] = {
```
skazi0: missing `:` ?

crowbar_framework/app/models/api/crowbar.rb
jsuchome: good point

crowbar_framework/app/models/api/upgrade.rb
jsuchome: ah!

jsuchome:
Yes, I'm aware that this might be rewritten/removed once we decide DB migration strategy.

jsuchome:
@jhesketh @rhafer new review please? I had to fix a bug pointed out by @skazi0

PR #1576, opened 2018-05-15
upgrade: Raise the default timeouts for most time consuming actions

jsuchome:
Latest jobs have been failing because of timeout during OS upgrade.

See
https://ci.suse.de/view/Cloud/view/Cloud8/job/cloud-mkcloud8-job-upgrade-nondisruptive-ha-mariadb-x86_64/

`E, [2018-05-15T07:06:38.294506 #3501:0x00000000bc2640] ERROR -- Error while executing OS upgrade script. Possible error during execution of /usr/sbin/crowbar-upgrade-os.sh at d52-54-77-77-01-01.vn2.cloud.suse.de. Action did not finish after 900 seconds. Check /var/log/crowbar/node-upgrade.log for details.`

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade_timeouts.rb
+++ b/crowbar_framework/lib/crowbar/upgrade_timeouts.rb
@@ -24,10 +24,10 @@ module Crowbar
       {
         prepare_repositories: @timeouts_config[:prepare_repositories] || 120,
         pre_upgrade: @timeouts_config[:pre_upgrade] || 300,
-        upgrade_os: @timeouts_config[:upgrade_os] || 900,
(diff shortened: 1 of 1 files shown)


jsuchome:
Update: just checked one job and it took 24 minutes to upgrade. I raised the timeout again to 1500 seconds

bbobrov:
Why does it take longer than it took before?

jsuchome:
I'm not sure @bbobrov . Maybe there's more packages, maybe zypper is slower, maybe the access to repositories is somewhat slow, or the mkcloud hosts are overloaded...

PR #1577, opened 2018-05-15
[4.0] upgrade: Adapt ceph-related pre-checks to 7-8 upgrade

jsuchome:
We do not allow the upgrade when ceph roles are present (that is,
if ceph was deployed using crowbar).

We no longer check the ceph cluster health. Unfortunatelly crowbar
does not have a knowledge where the ceph nodes are.


jsuchome:
Looks like there's also some related code in crowbar-ceph...

PR #1579, opened 2018-05-15
[4.0] upgrade: Correctly check for the presence of various upgrade indication

jsuchome:
On SOC7, there could be 2 files indicating running upgrade: one for 6-7
and one for 7-8. Fix the check to make sure we do not miss one of these
situations.


jsuchome:
@nicolasbock yes, the file is deleted when the upgrade is done.

But on SOC7 there could be 2 kinds of upgrades running (not at the same time of course): 6-7 or 7-8. Both are using different file to indicate the running upgrade

PR #1584, opened 2018-05-21
Revert "upgrade: Do not upgrade compute-related packages to Pike"

jsuchome:
This reverts commit 7d94b14fa3a9f902642493e5ec2cb293dcd90c02.

Turns out we might still need Pike client packages....see https://jira.prv.suse.net/browse/SCRD-3245 for more details.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
@@ -77,18 +77,29 @@ log "No HA setup found..."
 
 <% if @compute_node %>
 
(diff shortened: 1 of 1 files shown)


jsuchome:
@rhafer commit message updated

jsuchome:
@dirkmueller raised some concerns actually. It's possible we need to keep compute nodes completely untouched (until they are upgraded)

jsuchome:
One idea is to keep both options - some users might be fine after just starting the services, while some might need the services upgrade. See https://github.com/crowbar/crowbar-core/pull/1591

jsuchome:
Should not be needed after all. We have update fix our CI jobs so that cinder-volume is not placed on compute nodes

PR #1586, opened 2018-05-22
upgrade: Remove the database step

jsuchome:
Previously we had crowbar-init finishing admin upgrade step at the
point the admin server came back from reboot.
Now, crowbar-init is not needed any more during upgrade so let
crowabr itself make the end of step action once it is started.

### DIFF
--- a/crowbar_framework/config/puma.rb
+++ b/crowbar_framework/config/puma.rb
@@ -52,6 +52,18 @@ end
   FileUtils.mkdir_p File.join(ROOT, name)
 end
 
+# When starting the process during the upgrade (after reboot),
+# mark the end of "admin server upgrade" step.
+if File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-running")
+  CROWBAR_LIB_DIR = "/opt/dell/crowbar_framework/lib".freeze
+  $LOAD_PATH.push CROWBAR_LIB_DIR if Dir.exist?(CROWBAR_LIB_DIR)
+
+  require "logger"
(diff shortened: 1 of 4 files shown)


crowbar_framework/config/puma.rb
```diff
@@ -52,6 +52,17 @@
   FileUtils.mkdir_p File.join(ROOT, name)
 end
 
+CROWBAR_LIB_DIR = "/opt/dell/crowbar_framework/lib".freeze
+$LOAD_PATH.push CROWBAR_LIB_DIR if Dir.exist?(CROWBAR_LIB_DIR)
+
+# set the end_step status during the upgrade
+if File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-running")
```
nicolasbock: Do we need a shell pattern here as you did in a previous change?

jsuchome: Nope, this code is running on SOC8 and we're only interesting in SOC8 part of the upgrade

crowbar_framework/config/puma.rb
```diff
@@ -52,6 +52,17 @@
   FileUtils.mkdir_p File.join(ROOT, name)
 end
 
+CROWBAR_LIB_DIR = "/opt/dell/crowbar_framework/lib".freeze
```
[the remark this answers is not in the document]
jsuchome: Good point (this was copy and paste from crowbar-init's puma.rb)

jsuchome:
@skazi0 check the fresh changes

PR #1591, opened 2018-05-25
WIP upgrade: Split the preparation of compute nodes into 2 phases

jsuchome:
In the first phase (done after controllers upgrade), we start the
services at compute nodes. After that, live migration could work,
esp. when there are only nova-compute services located at compute nodes.

In the second phase (done at the start of compute node upgrade), we
upgrade services located at all compute nodes to make sure that
live migration works.

Possible case for such requirement is when cinder-volume is located
at compute node and instances are using volumes. cinder services need
to be upgraded, so we need to go to 2nd step in this setup.

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -742,11 +742,12 @@ module Api
           # Upgrade everything
           do_controllers_substep(substep)
           substep = :compute_nodes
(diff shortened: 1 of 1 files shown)


jsuchome:
Note - this PR is obviously missing some parts, but I think it's obvious how they should be written. My point so far is to offer the idea for discussion

PR #1592, opened 2018-05-28
[4.0] upgrade: Unlock crowbar-ui after completed upgrade

skazi0:
crowbar-ui package was locked before admin server upgrade. After the
whole upgrade process is completed, the lock can be removed to enable
crowbar-ui update before upgrade to next Cloud release.

Note https://github.com/crowbar/crowbar-core/pull/1593 should be merged first and this one rebased if needed.

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -763,6 +763,7 @@ module Api
         if status.progress[:remaining_nodes].zero?
           status.save_substep(substep, :finished)
           finalize_nodes_upgrade
+          unlock_crowbar_ui_package
           status.end_step
         end
       rescue ::Crowbar::Error::Upgrade::NodeError => e
@@ -1533,6 +1534,24 @@ module Api
           raise "Possible error during execution of #{script} at node(s) " \
             "#{running.join(",")}. " \
             "Action did not finish after #{seconds} seconds."
+        end
+      end
+
+      # remove crowbar-ui package lock for future upgrades
+      def unlock_crowbar_ui_package
+        ui_unlock = run_cmd("sudo zypper-retry removelock 'crowbar-ui*'")
+        unless ui_unlock[:exit_code].zero?
+          raise_node_upgrade_error(
+            "Removing crowbar-ui package lock has failed. " \
+            "Please re-try current step to finalize the upgrade."
+          )
+        end
+        ui_update = run_cmd("sudo zypper-retry -n update crowbar-ui")
+        unless ui_update[:exit_code].zero?
+          raise_node_upgrade_error(
+            "Updating crowbar-ui package has failed. " \
+            "Please re-try current step to finalize the upgrade."
+          )
         end
       end
 


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1536,6 +1537,18 @@ def execute_scripts_and_wait_for_finish(nodes, script, seconds)
         end
       end
 
+      # remove crowbar-ui package lock for future upgrades
+      def unlock_crowbar_ui_package
+        ui_unlock = run_cmd("sudo zypper-retry removelock 'crowbar-ui*'")
+        unless ui_unlock[:exit_code].zero?
+          Rails.logger.error(
```
jsuchome: Use `raise_node_upgrade_error` here

skazi0: Is this really node upgrade error?

jsuchome: In the wider sense :-)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1536,6 +1537,17 @@ def execute_scripts_and_wait_for_finish(nodes, script, seconds)
         end
       end
 
+      # remove crowbar-ui package lock for future upgrades
+      def unlock_crowbar_ui_package
+        ui_unlock = run_cmd("sudo zypper-retry removelock 'crowbar-ui*'")
+        unless ui_unlock[:exit_code].zero?
+          raise_node_upgrade_error(
+            "Removing crowbar-ui package lock has failed. " \
+            "Make sure to remove it before next upgrade."
```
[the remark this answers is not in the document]
Itxaka: I guess that if this fails it would require some manual action to fix it as a number of different things could prevent this from working. Maybe repeating + advice to manually check if the error persist is the way to go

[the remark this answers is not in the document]
skazi0: Message updated. Not sure if the wording makes sense...

rhafer:
@skazi0 I might be missing something, but shouldn't we explicitly install the updated crowbar-ui package after unlocking it? Or when is that supposed to happen?

jsuchome:
New version of package is not really needed after the upgrade is finished (that's like - after upgrade to SOC7, new version already available covers 7-8 upgrade). 
So it's enough to be installed once user starts new upgrade, I'd say

skazi0:
@rhafer we might but it can be risky as it may pull some additional dependencies etc. as @jsuchome wrote, we don't need it immediately after finished upgrade so in worst case it will be installed just before next upgrade. It should be detected by existing prechecks but we may need to add a special one to explicitly inform users that after upgrading crowbar-ui they need to refresh the UI.

rhafer:
> we might but it can be risky as it may pull some additional dependencies etc. as @jsuchome wrote, we don't need it immediately after finished upgrade so in worst case it will be installed just before next upgrade. It should be detected by existing prechecks but we may need to add a special one to explicitly inform users that after upgrading crowbar-ui they need to refresh the UI.

Hm, but isn't this also true the otherway around? Well have an outdated package hanging around, which because it was locked might have prevented other packages from being updated to the most recent version. Somehow I think we should make sure that after the upgrade completed all packages should be installed in their Cloud 8 version. But maybe I am fundamentally misunderstanding something here.

skazi0:
@rhafer you're fundamentally correct :o) This mechanism is not good for all cases but in our current scenario we get this just after admin server upgrade:
```
$ LANG=C zypper up
Loading repository data...
Reading installed packages...

The following item is locked and will not be changed by any action:
 Installed:
  crowbar-ui

The following package update will NOT be installed:
  crowbar-ui

Nothing to do.
```

So there seem to be no unwanted dependencies or other side effects so far.

rhafer:
@skazi0 But what's the risk of just explicitly updating crowbar-ui after the upgrade completed? If it needs additional packages, well it would be installed at some point anyway. If we do it after the upgrade, we have control over when that happens, and can be sure to have a "clean" system after the upgrade.

skazi0:
@rhafer I'm not sure if there is any "real" risk. I can add it if you insist :o)

skazi0:
Added update of crowbar-ui package after unlock. Please re-review.

PR #1598, opened 2018-06-02
upgrade: Unblock upgrade status API in Cloud8

skazi0:
After finished 7->8 upgrade, cloud API is disabled. One exception is
upgrade status API which should still be available to let users query
status of finished upgrade (i.e. after upgrade was completed).

### DIFF
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -16,7 +16,7 @@
 
 class Api::UpgradeController < ApiController
   # disable upgrade API until upgrade to next version is implemented
-  before_action do
+  before_action(except: :show) do
     # skip filter if we're in the middle of upgrade from previos version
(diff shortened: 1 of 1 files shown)


jack-likes-to-code:
I don't know Ruby to fully understand what's going on here, but Julie will try patching it into our system after the admin node upgrade and see how it goes.

skazi0:
@jack-likes-to-code the code is tricky indeed. It passes `except` parameter to `before_action` to say that the action should not be executed before `show()` which translates to simple GET to the `/api/upgrade` endpoint. Pure rails/ruby magic :o)

jsuchome:
Do we need a backport?

jack-likes-to-code:
Forgot to mention that I tried this change out and it indeed worked for me.  No more strange backend message.  Thanks for the fix @skazi0 !

skazi0:
I don't think we need a backport for this. The original filter was reverted on cloud7 side. It should only exist in the latest version.

PR #1601, opened 2018-06-11
upgrade: Fix file layout for rails' autoloading (bsc#1096759)

rhafer:
- Removed error.rb to remove the apparently unneeded use of ruby's
  autoload module for the various upgrade exceptions
- Move the UpgradeError class to its own file in a location where
  rails autoloading will actually find it.


rhafer:
> Do we need a backport?

Good question. Still need to check if this is also present with the Cloud7/SP2 code.

rhafer:
stable/4.0 doesn't seem to be directly affected, but I think it's better to backport this anyway.

PR #1603, opened 2018-06-12
upgrade: Check input is a valid node for nodes

s-t-e-v-e-n-k:
The else clause for upgrade nodes assumes input is valid, and if it
isn't, returns an unhelpful message about how controller nodes need to
be upgraded first. Validate the input to provide a helpful error.

### DIFF
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -183,6 +183,12 @@ class Api::UpgradeController < ApiController
           ::Crowbar::UpgradeStatus.new.start_step(:nodes)
         end
       else
+        # At this point params[:component] should be a node, if it is not,
+        # raise an error.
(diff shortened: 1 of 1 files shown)


jsuchome:
@s-t-e-v-e-n-k are you sure you have the latest changes? Current code looks different...

s-t-e-v-e-n-k:
@jsuchome I think you're looking at your unmerged PR that reflows that method, not master.

jsuchome:
I don't think so. This code is in master already:

https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/controllers/api/upgrade_controller.rb#L156

s-t-e-v-e-n-k:
@jsuchome I see you're right. However, it rebases without issue, which I'll push up after I've tested my changes for hounds whining.

rsalevsky:
Can you please create the backport?

PR #1607, opened 2018-06-13
upgrade: Adapt the rabbitmq client's config after rabbitmq setup change

jsuchome:
If rabbitmq is using DRBD in SOC7, it's switched to the clustered
setup automatically.
We need to update some configuration files on the nodes temporarily,
so that nova API works when doing live-migration (that is, before the
compute node is fully upgraded).

### DIFF
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -130,6 +130,11 @@ end
 
 swift_storage = roles.include? "swift-storage"
 
+rabbitmq_server = search(:node, "run_list_map:rabbitmq-server").first
+# original DRBD setup is switched to cluster one in prepare_nodes_for_os_upgrade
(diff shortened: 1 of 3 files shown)


susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

skazi0:
@jsuchome shouldn't it also update the proposal? what will happen when chef-client is run after the upgrade?

jsuchome:
@skazi0 that's already done https://github.com/crowbar/crowbar-core/pull/1557

skazi0:
@jsuchome wouldn't those client configs be updated by chef-run too?

jsuchome:
Of course - chef-client is run after the node is upgraded. 

But live migration of nova instances happens after upgrading the control plane, but before upgrading the compute nodes.

skazi0:
ah, true, forgot about this (again)...

jsuchome:
This does not seem to be needed after all

PR #1610, opened 2018-06-14
upgrade: Improve various error messages

skazi0:
Multiple error handling/message improvements. Please check commit messages for details.

Includes https://github.com/crowbar/crowbar-core/pull/1585

### DIFF
--- a/crowbar_framework/app/models/node.rb
+++ b/crowbar_framework/app/models/node.rb
@@ -903,6 +903,8 @@ class Node < ChefObject
     cmd = script
     cmd += " " + args.join(" ") unless args.empty?
 
+    raise "Script #{script} is not present on the node #{@node.name}." unless file_exist?(script)
+
     base = "/var/lib/crowbar/upgrade/" + File.basename(script, ".sh")
     ok_file = base + "-ok"
     failed_file = base + "-failed"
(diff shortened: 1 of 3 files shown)


jsuchome:
@skazi0 would you mind making this a bigger cleanup PR, adding also 

- https://github.com/crowbar/crowbar-core/pull/1585/commits/2a59894eb09369cb278cb519a3ed4388ac15cfb4

- http://pastebin.nue.suse.com/21339/src

skazi0:
@jsuchome  sure I can include the change from pastebin but the commit you linked is already in the repo, isn't it?

jsuchome:
I don't think so. I've just created https://github.com/crowbar/crowbar-core/pull/1585  some time ago and did not want to merge it yet as it was very little change

crowbar_framework/app/models/node.rb
```diff
@@ -903,6 +903,8 @@ def wait_for_script_to_finish(script, seconds, args = [])
     cmd = script
     cmd += " " + args.join(" ") unless args.empty?
 
+    raise "Script #{script} is not present at the node #{@node.name}." unless file_exist?(script)
```
rhafer: I think it's `s/at/on/` but you might want to check back with a native speaker.

skazi0: I would probably say "on" myself too, but I stole this message from @jsuchome :)

jgravel: I've asked Gary Smith who's a native speaker :-) He recommended using 'on' instead of 'at' in this case.

jsuchome: Oh, thanks for assessing my English ;-)

(Feel free to correct it of course)

PR #1612, opened 2018-06-18
[4.0] upgrade: Do not allow cinder-volume on compute nodes

jsuchome:
Compute nodes need to work even when they are not upgraded
(for live-migration from Newton-based node to Pike-based one).

But cinder has to be fully upgraded everywhere. So let's not allow
any cinder service to run on compute node.

https://jira.prv.suse.net/browse/SCRD-3543


skazi0:
@jsuchome would be good to mention somewhere that this is done as a precheck.

jsuchome:
@skazi0 commit message extended

PR #1613, opened 2018-06-19
[stable/4.0] upgrade: Check input is a valid node for nodes

s-t-e-v-e-n-k:
The else clause for upgrade nodes assumes input is valid, and if it
isn't, returns an unhelpful message about how controller nodes need to
be upgraded first. Validate the input to provide a helpful error.

(cherry picked from commit d3cd42d8816b892b23cb332772a773b1768cd125)

https://github.com/crowbar/crowbar-openstack/pull/1692 is the PR for master.

requests?**

### DIFF
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -151,6 +151,12 @@ class Api::UpgradeController < ApiController
           ::Crowbar::UpgradeStatus.new.start_step(:nodes)
         end
       else
+        # At this point params[:component] should be a node, if it is not,
+        # raise an error.
+        unless Node.all.map(&:name).include?(params[:component])
+          raise ::Crowbar::Error::UpgradeError, "Component must be 'all', "\
+            "'controllers' or a node name."
+        end
         if substep != :compute_nodes && status != :finished
           raise ::Crowbar::Error::UpgradeError.new(
             "Controller nodes must be upgraded first!"


crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -151,6 +151,12 @@ def nodes
           ::Crowbar::UpgradeStatus.new.start_step(:nodes)
         end
       else
+        # At this point params[:component] should be a node, if it is not,
+        # raise an error.
+        unless Node.all.map(&:name).include? params[:component]
+          raise ::Crowbar::Error::UpgradeError, "Component must be 'all', "\
+            "'controllers', 'resume', 'postpone' or a node name."
```
jsuchome: 'resume', 'postpone' are not available in previous upgrade version

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -151,6 +151,12 @@ def nodes
           ::Crowbar::UpgradeStatus.new.start_step(:nodes)
         end
       else
+        # At this point params[:component] should be a node, if it is not,
+        # raise an error.
+        unless Node.all.map(&:name).include? params[:component]
```
rsalevsky: Can you please use: `Node.all.map(&:name).include?(params[:component])`

skazi0:
please mention the original PR in the description

PR #1614, opened 2018-06-19
upgrade: Stop cron before stopping any other service

rhafer:
There was a small race condition that caused cron jobs (e.g.
glance-scrubber) to run while the openstack services were offline
already. This polluted the service's log file and confused some of the
test done by mkcloud. Stopping cron a bit earlier during the upgrade
shoud avoid that.


rhafer:
I guess we should backport this to stable/4.0 as well.

PR #1615, opened 2018-06-19
upgrade: Wait for services shutdown to finish

jsuchome:
The script was executed in the background and we did not wait for
it to finish. As a result, other scripts were started before this one was done.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
@@ -25,6 +25,15 @@ set -x
 systemctl stop cron
 systemctl disable cron
 
+mkdir -p $UPGRADEDIR
+
+if [[ -f $UPGRADEDIR/crowbar-shutdown-services-before-upgrade-ok ]] ; then
+    log "Services are already shut down."
+    exit 0
+fi
+
+rm -f $UPGRADEDIR/crowbar-shutdown-services-before-upgrade-failed
+
 <% if @nova_controller && (!@use_ha || @cluster_founder) %>
 # Remove nova-cert service which has been deprecated in Newton and is not present in Pike
 
@@ -50,16 +59,16 @@ for type in clone ms primitive; do
     do
         log "Stopping resource $resource"
         crm --force --wait resource stop $resource
+        ret=$?
+        if [ $ret != 0 ] ; then
+          log "Error occured during stopping resource $resource."
+          echo $ret > $UPGRADEDIR/crowbar-shutdown-services-before-upgrade-failed
+          exit $ret
+        fi
     done
 done
 
 <% end %>
-
-# stop cinder-volume at all nodes so the service could be removed from the database later
-if test -e /usr/sbin/rcopenstack-cinder-volume; then
-    log "Stopping cinder-volume service"
-    /usr/sbin/rcopenstack-cinder-volume stop
-fi
 
 <% else %>
 
@@ -80,4 +89,5 @@ done
 
 <% end %>
 
+touch $UPGRADEDIR/crowbar-shutdown-services-before-upgrade-ok
 log "$BASH_SOURCE is finished."
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -874,7 +874,7 @@ module Api
           "post-upgrade",
           "chef-upgraded"
         ].map { |f| "/usr/sbin/crowbar-#{f}.sh" }.join(" ")
-        scripts_to_delete << "/etc/neutron/lbaas-connection.conf"
+        scripts_to_delete << " /etc/neutron/lbaas-connection.conf"
         node.run_ssh_cmd("rm -f #{scripts_to_delete}")
       end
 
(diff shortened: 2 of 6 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -468,7 +460,9 @@ def services
           false,
           services: {
             data: e.message,
-            help: "Crowbar has failed. Check /var/log/crowbar/production.log for details."
+            help: "Check /var/log/crowbar/production.log at admin server for details." \
+              "If the action failed at a specific node, " \
+              "check /var/log/crowbar/node-upgrade.log at the node."
```
rhafer: Nit: Isn't it `s/at/on/`?

And  `check /var/log/crowbar/node-upgrade.log *on that* node.` ?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -468,7 +460,9 @@ def services
           false,
           services: {
             data: e.message,
-            help: "Crowbar has failed. Check /var/log/crowbar/production.log for details."
+            help: "Check /var/log/crowbar/production.log at admin server for details." \
+              "If the action failed at a specific node, " \
```
rhafer: nit: Actually I meant `s/at/on ` for all the occurrences here. (And in this PR in general).

jsuchome: Oh. Sure, I missed the other one :-)

chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
```diff
@@ -47,14 +57,15 @@ for type in clone ms primitive; do
     done
 done
 
-<% end %>
-
-# stop cinder-volume at all nodes so the service could be removed from the database later
-if test -e /usr/sbin/rcopenstack-cinder-volume; then
-    log "Stopping cinder-volume service"
-    /usr/sbin/rcopenstack-cinder-volume stop
+ret=$?
```
jhesketh: This will only check the result of the last crm call. It's possible an earlier one will have failed.

chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
```diff
@@ -19,6 +19,16 @@ log "Executing $BASH_SOURCE"
 
 set -x
 
+mkdir -p $UPGRADEDIR
+
+if [[ -f $UPGRADEDIR/crowbar-shutdown-services-before-upgrade-ok ]] ; then
```
jhesketh: I wonder if it is possible for services to somehow come back up after this script has been ran? (for example an external actor such as chef restarting them). If something else failed during the shutdown this file could still be touched. Perhaps we need to set -e and do some catching?

jsuchome: 1. chef is stopped. If user intervenes manually, he should know how to handle the status file.

2. the file is touched at the very end of the script. It's true that something new could have failed but then we do not report it anyway, so how can one know? Only by going through the logs after some mysterious behavior caused by unreported bug. And once user is already at it, they could see what's happening in the log file...

That said, maybe your idea about `-e` is correct, but that would be general improvement of all the upgrade scripts we're using, not just this one. See https://github.com/crowbar/crowbar-core/tree/master/chef/cookbooks/crowbar/templates/default

chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
jsuchome: Good point. I'll move the check to crm call.

jsuchome:
rebased after merging https://github.com/crowbar/crowbar-core/pull/1614

chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
```diff
@@ -50,17 +59,17 @@ for type in clone ms primitive; do
     do
         log "Stopping resource $resource"
         crm --force --wait resource stop $resource
```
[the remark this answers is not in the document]
jsuchome: I'm not sure if this command actually prints something, but the whole script has `-x` set so you should see everything in `/var/log/crowbar/node-upgrade.log`

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -881,7 +874,7 @@ def delete_upgrade_scripts(node)
           "post-upgrade",
           "chef-upgraded"
         ].map { |f| "/usr/sbin/crowbar-#{f}.sh" }.join(" ")
-        scripts_to_delete << "/etc/neutron/lbaas-connection.conf"
+        scripts_to_delete << " /etc/neutron/lbaas-connection.conf"
```
skazi0: maybe add some comment why this conf file is deleted together with scripts?

jsuchome: Good point, I have no idea. (That was not the point of the fix however)

rhafer: I think that file is specific to the lbaas migration tool. It contains the connection details to the neutron database.  But you're right that more something for a separate commit.

PR #1618, opened 2018-06-19
upgrade: Add a validation check for postponed nodes.

jsuchome:
- When the upgrade is not finished, show a link to wizard.
- Add a validation check to make sure postponed nodes can be skipped.
- Let skip_unready_nodes skip also nodes that are in crowbar_upgrade state 

https://jira.prv.suse.net/browse/SCRD-3745

### DIFF
--- a/crowbar_framework/app/controllers/application_controller.rb
+++ b/crowbar_framework/app/controllers/application_controller.rb
@@ -34,6 +34,10 @@ class ApplicationController < ActionController::Base
     Rails.env.test?
   }
   before_filter :upgrade, if: proc {
+    if File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-compute-nodes-postponed")
+      flash[:warning] = "Some nodes are not upgraded yet. " \
+        "Continue to <a href='/upgrade'>upgrade wizard</a> to complete the upgrade."
+    end
     File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-running") &&
       !File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-compute-nodes-postponed")
   }
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -692,7 +692,7 @@ class ServiceObject
   # the nodes can be skipped.
   #
   def validate_postponed_nodes(proposal)
-    return unless File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-compute-nodes-postponed")
+    return unless upgrade_postponed?
 
     skip_unready_nodes_enabled = Rails.application.config.experimental.fetch(
       "skip_unready_nodes", {}
@@ -1681,6 +1681,11 @@ class ServiceObject
 
   THREAD_POOL_SIZE = 20
 
+  # Is the upgrade of (some) compute nodes postponed
+  def upgrade_postponed?
+    Dir.glob("/var/lib/crowbar/upgrade/*-upgrade-compute-nodes-postponed").any?
+  end
+
   def wait_for_chef_daemons(node_list)
     return if node_list.empty?
 
@@ -1936,8 +1941,14 @@ class ServiceObject
         pre_cached_nodes[n] ||= Node.find_by_name(n)
         node = pre_cached_nodes[n]
         next if node.nil?
-        # skip if nodes are on ready or crowbar_upgrade state, we dont need to do anything
-        next if ["ready", "crowbar_upgrade"].include?(node.state)
+        # skip if nodes are on ready, we dont need to do anything
+        next if node.state == "ready"
+        # crowbar_upgrade is not an error state and some upgrade specific recipes are executed
+        # for the nodes in this state. So we do not want to skip it by default.
+        # The exceptional case is when the upgrade of compute nodes is postponed and
+        # user wants to apply some proposal. In such case we really want to skip such nodes
+        # so the proposal can be applied only to the nodes that are already upgraded and ready.
+        next if node.state == "crowbar_upgrade" && !upgrade_postponed?
         logger.warn(
           "Node #{n} is skipped until next chef run for #{bc}:#{inst} with role #{role}"
         )
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -682,7 +682,41 @@ class ServiceObject
   #
   def validate_proposal_after_save proposal
     validate_proposal_constraints proposal
+    validate_postponed_nodes proposal
     handle_validation_errors
+  end
+
+  #
+  # Make sure that when user wants to apply changes affecting the nodes that are postponed
+  # (= not upgraded while rest of the cloud is already upgraded)
+  # the nodes can be skipped.
+  #
+  def validate_postponed_nodes(proposal)
+    return unless File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-compute-nodes-postponed")
+
+    skip_unready_nodes_enabled = Rails.application.config.experimental.fetch(
+      "skip_unready_nodes", {}
+    ).fetch("enabled", false)
+
+    skip_unready_nodes_roles = Rails.application.config.experimental.fetch(
+      "skip_unready_nodes", {}
+    ).fetch("roles", [])
+
+    postponed_nodes = []
+    proposal["deployment"][@bc_name]["elements"].each do |role, nodes|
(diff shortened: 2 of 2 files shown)


susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please remove period from end of commit subject line.<br />
83d5aa7d2ed45039c068eecec2fa9508e5c4c2e5</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please remove period from end of commit subject line.<br />
7d2af29d3c05db741e7265cbd69509a35b2bf267</td>
    </tr>
  </tbody>
</table>

</p>

crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -34,6 +34,10 @@ class ApplicationController < ActionController::Base
     Rails.env.test?
   }
   before_filter :upgrade, if: proc {
+    if File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-compute-nodes-postponed")
+      flash[:warning] = "Some nodes are not upgraded yet. " \
```
jhesketh: are flash's translatable somewhere?

jsuchome: I'm not sure about flash messages, but validation ones are sometimes set for translation...
I chose not to this time, as other  validation strings in the same file are not translated.

It's a bit messy.

susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

crowbar_framework/app/models/service_object.rb
```diff
@@ -1902,8 +1931,8 @@ def skip_unready_nodes(bc, inst, new_elements, old_elements)
         pre_cached_nodes[n] ||= Node.find_by_name(n)
         node = pre_cached_nodes[n]
         next if node.nil?
-        # skip if nodes are on ready or crowbar_upgrade state, we dont need to do anything
-        next if ["ready", "crowbar_upgrade"].include?(node.state)
+        # skip if nodes are on ready, we dont need to do anything
+        next if node.state == "ready"
```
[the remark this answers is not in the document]
rhafer: @jsuchome Hm. Would it help to only skip the check for "crowbar_upgrade" if "7-to-8-upgrade-compute-nodes-postponed" exists?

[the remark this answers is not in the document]
jsuchome: It would. But don't you find it super ugly?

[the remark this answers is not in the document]
rhafer: Oh, yes for sure it's ugly! Unfortunately I am currently lacking any better idea. Might need to spend some additional thoughts on that.

You could dress it up a bit by introducing some library call for the check.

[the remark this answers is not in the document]
Itxaka: cant we check for the actual upgrade step here?

`next if node.state == "crowbar_upgrade" && node["crowbar_upgrade_step"] == :services`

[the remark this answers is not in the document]
jsuchome: We could, but that's just other way of going around it...

crowbar_framework/app/models/service_object.rb
```diff
@@ -1884,6 +1918,11 @@ def skip_unchanged_nodes(elements, old_role, role)
     cleaned_elements
   end
 
+  # if this is true, node should not be skipped even though it's in crowbar_upgrade state
+  def crowbar_upgrade_needs_to_run(node)
```
rhafer: Nit: I think this name is somewhat misleading. I'd suggest to reverse the logic here and call the method something like `upgrade_postponed?`

Also the `node` parameter is unused. Is it actually needed?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1902,8 +1941,9 @@ def skip_unready_nodes(bc, inst, new_elements, old_elements)
         pre_cached_nodes[n] ||= Node.find_by_name(n)
         node = pre_cached_nodes[n]
         next if node.nil?
-        # skip if nodes are on ready or crowbar_upgrade state, we dont need to do anything
-        next if ["ready", "crowbar_upgrade"].include?(node.state)
+        # skip if nodes are on ready, we dont need to do anything
+        next if node.state == "ready"
+        next if node.state == "crowbar_upgrade" && crowbar_upgrade_needs_to_run(node)
```
rhafer: Some comment on why this is actually needed would be helpful IMO. Remember how long it took you to understand the logic.

crowbar_framework/app/models/service_object.rb
jsuchome: True about `node` and comments. 
`upgrade_postponed?` - OK, I can change it, but then I'll have to negate the output in the check later which is somewhat less nice.

crowbar_framework/app/models/service_object.rb
```diff
@@ -682,9 +682,43 @@ def validate_proposal proposal
...
   end
 
+  #
+  # Make sure that when user wants to apply changes affecting the nodes that are postponed
+  # (= not upgraded while rest of the cloud is already upgraded)
+  # the nodes can be skipped.
+  #
+  def validate_postponed_nodes(proposal)
+    return unless upgrade_postponed?
+
```
Itxaka: I would streamline and exit fast from here if there is nothing to do:

```
skip_unready_nodes = Rails.application.config.experimental.fetch("skip_unready_nodes", {})

skip_unready_nodes_enabled = skip_unready_nodes.fetch("enabled", false)
raise Crowbar::NeedOptionEnabled("skip_unready_nodes") unless skip_unready_nodes_enabled

skip_unready_nodes_roles = skip_unready_nodes.fetch("roles", [])
return if skip_unready_nodes_roles.size == 0

```

crowbar_framework/app/models/service_object.rb
```diff
@@ -1647,6 +1681,11 @@ def run_remote_chef_client(node, command, logfile_name)
 
   THREAD_POOL_SIZE = 20
 
+  # Is the upgrade of (some) compute nodes postponed
+  def upgrade_postponed?
+    Dir.glob("/var/lib/crowbar/upgrade/*-upgrade-compute-nodes-postponed").any?
```
Itxaka: should this point to the exact versioning for the upgrade (i.e. "7-8") or is that file not versioned yet?

crowbar_framework/app/models/service_object.rb
```diff
@@ -682,9 +682,43 @@ def validate_proposal proposal
...
+      next if skip_unready_nodes_enabled && skip_unready_nodes_roles.include?(role)
+      nodes.each do |n|
+        next if is_cluster? n
+        node = NodeObject.find_by_name(n)
+        postponed_nodes << n unless node.upgraded?
+      end
+    end
+    return if postponed_nodes.empty?
+
+    validation_error "The upgrade of some nodes has been postponed: " +
```
skazi0: won't it generate duplicates if there are multiple roles in one barclamp?

crowbar_framework/app/models/service_object.rb
jsuchome: > raise Crowbar::NeedOptionEnabled("skip_unready_nodes") unless skip_unready_nodes_enabled

Hm, there's no reason to raise here until we find out if any postponed node has a role in this barclamp.

> return if skip_unready_nodes_roles.size == 0

Why would we do it? Again, I think we must compare `skip_unready_nodes_roles` with the roles being applied.

crowbar_framework/app/models/service_object.rb
skazi0: I don't have any example at hand... but it doesn't mean that it can't happen :)

jsuchome: I think it can't happen with our barclamps

jsuchome: @skazi0 anyway, i've added `uniq` now.

crowbar_framework/app/controllers/application_controller.rb
nicolasbock: Are flash message shown to the user in the UI? Then we might want to localize the message I would say.

jsuchome: That's the same as for validation strings - in the end, they are presented by flash messages.

And those defined in this file (the old ones, not the one I've added) are not set for translation...

PR #1621, opened 2018-06-22
upgrade: Add missing brackets checking for nodes

s-t-e-v-e-n-k:
Forward port from stable/4.0 adding brackets to the check that the
component for upgrade node is a valid node name.

requests?**


jsuchome:
That's just for consistency, right? No error happened because of this?

s-t-e-v-e-n-k:
Correct, it's just to have the code the same for consistency.

PR #1623, opened 2018-06-28
upgrade: Wait until all nova-compute services are up before evacuation

jsuchome:
As nova-compute services were started on compute nodes just shortly before
this step, let's wait for them to correctly report their status before
we start with live-migrations.
 
This should help with https://jira.prv.suse.net/browse/SCRD-3750

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-evacuate-host.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-evacuate-host.sh.erb
@@ -45,6 +45,19 @@ ORIGINAL_INSTANCES=$UPGRADEDIR/original_instances
 # List active instances on host before the live-migration starts
 openstack server list --insecure --all-projects --status active --host $host -f value -c ID > $ORIGINAL_INSTANCES
 
+i=0
+while [[ $i -lt 10 ]] && openstack --insecure compute service list --service nova-compute -f value -c State | grep down ; do
+    log "Some nova-compute service is still down..."
+    sleep 6
+    i=$(($i + 1))
+done
+
+if openstack --insecure compute service list --service nova-compute -f value -c State | grep down ; then
+    log "Some nova-compute service is still down. Check the state of compute nodes before proceeding with the live-migration."
+    echo 4 > $UPGRADEDIR/crowbar-evacuate-host-failed
+    exit 4
+fi
+
 if ! nova --insecure host-evacuate-live "$host" $blockmigrate; then
     log "Live migration of instances from host $host has failed!"
     echo 2 > $UPGRADEDIR/crowbar-evacuate-host-failed


susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

chef/cookbooks/crowbar/templates/default/crowbar-evacuate-host.sh.erb
```diff
@@ -45,6 +45,11 @@ ORIGINAL_INSTANCES=$UPGRADEDIR/original_instances
 # List active instances on host before the live-migration starts
 openstack server list --insecure --all-projects --status active --host $host -f value -c ID > $ORIGINAL_INSTANCES
 
+while openstack --insecure compute service list --service nova-compute -f value -c State | grep down ; do
```
nicolasbock: Does it make sense to wait indefinitely?

matthewoliver: Yeah, if there is some other problem we don't want the upgrade to stall infinitely here. I'm not too worried about what that number should be, it could be long if we really need to wait awhile. On my last few upgrades, it didn't seem to take too long. Maybe 5 or 10 minutes? Or is that way to small a time? How long could a potential compute reboot take? jump it to 30 mins?

jsuchome: Well, there's a timeout on the UI side that tells the user there's an error when the scripts does not finish in X minutes.

But maybe we could add some timeout also into the script itself...

matthewoliver: The loop down on line 59 is ok, because if there is an error the status won't be ACTIVE or MIGRATING anymore I suspect. We cloud potentially look at a similar thing if we don't want the timeout approach.

jsuchome:
@matthewoliver @nicolasbock updated

matthewoliver:
Cool, looks good. I'll give it a run on my cloud just to make sure.

chef/cookbooks/crowbar/templates/default/crowbar-evacuate-host.sh.erb
```diff
@@ -45,6 +45,19 @@ ORIGINAL_INSTANCES=$UPGRADEDIR/original_instances
 # List active instances on host before the live-migration starts
 openstack server list --insecure --all-projects --status active --host $host -f value -c ID > $ORIGINAL_INSTANCES
 
+i=0
+while [[ $i -lt 10 ]] && openstack --insecure compute service list --service nova-compute -f value -c State | grep down ; do
```
nicolasbock: You could write this as `(( $i < 10 ))` which makes it look a lot less like Fortran 77 :wink: but that's really just a minor stylistic thing :smile:.

PR #1624, opened 2018-07-03
[4.0] upgrade: Do not allow manila-share on compute nodes

jsuchome:
Non-disruptive upgrade cannot have manila shares on compute nodes.
It seems upgraded (Pike based) control plane cannot communicate with
old (Newton based) shares on (still not upgraded) compute nodes.

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -191,6 +191,7 @@ module Api
           "heat-server",
           "horizon-server",
           "manila-server",
+          "manila-share",
           "trove-server"
         ]
         ret = {}


s-t-e-v-e-n-k:
Confirmed. This is the correct solution, once I had forced the service VM back up, manila-share 3.0 on the compute hosts could not query the manila 5.0 database on the controller node.

2018-07-04 04:29:03.385 11325 ERROR oslo_db.sqlalchemy.exc_filters InternalError: (1054, u"Unknown column 'shares.consistency_group_id' in 'field list'")
2018-07-04 04:29:03.385 11325 ERROR oslo_db.sqlalchemy.exc_filters 
2018-07-04 04:29:03.402 11325 ERROR oslo_service.service [req-c2ed5e31-f641-4796-b73f-4dc8320f8719 - - - - -] Error starting thread.

jsuchome:
> I'm also unsure if this affects any requirements. ie, do we have any customers who may have a 
> deployment like this that are expecting to be able to do non-distruptive upgrades

I've updated the bug for upgrade documentation, so at least users would know about it in advance

PR #1628, opened 2018-07-09
dns: Remove duplicate records check for dns barclamp

dirkmueller:
It is valid to set the same hostname for multiple "A" records,
so we should allow it.

requests?**


dirkmueller:
this doesn't work

PR #1630, opened 2018-07-21
bind: Allow new zones configured via rndc

dirkmueller:
This is a precondition for using rndc to manage bind and also
enables designate integration. Enabling it should not be harmful
as it listens to localhost only and requires a shared secred
from (/etc/rndc.key) to be available to both the client and the server
for proper communication. And that file is only readable by root.

requests?**


nicolasbock:
For reference, this is needed for https://github.com/crowbar/crowbar-openstack/pull/1707

PR #1634, opened 2018-07-26
[4.0] upgrade: Check for the presence of correct database role

jsuchome:
Database roles were split into database-server and mysql-server.
For the purpose of the upgrade, correct database should be the one with
mysql-server role.

Requires: https://github.com/crowbar/crowbar-openstack/pull/1746

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -136,7 +136,7 @@ module Api
 
         # Check if roles important for non-disruptive upgrade are deployed in the cluster
         clustered_roles = [
-          "database-server",
+          "mysql-server",
           "rabbitmq-server",
           "keystone-server",
           "glance-server",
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -250,9 +250,8 @@ def ha_config_check
       def deployment_check
         ret = {}
         # Only allow the upgrade for mariadb setup
-        # TODO: update the search if we allow multiple proposals
-        db_node = ::Node.find("roles:database-server").first
-        ret[:wrong_sql_engine] = true if db_node[:database][:sql_engine] != "mysql"
+        db_nodes = ::Node.find("roles:mysql-server")
+        ret[:wrong_sql_engine] = true if db_nodes.empty?
```
rhafer: Even if the "mysql-server" role is assigned to a node. Shouldn't we explicitly check the value of the "sql_engine" attribute? It might still be postgreql,

We might also want to add an explicit check for the "database-server" to be unassigned to make sure there is not running postgresql somewhere anymore.

jsuchome: We did not exactly agree on exact steps of changing sql_engine - unassigning datavase-server before/during db migration. So I think once we have a clear idea here, we could adapt the prechecks again.

But for sql_engine value, I think you are right

jsuchome: Added check for sql_engine + a test

PR #1635, opened 2018-07-26
upgrade: Before stopping services, put the openvswitch agents into admin_state_down

jsuchome:
Please check https://jira.prv.suse.net/browse/SCRD-3751 for the problem description


rhafer:
Unfortunately this doesn't seem to work as a workaround for lp#1780453 at least when testing in devstack I get the very same failure behavior regardless of what the "admin_state_up" value of the ovs-agent is.

(side note, but it doesn't matter much in scope of the but we're trying to address: if we disable admin_state_up we would also need to re-enable it after the upgrade)

PR #1639, opened 2018-08-02
upgrade: enable clone_stateless_services on upgrade

Itxaka:
clone_stateless_services was introduced in order to not use pacemaker
for stateless services[0]

This alleviates the chef run as those resources do not need to be
checked/set and also they dont need any features that pacemaker
provides.

This is set as default false for new installs but if we are
upgrading we maintain the old style of having pacemaker
manage everything.
This makes no sense because during upgrade we remove all
resources so we can take advantage of that and move to the
new style which is simpler.

This also fixes some of the upgrade jobs that were failing after
the upgrade as the chef run was hitting the 900s time limit
during the "change keystone password test" by making the chef
runs shorter as we dont check/set that many pacemaker resources.

[0] https://github.com/crowbar/crowbar-ha/commit/ada1d41565c6a08a3d499f141fae7cb5c45fd052

requests?**


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -945,12 +947,23 @@ def upgrade_first_cluster_node(node, other_node)
...
         node_api.join_and_chef
         node_api.save_node_state("controller", "upgraded")
       end
 
+      def enable_clone_stateless_services(node)
+        pacemaker_role_name = node.roles.grep(/^pacemaker-config-.*/).first
+        return if pacemaker_role_name.nil?
+        pacemaker_role = RoleObject.find_role_by_name(pacemaker_role_name)
+        return if pacemaker_role.default_attributes["pacemaker"]["clone_stateless_services"]
+        pacemaker_role.default_attributes["pacemaker"]["clone_stateless_services"] = false
```
s-t-e-v-e-n-k: The method name says enable, but this sets it to false?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -945,12 +947,23 @@ def upgrade_first_cluster_node(node, other_node)
...
+        pacemaker_role.save
```
matthewoliver: So if I understand ruby correctly, this is setting 'clone_stateless_services' to false if 'clone_stateless_services' hasn't been defined. In which case isn't it disabling clone stateless services? And only when it isn't defined.
If all that is correct, then maybe the name "disable_clone_stateless_services_by_default" or something, to indicate that it disables, but only if not already defined.

crowbar_framework/app/models/api/upgrade.rb
matthewoliver: Lol, turns out I need to refresh the page before I press 'add comment'. As I was called out of my office for about an hour between writing a comment and pressing save :P

Itxaka: Indeed the name is wrong. I changed it quickly as hound complained about it being named `set_*` but the current name has no sense.

crowbar_framework/app/models/api/upgrade.rb
Itxaka: Only changes it if its set to true, but I like the name `disable_clone_stateless_services`

jsuchome:
@Itxaka what about putting it to https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/crowbar_service.rb#L313 where we already change some other defaults during the upgrade (like rabbitmq cluster config).

Additionally, you could change the value in the proposal, like @skazi0 mentioned - same way as it is done for that rabbitmq setup...

crowbar_framework/app/models/api/upgrade.rb
jsuchome: I am not sure if it is correct ... I have a default SOC7 setup where I did not edit this option, and it `clone_stateless_services` is set to true in pacemaker-config-data role.
Shouldn't we rather change it to false aways?

jsuchome:
@s-t-e-v-e-n-k , I assume your original request is not valid any more?

dirkmueller:
LGTM in principle, but this should be proposed against stable/5.0-pike rather than master..

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -342,6 +342,16 @@ def prepare_nodes_for_os_upgrade
       prop.save
     end
 
+    # change pacemaker setting to not use pacemaker for stateless services
+    pacemaker_proposals = Proposal.all.where(barclamp: "pacemaker")
+    pacemaker_proposals.each do |proposal|
+      role = proposal.role
+      proposal.raw_data["attributes"]["pacemaker"]["clone_stateless_services"] = false
```
dirkmueller: can we reference the schema default here instead of hardcoding it? might be difficult..

Itxaka:
> LGTM in principle, but this should be proposed against stable/5.0-pike rather than master..

Should we do master then backport? or has that changed?

crowbar_framework/app/models/crowbar_service.rb
Itxaka: why the schema? the schema values are default values but here we want to make clean that we want to set this value to false explicitly.

I wonder how would that make it more clear that what we have right now.

jsuchome:
and backport is here https://github.com/crowbar/crowbar-core/pull/1642

PR #1643, opened 2018-08-08
upgrade: Make sure OpenStack services are stopped

jsuchome:
Even for HA setup, they might not be managed by pacemaker


chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
```diff
@@ -70,10 +70,22 @@ done
 
 <% end %>
 
+# It's possible services are not managed by pacemaker (if clone_stateless_services was not set),
+# so to make sure that services are stopped, use additional systemctl commands.
+log "Stopping OpenStack services..."
+
+for i in /usr/sbin/rcopenstack-* ; do
```
Itxaka: cant you just move the `grep -v neutron` to this line so you get only the rest of the services?

chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
```diff
@@ -70,10 +70,22 @@ done
 
 <% end %>
 
+# It's possible services are not managed by pacemaker (if clone_stateless_services was not set),
+# so to make sure that services are stopped, use additional systemctl commands.
+log "Stopping OpenStack services..."
```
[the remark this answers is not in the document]
jsuchome: This part is limited, we do not want to shutdown everything

chef/cookbooks/crowbar/templates/default/crowbar-shutdown-services-before-upgrade.sh.erb
jsuchome: I guess I could

PR #1646, opened 2018-08-14
upgrade: disable network agents during upgrade

rhafer:
As a workaround for https://bugs.launchpad.net/neutron/+bug/1780453 we
are now setting the "ADMIN_STATE_UP" attribute to "down" before
upgrading a (network)-node. The attribute is reset to "up" after the
node was successfully upgrade and all services are running again. This
should e.g. avoid neutron/neutron-dhcp-agent to try to bring up e.g. a
dhcp port on a node where the Layer 2 agent (openvswitch-/linuxbridge-agent)
is not fully up yet.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -175,6 +175,14 @@ template "/usr/sbin/crowbar-router-migration.sh" do
   action :create
 end
 
+template "/usr/sbin/crowbar-set-network-agents-state.sh" do
+  source "crowbar-set-network-agents-state.sh.erb"
+  mode "0755"
+  owner "root"
+  group "root"
+  action :create
+end
+
 neutron_server = search(:node, "run_list_map:neutron-server").first
 template "/etc/neutron/lbaas-connection.conf" do
   source "lbaas-connection.conf"
--- a/chef/cookbooks/crowbar/templates/default/crowbar-set-network-agents-state.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-set-network-agents-state.sh.erb
@@ -1,0 +1,64 @@
+#!/bin/bash
+#
+# This script disables the l3, dhcp and l2 neutron agents on a certain
+# network node to avoid neutron to schedule any resource to this node
+# while it is being upgraded. After the upgrade completed the agents need
+# to be enabled again.
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+UPGRADEDIR=/var/lib/crowbar/upgrade
+mkdir -p "`dirname "$LOGFILE"`"
+exec >>"$LOGFILE" 2>&1
+
+hostname=$1
+mode=$2
+
+log()
+{
+    set +x
+    echo "[$(date --iso-8601=ns)] [$$] $@"
+    set -x
+}
+
+log "Executing $BASH_SOURCE"
+
+set -x
+
+mkdir -p $UPGRADEDIR
+rm -f $UPGRADEDIR/crowbar-set-network-agents-state-failed
+
+agent_order="dhcp l3 openvswitch linuxbridge"
+# Enable in the reverse order of disabling (e.g. l2-agents should to be
+# active before l3 and dhcp)
+if [[ "$mode" = "enable" ]]; then
+    agent_order="linuxbridge openvswitch l3 dhcp"
+elif [[ "$mode" != "disable" ]]; then
+    echo "Invalid mode \"$mode\" for crowbar-set-network-agents-state."
+    echo 255 > $UPGRADEDIR/crowbar-set-network-agents-state-failed
+    exit 255
+fi
+
+set +x
+source /root/.openrc
+set -x
+
+for agenttype in $agent_order; do
+    # Using "neutron" here as the "openstack" client doesn't support
+    # listing agents by type in Newton (got added with later versions)
+    id=$(/usr/bin/neutron --insecure agent-list --binary neutron-$agenttype-agent --host $hostname -c id -f value)
+    if [ -n "$id" ]; then
+        log "Setting state of $agenttype agent ($id) on node $hostname to $mode"
+        /usr/bin/openstack --insecure network agent set --$mode $id
+        ret=$?
+        if [ $ret != 0 ] ; then
+            echo "Failed to set state of $agenttype agent ($id) on host: $hostname"
+            echo $ret > $UPGRADEDIR/crowbar-set-network-agents-state-failed
+            exit $ret
+        fi
+    else
+        log "Node $hostname doesn't run $agenttype."
+    fi
+done
+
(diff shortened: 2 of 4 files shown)


rhafer:
Note: This will likely have some merge-conflict with https://github.com/crowbar/crowbar-core/pull/1643, but we can solve this once the first PR is merged.

chef/cookbooks/crowbar/templates/default/crowbar-set-network-agents-state.sh.erb
```diff
@@ -0,0 +1,64 @@
...
+fi
+
+set +x
+source /root/.openrc
+set -x
+
+for agenttype in $agent_order; do
+    # Using "neutron" here as the "openstack" client doesn't support
+    # listing agents by type in Newton (got added with later versions)
+    id=$(/usr/bin/neutron agent-list --binary neutron-$agenttype-agent --host $hostname -c id -f value)
```
jsuchome: Please add `--insecure`

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -175,6 +175,14 @@
   action :create
 end
 
+template "/usr/sbin/crowbar-set-network-agents-state.sh" do
+  source "crowbar-set-network-agents-state.sh.erb"
+  mode "0755"
+  owner "root"
+  group "root"
+  action :create
+end
```
jsuchome: Do we need to place the script on all nodes? (Not that it really hurts)

chef/cookbooks/crowbar/templates/default/crowbar-set-network-agents-state.sh.erb
```diff
@@ -0,0 +1,64 @@
...
+    if [ -n "$id" ]; then
+        log "Setting state of $agenttype agent ($id) on node $hostname to $mode"
+        /usr/bin/openstack network agent set --$mode $id
+        ret=$?
+        if [ $ret != 0 ] ; then
+            echo "Failed set state of $agenttype agent ($id) on host: $hostname"
```
jsuchome: _Failed to_ ?

chef/cookbooks/crowbar/templates/default/crowbar-set-network-agents-state.sh.erb
```diff
@@ -0,0 +1,64 @@
...
+source /root/.openrc
+set -x
+
+for agenttype in $agent_order; do
+    # Using "neutron" here as the "openstack" client doesn't support
+    # listing agents by type in Newton (got added with later versions)
+    id=$(/usr/bin/neutron agent-list --binary neutron-$agenttype-agent --host $hostname -c id -f value)
+    if [ -n "$id" ]; then
+        log "Setting state of $agenttype agent ($id) on node $hostname to $mode"
+        /usr/bin/openstack network agent set --$mode $id
```
jsuchome: `insecure` please

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1086,11 +1088,41 @@ def evacuate_network_node(controller, network_node, delete_namespaces = false)
...
+        controller.delete_script_exit_files("/usr/sbin/crowbar-set-network-agents-state.sh")
+      rescue StandardError => e
+        raise_node_upgrade_error(
+          e.message +
+          " Check /var/log/crowbar/node-upgrade.log at #{controller.name} for details."
+        )
+      end
+
+      # Re-enable neutron network agents on a just upgraded node
+      def re_enable_network_agents(controller, network_node)
```
jsuchome: Wouldn't it make sense to put this method into `api/node.rb` ?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1086,11 +1088,41 @@ def evacuate_network_node(controller, network_node, delete_namespaces = false)
...
+        save_node_action("Enabling neutron networking agents on #{hostname}")
+        args = [hostname, "enable"]
+        save_node_action("re-enabling network agents")
+        controller.wait_for_script_to_finish(
+          "/usr/sbin/crowbar-set-network-agents-state.sh",
+          timeouts[:set_network_agents_state],
+          args
+        )
+        Rails.logger.info("Enabling network agents on #{hostname} was successful.")
+        controller.delete_script_exit_files("/usr/sbin/crowbar-set-network-agents-state.sh")
```
jsuchome: Is there a reason to delete those files now? Do we still wanna do enabling/disabling again once the node is upgraded?

crowbar_framework/lib/crowbar/upgrade_timeouts.rb
```diff
@@ -31,6 +31,7 @@ def values
         chef_upgraded: @timeouts_config[:chef_upgraded] || 1200,
         router_migration: @timeouts_config[:router_migration] || 600,
         lbaas_evacuation: @timeouts_config[:lbaas_evacuation] || 600,
+        set_network_agents_state: @timeouts_config[:set_network_agents_state] || 120,
```
jsuchome: Recent experiences says: start with bigger timeout so we do not have to raise it later... :-)

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
rhafer: Not really. But the same is true for the `crowbar-router-migration.sh` script above. Both are only needed on the "controller" nodes. (Not neutron-network)

crowbar_framework/app/models/api/upgrade.rb
rhafer: Hm, this method is somewhat similar to `evacuate_network_node` in that is calls something on a Node that is different from the node that is currently being upgrade. And currently not all callers of this method have a node_api object from the target node available. If we want to move this to api/node.rb we should move `evacuate_network_node` as well.

crowbar_framework/app/models/api/upgrade.rb
rhafer: Sure. Yes we're using the same node multiple times. For enabling/disabling agents on the various network-nodes.

chef/cookbooks/crowbar/templates/default/crowbar-set-network-agents-state.sh.erb
rhafer: fixed

chef/cookbooks/crowbar/templates/default/crowbar-set-network-agents-state.sh.erb
rhafer: fixed

chef/cookbooks/crowbar/templates/default/crowbar-set-network-agents-state.sh.erb
rhafer: fixed

jsuchome:
I've merged https://github.com/crowbar/crowbar-core/pull/1643, seems like it could work (also manually it worked for me without a conflict)

PR #1650, opened 2018-08-15
ohai: fix path to check for aacraid driver (bsc#1085170 bsc#1103882)

JanZerebecki:
The check never worked before. On the 2 variants of machines I have seen
these are nested one deeper. While this may be different on different
hardware, checking every driver in the tree is more expensive.

Follow up for: a7312a8274312a0b1b0621e2578cf2290748b61c

### DIFF
--- a/chef/cookbooks/ohai/files/default/plugins/linux/block_device_extended.rb
+++ b/chef/cookbooks/ohai/files/default/plugins/linux/block_device_extended.rb
@@ -31,11 +31,11 @@ if File.exist?("/sys/block")
     parts = File.realpath(dir, "/sys/block/").split("/")
     # example:
     # /sys/devices/pci0000:00/0000:00:02.0/0000:03:00.0/host0/target0:1:4/0:1:4:0/block/sdb
-    parts = parts[0..4] + ["driver"]
+    parts = parts[0..5] + ["driver"]
     path = parts.join("/")
     # example:
-    # /sys/devices/pci0000:00/0000:00:02.0/driver
-    # -> ../../../bus/pci/drivers/aacraid
+    # /sys/devices/pci0000:00/0000:00:02.0/0000:03:00.0/driver
+    # -> ../../../../bus/pci/drivers/aacraid
     next unless File.exist?(path) && File.readlink(path).split("/")[-1] == "aacraid"
     block[dir]["removable"] = "0"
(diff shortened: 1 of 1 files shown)


chef/cookbooks/ohai/files/default/plugins/linux/block_device_extended.rb
```diff
@@ -31,11 +31,11 @@
     parts = File.realpath(dir, "/sys/block/").split("/")
     # example:
     # /sys/devices/pci0000:00/0000:00:02.0/0000:03:00.0/host0/target0:1:4/0:1:4:0/block/sdb
-    parts = parts[0..4] + ["driver"]
```
nicolasbock: There can be a driver entry in both locations:
~~~
# ll /sys/devices/pci0000:00/0000:00:02.2/
driver -> ../../../bus/pci/drivers/pcieport

# ll /sys/devices/pci0000:00/0000:00:02.2/0000:02:00.0
driver -> ../../../../bus/pci/drivers/hpsa
~~~
The driver at the deeper level seems more appropriate though.

JanZerebecki:
Yes both exist, see the comment above from Nicolas. I chose one of them (I just chose the wrong one so the previous version never worked and this fixes that). (However all location are correct, as the location depends on how the hardware is wired internally. Both and more are possible. However what this patch uses is the way that the machines we have seen are built. If we wanted to cover all ways a machine can be built we would need to check every level.)

nicolasbock:
> But are you sure both paths actually exist? Shouldn't we pick one if we no the other one is not there?

@jsuchome They both exist. It looks like the shallower driver is the PCI driver and the deeper one is the storage driver. We likely don't care about the PCI driver but presumably want to know about the storage driver. I don't know how many more layers there could be. @JanZerebecki I wonder whether we ever use this information anywhere.

JanZerebecki:
> I wonder whether we ever use this information anywhere.

Which information? On which level of the hierarchy the aacraid driver is? Not in this area of the code. No idea about totally unrelated code.

> I don't know how many more layers there could be.

I know PCIE can have one less and SATA can have at least another layer. So in both direction the hardware specs don't provide for a fixed offset.

JanZerebecki:
This patch was tested successfully on an affected machine.

PR #1651, opened 2018-08-15
upgrade: Port openstack db backup to mysql

rhafer:
This was still relying on postgresql being the database backend for openstack. It went unnoticed for so long because the default values for the postgresql backend related chef attributes pointed the upgrade     procedure to "localhost". So it created a dump of the admin node's database

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -484,6 +484,9 @@ module Api
           ::Crowbar::UpgradeStatus.new.end_step
           return
         end
+        # Note: Trying to (gu)estimate the size of the (compressed) database
+        #       SQL dump from the sizes report via the below query is probably a
+        #       bit far fetched. Is there a more realistic way for this?
         query = "SELECT SUM(pg_database_size(pg_database.datname)) FROM pg_database;"
         cmd = "PGPASSWORD=#{psql[:pass]} psql -t -h #{psql[:host]} -U #{psql[:user]} -c '#{query}'"
 
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -477,27 +477,34 @@ module Api
           return
         end
 
-        psql = postgres_params
-        if psql.nil?
+        db_node = openstack_db_node
+        if db_node.nil?
           # This can happen if only the crowbar node was deployed and will be upgraded
           Rails.logger.warn("Can not get database parameters for OpenStack backup. Skipping...")
           ::Crowbar::UpgradeStatus.new.end_step
           return
         end
+
+        db_user = "root"
+        db_password = db_node["database"]["mysql"]["server_root_password"]
+
         # Note: Trying to (gu)estimate the size of the (compressed) database
         #       SQL dump from the sizes report via the below query is probably a
         #       bit far fetched. Is there a more realistic way for this?
-        query = "SELECT SUM(pg_database_size(pg_database.datname)) FROM pg_database;"
-        cmd = "PGPASSWORD=#{psql[:pass]} psql -t -h #{psql[:host]} -U #{psql[:user]} -c '#{query}'"
-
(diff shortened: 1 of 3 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -507,15 +517,20 @@ def openstackbackup
...
         Rails.logger.debug("Creating OpenStack database dump")
-        db_dump = run_cmd(
-          "PGPASSWORD=#{psql[:pass]} pg_dumpall -h #{psql[:host]} -U #{psql[:user]} | " \
-            "gzip > #{dump_path}"
+
+        # Note: We need to run this on a database node since the mysql root user
+        #       doesn't have remote access to the mysql server. But we can't
+        #       use Node.run_ssh_cmd her because we want to redirect to a file
+        #       on the crowbar node.
+        db_dump = run_cmd("sudo ssh -o ConnectTimeout=10 root@#{db_node.name} " \
```
jsuchome: Can't you use `run_ssh_cmd` here?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -507,15 +517,20 @@ def openstackbackup
...
         end
 
         Rails.logger.debug("Creating OpenStack database dump")
-        db_dump = run_cmd(
-          "PGPASSWORD=#{psql[:pass]} pg_dumpall -h #{psql[:host]} -U #{psql[:user]} | " \
-            "gzip > #{dump_path}"
+
+        # Note: We need to run this on a database node since the mysql root user
+        #       doesn't have remote access to the mysql server. But we can't
+        #       use Node.run_ssh_cmd her because we want to redirect to a file
```
jsuchome: Typo _her_

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1793,17 +1808,13 @@ def admin_architecture
       #
       # openstackbackup helpers
       #
-      def postgres_params
+      def openstack_db_node
         db_node = ::Node.find("roles:database-config-default").first
```
jsuchome: Maybe you can already check if there's a node with mysql-server role, otherwise we'd need to change this code later

rhafer: @jsuchome Why? The nodes with the "mysql-server" role will also have the "database-config-default" role. And making sure that the database-cluster is migrated to "mysql" should be part of the prechecks, shouldn't it?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -507,15 +517,20 @@ def openstackbackup
...
-        db_dump = run_cmd(
-          "PGPASSWORD=#{psql[:pass]} pg_dumpall -h #{psql[:host]} -U #{psql[:user]} | " \
-            "gzip > #{dump_path}"
+
+        # Note: We need to run this on a database node since the mysql root user
+        #       doesn't have remote access to the mysql server. But we can't
+        #       use Node.run_ssh_cmd her because we want to redirect to a file
+        #       on the crowbar node.
+        db_dump = run_cmd("sudo ssh -o ConnectTimeout=10 root@#{db_node.name} " \
+                          "\'mysqldump -u root -pkQYaTi7yWzS3 --all-databases | gzip\' " \
```
jsuchome: Ouch, you have your password here

rhafer: Ouch indeed ...

crowbar_framework/app/models/api/upgrade.rb
jsuchome: Ah, I missed that it's database-config-default and not database-server role you're looking for.

crowbar_framework/app/models/api/upgrade.rb
rhafer: fixed

crowbar_framework/app/models/api/upgrade.rb
rhafer: fixed

PR #1653, opened 2018-08-20
[5.0] upgrade: Remove nova-cert service after the node upgrade

jsuchome:
nova-cert service got deprecated and needs to be removed when upgrading
to Pike. Let's do the removal after the node is fully upgraded and
service's binary is not present, so its accidental (re)start does not cause
recreating deleted nova-cert.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-chef-upgraded.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-chef-upgraded.sh.erb
@@ -45,5 +45,19 @@ rm -f /usr/sbin/crowbar_join_tmp
 systemctl start cron
 systemctl enable cron
 
+# Remove nova-cert service which has been deprecated in Newton and is not present in Pike
+<% if @nova_controller %>
+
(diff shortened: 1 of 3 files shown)


chef/cookbooks/crowbar/templates/default/crowbar-chef-upgraded.sh.erb
```diff
@@ -45,5 +45,19 @@ rm -f /usr/sbin/crowbar_join_tmp
 systemctl start cron
 systemctl enable cron
 
+# Remove nova-cert service which has been deprecated in Newton and is not present in Pike
+<% if @nova_controller %>
+
```
rhafer: This of course means we're now running this on every single nova-controller, while it would be enough to just run it once. But I guess we can live with that.

jsuchome: Yes, I know, I wanted to be sure :-)
And if no service is found, no deleting will happen

PR #1661, opened 2018-09-18
[5.0] upgrade: Upgrade former DRBD-cluster correctly

jsuchome:
posgresql and drbd resources no longer need to be excluded from
the list of services that are going to be shut down.

Neither of those is supported in SOC8 and neither of them should
be used by SOC7 when the upgrade is started (if there was PostgreSQL
setup initially, it had to be migrated to MariaDB).


jsuchome:
@rhafer indeed. Some kind of tear down is actually handled here - at least deleting and not recreating pacemaker resources.
That does not say anything about the stored data, of course. Currently we suggest user to do it manually (see Stop/Uninstall psql in https://etherpad.nue.suse.com/p/cloud-upgrade-so7-to-socc8-preparations)

PR #1672, opened 2018-10-05
network: Limit number of openvswitch threads per port (bsc#1110865)

dirkmueller:
Open vSwitch older than 2.11 is starting 2*$(nproc)*3/4 netlink sockets per
number of cpu cores and per port. On machines with high core
count (e.g. 56 or higher) this can quickly exceed the total fd
limit set by ovs on itself on startup of 65536.

By Limit the handler threads to 8 per port we can handle ~ 4000 ports
instead of just ~ 700.
see https://mail.openvswitch.org/pipermail/ovs-dev/2018-September/352402.html

### DIFF
--- a/chef/cookbooks/network/attributes/default.rb
+++ b/chef/cookbooks/network/attributes/default.rb
@@ -66,3 +66,14 @@ default[:network][:ovs_module] = "openvswitch"
 # This flag be overridden on the node/role level (e.g. by the neutron
 # barclamp) to indicate that a node needs openvswitch installed and running
 default[:network][:needs_openvswitch] = false
+
+# Open vSwitch older than 2.11 is starting 2*$(nproc)*3/4 netlink sockets per
+# number of cpu cores and per port. On machines with high core
+# count (e.g. 56 or higher) this can quickly exceed the total fd
+# limit set by ovs on itself on startup of 65536.
+
+# By Limit the handler threads to 8 per port we can handle ~ 4000 ports
+# instead of just ~ 700.
+# see https://mail.openvswitch.org/pipermail/ovs-dev/2018-September/352402.html
+# see bsc#1110865
+default[:network][:ovs_max_handler_threads] = 8
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -44,6 +44,10 @@ if node[:network][:needs_openvswitch]
   s.run_action :enable
   s.run_action :start
 
+  # limit the number of handler threads to avoid exceeding
+  # hardcoded 65536 limit in the ovs-start script (bsc##1110865)
+  ::Kernel.system("ovs-vsctl set Open_vSwitch . other_config:n-handler-threads=#{node[:network][:ovs_max_handler_threads]}")
+
   # Cleanup on SLE12. Disable (NOT stop) old sysvinit service for ovs to avoid
   # issues (https://bugzilla.suse.com/show_bug.cgi?id=935912). We use the
   # (differently named) systemd unit on newer suse platforms now.


dirkmueller:
(needs testing)

aojea:
My concern is if we should fix it in crowbar only or in the openvswitch-package.
I'm more inclined to the later but don't have a strong opinion, any thoughts @hwoarang @rossella ?

hwoarang:
I have no strong opinion either.

chef/cookbooks/network/templates/default/ovs-pre-up.sh.erb
```diff
@@ -12,3 +12,17 @@ ovs-vsctl del-fail-mode <%= @bridgename %>
...
+# count (e.g. 56 or higher) this can quickly exceed the total fd
+# limit set by ovs on itself on startup of 65536.
+
+# By Limit the handler threads to 8 per port we can handle ~ 4000 ports
+# instead of just ~ 700.
+# see https://mail.openvswitch.org/pipermail/ovs-dev/2018-September/352402.html
+# see bsc#1110865
+max_handler_threads=8
+if [ $(nproc) -gt $max_handler_threads ]; then
+    ovs-vsctl set Open_vSwitch $(ovs-vsctl show | head -n 1) other_config:n-handler-threads=$max_handler_threads
```
aojea: I'd prefer to take an approach as in line 4 instead of relying on bash commands

dirkmueller: Any idea how to retrieve the Open_vSwitch id other than show | head -n 1 ? I tried several things, and was even desparate enough to grep the documentation and couldn't find a hint.

hwoarang: @dirkmueller 
`ovs-vsctl get Open_vSwitch . _uuid`

aojea: didn't want to nitcpick sorry, just thought that there was something like `<%= @datapath_id %>` already that we can leverage

rossella:
Cool finding Dirk!
About @aojea's questions, I think he has a valid point, we might incur in the same problem when using ovs in a different context (ardana, caasp, etc.). It's worth fixing it in the package.

chef/cookbooks/network/templates/default/ovs-pre-up.sh.erb
rhafer: I think you could just use

`ovs-vsctl set Open_vSwitch . other_config:n-handler-threads=$max_handler_threads`

chef/cookbooks/network/templates/default/ovs-pre-up.sh.erb
```diff
@@ -12,3 +12,17 @@ ovs-vsctl del-fail-mode <%= @bridgename %>
...
+# Open vSwitch older than 2.11 is starting 2*$(nproc)*3/4 netlink sockets per
+# number of cpu cores and per port. On machines with high core
+# count (e.g. 56 or higher) this can quickly exceed the total fd
+# limit set by ovs on itself on startup of 65536.
+
+# By Limit the handler threads to 8 per port we can handle ~ 4000 ports
+# instead of just ~ 700.
+# see https://mail.openvswitch.org/pipermail/ovs-dev/2018-September/352402.html
+# see bsc#1110865
+max_handler_threads=8
```
rhafer: Should we really hard code this here? I guess we'd better make this a chef attribute (not a proposal attribute just default attribute defined in the recipe that can be overridden node the node level)  and pass via the template.

dirkmueller:
@rossella I agree that a fix in SLES would b epreferable, but SLES packages can't set config tunables I guess.. the SLES fix would be the epoll backport (which I tested successfully in the bug that is referenced) and that is a magnitude better than this "hack" in the scalability front

dirkmueller:
For SOC7 the OVS backport included the socket handling fixes. need to check if that is fixed for soc8+ as well, then we can close this.

jdsn:
Closing now as it blocks our CI currently.
Please feel free to reopen when you push a new commit to this PR or force-push or rebase.
The latest commit needs to change as it reached the maximum number of status values (1000).

PR #1674, opened 2018-10-05
Remove SES references

rsalevsky:
SES droped Crowbar support some time ago, so let's remove the old
references.


crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -26,11 +26,7 @@ def load!
 
         etc_yml = "/etc/crowbar/repos.yml"
 
-        if Crowbar::Product.is_ses?
-          @all_repos = YAML.load_file(Rails.root.join("config/repos-ses.yml"))
```
dirkmueller: remove the file as well?

rsalevsky: Done

PR #1675, opened 2018-10-08
Fix Crowbar Version 6.0

rsalevsky:


rsalevsky:
Travis Passed so merging it

PR #1676, opened 2018-10-11
crowbar: move skip unready nodes out of experimental

Itxaka:
Looks like this feature has been tested extensively on prod envs
with little to no failures so its time to move it out of the
experimental umbrella, make a normal config for it and document
it properly.

 * Creates a new config file (crowbar.yml) for normal config switches
 * Moves the config settings from the experimental to the normal config
 * Changes the calls to experimental settings to point tot he new file

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -694,11 +694,11 @@ class ServiceObject
   def validate_postponed_nodes(proposal)
     return unless upgrade_postponed?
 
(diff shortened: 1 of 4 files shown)


Itxaka:
[Move "Do not require all nodes to be ready for applying proposals" feature out of experimental](https://trello.com/c/ANEIYAIL/328-move-do-not-require-all-nodes-to-be-ready-for-applying-proposals-feature-out-of-experimental)

PR #1679, opened 2018-10-17
[4.0] upgrade: Exclude removed nodes from upgrade

skazi0:
When some node is removed from crowbar (e.g. using "forget" operation) and
not all proposals are saved, this node was still included in some parts of
the upgrade process causing errors when trying to access node attributes.

Additional filtering was added to make sure we don't try to upgrade
non-existent nodes.

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -673,7 +673,10 @@ module Api
             end
           end
         end
-        to_upgrade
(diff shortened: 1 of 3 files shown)


skazi0:
@Itxaka and now they are fixed :o)

jsuchome:
So ... non-disruptive path is not affected by such (IMHO broken) setup?

skazi0:
@jsuchome true, looking at the code, this seems to affect only disruptive option.

skazi0:
Closed in favor of https://github.com/crowbar/crowbar-core/pull/1684

PR #1684, opened 2018-10-18
upgrade: Precheck for unsaved proposals

skazi0:
If some nodes are removed from crowbar (e.g. using "forget" function),
they could still be present in some proposals' role assignments.
In such cases the upgrade process can fail later when trying to access
attributes of non-existing nodes.


crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -96,6 +96,10 @@ def health_check
         ret[:nodes_not_ready] = unready unless unready.empty?
         failed = Proposal.all.select { |p| p.active? && p.failed? }
         ret[:failed_proposals] = failed.map(&:display_name) unless failed.empty?
+        # Not using Node.all here as we only need node names
+        all_node_names = ChefObject.fetch_nodes_from_cdb.first.map(&:name)
+        unsaved = Proposal.all.reject { |p| (p.all_elements - all_node_names).empty? }
```
jsuchome: Can't we actually save the result of previous Proposal.all so we don't have to do the same search again?

skazi0:
tests to be updated

crowbar_framework/app/models/api/crowbar.rb
skazi0: done

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -94,8 +94,13 @@ def health_check
           unready << node.name unless node.ready?
         end
         ret[:nodes_not_ready] = unready unless unready.empty?
-        failed = Proposal.all.select { |p| p.active? && p.failed? }
+        proposals = Proposal.all
+        failed = proposals.select { |p| p.active? && p.failed? }
         ret[:failed_proposals] = failed.map(&:display_name) unless failed.empty?
+        # Not using Node.all here as we only need node names
+        all_node_names = ChefObject.fetch_nodes_from_cdb.first.map(&:name)
+        unsaved = proposals.reject { |p| (p.all_elements - all_node_names).empty? }
```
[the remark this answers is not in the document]
skazi0: I got confused by the code which I'm trying to fix with this change (in upgrade). The roles are used for proper ordering of nodes there.

skazi0:
@jsuchome do we need forward ports for this?

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -801,6 +801,9 @@ en:
         failed_proposals:
           error: 'These proposals report deployment failures: %{proposals}.'
           help: 'Check the state of the specific barclamps and redeploy them correctly or disable failing proposals.'
+        unsaved_proposals:
+          error: 'These proposals contain nodes which were deleted: %{proposals}.'
```
jsuchome: Are we sure that the case is only "nodes that were deleted" ? I think it might be safer to say something like "which are not known to crowbar"...

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -94,8 +94,15 @@ def health_check
...
-        failed = Proposal.all.select { |p| p.active? && p.failed? }
+        proposals = Proposal.all
+        failed = proposals.select { |p| p.active? && p.failed? }
         ret[:failed_proposals] = failed.map(&:display_name) unless failed.empty?
+        # Not using Node.all here as we only need node names
+        nodes = ChefObject.fetch_nodes_from_cdb.first.map(&:name)
+        clusters = ServiceObject.available_clusters.keys
+        valid_elements = nodes + clusters
+        unsaved = proposals.reject { |p| (p.elements.values.flatten - valid_elements).empty? }
+        ret[:unsaved_proposals] = unsaved.map(&:display_name) unless unsaved.empty?
```
jsuchome: What about showing the problematic value(s) ( = unknown nodes) as well?

PR #1685, opened 2018-10-19
crowbar: move disallow chef restarts out of experimental

ilausuch:
Looks like this feature has been tested extensively on prod envs
with little to no failures so its time to move it out of the
experimental umbrella, make a normal config for it and document
it properly.

Moves the config settings from the experimental to the normal config
Changes the calls to experimental settings to point to the new file


crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -781,6 +781,7 @@ en:
     error:
       destroy_failed: 'Failed to destroy %{component}'
     crowbar:
+      disabled: 'This API is currently disabled in crowbar.yml config file (%{option} option).'
```
Itxaka: should remove this text from the eperimental part in line 780, probably also remove the whole experimental section starting on line 779 as we have no longer text that is part of experimental.

Itxaka: and probably use a more descriptive name for the key as `disabled` is really generic but it worked under the experimental key. Maybe `disallow_chef_restart_disabled`?

ilausuch: I agree

ilausuch:
Removed experimental.yml won't be necessary

crowbar_framework/config/application.rb
```diff
@@ -65,7 +65,5 @@ class Application < Rails::Application
     end
     # normal options
     config.crowbar = config_for(:crowbar)
-    # experimental options
```
skazi0: is this the last experimental option we have?

Itxaka: yup, all others have been moving out of experimental and are now normal features :)

PR #1695, opened 2018-10-25
[4.0] upgrade: Skip powered off compute nodes during the upgrade prechecks

jsuchome:


crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -91,7 +91,12 @@ def health_check
         unready = []
         # We are ignoring the ceph nodes, as they should already be in crowbar_upgrade state
         NodeObject.find("NOT roles:ceph-*").each do |node|
-          unready << node.name unless node.ready?
+          next if node.ready?
+          if node.state == "shutdown" && node.roles.include?("nova-compute-kvm")
```
skazi0: do we want this only for kvm?

jsuchome: I think we support only kvm compute nodes during the upgrade ... but as this is ECP specific anyway - are there any different ones?

skazi0: True, ECP only has KVM.

jsuchome:
Closing in favor of https://github.com/crowbar/crowbar-core/pull/1749

PR #1698, opened 2018-10-31
Add ping support for IPv6

matthewoliver:
Currently the SLE12 iptuils package is using iputils 20121221 which
has separate ping and ping6 binaries. Which is fine, but the ping
command only works with ipv4, even if given a FQDN the qaulifies to
ipv6.

SLE15's iputils is using 20161105, which has both ipv4 and ipv6 in
support in the ping command.

This change modifies the check to test with ping6 if ping fails, so
the check will work on legecy systems.

The other alternative is to update SLE12's iputils package which
may be a better alternative, but this change bypasses the problem.

### DIFF
--- a/crowbar_framework/lib/crowbar/checks/network.rb
+++ b/crowbar_framework/lib/crowbar/checks/network.rb
@@ -60,16 +60,17 @@ module Crowbar
           unless ipv4_configured || ipv4_addrs.empty?
             return false
           end
-          # we don't really depend on IPv6, so no big deal
-          # unless ipv6_configured || ipv6_addrs.empty?
-          #   return false
-          # end
+
+          unless ipv6_configured || ipv6_addrs.empty?
+            return false
+          end
 
           true
         end
 
         def ping_succeeds?
-          system("ping -c 1 #{fqdn} > /dev/null 2>&1 || ping6 -c 1 #{fqdn} > /dev/null 2>&1")
+          return system("ping6 -c 1 #{fqdn} > /dev/null 2>&1") unless ipv6_addrs.empty?
+          system("ping -c 1 #{fqdn} > /dev/null 2>&1")
         end
 
         def hostname
--- a/crowbar_framework/lib/crowbar/checks/network.rb
+++ b/crowbar_framework/lib/crowbar/checks/network.rb
@@ -104,7 +104,9 @@ module Crowbar
               else
                 next unless ip_addr.ipv4?
               end
-              addresses.push address
+              # Sanitise the ip string, as ipv6 writing shortcuts can make
+              # string comparisons fail.
+              addresses.push ip_addr.to_s
             end
           end
         end
--- a/crowbar_framework/lib/crowbar/checks/network.rb
+++ b/crowbar_framework/lib/crowbar/checks/network.rb
@@ -69,7 +69,7 @@ module Crowbar
         end
 
         def ping_succeeds?
-          system("ping -c 1 #{fqdn} > /dev/null 2>&1")
+          system("ping -c 1 #{fqdn} > /dev/null 2>&1 || ping6 -c 1 #{fqdn} > /dev/null 2>&1")
         end
 
         def hostname
--- a/crowbar_framework/spec/lib/crowbar/checks/network_spec.rb
+++ b/crowbar_framework/spec/lib/crowbar/checks/network_spec.rb
@@ -71,7 +71,15 @@ describe Crowbar::Checks::Network do
     it "returns true if fqdn is pingable" do
       allow_any_instance_of(Kernel).to(
         receive(:system).with(
-          "ping -c 1 #{subject.fqdn} > /dev/null 2>&1 || "\
+          "ping -c 1 #{subject.fqdn} > /dev/null 2>&1"
+        ).and_return(true)
+      )
+      expect(subject.ping_succeeds?).to be true
+    end
+    it "returns true if fqdn is pingable and ipv6" do
+      allow(subject).to receive(:ipv6_addrs).and_return(["fd00:0:0:3::1"])
+      allow_any_instance_of(Kernel).to(
+        receive(:system).with(
           "ping6 -c 1 #{subject.fqdn} > /dev/null 2>&1"
         ).and_return(true)
       )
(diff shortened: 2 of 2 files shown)


barclaac:
Could this get you into a situation where a host has operable IPv4 associated to the FQDN but maybe the IPv6 is specified against the FQDN but not reachable? What should the correct action be in that case? With the patch as written things would succeed but should they?

rktidwell:
@barclaac That's a really good question. It does seem like there is some special-casing for handling IPv6 that is required. This change unblocks IPv6 development. I wonder if it is safe to merge this, then come back and really polish this up later as we teach crowbar about IPv6 more broadly.

matthewoliver:
True, the other option is to upgrade the iputils package, to be the same as SLE15. Where the ping works on both version 4 and 6.. but then don't we have the same problem? Though I guess we cloud assume it'll do a AAAA first. 

In this patch, it'll attempt an IPv4 ping and only if that fails try the ipv6. So if it had a working ipv4, great. If you expect the ipv4 but it fails but succeeds on ipv6 then, well you asked to see if it's pingable via FQDN and it is (or visa versa). If that is a problem maybe doing a DNS looking and pinging the required ip would be better, but not what we are trying to do in this function.. well at least that's my opinion.

All I know is, currently we _can't_ ping a ipv6 FQDN currently, which actually stops us deploying a single stack ipv6 crowbar admin node. So this unblocks us so we can continue to try and get some ipv6 mkcloud testing in place.

barclaac:
I would definitely merge this as it's almost certainly fit for purpose now as you need it to be "good enough." You may want to capture that it's not a complete solution for later. The true case is likely "if (A record || ipv4 numeric) then pingv4 _and_ if (AAAA record || ipv6 numeric) then pingv6", which is saying that if you're dumb enough to define a DNS entry then at least have the decency to make sure it's correct and reachable. For this the new iputils from SLE15 would be insufficient because it's still a v4 or v6 test and you potentially want both.

bmwiedemann:
I'd also wish for the commit message to explain that the goal is to support nodes with only IPv6 addrs.
Dual-stack deployments should already work without the patch.

matthewoliver:
The other option as I have already mentioned is to upgrade the iputils package in SLES 12 to that of SLES15, as that iputils version (from 2016 I think), supports both ipv4 and ipv6 in the single ping binary.

But sure I'll have a play and also update the commit message

crowbar_framework/lib/crowbar/checks/network.rb
```diff
@@ -69,7 +69,7 @@ def ip_configured?
         end
 
         def ping_succeeds?
-          system("ping -c 1 #{fqdn} > /dev/null 2>&1")
+          system("ping -c 1 #{fqdn} > /dev/null 2>&1 || ping6 -c 1 #{fqdn} > /dev/null 2>&1")
```
rhafer: This goes into a similar direction to what @dirkmueller already wrote:

Shouldn't we just call `ping` vs. `ping6` depending on the return values of the `ipv4_addrs` and `ipv6_addrs` methods (see line 87 and following)? I.e. call ping6 if just ipv6_addrs is not empty, call ping if only ipv4_addrs is not empty.

(Call both if this is a dual stack setup?)

matthewoliver: ahh, cool, thanks Ralf. Let me look at that. Thanks, the gut's of crowbar (and ruby) is still rather new to me :)

crowbar_framework/spec/lib/crowbar/checks/network_spec.rb
```diff
@@ -71,7 +71,15 @@
...
       allow_any_instance_of(Kernel).to(
         receive(:system).with(
-          "ping -c 1 #{subject.fqdn} > /dev/null 2>&1 || "\
+          "ping -c 1 #{subject.fqdn} > /dev/null 2>&1"
+        ).and_return(true)
+      )
+      expect(subject.ping_succeeds?).to be true
+    end
+    it "returns true if fqdn is pingable and ipv6" do
+      allow(subject).to receive(:ipv6_addrs).and_return(["fd00:0:0:3::1"])
```
bmwiedemann: What is the meaning of this IPv6 addr here? I hope we dont actually use these, because the 40 bits following "FD" are supposed to be random.

bmwiedemann:
Yes, the localhost addr `::1` might be better.

matthewoliver:
@rhafer to be honest I've been trying to avoid a "this is ipv6 setup" flag in crowbar. Because it feels then  ipv6 is treated as a second class citizen, is there a "this is an ipv4 setup" flag? I'd rather crowbar to say cool an IP I can deal with this. But maybe I'm to idealistic :) 

Seeing as I haven't gotten far enough along yet to start worrying potentially about ra's and will we potentually need radvd deployed with crowbar etc.

Currently I'm working my way though a mkcloud deployment getting through each step so I can set up some testing around it all. Then can take a look at the big picture. As frankly I don't have enough crowbar knowledge or enough of the picture in my head yet. You probably do, so you'll probably end up being right. ;)

matthewoliver:
@bmwiedemann Changed to ::1

crowbar_framework/lib/crowbar/checks/network.rb
```diff
@@ -60,15 +60,16 @@ def ip_configured?
           unless ipv4_configured || ipv4_addrs.empty?
             return false
           end
-          # we don't really depend on IPv6, so no big deal
-          # unless ipv6_configured || ipv6_addrs.empty?
-          #   return false
-          # end
+
+          unless ipv6_configured || ipv6_addrs.empty?
```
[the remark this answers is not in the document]
rhafer: Hm, I guess hound wants you to write `return false unless ipv6_configured || ipv6_addrs.empty?`. I thing I'd rather keep what you have here for now as it matches what is used elsewhere throughout this file. (If at all, this should be addressed in a separate change)

crowbar_framework/lib/crowbar/checks/network.rb
```diff
@@ -60,16 +60,17 @@ def ip_configured?
...
+            return false
+          end
 
           true
         end
 
         def ping_succeeds?
-          system("ping -c 1 #{fqdn} > /dev/null 2>&1 || ping6 -c 1 #{fqdn} > /dev/null 2>&1")
+          return system("ping6 -c 1 #{fqdn} > /dev/null 2>&1") unless ipv6_addrs.empty?
```
toabctl: So if a IPv6 ping fails, we do not even try the IPv4 ping? Is that really what we want?

crowbar_framework/spec/lib/crowbar/checks/network_spec.rb
toabctl: @bmwiedemann This is in the _spec.rb file which are unittests afaik

PR #1699, opened 2018-10-31
assign host to hostless keys

sjamgade:
"nodename" in keys is just a comment; so create a dummy because in case
of no(empty) comment 'keys.sort' in the template fails because in ruby 'nil'
does not have any order

### DIFF
--- a/chef/cookbooks/provisioner/recipes/keys.rb
+++ b/chef/cookbooks/provisioner/recipes/keys.rb
@@ -37,10 +37,11 @@ root_pub_key = `cat /root/.ssh/id_rsa.pub`.chomp
 access_keys[node.name] = root_pub_key
 
 # Add additional keys
+nullnodecounter = 0
 node["provisioner"]["access_keys"].strip.split("\n").each do |key|
   key.strip!
   unless key.empty?
-    nodename = key.split(" ")[2]
+    nodename = key.split(" ")[2] || "hostless-key-#{nullnodecounter += 1}"
     access_keys[nodename] = key
   end
 end


sjamgade:
```
╰─ tail *
==> t.erb <==
<% a.sort.each do |k,v| %>
<%= v %>
<% end %>

==> test.rb <==
require 'erb'
a  = {"one"=>111, "two"=>222, nil=>3333}
ERB.new(File.read('t.erb')).result(binding)
╰─ irb test.rb
test.rb(main):001:0> require 'erb'
=> true
test.rb(main):002:0> a  = {"one"=>111, "two"=>222, nil=>3333}
=> {"one"=>111, "two"=>222, nil=>3333}
test.rb(main):003:0> ERB.new(File.read('t.erb')).result(binding)
ArgumentError: comparison of Array with Array failed
	from (erb):1:in `sort'
	from (erb):1
	from /usr/lib64/ruby/2.1.0/erb.rb:850:in `eval'
	from /usr/lib64/ruby/2.1.0/erb.rb:850:in `result'
	from test.rb:3
	from /usr/bin/irb:11:in `<main>'
test.rb(main):004:0>

File: chef/cookbooks/provisioner/recipes/keys.rb [hunk 1]
```diff
@@ -37,10 +37,11 @@
 access_keys[node.name] = root_pub_key
 
 # Add additional keys
+nullnodecounter = 0
 node["provisioner"]["access_keys"].strip.split("\n").each do |key|
   key.strip!
   unless key.empty?
-    nodename = key.split(" ")[2]
+    nodename = key.split(" ")[2] || "hostless-key-#{nullnodecounter += 1}-on-admin"
```
sjamgade:
```ruby
irb(main):043:0> a = 0
=> 0
irb(main):044:0> n = "in#{a += 1}side"
=> "in1side"
irb(main):045:0> n = "in#{a += 1}side"
=> "in2side"
irb(main):046:0> n = "in#{a += 1}side"
=> "in3side"
irb(main):047:0> n = "in#{a += 1}side"
=> "in4side"
irb(main):048:0> a
=> 4
irb(main):049:0> 

```

skazi0: What do you mean by the "on-admin" part? Those keys come from the provisioner proposal, right?

sjamgade: I was under the impression that `provisioner` barclamp picked up keys from the admin node, but after you pointed, I could not find the corresponding code. And now I wonder how did that key end up getting into my deployment which led me to this fix.

So "on-admin" then does not really make sense. I will remove that, but do you think I also change "hostless" to "comment-less"

skazi0: I think "hostless" is as good as anything else here :o)

sjamgade:
@skazi0 could you please re-review ?

PR #1702, opened 2018-11-06
upgrade: Introduce parallelism into the upgrade of compute nodes

jsuchome:
Originally we upgraded compute nodes one by one, for non-disruptive mode.

But it might be possible to live-evacuate more than one compute node at a time.
In such case we can upgrade all the nodes that currently do not host
running instance.


jsuchome:
From `/usr/sbin/crowbar-evacuate-host.sh` script

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1351,10 +1354,34 @@ def upgrade_compute_nodes(virt)
...
+        while compute_nodes.any?
+          # 1. Evacuate as many as possible, create a subset of nodes without instances.
+          nodes_to_upgrade = []
+          compute_nodes.each do |node|
+            hostname = node[:hostname]
+            begin
+              save_nodes_state(nodes_to_upgrade + [node], "compute", "upgrading")
+              live_evacuate_compute_node(controller, hostname)
+              nodes_to_upgrade << node
+            rescue StandardError
```
jsuchome: Well, this is problematic. If `live_evacuate_compute_node` fails for some other reason than we expect (= no free host) user may never find out.

jsuchome:
If the OS installation script fails, nova-compute services are stopped. They should be started with chef-client run, but if we just repeat the step it will go to live-evacuation which is unable to proceed as there's a check ensuring all nova-compute services are running...

jsuchome:
Another issue is that I got multiple running `/usr/sbin/crowbar-evacuate-host.sh` scripts at the same time, which should not happen... there might be something wrong in the way we execute those...

crowbar_framework/app/models/api/upgrade.rb
jsuchome: I've tried to address this in later commits

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1364,6 +1401,58 @@ def upgrade_compute_nodes(virt)
...
+          # 1. Evacuate as many as possible, create a subset of nodes without instances.
+          nodes_to_upgrade = []
+          already_upgrading = false
+          compute_nodes.each do |node|
+            # if nodes are already in upgrading state, we can skip the live-migration part
+            already_upgrading = true if node.upgrading?
+            break if already_upgrading && !node.upgrading?
+            hostname = node[:hostname]
+            begin
+              live_evacuate_compute_node(controller, hostname) unless already_upgrading
```
rhafer: Isn't there the "risk" that this will move instances to a host that is also supposed to be upgraded within the current batch? I think the easiest way to address this to require the operator to manually disable and evacuate the compute nodes that are about to be upgraded.

Alternatively we could disable the whole batch of computes before actually starting the evacuation. To avoid instances being move around too many times.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1364,6 +1401,58 @@ def upgrade_compute_nodes(virt)
...
+              nodes_to_upgrade << node
+            rescue StandardError => e
+              # We could safely ignore the error when the execution failed with a timeout,
+              # raise it otherwise.
+              failed_file = "/var/lib/crowbar/upgrade/crowbar-evacuate-host-failed"
+              raise e if controller.file_exist? failed_file
+              Rails.logger.info("attempt to live evacuate #{hostname} took too long, exiting loop")
+              # We need to enable nova-service on the failed node again so its compute powers could
+              # be used when upgrading next bunch of compute nodes
+              enable_compute_service(controller, node, true)
```
rhafer: Is this a good idea? We don't exactly know the evacuation failed on this node. Is it smart to put it back into the pool?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1316,21 +1322,52 @@ def parallel_upgrade_compute_nodes(compute_nodes)
             "Error while upgrading compute nodes. " + e.message
           )
         end
-        # FIXME: should we paralelize this as well?
+        # reboot block (reboot could be just initiated)
```
rhafer: What do you mean with "reboot could be just initiated" ?

jsuchome: I just mean that we do not need to wait for the node to be back online (as it seems to be confusing, I should reword it)

crowbar_framework/app/models/api/upgrade.rb
jsuchome: That's indeed problematic, I know. However I'm trying to rule out the case of an error on lines 1436-1437. If script itself found an error that it should have created the failed file.

If there is no failed file, than the script was killed by timeout which (likely, I know) means that live evacuation failed because there was no other host available.

Again, I know I cannot be 100% sure about this. But _not_ putting the node back into the pool could bring a risk of having increased number of nodes with nova-compute service disabled which would reduce a chance of live-migration in the next iteration.

crowbar_framework/app/models/api/upgrade.rb
jsuchome: The risk is there, indeed. But as I stated elsewhere, this methods is only used for nodes that are explicitly selected by operator. So we could really require them (by documentation maybe) to manually disable (or even evacuate, but that's not necessary) the services before the action.

jsuchome: There is also something to the other idea, yes. Potential problem is that the set might be too big and evacuation of all of the host might not be possible if all of them are disabled (same problem as in the discussion of the other comment)

rhafer: > There is also something to the other idea, yes. Potential problem is that the set might be too big and evacuation of all of the host might not be possible if all of them are disabled

This is also true for the first idea. That's why I suggested to make this the operator's "problem" (including the evacuation). He should know much better than our code what's going on.

susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -224,7 +229,7 @@ def nodes
         errors: {
           nodes: {
             data: "No component parameter has been specified. " \
-              "Pass 'all', 'controllers' or a node name for upgrade actions. " \
+              "Pass 'all', 'controllers' or a node name(s) for upgrade actions. " \
```
skazi0: please add info about separators (and quoting for space)

PR #1703, opened 2018-11-09
[5.0] upgrade: If pacemaker is used to manage the services, don't use systemctl

jsuchome:
Usage of "systemctl stop" for shutting down the services should be limited
to the case when pacemaker is not managing those services, i.e. clone_stateless_services
is set to false. However, we're changing the value of this variable during the upgrade
so we must save the original one first.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -81,6 +81,12 @@ use_ha = roles.include? "pacemaker-cluster-member"
 remote_node = roles.include? "pacemaker-remote"
 is_cluster_founder = use_ha && node["pacemaker"]["founder"] == node[:fqdn]
 
+clone_stateless = if node["pacemaker"]
+  node["pacemaker"]["clone_stateless_services_orig"] || true
+else
+  true
+end
+
 template "/usr/sbin/crowbar-shutdown-services-before-upgrade.sh" do
   source "crowbar-shutdown-services-before-upgrade.sh.erb"
   mode "0755"
@@ -89,7 +95,8 @@ template "/usr/sbin/crowbar-shutdown-services-before-upgrade.sh" do
   action :create
   variables(
     use_ha: use_ha || remote_node,
-    cluster_founder: is_cluster_founder
+    cluster_founder: is_cluster_founder,
+    clone_stateless: clone_stateless
   )
 end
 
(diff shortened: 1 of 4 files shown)


susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -81,6 +81,12 @@
 remote_node = roles.include? "pacemaker-remote"
 is_cluster_founder = use_ha && node["pacemaker"]["founder"] == node[:fqdn]
 
+clone_stateless = if node["pacemaker"]
+  node["pacemaker"]["clone_stateless_services_orig"] || true
+else
+  true
+end
```
rhafer: Is true really the correct default to assume here? I am actually not sure which one is "more correct".

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -349,6 +349,10 @@ def prepare_nodes_for_os_upgrade
       role = proposal.role
       proposal.raw_data["attributes"]["pacemaker"]["clone_stateless_services"] = false
       proposal.raw_data["attributes"]["pacemaker"]["drbd"]["enabled"] = false
+      # Save the original value of clone_stateless_services, we need it for determining
+      # some actions from recipes.
+      role.default_attributes["pacemaker"]["clone_stateless_services_orig"] =
+        role.default_attributes["pacemaker"]["clone_stateless_services"]
```
rhafer: We should delete  ` role.default_attributes["pacemaker"]["clone_stateless_services_orig"]` at some point again.

The earliest opportunity would be after the `commit_and_check_proposal` call on line 365. But maybe it's better to actually do it once the upgrade completed.

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
jsuchome: I think the default for SOC7 is still pacemaker

crowbar_framework/app/models/crowbar_service.rb
jsuchome: Yeah, I've still not decided about the right place...

jsuchome: Actually right after the `commit_and_check_proposal` might be a good idea, later we do not have any nice place where we tackle the roles again...

jsuchome: update: actually I moved it to the very end of `services` method

PR #1704, opened 2018-11-15
Ipv6 support in the crowbar_validator

matthewoliver:
The crowbar validator uses the barclamp schema files to determine
if the current values are correct. When attempting to validate the
network.json file for the network barclamp that validator doesn't
understand ipv6 addresses.

This patch adds some extra regexes to the IpAddress validation to
understand not just IPv4 addresses but also IPv6 and CIDR netmasks.

This is one step closer to getting IPv6 support in crowbar.

### DIFF
--- a/crowbar_framework/app/models/crowbar_validator.rb
+++ b/crowbar_framework/app/models/crowbar_validator.rb
@@ -17,6 +17,7 @@
 
 require "kwalify"
 require "uri"
+require "ipaddr"
 
 class CrowbarValidator < Kwalify::Validator
    def initialize(schema_filename)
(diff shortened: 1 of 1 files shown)


matthewoliver:
OK, how about something like this? this is using the ipaddr module, which we use in the network-json-validator, which will error on an incorrect IP address (v4 or v6) so it takes all the ip regexs out of the picture.

crowbar_framework/app/models/crowbar_validator.rb
```diff
@@ -101,7 +101,8 @@ def validate_hook(value, rule, path, errors)
       when "IpAddress"
         cidr = (1..128).cover? Integer(value) rescue false
         begin
-          IPAddr.new(value) if !(path.include?("broadcast") && value.empty?) && !cidr
+          IPAddr.new(value) if !(path.is_a?(String) && path.include?("broadcast") \
+                                 && value.empty?) && !cidr
```
s-t-e-v-e-n-k: This is getting unwieldy, can we restructure this to 1) get rid of the \ and 2) not have 4 conditions, 3 of which are nested.

crowbar_framework/app/models/crowbar_validator.rb
```diff
@@ -98,15 +99,20 @@ def validate_hook(value, rule, path, errors)
            errors << Kwalify::ValidationError.new(msg, path)
          end
       when "IpAddress"
-         regex = /\A(?:25[0-5]|(?:2[0-4]|1\d|[1-9])?\d)(?:\.(?:25[0-5]|(?:2[0-4]|1\d|[1-9])?\d)){3}\z/
-         if value[regex].nil?
-           msg = "Should be an IP Address: #{value}"
-           errors << Kwalify::ValidationError.new(msg, path)
-         end
+        cidr = (1..128).cover? Integer(value) rescue false
```
rktidwell: What if an IPv4 address is passed with a prefix larger than 32? I would think that should error out somewhere, should it be here?

matthewoliver: The problem is at this point, we're only validating the single input sent it. So we have no context for the other data. So here we get a string, and it is suppose to be IpAddress, this could be ipv4, ipv6, full netmask, or a cidr. So we're validating only the input for this one value, and in that context alone we have no way of knowing whether it should be range 1-32 or 1-128.

The other checks, network-json-validator and in the network barclamps we do get this context. I.E the cidr along with an IP and/or subnet so we also validate the range is the correct range for the IP version.
See: https://github.com/crowbar/crowbar/pull/2363 and https://github.com/crowbar/crowbar-core/pull/1716

PR #1705, opened 2018-11-15
[5.0] upgrade: Revert parts of 826d376d9c7a6aad6ae019665da674849431e951

jsuchome:
Looks like wee need a workaround for a situation when neutron-ovs-agent
cuts itself of the network (bsc#1067482) and temporary change of of_interface
value seems to help.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
@@ -87,10 +87,10 @@ log "No HA setup found..."
 
 <% if @compute_node %>
 
-<% if @rabbitmq_cluster %>
-
-# update rabbitmq config for clients if it was changed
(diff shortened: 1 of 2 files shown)


jsuchome:
It seems to help - I added it manually to this build: https://ci.suse.de/view/Cloud/view/Cloud8/job/cloud-mkcloud8-job-upgrade-nondisruptive-ha-mariadb-x86_64/288/

rhafer:
I would prefer not merging this for now. I think a cleaner fix would be to:
- shutdown and disable ovs-agent
- remove br-int
and
- install default flows (or del-failmode)

before starting the os-upgrade. That way we'd  avoid any uncontrolled agent restarts during the upgrade.

But currently the problem seems to have gone anyways. So it's still unclear what actually caused it. (It might also have been some defective update in the test-updates channel)

jsuchome:
I agree. Closing this for now and we'll see if alternative is needed.

PR #1707, opened 2018-11-16
apache2: use apache2-worker

dirkmueller:
The prefork module is being deprecated and has a higher
memory/performance overhead than the default worker module.
Also prefork prevents from using more modern technologies
like HTTP/2. Switch to worker instead.

### DIFF
--- a/chef/cookbooks/apache2/recipes/default.rb
+++ b/chef/cookbooks/apache2/recipes/default.rb
@@ -22,7 +22,7 @@ package "apache2" do
   when "rhel", "fedora"
     package_name "httpd"
   when "debian", "suse"
-    package_name "apache2"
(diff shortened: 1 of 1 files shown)


sjamgade:
so iiuc if the package is not already there apache2-worker will be installed but customer wont be able use mod-php (for something else maybe), which internet tells me is not (yet?) compatible with apache2-worker. 
So the customer will have to install apache2 before deploying crowbar ?

sjamgade:
>  customer is not supposed to use mod_php on a crowbar managed installation

then its fine for me.

JanZerebecki:
Due this being a discontinued product, this probably doesn't make sense to merge any more. Please reopen if you feel otherwise.

PR #1714, opened 2018-11-22
[4.0] upgrade: Improve handling of zypper prompts (bsc#1116853)

skazi0:
Some zypper prompts include important information in <text> while
others also need <message> items to give proper feedback to the
user.

The display is still very ugly but at least it shows more details.


skazi0:
Sample outputs:
zrzut ekranu z 2018-11-22 17-05-29
zrzut ekranu z 2018-11-22 17-17-02

skazi0:
I added handling of more corner cases: no <message> and single <message>.

PR #1715, opened 2018-11-23
[5.0] upgrade: reduce keystone downtime

rhafer:
This tries to reduce the required downtime of keystone during the upgrade. Basically we achieve this by keep keystone running on the non-upgraded nodes for a while longer and by preventing it from being started by crowbar_join on the 1st upgraded node. By doing this we have more control over when we actually run the db migrations for keystone and do the switchover. 

There's a couple of things missing for the PR to be ready for merge:

- [x] Currently this only works for setups not using cloned pacemaker resources for the stateless openstack services. Making this work with cloned resources is going to be complex as we switch from cloned to non-cloned during the upgrade (which means additional downtime)
- [x] This might need additional changes for the multicluster setups, which I completely ignored for now

There is some potential for reducing the downtime even more, by utilizing keystone's rolling upgrade features  (https://docs.openstack.org/keystone/pike/admin/identity-upgrading.html "upgrading without downtime" section). But it's uncertain that this actually works for a newton -> pike upgrade. (It would at least need increased db privileges for the keystone user to be able to install triggers).

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-migrate-keystone-and-start.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-migrate-keystone-and-start.sh.erb
@@ -1,0 +1,48 @@
+#!/bin/bash
+#
+# This will execute the database schema migrations for keystone and start the
+# service by re-enabling the vhost config
+#
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+UPGRADEDIR=/var/lib/crowbar/upgrade
+mkdir -p "`dirname "$LOGFILE"`"
+exec >>"$LOGFILE" 2>&1
+
+log()
+{
+    set +x
+    echo "[$(date --iso-8601=ns)] [$$] $@"
+    set -x
+}
+
+log "Executing $BASH_SOURCE"
+
+set -x
+
+mkdir -p $UPGRADEDIR
+rm -f $UPGRADEDIR/crowbar-migrate-keystone-and-start-failed
+
+if [[ -f $UPGRADEDIR/crowbar-migrate-keystone-and-start-ok ]] ; then
+    log "database migration for keystone and restart of the vhost was already done."
+    exit 0
+fi
+
+# flush memcached caches
+systemctl restart memcached
+
+log "Running database migrations for keystone"
+keystone-manage db_sync
+log "Keystone database migrations completed"
+
+for vhost in admin public; do
+    log "Enabling apache vhost for keystone"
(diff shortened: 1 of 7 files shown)


rhafer:
To fully work this needs: https://github.com/crowbar/crowbar-openstack/pull/1906

chef/cookbooks/crowbar/templates/default/crowbar-migrate-keystone-and-start.sh.erb
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+#
+# This makes sure openstack related services are really shut down
```
jsuchome: You might wanna edit these copied-and-pasted content (same for the other script)

chef/cookbooks/crowbar/templates/default/crowbar-migrate-keystone-and-start.sh.erb
```diff
@@ -0,0 +1,48 @@
...
+    echo "[$(date --iso-8601=ns)] [$$] $@"
+    set -x
+}
+
+log "Executing $BASH_SOURCE"
+
+set -x
+
+mkdir -p $UPGRADEDIR
+rm -f $UPGRADEDIR/crowbar-migrate-keystone-and-startfailed
```
jsuchome: I know that you are actually not creating `failed` file in any case (at least for now), but it should be called `crowbar-migrate-keystone-and-start-failed`

chef/cookbooks/crowbar/templates/default/crowbar-migrate-keystone-and-start.sh.erb
```diff
@@ -0,0 +1,48 @@
...
+
+if [[ -f $UPGRADEDIR/crowbar-migrate-keystone-start-ok ]] ; then
```
jsuchome: `crowbar-migrate-keystone-and-start-ok` (missing **-and-**)

chef/cookbooks/crowbar/templates/default/crowbar-migrate-keystone-and-start.sh.erb
```diff
@@ -0,0 +1,48 @@
...
+    log "Shut down of keystone was already done."
```
jsuchome: That's probably not what you wanted to write

rhafer:
I've tested this in a multi cluster setup now as well. (There is one additional small downtime in that case, when the additional haproxy for galera needs to be reconfigured, from group to clone)

rhafer:
Updated to only install the new crowbar scripts on keystone nodes.

PR #1716, opened 2018-11-26
network: add IPv6 and CIDR support

matthewoliver:
The 'network:default' chef recipe attempts to ping the
the admin IP address but fails when the admin node is using
an IPv6 address. It fails because SLE12.* ships an older version
of iputils, whose ping only supports IPv4 addresses and requires
the use of ping6 for IPv6 addresses.
This patch checks the admin ip version and uses the correct ping
binary.

This patch also adds CIDR support to
`crowbar_framework/app/models/network_service.rb`. As without it
it fails to find the admin IP to apply to the chef node causing
the admin node's eth0 interface to be de-configured.

CIDR notation is most common in IPv6 as the writing it as a full
IPv6 mask is alot more work. This patch detects whether the netmask
is in CIDR notation, if it is it'll convert it to the full notation
so it can be used in the all the IPAddr bitwise checks.

### DIFF
--- a/crowbar_framework/app/models/network_service.rb
+++ b/crowbar_framework/app/models/network_service.rb
@@ -71,6 +71,7 @@ class NetworkService < ServiceObject
       lock = acquire_ip_lock
       db = Chef::DataBag.load("crowbar/#{network}_network") rescue nil
       net_info = build_net_info(role, network)
+      netmask = network_to_netmask(net_info)
 
       # Did we already allocate this, but the node lost it?
       if db["allocated_by_name"].key?(name)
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/network_service.rb
```diff
@@ -438,4 +439,21 @@ def build_net_info(role, network)
     end
     net_info
   end
+
+  def cidr_to_netmask(cidr, ipv6 = false)
```
rktidwell: I'm seeing a lot of this in a number of PR's and gerrit reviews: **ipv6 = false/true**, or **want_ipv6**. I'm wondering if we ought to be thinking about this in terms of ip_version, where we pass around the address family, operate on that, and use it as a suffix for files and things. One may argue this is a nit, but just thinking out loud I wonder if it will keep things more maintainable in the long run.

matthewoliver: Yeah, that isn't a bad idea. I think ipv6=true/flase and want_ipv6 came from the way these changes came to be. Ie. the pulling the loose thread model :P 
In hindsight address family would make sense.

nicolasbock: We could merge it for now and refactor the ip version in a follow up. What do you think @rktidwell ?

rktidwell: I don't want to stand in the way of progress here, we're following precedent here which is understandable. I'm just putting this out there as a suggestion going forward, no need to block this I suppose.

PR #1720, opened 2018-11-28
Update Rack in the barclamp Gemfile

rsalevsky:
https://nvd.nist.gov/vuln/detail/CVE-2018-16471


toabctl:
@rsalevsky travis fails with:
```
Bundler could not find compatible versions for gem "ruby":
  In Gemfile:
    ruby
    rack (~> 2.0) was resolved to 2.0.6, which depends on
      ruby (>= 2.2.2)
```

PR #1721, opened 2018-11-29
upgrade: Increase delayed_job time limit

skazi0:
The longest method started in the background seems to be Api::Upgrade.nodes.
Assuming 1h per node we could upgrade 168 nodes with 7 days limit.
Default limit is 4 hours which is way too low for this task.


jsuchome:
(Adding Rick, I think he was there when we originally designed this delayed_job)

PR #1723, opened 2018-12-05
IPv6: dhcp/provisioner: make ipv6 aware

matthewoliver:
If the admin network is IPv6 setup the ISC DHCPD server to configure
and use the IPv6 daemon.

Make sure tftp is listening on both IPv4 and v6.


susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please remove period from end of commit subject line.<br />
56985b908c3636fad4baef35e2df427f747b4673</td>
    </tr>
  </tbody>
</table>

</p>

chef/cookbooks/dhcp/attributes/default.rb
```diff
@@ -10,4 +10,16 @@
     "option dhcp-client-debug code 226 = unsigned integer 16",
     "option dhcp-client-debug 0"
 ]
+default[:dhcp][:options][:v6] = [
+    "ddns-update-style none",
```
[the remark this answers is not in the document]
matthewoliver: yeah, it's times like this I wonder should I just fix the hound "error" everywhere in the file to get it to pass or just ignore it. I mean we use 4 spaces everywhere here, so am keeping with the style.

chef/cookbooks/dhcp/providers/subnet.rb
```diff
@@ -13,11 +13,20 @@
 # limitations under the License.
 #
 
+require "ipaddr"
+
 action :add do
+  if IPAddr.new(new_resource.network["subnet"]).ipv6?
```
rktidwell: I've seen this sort of thing in a number of places. I wonder if it might be cleaner to use an ip_version variable to store 4 vs. 6, and then you can simply use ip_version as a suffix and do away with some if/else blocks and make any file names explicit (ie subnet4.conf.erb vs. subnet.conf.erb)

chef/cookbooks/dhcp/recipes/default.rb
```diff
@@ -63,18 +78,27 @@
 intfs = [Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").interface]
 address = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
 
-d_opts = node[:dhcp][:options]
+require "ipaddr"
+admin_addr = IPAddr.new(address)
+
+if admin_addr.ipv4?
```
rktidwell: See my previous comment about maybe referencing an ip_version variable. I also wonder if this precludes us from doing dual-stack, meaning we would have to come back to this and refactor it again.

nicolasbock:
Rebased

chef/cookbooks/dhcp/providers/subnet.rb
matthewoliver: The way that crowbar updates and removes subnets from the list does make we worry about it from an upgrade perspective. I.e an existing env with subnet_list.conf with a bunch of subnets and we suddently change it. Will only "new" subnet source config lines be added.

I'll re-roll 1/2 way between. Leave ipv4 and subnet_list.conf and use subnet6_list.conf. But change to just adding a 6 or not. It does look much cleaner. Though ip_version = '' on ipv4 but. still cleaner then it was.

chef/cookbooks/dhcp/recipes/default.rb
matthewoliver: Yeah, I have a feeling there might be a bunch or refactoring required for true dual stack. Currently crowbar only expects 1 network of each type. So I think a bunch of changes maybe be required for it to have more then 1 network for each type. I think once we're gone through single stack, we'll have a very good idea of how much of crowbar we need to touch.

Unfortunately if I was more experienced with Crowbar, maybe I'd be able to tell you outright. But this is a new journey into the crux of crowbar for me too ;)

susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please rebase to get rid of the merge commits in this PR</td>
    </tr>
  </tbody>
</table>

</p>

chef/cookbooks/provisioner/recipes/setup_base_images.rb
```diff
@@ -525,7 +530,7 @@
         owner "root"
         group "root"
         source "crowbar_register.erb"
-        variables(admin_ip: admin_ip,
+        variables(admin_ip: admin_ip.to_s,
                   admin_broadcast: admin_net.broadcast,
```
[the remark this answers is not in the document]
matthewoliver: it'll be an empty string in IPv6 I believe.. which might make the crowbar_register script fail to find the correct interface (from a very quick look at the script) when attempting to register a new node with crowbar.

So I have a feeling we'll need to modify the register script some more. But that comes into play when were creating and registering nodes. So in the next few mkcloud steps after this one. But a good candidate for a new card in our backlog:

  Test and modify crowbar_register to work with IPv6 addresses, as this script has comes reliance on broadcast addresses to find interfaces.

chef/cookbooks/dhcp/recipes/default.rb
```diff
@@ -48,6 +48,21 @@
...
+end
+file "/etc/dhcp3/subnets.d/subnet6_list.conf" do
+  owner "root"
+  group "root"
+  mode 0644
+end
+file "/etc/dhcp3/hosts.d/host6_list.conf" do
+  owner "root"
+  group "root"
+  mode 0644
```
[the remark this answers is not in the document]
matthewoliver: True, when originally writing this I was thinking about an environment that was dual stack. But your right, I'm already making assumptions based on  the admin network and what version it is in order to create the correct dhcpd[6] config. I'll make the change to only write out for the ip_version of the admin net.

chef/cookbooks/dhcp/recipes/default.rb
```diff
@@ -60,21 +75,23 @@
 end
 
 # This needs to be evaled.
-intfs = [Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").interface]
-address = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+admin_network = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin")
+intfs = [admin_network.interface]
```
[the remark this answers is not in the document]
matthewoliver: it is unused.. but the comment above say. "# this needs to be evaled." so maybe it's where to make sure an address exists? All I know if hound complained because it was an unused variable. Will add it back as it was there before.

chef/cookbooks/dhcp/templates/default/dhcpd.conf.erb
```diff
@@ -21,6 +21,34 @@ log-facility local7;
...
+
+# The delay before information-request refresh
+#  (minimum is 10 minutes, maximum one day, default is to not refresh)
+#  (set to 6 hours)
+#option dhcp6.info-refresh-time 21600;
+
+include "/etc/dhcp3/groups.d/group6_list.conf";
+include "/etc/dhcp3/subnets.d/subnet6_list.conf";
+include "/etc/dhcp3/hosts.d/host6_list.conf";
+<% else -%>
```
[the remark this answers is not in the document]
matthewoliver: They are the default values from the "sample" config, there commented out there too. Really there there because maybe there something we may or may not what to tweak and left them here seeing as I haven't really testing the actual running dhcpd6 yet.. other then it starts.

Also it was WIP so figured doesn't hurt. I'll remove them for now.

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -346,10 +346,15 @@
...
 end
 
+require "ipaddr"
 crowbar_node = node_search_with_cache("roles:crowbar").first
 address = crowbar_node["crowbar"]["network"]["admin"]["address"]
 protocol = crowbar_node["crowbar"]["apache"]["ssl"] ? "https" : "http"
-server = "#{protocol}://#{address}"
+server = if IPAddr.new(address).ipv6?
+  "#{protocol}://[#{address}]"
+else
```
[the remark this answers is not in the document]
matthewoliver: did we? I don't remember it. But doesn't mean there isn't... I do have young girls who don't sleep.. which means I don't :P

If we don't I'll add an address_wrapped to the network class.

matthewoliver:
This time push after testing it in my mkcloud env :P

matthewoliver:
This time forgot to remove the address_wrapped from the network class, which isn't required now that we have the NetworkHelper.

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
```diff
@@ -114,7 +114,11 @@ wait_for_hostname() {
 wait_for_admin_server() {
     # wait for admin server to become pingable
     tries_left=120
-    while ! ping -q -c1 $IP > /dev/null; do
+    local ping="ping"
+    if (( $IPV6 > 0 )); then
```
s-t-e-v-e-n-k: Where is that defined?

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
```diff
@@ -398,7 +402,17 @@ EXVAL=0
 export HOME=/root
 
 IP="<%= @admin_ip %>"
-HTTP_SERVER="<%= @admin_ip %>:<%= @web_port %>"
+<% if @ip_version == "6" -%>
+IPV6=1
+<% else  -%>
+IPV6=0
```
s-t-e-v-e-n-k: Below where it is first used? I'm so confused.

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -122,7 +122,17 @@ add_user() {
...
+<% if @ip_version == "6" -%>
+IPV6=1
+<% else -%>
+IPV6=0
+<% end -%>
+if (( $IPV6 > 0 )); then
+  ADMIN_IP_WRAPPED="[ADMIN_IP]"
+else
+  ADMIN_IP_WRAPPED="ADMIN_IP"
+fi
```
s-t-e-v-e-n-k: Could we collapse these into one if block? They will effectively do the same thing anyway

chef/cookbooks/dhcp/libraries/helpers.rb
```diff
@@ -0,0 +1,9 @@
+module DhcpHelper
+  def self.config_filename(base, ip_version, extension = ".conf")
+    if ip_version == "4"
+      "#{base}#{extension}"
+    else
+      "#{base}#{ip_version}#{extension}"
+    end
```
s-t-e-v-e-n-k: To be fair I had envisioned this as always returning "#{base}#{extra}#{extension}" where extra gets set to 6 if ip_version is 6, but this is very much a nit.

chef/cookbooks/dhcp/attributes/default.rb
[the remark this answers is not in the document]
s-t-e-v-e-n-k: You have two choices: Ignore Hound, or push a fix-up commit and then base your changes on top of that.

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
```diff
@@ -163,9 +167,9 @@ do_setup() {
     # Make sure that the client knows how to talk to the server.
     local cfg=/etc/chef/client.rb
     if ! [ -f $cfg ] || \
-       ! grep -q "^\s*chef_server_url\s*[\"\']http://$IP:4000[\"\']" $cfg; then
+       ! grep -q "^\s*chef_server_url\s*[\"\']http://$IP_WRAPPED:4000[\"\']" $cfg; then
```
s-t-e-v-e-n-k: And this? Because below is ADMIN_IP_WRAPPED, not IP_WRAPPED ...

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
matthewoliver: yeah it is confusing, but really the code here is run first because it's outside of functions. Then it's used in functions. So was just following the pattern that was previously used for the $IP. As I figured once this is templated out into a bash script I wanted it to look like the old version as how knows in anyone ever goes in an hacks it. I guess they shouldn't.

I was lazy, so figured following an existing pattern and minimalise the change to the crowbar_join rather then refactor and probably cause or introduce bugs.

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
matthewoliver: yup, good point. I guess I shouldn't try and write back code that allows someone to easy hack on a machine (turn IPV6 to 0 or 1) cause it's not really making it any better.
DONE

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
matthewoliver: huh, it's all IP_WRAPPED in this script. Unless I'm missing something.. which might be possible :P

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -350,7 +350,8 @@
 crowbar_node = node_search_with_cache("roles:crowbar").first
 address = crowbar_node["crowbar"]["network"]["admin"]["address"]
 protocol = crowbar_node["crowbar"]["apache"]["ssl"] ? "https" : "http"
-server = "#{protocol}://#{address}"
+server = "#{protocol}://#{NetworkHelper.wrap_ip(address)}"
+password = crowbar_node["crowbar"]["users"]["crowbar"]["password"]
```
s-t-e-v-e-n-k: Where did this line come from? This is the cause of the current CI failure, and on current code, is roughly replicated and replaced a few lines down. Can you remove this, rebase and push it up?

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -219,19 +222,33 @@ def find_node_boot_mac_addresses(node, admin_data_net)
...
-  append = []
-  mac_list.each_index do |i|
-    dhcp_host "#{mnode.name}-#{i}" do
-      hostname mnode.name
-      macaddress mac_list[i]
-      if admin_mac_addresses.include?(mac_list[i])
-        ipaddress admin_ip_address
-        options [
-          "if exists dhcp-parameter-request-list {
+  if admin_data_net.ip_version == "6"
```
s-t-e-v-e-n-k: This looks to be very much duplicating the next block. I'm not going to block on it, but can comment that it could be refactored to be IP version agnostic.

matthewoliver:
Added to more commits to this PR. First adds the adminip to the kernel commandline so the new version of the sleshammer script can pick it up. And the second fixes the ntp configuration.
I've only roughly tested on my laptop. Tomorrow I'll work through hounds grips and clean up a little :)

nicolasbock:
Rebased

PR #1727, opened 2018-12-08
[4.0] provisioner: validate format of additional ssh keys section

dirkmueller:
The chef recipes crash with a very weird error message when
the additional ssh keys section is malformatted, which can easily
be the case given that ssh keys are usually copy&pasted. Provide
some useful hints for troubleshooting formatting errors and enforce
that we have at least 3 sections because the chef recipes
are doing unification based on the "comment" field.

Backport of: https://github.com/crowbar/crowbar-core/pull/1696
(cherry picked from commit 33b52622dbc23e82a411209e897161bd8cc81899)


nicolasbock:
Isn't that the same as #1724 @dirkmueller ?

PR #1728, opened 2018-12-08
conduits: match base_board.serial_number as well

dirkmueller:
Some machines, like for example HPE Proliant Gen8/9 do not report
their serial number under dmi.system but under dmi.base_board.

$ knife node show $node -a node.dmi.system.serial_number
node.dmi.system.serial_number:  Not Specified
$ knife node show $node -a node.dmi.base_board.serial_number
node.dmi.base_board.serial_number:  2M2528179N

Treat that as equivalent for matching the serial_number bus_order
specifier.

### DIFF
--- a/crowbar_framework/lib/crowbar/conduit_resolver.rb
+++ b/crowbar_framework/lib/crowbar/conduit_resolver.rb
@@ -34,8 +34,11 @@ module Crowbar
         if cr_network_config.key?("interface_map")
           cr_network_config["interface_map"].each do |data|
             next unless cr_dmi_system["product_name"] =~ /#{data["pattern"]}/
-            next if data.key?("serial_number") &&
-                cr_dmi_system["serial_number"].strip != data["serial_number"].strip
+            if data.key?("serial_number")
+              data_serial = data["serial_number"].strip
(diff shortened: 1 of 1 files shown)


crowbar_framework/lib/crowbar/conduit_resolver.rb
```diff
@@ -34,8 +34,10 @@ def bus_order
         if cr_network_config.key?("interface_map")
           cr_network_config["interface_map"].each do |data|
             next unless cr_dmi_system["product_name"] =~ /#{data["pattern"]}/
-            next if data.key?("serial_number") &&
-                cr_dmi_system["serial_number"].strip != data["serial_number"].strip
+            if data.key?("serial_number")
+              next if data["serial_number"].strip != cr_dmi_system["serial_number"].strip &&
+                  cr_dmi_base_board["serial_number"].strip != data["serial_number"].strip
```
toabctl: the function `cr_dmi_base_board` can return `{}` and then the `strip` will fail:
```
irb(main):008:0> {}['foo'].strip
Traceback (most recent call last):
        2: from /usr/bin/irb:11:in `<main>'
        1: from (irb):8
NoMethodError (undefined method `strip' for nil:NilClass)

```

PR #1732, opened 2018-12-12
salt: Make the barclamp visible in the UI and adjust order

toabctl:
To be able to deploy salt on the admin node (and use it via salt-ssh),
the baclamp must be visible. That's done now.
Also move the barclamp in the list down (given that it is not that
important and more an optional feature).
Also change the "proposal_schema_version" to 3 which seems to be the
default in other barclamps.

### DIFF
--- a/salt.yml
+++ b/salt.yml
@@ -19,13 +19,13 @@ barclamp:
   display: 'Salt'
   description: 'Salt Management'
   version: 0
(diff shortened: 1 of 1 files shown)


toabctl:
The documentation update can be found here: https://github.com/SUSE-Cloud/doc-cloud/pull/668

PR #1734, opened 2018-12-13
[5.0] upgrade: Signal all nova services after the upgrade is complete 

jsuchome:
There are operations (like attaching volume using reserve_block_device_name
method) that cannot be performed with mixed nova setup. So we pin the RPC
version to the latest one that works for Newton. That is achieved with https://github.com/crowbar/crowbar-openstack/pull/1941

After the upgrade, we need to reload (and restart?) the services so they can switch to the latest RPC version.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-run-nova-online-migrations.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-run-nova-online-migrations.sh.erb
@@ -1,0 +1,50 @@
+#!/bin/bash
+#
+# After the upgrade of all services on all nodes is finished and all nova services
+# run the new version, it's recommended to run nova online data migrations
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+UPGRADEDIR=/var/lib/crowbar/upgrade
+mkdir -p "`dirname "$LOGFILE"`"
+exec >>"$LOGFILE" 2>&1
+
+log()
+{
+    set +x
+    echo "[$(date --iso-8601=ns)] [$$] $@"
+    set -x
+}
+
+log "Executing $BASH_SOURCE"
+
+set -x
+
+mkdir -p $UPGRADEDIR
+rm -f $UPGRADEDIR/crowbar-run-nova-online-migrations-failed
+
+if [[ -f $UPGRADEDIR/crowbar-run-nova-online-migrations-ok ]] ; then
+    log "Nova migrations were already executed"
+    exit 0
+fi
+
+nova-manage db online_data_migrations
+
+# From the manual:
+# "Returns exit code 0 if migrations were successful or exit code 1 for partial updates.
+# If the command exits with partial updates (exit code 1) the command will need to be called again."
+
+# So let's call it again right away. If it fails for second time, throw an error and give
+# control to user.
+
+nova-manage db online_data_migrations
+
+ret=$?
+if [ $ret != 0 ] ; then
+    log "Error occured during online_data_migrations run"
+    echo $ret > $UPGRADEDIR/crowbar-run-nova-online-migrations-failed
+    exit $ret
+fi
+
+
+touch $UPGRADEDIR/crowbar-run-nova-online-migrations-ok
+log "$BASH_SOURCE is finished."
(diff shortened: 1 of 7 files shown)


susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

jsuchome:
@skazi0 do you think we should restart nova services at the end? Is it possible just to reload it?

rhafer:
> No, reloading does not work. But I'm bit of unsure we should just restart the service remotely (and not via chef client)

chef-client won't restart them. As that override file is not managed by chef. So I guess it has to happen remotely somehow.

jsuchome:
@skazi0 @rhafer I've added the restart code. Should be correct but is totally untested

skazi0:
@jsuchome travis doesn't like you changes :o)

jsuchome:
@skazi0 it does now

dirkmueller:
Please at least cobsider add a bug reference to the commit message which links to documentation update request.

skazi0:
@dirkmueller the cleanup is there (in `delete_upgrade_scripts()`). In addition the API is restarted to make sure the changes are picked up.

chef/cookbooks/crowbar/templates/default/crowbar-run-nova-online-migrations.sh.erb
```diff
@@ -0,0 +1,41 @@
...
+
+log "Executing $BASH_SOURCE"
+
+set -x
+
+mkdir -p $UPGRADEDIR
+rm -f $UPGRADEDIR/crowbar-run-nova-online-migrations-failed
+
+if [[ -f $UPGRADEDIR/crowbar-run-nova-online-migrations-ok ]] ; then
+    log "Restart nova script was already successfully executed"
```
skazi0: old message?

jsuchome: fixed

jsuchome:
> do we need to delete those new scripts?

I think we do. It's done in `delete_upgrade_scripts` method

chef/cookbooks/crowbar/templates/default/crowbar-restart-nova-after-upgrade.sh.erb
```diff
@@ -0,0 +1,44 @@
...
+if [[ -f $UPGRADEDIR/crowbar-restart-nova-after-upgrade-ok ]] ; then
+    log "Restart nova script was already successfully executed"
+    exit 0
+fi
+
+<% if @nova_controller %>
+for service in conductor consoleauth scheduler novncproxy serialproxy api; do
+    fullname="openstack-nova-$service"
+    if systemctl --quiet is-active $fullname 2>/dev/null ; then
+        kill -s HUP $fullname
```
jsuchome: should be `systemctl kill` (also in the other place)

jsuchome:
@skazi0 @rhafer I've adapted the call to `nova-manage db online_data_migrations`, new review is necessary :-(

skazi0:
@jsuchome I was lucky enough to grab the latest version for my testing :)

chef/cookbooks/crowbar/templates/default/crowbar-run-nova-online-migrations.sh.erb
```diff
@@ -0,0 +1,50 @@
...
+nova-manage db online_data_migrations --max-count 500
+
+# From the manual:
+# "Returns exit code 0 if migrations were successful or exit code 1 for partial updates.
+# If the command exits with partial updates (exit code 1) the command will need to be called again."
+
+# So let's call it again right away. If it fails for second time, throw an error and give
+# control to user.
+
+nova-manage db online_data_migrations --max-count 500
```
skazi0: does that mean that we will migrate at most 1000 records?

jsuchome: I think it means it goes in batches of 500, but have to check the docs..

skazi0:
I've just finished my test run with this change. All nova-compute services reported new RPC version :)

PR #1735, opened 2018-12-13
[5.0] important patches for ECP upgrade

jsuchome:
These were not merged yet, but we need them now for the ECP upgrade


toabctl:
@jsuchome first one (validate format ....) got merged: https://github.com/crowbar/crowbar-core/pull/1725

PR #1741, opened 2018-12-18
[5.0]upgrade: Skip already upgraded nodes+fail with no upgradeable nodes

skazi0:
**Skip already upgraded nodes**
If list of nodes is requested for upgrade, already upgraded nodes will be
skipped. Re-running upgrade procedure on an already upgraded node could
trigger various problems.
+
**Fail if no nodes can be upgraded**
This can happen e.g. if some live evacuation doesn't succeed.
The added condition should ensure non_disruptive_upgrade_compute_nodes()
will not loop forever.

### DIFF
--- a/crowbar_framework/app/controllers/api/upgrade_controller.rb
+++ b/crowbar_framework/app/controllers/api/upgrade_controller.rb
@@ -187,14 +187,23 @@ class Api::UpgradeController < ApiController
         # At this point params[:component] should be a node, if it is not,
         # raise an error.
         nodes_names = params[:component].split(/[\s,;]/)
+        upgraded_nodes_names = []
         nodes_names.each do |name_or_alias|
           node = ::Node.find_node_by_name_or_alias(name_or_alias)
-          next unless node.nil?
-          raise ::Crowbar::Error::UpgradeError,
-            "Component must be 'all', 'controllers', 'resume', "\
-            "'postpone' or a node name(s). "\
-            "No node with '#{name_or_alias}' name or alias was found. "
+          if node.nil?
+            raise ::Crowbar::Error::UpgradeError,
+              "Component must be 'all', 'controllers', 'resume', "\
+              "'postpone' or a node name(s). "\
+              "No node with '#{name_or_alias}' name or alias was found. "
+          end
+          upgraded_nodes_names << name_or_alias if node.upgraded?
         end
+
+        if (nodes_names - upgraded_nodes_names).empty?
+          raise ::Crowbar::Error::UpgradeError,
+            "All requested nodes are already upgraded."
+        end
+
         if substep != :compute_nodes && status != :finished
           raise ::Crowbar::Error::UpgradeError.new(
             "Controller nodes must be upgraded first!"
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -1442,6 +1442,12 @@ module Api
           ::Node.find_node_by_name_or_alias(name)
         end
 
+        # make sure we upgrade each node only once
+        compute_nodes.uniq!(&:name)
+
+        # remove nodes which were already upgraded
+        compute_nodes.reject!(&:upgraded?)
+
         controller = fetch_nova_controller
 
         # If there's a compute node which we already started to upgrade,
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -960,7 +960,7 @@ module Api
           "run_list_map:pacemaker-cluster-member AND " \
           "NOT fqdn:#{founder[:fqdn]}"
         )
-        non_founder_nodes.select! { |n| !n.upgraded? }
+        non_founder_nodes.reject!(&:upgraded?)
 
         if founder.upgraded? && non_founder_nodes.empty?
           Rails.logger.info("All nodes in cluster #{cluster} have already been upgraded.")
@@ -1294,7 +1294,7 @@ module Api
         end
 
         # remove upgraded compute nodes
-        compute_nodes.select! { |n| !n.upgraded? }
+        compute_nodes.reject!(&:upgraded?)
         if compute_nodes.empty?
           Rails.logger.info(
             "All compute nodes of #{virt} type are already upgraded."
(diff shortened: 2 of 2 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1442,6 +1442,9 @@ def non_disruptive_upgrade_compute_nodes(names)
           ::Node.find_node_by_name_or_alias(name)
         end
 
+        # remove nodes which were already upgraded (it might leave an empty list)
+        compute_nodes.reject!(&:upgraded?)
```
jsuchome: Could you add a log message (and exit from the method) after you check that the list is empty? It does not have to be _completely_ silent

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1473,6 +1476,12 @@ def non_disruptive_upgrade_compute_nodes(names)
               break
             end
           end
+          if nodes_to_upgrade.empty?
+            raise_node_upgrade_error(
+              "None of the nodes selected for upgrade can be upgraded now. " \
+              "Check /var/log/crowbar/production.log for details."
```
jsuchome: Hm, although this is true in fact, we should somehow direct user to the node with a failure, don't you think?

jsuchome: I really think we could end in this place only by failing the live migration of the first node in the `compute_nodes` list. So we know what happened and should be able to tell more.

crowbar_framework/app/models/api/upgrade.rb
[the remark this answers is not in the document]
skazi0: I changed the other places in separate commit.

skazi0: done

crowbar_framework/app/models/api/upgrade.rb
skazi0: But there should be a line about live migration already in the log, right?

jsuchome: It is, but I think we should mention it in the error message if we know it. Going to the logs is next level.

skazi0: I added another line with hint on the live evacuation.

skazi0: I added name of first node to the message. Please test this if you have some env where this scenario can be triggered.

jsuchome: Wouldn't it be simpler to just save the node name at the time we catch the timeout exception and exit the loop?

skazi0: Hmm... adding another variable just to print it in some corner case... not sure which is simpler really...

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1442,6 +1442,15 @@ def non_disruptive_upgrade_compute_nodes(names)
...
 
+        # remove nodes which were already upgraded (it might leave an empty list)
+        compute_nodes.reject!(&:upgraded?)
+        if compute_nodes.empty?
+          Rails.logger.info(
+            "All requested compute nodes are already upgraded."
+          )
+          return
+        end
+
```
jsuchome: So, I think I've changed my mind on this :-)

I think the `reject` call above is correct and we should keep it to (silently) filter out upgraded nodes. However if all provided nodes are already upgraded, we should raise an error. Preferably in controller already, as you suggested earlier. This place https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/controllers/api/upgrade_controller.rb#L189 seems good for it as we are already checking all nodes data.

crowbar_framework/app/models/api/upgrade.rb
jsuchome: In fact, the "None of the nodes selected for upgrade can be upgraded now." message could be wrong.
Consider 2 nodes selected for upgrade. live evacuation fails for node B but not for node A, never mind, we just upgrade A. `compute_nodes` is just one node now and it goes to next cycle. live evacuation again fails for node B and as there's no other node to upgrade left, we exit with an error (correctly). But we should not say nothing got upgraded, because actually one node was.

skazi0: The "now" was supposed to indicate that this includes changes done between the request and now :o)
I reworded the message. Any better?

crowbar_framework/app/models/api/upgrade.rb
skazi0: done

crowbar_framework/app/controllers/api/upgrade_controller.rb
```diff
@@ -187,14 +187,23 @@ def nodes
...
+          if node.nil?
+            raise ::Crowbar::Error::UpgradeError,
+              "Component must be 'all', 'controllers', 'resume', "\
+              "'postpone' or a node name(s). "\
+              "No node with '#{name_or_alias}' name or alias was found. "
+          end
+          upgraded_node_names << name_or_alias if node.upgraded?
+        end
+
+        if (node_names - upgraded_node_names).empty?
```
jsuchome: It's `nodes_names` at line 189

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1442,6 +1442,9 @@ def non_disruptive_upgrade_compute_nodes(names)
           ::Node.find_node_by_name_or_alias(name)
         end
 
+        # remove nodes which were already upgraded
+        compute_nodes.reject!(&:upgraded?)
```
jsuchome: Please also add a `unique` call to remove duplicates from `compute_nodes`. I know that's a different problem, but fix kind of fits in this PR

crowbar_framework/app/controllers/api/upgrade_controller.rb
skazi0: renamed all `*node*` to `*nodes*` for consistency.

crowbar_framework/app/models/api/upgrade.rb
skazi0: added but I haven't tested this part. please re-review.

PR #1742, opened 2018-12-19
[5.0] upgrade: Exit the upgrading loop if there is only one node left

jsuchome:
If the live-migration is failing for some reason that cannot
be fixed by repeating the operation, we have to exit the upgrading loop.
Otherwise it may loop trying to live migrate the only remaining node
over and over.


jsuchome:
https://github.com/crowbar/crowbar-core/pull/1741 seems to be a better solution

PR #1746, opened 2018-12-20
upgrade: Do not mark nodes as upgraded before the last action

jsuchome:
Last action for compute nodes is re-enabling the compute service.
If it fails, there should be an option to run the step again, but
that is not possible if the nodes are already marked as upgraded.

Additionally, make the check for ssh command run a bit more robust.

(cherry picked from commit bef6800be1dcc739f497f5c610fbad08529c6bf7)

port of https://github.com/crowbar/crowbar-core/pull/1740


jsuchome:
@rhafer @skazi0 I had to rebase, could you please review again?

PR #1739, opened 2018-12-18
Revert "upgrade: Deleting cinder services from database no longer nee…

jsuchome:
…ded"

This reverts commit b27b6d72a56bf7e9a22a4d7d93ce6df6f22db8ba.

(cherry picked from commit 47e531266283bde3616c3ed7adf160d6e852868c)

Port of https://github.com/crowbar/crowbar-core/pull/1736

PR #1748, opened 2018-12-20
upgrade: reduce keystone downtime 

rhafer:
This tries to reduce the required downtime of keystone during the upgrade. Basically we achieve this by keep keystone running on the non-upgraded nodes for a while longer and by preventing it from being started by crowbar_join on the 1st upgraded node. By doing this we have more control over when we actually run the db migrations for keystone and do the switchover.

Forward port of #1715

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -1009,9 +1009,26 @@ module Api
         # delete old pacemaker resources (from the node where old pacemaker is running)
         delete_pacemaker_resources other_node
         shutdown_all_services_in_cluster node
+
+        # prevent first upgrade node to start keystone during the
+        # crowbar join call
+        node = ::Node.find_by_name(node.name)
+        node["keystone"]["disable_vhost"] = true
+        node.save
+
         # start crowbar-join at the first node
         node_api.post_upgrade
         node_api.join_and_chef
+
+        # stop keystone on non-upgraded nodes, run db migrations
+        # and start on the upgrade node
+        keystone_migrate_and_restart node
+
+        # allow keystone to be handled by chef again
+        node = ::Node.find_by_name(node.name)
+        node["keystone"].delete "disable_vhost"
+        node.save
+
         re_enable_network_agents(node, node)
         node_api.save_node_state("controller", "upgraded")
       end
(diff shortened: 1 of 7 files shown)


rhafer:
Needs to be rebased after #1739 got merged

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -946,6 +947,15 @@ def upgrade_cluster(founder, cluster)
...
         end
 
+        # After the cluster was upgraded, remove the temporary attribute from
+        # the proposal role that was used for tracking the original (pre-upgrade)
+        # value of "clone_stateless_services"
+        pacemaker_proposal_role = RoleObject.find_roles_by_name(cluster).first
+        pacemaker_proposal_role.default_attributes["pacemaker"].delete(
+          "clone_stateless_services_orig"
+        )
+        pacemaker_proposal_role.save
```
jsuchome: We should be able to drop this, right?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -974,9 +984,33 @@ def upgrade_first_cluster_node(node, other_node)
...
+        clone_stateless_pre_upgrade = node["pacemaker"]["clone_stateless_services_orig"].nil? ||
+          node["pacemaker"]["clone_stateless_services_orig"]
+
+        if !clone_stateless_pre_upgrade && node[:run_list_map].key?("keystone-server")
+          # prevent first upgrade node to start keystone during the
+          # crowbar join call
+          node = ::Node.find_by_name(node.name)
+          node["keystone"]["disable_vhost"] = true
+          node.save
+        end
```
jsuchome: And all these `clone_stateless_pre_upgrade` related code

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -974,9 +984,33 @@ def upgrade_first_cluster_node(node, other_node)
...
+
         # start crowbar-join at the first node
         node_api.post_upgrade
         node_api.join_and_chef
+
+        if !clone_stateless_pre_upgrade && node[:run_list_map].key?("keystone-server")
```
jsuchome: And here

dirkmueller:
@rhafer rebase?

PR #1693, opened 2018-10-23
[4.0] Bring back the suport for ceph nodes during the upgrade

jsuchome:
This is experimental code that attempts to use the same way of handling ceph nodes from 6-7 upgrade to 7-8.

This includes:
- Port of https://github.com/crowbar/crowbar-core/pull/1077
- Port of https://github.com/crowbar/crowbar-core/pull/1079
- Revert of https://github.com/crowbar/crowbar-core/pull/1577

PR #1751, opened 2018-12-21
Bring back the suport for ceph nodes during the upgrade

jsuchome:
Originally https://github.com/crowbar/crowbar-core/pull/1693. Only to archive the work.


jsuchome:
This won't pass the testbuild ... it has one extra patch that is already in official 4.0 branch

PR #1754, opened 2018-12-21
upgrade: Delete nova services that are registered to unknown nodes

jsuchome:
If such (presumably old) services would stay in database, it would not be possible
to convince fully upgraded nova to use the latest RPC API.


chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -112,6 +112,20 @@
   only_if { cinder_controller && (!use_ha || is_cluster_founder) }
 end
 
+all_nodes = search(:node, "state:crowbar_upgrade").map { |n| n["hostname"] }
```
jsuchome: It might be a good idea to do 2 searches, one for controller nodes and one for compute

PR #1762, opened 2019-01-08
upgrade: Signal all nova services after the upgrade is complete 

jsuchome:
Port of https://github.com/crowbar/crowbar-core/pull/1734


jsuchome:
@skazi0 could you please approve so we can merge it?

PR #1765, opened 2019-01-09
Added ability to include recipes smartly

sjamgade:
Roles can get smart by making recipe inform the role about its dependencies
across the proposal.

So a recipe will be included by the role only if the attributes
registered by the recipe are actually changed in current chef-client
run.

a recipe can register its dependencies as
```
mydependson = {
    "horizon::server" => [
        ['glance','api','bind_port']
    ]
}

BarclampLibrary::Barclamp::DependsOn.add(mydependson)
```
So here the "horizon::server"  is informing about its
dependencies.

This is accomplished by comparing the value of the current attributes of node
against the one already stored in the databag after the previous successful
chef-client run.

BarclampLibrary::Barclamp::DependsOn.add:
takes in a map:
```
     recipe_name => [
                [barclampname,drill,down,till,value],
                [otherbarclampname,onlyhere],
               ]
```
This map is flushed and recreated only everyrun, however like resources
this could also be cached.

this behaviour can be altered by adding flag to config and using it
'include_recipe_smartly', however that is an extension to this behavior
and can be addressed in subsequent commits

### DIFF
--- a/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+++ b/chef/cookbooks/barclamp/libraries/barclamp_library.rb
@@ -417,9 +417,52 @@ module BarclampLibrary
       end
     end
 
+    class DependsOn
+      class << self
+        def add(dependency)
+            @recipe_depedancies ||= Mash.new
+            @recipe_depedancies.merge!(dependency)
+        end
+        def get(recipe)
+            @recipe_depedancies.fetch(recipe, [])
+        end
+      end
+    end
+
     class Config
       class << self
         attr_accessor :node
+
+        def last_two_configs(group, barclamp, instance)
+          prev_cfg = load(group, barclamp, instance)
+          cur_cfg = node[barclamp.to_sym].current_default
+          return prev_cfg, cur_cfg
+        end
+
+        def loadattr(hash, attrlist)
+          i=0
+          while hash.is_a?(Hash)
+            if hash.key?(attrlist[i])
+              hash = hash[attrlist[i]]
+              Chef::Log.debug("smart loadattr #{attrlist[i]}, #{hash}")
+              i+=1
+              return hash if attrlist.length == i
+            else
+              return nil
+            end
+          end
+        end
+
+        def has_changes_to_apply(depedency, group="openstack" ,instance = nil)
+          barclamp = depedency.shift
+          prev_cfg, curr_cfg = Barclamp::Config.last_two_configs(group, barclamp, instance)
+          old = loadattr(prev_cfg, depedency)
+          new = loadattr(curr_cfg, depedency)
+          Chef::Log.debug("smart loadattr prev #{depedency}, #{prev_cfg}")
+          Chef::Log.debug("smart loadattr curr #{depedency}, #{curr_cfg.inspect}")
+          Chef::Log.debug("smart comparisoin #{old}, #{new}")
+          return old != new
+        end
 
         def load(group, barclamp, instance = nil)
           # If no instance is specified, see if this node uses an instance of
--- a/chef/cookbooks/barclamp/libraries/smartroles.rb
+++ b/chef/cookbooks/barclamp/libraries/smartroles.rb
@@ -1,0 +1,24 @@
+require 'chef/mixin/language_include_recipe'
+
+class Chef
+  module Mixin
+    module LanguageIncludeRecipe
+      def include_recipe_smartly(*list_of_recipes)
+        for recipe in list_of_recipes
+          included=false
+          for dependency in BarclampLibrary::Barclamp::DependsOn.get(recipe)
+            if BarclampLibrary::Barclamp::Config.has_changes_to_apply(dependency)
+              Chef::Log.info("smart including recipe: #{recipe}")
+              Chef::Log.debug("due to change in: #{dependency}")
+              include_recipe recipe
+              included=true
+              break
+            end # if
+          end # for
+          Chef::Log.info("recipe excluded: #{recipe}") unless included
+        end # for
+      end # def include_recipe_smartly
+    end
+  end
+end
+


sjamgade:
I did do some initial testing, here is a smple of how it could look like : https://github.com/crowbar/crowbar-openstack/commit/4fc8662e454b5f7464fb266baab17acfe8cffc99

This is not about replacing `include_recipe`, these both can live together side by side,

nicolasbock:
What I was trying to get at was that by replacing an `include_recipe` with an `include_recipe_smartly` in a recipe you are expecting a performance gain because the dependency is _not_ included.

Besides testing whether a dependency is really _not_ included when its attributes have not changed, we need to ask the following question:

For maximizing performance gain, the list of attributes of a dependency needs to be minimal in a sense that it contains only necessary attributes. However, a missing attribute introduces a regression.

How do we test that this set does not introduce a regression because it's missing something?

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -417,10 +417,52 @@ def self.size_to_bytes(s)
...
+      end
+    end
+
     class Config
       class << self
         attr_accessor :node
 
+        def last_two_configs(group, barclamp, instance)
+          prev_cfg = load(group, barclamp, instance)
+          cur_cfg = node[barclamp.to_sym].current_default
```
sjamgade: I had to choose/use `current_default` attribute here to get the underlying hash, however this includes the default keys, since some of these are never saved to databag, they do not show up in `prev_cfg` but always in `cur_cfg`. 

can someone suggest and way to not have these used for `prev_cfg, cur_cfg` comparison  ?

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -417,21 +418,96 @@ def self.size_to_bytes(s)
...
+              return nil
+            end
+            hash = hash[attrlist[i]]
+            i += 1
+            if attrlist.length == i
+              # Chef::Log.debug("[smart] loadattr return #{hash}")
+              return hash
+            end
+          end
+          Chef::Log.debug("[smart] hash does not respond to []")
```
skazi0: this (and the variable name) is misleading. if variable doesn't respond to `[]` it's most probably not a hash.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -417,21 +418,96 @@ def self.size_to_bytes(s)
...
+          i = 0
+          return hash if attrlist.empty?
+          while hash.respond_to?(:[])
+            unless hash.key?(attrlist[i])
+              Chef::Log.debug("[smart] hash is missing #{attrlist[i]}")
+              return nil
+            end
+            hash = hash[attrlist[i]]
+            i += 1
+            if attrlist.length == i
```
skazi0: wouldn't it be cleaner to shift items out of this list and forget about `i`? or do `each` on the list and check `respond_to` as normal condition inside the loop?

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -417,21 +418,96 @@ def self.size_to_bytes(s)
...
           end
 
-          # Accept environments passed as instances
           if instance =~ /^#{barclamp}-config-(.*)/
             instance = $1
           end
+          instance
+        end
+
+        def infer_instance(group, barclamp, instance)
```
skazi0: imho, in software "guess" and "infer" are pretty similar in meaning. I think the names of those two functions should give clear indication on how the instance is "found out".

chef/cookbooks/barclamp/libraries/barclamp_library.rb
sjamgade: hope this is cleaner now

chef/cookbooks/barclamp/libraries/barclamp_library.rb
skazi0: it could also be a single value. this is exactly when the log would trigger. I generally think that this message could be more informative (e.g. "<param name> is not a collection" or sth like this).

chef/cookbooks/barclamp/libraries/barclamp_library.rb
skazi0: it is :o)

chef/cookbooks/barclamp/libraries/barclamp_library.rb
sjamgade: exactly, it could be a single value or a collection ( i could not think of a word for that)
and that is why the log says the `collection did not respond to ...` because it is a int/string/....

sjamgade:
Modified the example test to include proposal level depedancy : https://github.com/sjamgade/crowbar-openstack/commit/954354efa52cd6058bee4d6fa2fded4e07b5c612

vuntz:
One key thing to be aware of is that not all barclamps save something to the data bags, and among the ones that do, not all of them simply copy the content of the proposal (eg: dns and ntp).

So the proposed approach here only works for the openstack barclamps. Which is still not too bad, but still something to be aware of.

sjamgade:
I didnot know that, thanks @vuntz 
Non-openstack barclamps are relatively fast and I dont think we need to consider them with respect to this PR.

any more reviews ? @skazi0

chef/cookbooks/barclamp/libraries/barclamp_library.rb
```diff
@@ -417,21 +418,93 @@ def self.size_to_bytes(s)
...
+        def changes_to_apply?(depedency, group = "openstack", instance = nil)
+          barclamp = depedency.shift
+          instance = guess_instance(barclamp, instance)
+          prev_cfg, curr_cfg = last_two_configs(group, barclamp, instance)
+          old = loadattr(prev_cfg, depedency)
+          new = loadattr(curr_cfg, depedency)
+          # Chef::Log.info("[smart] loadattr prev #{depedency}, #{prev_cfg}")
+          # Chef::Log.info("[smart] loadattr curr #{depedency}, #{curr_cfg.inspect}")
+          # Chef::Log.info("[smart] comparision #{old}, #{new}")
+          old != new
```
rhafer: Does this really work on "non-leave" attributes? E.g. if you specify  ["glance"]["api"] as a dependency. ICBW but my guess is that this will always be `true` in that case. Even if the attribute values didn't change.

sjamgade:
> > Non-openstack barclamps are relatively fast and I dont think we need to consider them with respect to this PR.
> 
> But I guess it means that if e.g. something (anything) changes in a barclamp that doesn't maintain the config databags in the way we expect it, we need to always run the full set of recipes, right? As far as I can see the current code doesn't really handle that case. Also keep in mind that even not all barclamps from crowbar-openstack are writing the config databags yet. At least database is not using it currently I think.

you are right, the code does not handle this case (recipes dependent on stuff outside the databags), but such recipes can continue to use `include_recipe` and **only** those which rely **entirely** on databag can use `include_recipe_smartly`
This PR just adds support for something like this and does not alter the current behavior.

> I haven't fully made up my mind on this change yet. On the one hand it introduces a bit of magic that can have weird side-effects (e.g. a local chef-client run will not revert up any manual config change that were done or crowbar_join might be missing stuff on boot up).  \<snip\>

local chef run can be asked to always run via a "config" option, which I will add if this PR sees the master branch. 
> 
> In general I think the code could benefit from a few more comments here and there to make it easier for reviewers to understand what it actually does.
Will do. Thanks.
> 
> ```
> mydependson = {
>     "horizon::server" => [
>         ["horizon"]       # depend on the entire proposal
> ```
> 
> Does this part really work with the current code?

yes I have been testing it. :-)
 
> BTW, I didn't really check this, but are you sure that this PR also works when e.g. the chef-client run on a node fails during apply. (Is the databag only written after a successful run?)

yes :-)

chef/cookbooks/barclamp/libraries/barclamp_library.rb
sjamgade: your skepticism (thanks for that) got me thinking, hth 

```ruby
irb(main):019:0> a = {'glance'=>{'api'=>{'aport'=>1111, 'bport'=>2222}}}
=> {"glance"=>{"api"=>{"aport"=>1111, "bport"=>2222}}}

irb(main):020:0> b = {'glance'=>{'api'=>{'aport'=>1111}}}
=> {"glance"=>{"api"=>{"aport"=>1111}}}

irb(main):021:0> a == b
=> false

irb(main):022:0> a != b
=> true

irb(main):023:0> aa = a['glance']['api']
=> {"aport"=>1111, "bport"=>2222}

irb(main):024:0> bb = b['glance']['api']
=> {"aport"=>1111}

irb(main):025:0> aa != bb
=> true

irb(main):026:0> b = {'glance'=>{'api'=>{'aport'=>1111, 'bport'=>2222}}}
=> {"glance"=>{"api"=>{"aport"=>1111, "bport"=>2222}}}

irb(main):027:0> bb = b['glance']['api']
=> {"aport"=>1111, "bport"=>2222}

irb(main):028:0> aa != bb
=> false
```

sjamgade:
I understand there is a lot of uncertainty about the behavior of almost all barclamps as everyone here is suggesting. 

But this PR is merely adding support for skipping unnecessary stuff and not actually affecting the real workflow. And as followup PR, I will add the option to disable skipping entirely.

Is there anything else we can do to move things forward here.

chef/cookbooks/barclamp/libraries/barclamp_library.rb
rhafer: Hm, fascinating. And it even seems to work when comparing a `Hash` as returned by Chef::Role with a `Mash` as returned by Chef::DataBagItem.

rhafer:
> I understand there is a lot of uncertainty about the behavior of almost all barclamps as everyone here is suggesting.

There is also some uncertainty (at least from me) about how much additional maintenance this requires. This finally adds another place where we carefully need to think and maintain dependencies between barclamps (at least of the barclamps for which we actually want to use it).

We haven't been particularly good at that in the past IMO :)
You're obviously right, though that this PR itself doesn't change any behavior for now until we start adapting the role_recipes. Maybe it's fine to just merge it and continue with experitentally enabling some roles for this.

> Is there anything else we can do to move things forward here.

I guess some more reviews/opinions form others might help. @aspiers maybe?

PR #1770, opened 2019-01-17
[5.0] upgrade-8-9: Revert "Disable upgrade API in Cloud8"

jsuchome:
This reverts commit 614c4e2f21a2a13ef9d74d09e8639cf7c6332e46.

For now this is just a test PR to see if upgrade of 8 to 9 is already possible. DO NOT MERGE YET


jsuchome:
@dirkmueller BTW once we merge this, we can get SOC8-SOC9 upgrade job to the point where chef-client fails on first upgraded node

dirkmueller:
lets merge it then..

PR #1773, opened 2019-01-17
upgrade: Adapt crowbar to handle 8-9 upgrade correctly

jsuchome:


crowbar_framework/app/controllers/application_controller.rb
```diff
@@ -34,12 +34,12 @@ class ApplicationController < ActionController::Base
...
   }
   before_filter :upgrade, if: proc {
-    if File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-compute-nodes-postponed")
+    if File.exist?("/var/lib/crowbar/upgrade/8-to-9-upgrade-compute-nodes-postponed")
       flash[:warning] = "Some nodes are not upgraded yet. " \
         "Continue to <a href='/upgrade'>upgrade wizard</a> to complete the upgrade."
     end
-    File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-running") &&
-      !File.exist?("/var/lib/crowbar/upgrade/7-to-8-upgrade-compute-nodes-postponed")
+    File.exist?("/var/lib/crowbar/upgrade/8-to-9-upgrade-running") &&
```
jsuchome: Hm, we might need `Dir.glob("/var/lib/crowbar/upgrade/*-upgrade-running").any?` instead

jsuchome: Done

PR #1774, opened 2019-01-18
[4.0] upgrade: Allow backup list and download during upgrade

skazi0:
When upgrade is active user should still be able to list and download
crowbar/admin server backups.


skazi0:
closed in favor of #1775

PR #1779, opened 2019-01-22
[5.0] Migrations after upgrade

jsuchome:
This is how it went during my test:
```
+ nova-manage db online_data_migrations --max-count 200
3 rows matched query populate_missing_availability_zones, 3 migrated
+---------------------------------------------+--------------+-----------+
|                  Migration                  | Total Needed | Completed |
+---------------------------------------------+--------------+-----------+
| delete_build_requests_with_no_instance_uuid |      0       |     0     |
|    migrate_aggregate_reset_autoincrement    |      0       |     0     |
|              migrate_aggregates             |      0       |     0     |
|      migrate_flavor_reset_autoincrement     |      0       |     0     |
|               migrate_flavors               |      0       |     0     |
|      migrate_instance_groups_to_api_db      |      0       |     0     |
|          migrate_instance_keypairs          |      0       |     0     |
|      migrate_instances_add_request_spec     |      0       |     0     |
|          migrate_keypairs_to_api_db         |      0       |     0     |
|       migrate_quota_classes_to_api_db       |      0       |     0     |
|        migrate_quota_limits_to_api_db       |      0       |     0     |
|     populate_missing_availability_zones     |      3       |     3     |
|     service_uuids_online_data_migration     |      0       |     0     |
+---------------------------------------------+--------------+-----------+
+ ret=1
+ [[ 1 = 1 ]]
+ nova-manage db online_data_migrations --max-count 200
+---------------------------------------------+--------------+-----------+
|                  Migration                  | Total Needed | Completed |
+---------------------------------------------+--------------+-----------+
| delete_build_requests_with_no_instance_uuid |      0       |     0     |
|    migrate_aggregate_reset_autoincrement    |      0       |     0     |
|              migrate_aggregates             |      0       |     0     |
|      migrate_flavor_reset_autoincrement     |      0       |     0     |
|               migrate_flavors               |      0       |     0     |
|      migrate_instance_groups_to_api_db      |      0       |     0     |
|          migrate_instance_keypairs          |      0       |     0     |
|      migrate_instances_add_request_spec     |      0       |     0     |
|          migrate_keypairs_to_api_db         |      0       |     0     |
|       migrate_quota_classes_to_api_db       |      0       |     0     |
|        migrate_quota_limits_to_api_db       |      0       |     0     |
|     populate_missing_availability_zones     |      0       |     0     |
|     service_uuids_online_data_migration     |      0       |     0     |
+---------------------------------------------+--------------+-----------+
+ ret=0
```

### DIFF
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -308,10 +308,18 @@ template "/usr/sbin/crowbar-reload-nova-after-upgrade.sh" do
   only_if { nova_node }
 end
 
-template "/usr/sbin/crowbar-run-nova-online-migrations.sh" do
-  source "crowbar-run-nova-online-migrations.sh.erb"
+template "/usr/sbin/crowbar-nova-migrations-after-upgrade.sh" do
+  source "crowbar-nova-migrations-after-upgrade.sh.erb"
   mode "0775"
   owner "root"
   group "root"
   only_if { roles.include?("nova-controller") }
+end
+
+template "/usr/sbin/crowbar-heat-migrations-after-upgrade.sh" do
+  source "crowbar-heat-migrations-after-upgrade.sh.erb"
+  mode "0775"
+  owner "root"
+  group "root"
+  only_if { roles.include?("heat-server") }
 end
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -981,24 +981,37 @@ module Api
       # Once all nova services are running Pike, it's possible to execute
       # online db migrations
       def run_online_migrations
-        nova_controllers = ::Node.find("roles:nova-controller")
-        return if nova_controllers.empty?
+        nova_nodes = ::Node.find("roles:nova-controller")
+        heat_nodes = ::Node.find("roles:heat-server")
+        return if nova_nodes.empty? && heat_nodes.empty?
 
-        if nova_controllers.size > 1
-          nova_controllers.select! { |n| n[:fqdn] == n[:pacemaker][:founder] }
+        nova_nodes.select! { |n| n[:fqdn] == n[:pacemaker][:founder] } if nova_nodes.size > 1
+        heat_nodes.select! { |n| n[:fqdn] == n[:pacemaker][:founder] } if heat_nodes.size > 1
+
+        if nova_nodes.any?
+          node = nova_nodes.first
+          save_node_action("Executing nova online migrations on #{node.name}...")
+          node.wait_for_script_to_finish(
+            "/usr/sbin/crowbar-nova-migrations-after-upgrade.sh",
+            timeouts[:online_migrations]
+          )
+          Rails.logger.info("Nova online migrations finished.")
         end
-        node = nova_controllers.first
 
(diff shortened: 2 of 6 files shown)


chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -308,10 +308,13 @@
...
-  source "crowbar-run-nova-online-migrations.sh.erb"
+template "/usr/sbin/crowbar-migrations-after-upgrade.sh" do
+  source "crowbar-migrations-after-upgrade.sh.erb"
   mode "0775"
   owner "root"
   group "root"
+  variables(
+    heat_server: roles.include?("heat-server")
+  )
   only_if { roles.include?("nova-controller") }
```
skazi0: what if we have heat-server on a node which doesn't have nova-controller?

jsuchome: Hm, I did not expect it , but it probably could happen.

So I guess I should just check if heat is deployed (and still call the action from nova controller to make it simpler)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -991,14 +991,30 @@ def run_online_migrations
...
-          "/usr/sbin/crowbar-run-nova-online-migrations.sh",
-          timeouts[:nova_online_migrations]
+          "/usr/sbin/crowbar-nova-migrations-after-upgrade.sh",
+          timeouts[:online_migrations]
         )
         Rails.logger.info("Nova online migrations finished.")
-        save_node_action("Nova online migrations finished.")
+
+        heat_servers = ::Node.find("roles:heat-server")
+        return if heat_servers.empty?
```
skazi0: is it possible that we have no `roles:nova-controller` but nodes with `roles:heat-server` are present? if so, then heat code will never run because of return above.

jsuchome: Is it possible to have heat without nova?

jsuchome: Probably is (although unlikely). PR adapted.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -982,23 +982,41 @@ def reload_nova_services
...
       def run_online_migrations
         nova_controllers = ::Node.find("roles:nova-controller")
-        return if nova_controllers.empty?
+        heat_servers = ::Node.find("roles:heat-server")
+        return if nova_controllers.empty? && heat_servers.empty?
 
-        if nova_controllers.size > 1
-          nova_controllers.select! { |n| n[:fqdn] == n[:pacemaker][:founder] }
+        if nova_controllers.any?
+          if nova_controllers.size > 1
```
skazi0: maybe simpler like this?:
before:
```
return if nova_controllers.empty? && heat_server.empty?

if nova_controller.any?
  if nova_controller.size > 1
    nova_controllers.select! ...
  end
  node = nova_controllers.first
  ...
end
...
```

after:
```
nova_controllers.select! ... if nova_controllers.size > 1 # might result in long lines
heat_controllers.select! ... if heat_controllers.size > 1

# at this point *_controllers should contain non-ha node/founder or nothing
if nova_controllers.any?
  node = nova_controllers.first
  ...
end
...
```

jsuchome: OK, good one

PR #1781, opened 2019-01-28
[5.0] upgrade: Restart nova-compute instead of relaoding via SIGHUP

rhafer:
To workaround a bug in nova-compute, we need to restart the server after
the upgrade instead of just sending a SIGHUP until a fix for
https://launchpad.net/bugs/1715374 is available.


rhafer:
There is an upstream fix under review currently (https://review.openstack.org/#/c/420026/) bug I guess it will still take a while until that lands and we can backport it to pike. So I guess we should try the workaround in parallel.

PR #1782, opened 2019-01-28
[4.0] upgrade: Add default upgrade timeouts file

skazi0:
By default there was no upgrade_timeouts.yml file installed so users
had to create one if they wanted to adjust some timeout. It's easier
for users to modify values in an existing file so file with defaults
was added.

Packaging SR will follow when this PR is merged.

### DIFF
--- a/configs/upgrade_timeouts.yml
+++ b/configs/upgrade_timeouts.yml
@@ -1,0 +1,17 @@
+:prepare_repositories: 120
+:pre_upgrade: 300
+:upgrade_os: 1500
+:post_upgrade: 600
+:shutdown_services: 600
+:shutdown_remaining_services: 600
+:evacuate_host: 300
+:chef_upgraded: 1200
+:router_migration: 600
+:lbaas_evacuation: 600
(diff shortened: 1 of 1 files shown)


configs/upgrade_timeouts.yml
```diff
@@ -0,0 +1,11 @@
...
+:pre_upgrade: 300
+:upgrade_os: 1500
+:post_upgrade: 600
+:evacuate_host: 300
+:chef_upgraded: 1200
+:router_migration: 600
+:lbaas_evacuation: 600
+:delete_pacemaker_resources: 300
+:delete_cinder_services: 300
+:wait_until_compute_started: 60
```
jsuchome: You're missing at least `reload_nova_services` and `online_migrations` ... and `shutdown_services`, `shutdown_remaining_services`

jsuchome: And `delete_nova_services`

skazi0: updated to SOC8 set of timeouts

PR #1790, opened 2019-01-31
[5.0] upgrade: Delay status switch after upgrade ends

skazi0:
Previously the "next upgrade" status file was used as soon as "current"
upgrade was finished. This sometimes caused problems when "final" status
was missed because it was already covered with new "not started yet" status.

With this change, "old" status will be returned for 24hrs after upgrade was
finished. After that subsequent calls to status API would return "new" status.

### DIFF
--- a/crowbar_framework/lib/crowbar/upgrade_status.rb
+++ b/crowbar_framework/lib/crowbar/upgrade_status.rb
@@ -30,13 +30,7 @@ module Crowbar
     # We're keeping the information in the file so is accessible by
     # external applications and different crowbar versions.
     def initialize(logger = Rails.logger, yaml_file = nil)
-      # If no upgrade is currently running, the default behavior
-      # is to start 8-9 upgrade.
-      # 7-8 upgrade can be only running because it was already started
-      # from Cloud7 (before admin server package upgrade)
-      if yaml_file.nil? || yaml_file.empty?
-        yaml_file = File.exist?(running_file_7_8) ? yaml_file_7_8 : yaml_file_8_9
-      end
+      yaml_file = current_status_file if yaml_file.nil? || yaml_file.empty?
 
       @running_file_location =
         if yaml_file == yaml_file_7_8
@@ -167,6 +161,8 @@ module Crowbar
           progress[:current_substep_status] = :finished
           progress[:current_nodes] = {}
           progress[:current_node_action] = "finished"
+          # Store time when the upgrade was finished
+          File.write(finished_file, Time.now.to_i)
         end
         next_step
         save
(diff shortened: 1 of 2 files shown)


crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -420,6 +416,26 @@ def next_step_to_execute
       upgrade_steps[i + 1]
     end
 
+    def upgrade_finished_recently?
+      return false unless File.exist?(finished_file)
+      # here "recently" means within last 24hrs
+      (Time.now.to_i - File.read(finished_file).to_i) < 24 * 60 * 60
```
jsuchome: So we know that the file exists, but is it readable and does it have the right content? Should we add some exception handling?

skazi0: in the end the question is: what should we return from this function in case of any problem with the file? can we assume that "upgrade didn't finish recently" in those cases?

jsuchome: That's the right question indeed, but I'm not sure about the answer.
If the upgrade has failed for some reason, we should not even get to this place, and `running` file should still exist.
It could be broken when writing (let's say, because of full disk), but that does not mean upgrade itself is unfinished.
Or user could have modified it.

skazi0: if the file was not corrupted or modified, the mtime should more or less match the contents.

jsuchome: Make sense.

crowbar_framework/lib/crowbar/upgrade_status.rb
```diff
@@ -420,6 +416,34 @@ def next_step_to_execute
...
     end
 
+    def upgrade_finished_recently?
+      return false unless File.exist?(finished_file)
+      finish_time = nil
+      begin
+        # read timestamp stored in the file
+        finish_time = File.read(finished_file).to_i
+      rescue StandardError
+        # if anything waw wrong with the file, fall back to last modification time
```
jsuchome: Typo: "waw"

skazi0:
recording from my `record_status.sh` script.

```
-rw-r--r-- 1 skazi users  725 2019-02-01 11:46:54.297291973 +0100 000087.json
-rw-r--r-- 1 skazi users  667 2019-02-02 11:46:54.160249282 +0100 000088.json
```

87 is the final status from 7-8 upgrade
88 is the "prechecks" status for 8-9 upgrade.

everything seems to work exactly as expected :o)

PR #1793, opened 2019-02-04
upgrade: Create update repositories with refresh enabled

dirkmueller:
We want the Update repositories to autorefresh because
they frequently change, and it is convenient if zypper
does the check for us rather than having to manually
"zypper ref -f" all the time.


dirkmueller:
(untested)

PR #1794, opened 2019-02-04
[5.0] upgrade: Create update repositories with refresh enabled

dirkmueller:
We want the Update repositories to autorefresh because
they frequently change, and it is convenient if zypper
does the check for us rather than having to manually
"zypper ref -f" all the time.

backport of #1793 
(cherry picked from commit 6d3b3549e73a43f233b18344f4e26033a09bc7da)


dirkmueller:
(untested)

PR #1800, opened 2019-02-27
upgrade: Make sure all compute nodes get compute related scripts

jsuchome:
There may be also other compute nodes than just kvm one , see https://bugzilla.suse.com/show_bug.cgi?id=1127129


skazi0:
@jsuchome how about `compute_node = roles.any? {|r| r.start_with? "nova-compute"}
`?

jsuchome:
Yeah, but `compute_nodes` are already present at this time (line 116), so why not use it?

PR #1802, opened 2019-03-04
crowbar: Restrict admin access (bsc#1117080) 

cmurphy:
Without this patch, crowbar-managed nodes have full admin access to the
crowbar API via both /etc/crowbarrc and /etc/crowbar.install.key. This
means users with root access on a crowbar-managed machine can modify
crowbar proposals and gain access to the crowbar node itself, if they
didn't already have access. This patch creates a special client user
that can only access the crowbar transition and node status APIs and
cannot modify proposals or other node attributes. This means that
crowbar-managed nodes can still change the transition state of other
nodes, which is unavoidable as long as the crowbar server is limited to
using digest auth, but at least this way they cannot change proposals or
do other destructive actions.

This change also changes the machine-install user's password once the
initial deployment is completed. This means that the password found in
/etc/crowbar.install.key as well as in the grub config and all over the
logs will no longer be able to be used to gain access to the crowbar
admin server. When the password is changed, new node deployments will
have the new key injected into their grub configs during provisioning,
and the key will be changed again on the next chef run.

Co-authored-by: Sumit Jamgade <sjamgade@suse.com>

### DIFF
--- a/chef/cookbooks/crowbar/recipes/default.rb
+++ b/chef/cookbooks/crowbar/recipes/default.rb
@@ -242,6 +242,34 @@ unless node["crowbar"].nil? or node["crowbar"]["users"].nil? or node["crowbar"][
     "/var/cache/chef/solr/data"
   end
 
+  # If all nodes are done installing, change the machine-install password so it
+  # can't be reused. This will change it on every chef run.  This only changes
+  # the install key on the admin node. This will be injected into future nodes
+  # that are added later.
+  all_nodes = node_search_with_cache("*:*")
+  unready_nodes = node_search_with_cache("NOT state:ready")
+  nodes_ready = all_nodes.length > 1 && unready_nodes.empty?
+  live_nodes = all_nodes.length
+  # Change the install key when all nodes are ready and the number of nodes in
+  # the deployment has changed since the last time it was set
+  if nodes_ready && live_nodes != node["crowbar"]["known_nodes"]
+    node["crowbar"]["known_nodes"] = live_nodes
+    secret = SecureRandom.hex(64)
+    node["crowbar"]["users"]["machine-install"]["password"] = secret
+    file "/etc/crowbar.install.key" do
+      content "machine-install:#{secret}\n"
+      owner "root"
+      group "root"
+      mode "0o600"
+    end
+    execute "update crowbar proposal with new machine-install password" do
+      command "crowbarctl proposal edit crowbar default \
+               --data \"$(crowbarctl  proposal show crowbar default --json | \
+               /opt/dell/bin/json-edit \
+                 --attribute attributes.crowbar.users.machine-install.password \
+                 --value #{secret} - )\""
+    end
+  end
   users = {}
   node["crowbar"]["users"].each do |k,h|
     next if h["disabled"]
@@ -253,6 +281,20 @@ unless node["crowbar"].nil? or node["crowbar"]["users"].nil? or node["crowbar"][
   template "/opt/dell/crowbar_framework/htdigest" do
     source "htdigest.erb"
     variables(users: users, realm: realm)
+    owner "root"
+    group node[:apache][:group]
+    mode "0640"
+  end
+
+  client_users = users.dup
+  client_username = node["crowbar"]["client_user"]["username"]
+  # Fix passwords into digests.
+  client_password = node["crowbar"]["client_user"]["password"]
+  client_digest = Digest::MD5.hexdigest("#{client_username}:#{realm}:#{client_password}")
+  client_users[client_username] = { "digest" => client_digest }
+  template "/opt/dell/crowbar_framework/htdigest-clients" do
+    source "htdigest.erb"
+    variables(users: client_users, realm: realm)
     owner "root"
     group node[:apache][:group]
     mode "0640"
--- a/chef/cookbooks/provisioner/recipes/base.rb
+++ b/chef/cookbooks/provisioner/recipes/base.rb
@@ -350,7 +350,14 @@ crowbar_node = node_search_with_cache("roles:crowbar").first
 address = crowbar_node["crowbar"]["network"]["admin"]["address"]
 protocol = crowbar_node["crowbar"]["apache"]["ssl"] ? "https" : "http"
 server = "#{protocol}://#{address}"
-password = crowbar_node["crowbar"]["users"]["crowbar"]["password"]
+is_admin = CrowbarHelper.is_admin?(node)
+if is_admin
+  username = "crowbar"
+  password = crowbar_node["crowbar"]["users"][username]["password"]
+else
+  username = crowbar_node["crowbar"]["client_user"]["username"]
+  password = crowbar_node["crowbar"]["client_user"]["password"]
+end
 verify_ssl = !crowbar_node["crowbar"]["apache"]["insecure"]
 
 package "ruby2.1-rubygem-crowbar-client"
@@ -359,6 +366,7 @@ template "/etc/crowbarrc" do
   source "crowbarrc.erb"
   variables(
     server: server,
+    username: username,
     password: password,
     verify_ssl: verify_ssl
   )
--- a/chef/data_bags/crowbar/migrate/crowbar/301_add_client_password.rb
+++ b/chef/data_bags/crowbar/migrate/crowbar/301_add_client_password.rb
@@ -1,0 +1,13 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  unless attrs.key?("client_user")
+    attrs["client_user"] = template_attrs["client_user"]
+    service = ServiceObject.new "fake-logger"
+    attrs["client_user"]["password"] = service.random_password
+  end
+  return attrs, deployment
+end
+
+def downgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs.delete("client_user") if attrs.key?("client_user")
+  return attrs, deployment
+end
--- a/chef/data_bags/crowbar/template-crowbar.schema
+++ b/chef/data_bags/crowbar/template-crowbar.schema
@@ -23,6 +23,12 @@
                 }
               }
             },
+            "client_user": { "type": "map", "required": true,
+              "mapping": {
+                "username": { "type": "str"},
+                "password": { "type": "str" }
+              }
+            },
             "upgrade": {
               "type": "map",
               "required": true,
--- a/bin/gather_logs.sh
+++ b/bin/gather_logs.sh
@@ -18,7 +18,8 @@
 
 if [[ -f /etc/crowbar.install.key ]]; then
     export CROWBAR_KEY=$(cat /etc/crowbar.install.key)
-    export CROWBAR_PASS="$(sed -e 's/^machine-install://' <<< $CROWBAR_KEY)"
+    export CROWBAR_USER="$(sed -e 's/:[^:]*$//' <<< $CROWBAR_KEY)"
+    export CROWBAR_PASS="$(sed -e 's/^.*://' <<< $CROWBAR_KEY)"
 fi
 mkdir -p /tmp/crowbar-logs
 tarname="${1-$(date '+%Y%m%d-%H%M%S')}"
@@ -48,10 +49,10 @@ sort_by_last() {
 	-o 'UserKnownHostsFile /dev/null')
     logs=(/var/log /etc)
     logs+=(/var/chef/cache /var/cache/chef /opt/dell/crowbar_framework/db)
-    crowbarctl node list -U machine-install -P $CROWBAR_PASS --no-verify-ssl
+    crowbarctl node list -U $CROWBAR_USER -P $CROWBAR_PASS --no-verify-ssl
 
     for to_get in proposals roles; do
-        crowbarctl $to_get proposal list crowbar -U machine-install -P $CROWBAR_PASS --no-verify-ssl
+        crowbarctl $to_get proposal list crowbar -U $CROWBAR_USER -P $CROWBAR_PASS --no-verify-ssl
     done
     for node in $(sudo -H knife node list); do
 	tarfile="${node%%.*}-${tarname}.tar.gz"
--- a/bin/transition.sh
+++ b/bin/transition.sh
@@ -22,7 +22,8 @@ if [[ $(cat /proc/cmdline) =~ $key_re ]]; then
 elif [[ -f /etc/crowbar.install.key ]]; then
     export CROWBAR_KEY="$(cat /etc/crowbar.install.key)"
 fi
-export CROWBAR_PASS="$(sed -e 's/^machine-install://' <<< $CROWBAR_KEY)"
+export CROWBAR_USER="$(sed -e 's/:[^:]*$//' <<< $CROWBAR_KEY)"
+export CROWBAR_PASS="$(sed -e 's/^.*://' <<< $CROWBAR_KEY)"
 
 if [ "$1" == "" ]
 then
@@ -41,7 +42,7 @@ do
   if [ $1 == $line ]
   then
     echo "Transitioning node $1 to state $2"
-    crowbarctl node transition $1 $2 -U machine-install -P $CROWBAR_PASS --no-verify-ssl
+    crowbarctl node transition $1 $2 -U $CROWBAR_USER -P $CROWBAR_PASS --no-verify-ssl
     break
   fi
 done
--- a/chef/cookbooks/crowbar/files/default/crowbar
+++ b/chef/cookbooks/crowbar/files/default/crowbar
@@ -23,7 +23,8 @@ IP="127.0.0.1"
 
 if [[ -f /etc/crowbar.install.key ]]; then
     export CROWBAR_KEY="$(cat /etc/crowbar.install.key)"
-    export CROWBAR_PASS="$(sed -e 's/^machine-install://' <<< $CROWBAR_KEY)"
+    export CROWBAR_USER="$(sed -e 's/:[^:]*$//' <<< $CROWBAR_KEY)"
+    export CROWBAR_PASS="$(sed -e 's/^.*://' <<< $CROWBAR_KEY)"
 fi
 
 case $1 in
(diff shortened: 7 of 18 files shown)


jsuchome:
Well... credentials stay in that `crowbar_join` script, am I right?

cmurphy:
[**@jsuchome**](https://github.com/jsuchome) commented on [Mar 4, 2019, 2:24 PM GMT+1](https://github.com/crowbar/crowbar-core/pull/1802#issuecomment-469251537 "2019-03-04T13:24:28Z - Replied by Github Reply Comments"):
> Well... credentials stay in that `crowbar_join` script, am I right?

Hmm yes :/

cmurphy:
Trying a different approach (not tested yet)

sjamgade:
there are some `Power off/on/restart` states too. 
crowbarctl will not be able to inform server when a node is rebooting

cmurphy:
[**@sjamgade**](https://github.com/sjamgade) commented on [Mar 8, 2019, 11:01 AM GMT+1](https://github.com/crowbar/crowbar-core/pull/1802#issuecomment-470873152 "2019-03-08T10:01:16Z - Replied by Github Reply Comments"):
> there are some `Power off/on/restart` states too. 
> crowbarctl will not be able to inform server when a node is rebooting
> 
> edit: unless they are unauthticated transitions

The non-admin nodes don't need crowbarctl to power themselves on and off, that should only be done from the admin node.

vuntz:
I think @sjamgade's point is that the node is reporting to crowbar when it's is powering off or restarting, through crowbar_join. And we should not break that.

Also, for the record, removing `/etc/crowbarrc` unfortunately doesn't help much if `/etc/crowbar.install.key` is still there (which I guess is still true nowadays).

sjamgade:
> I think @sjamgade's point is that the node is reporting to crowbar when it's is powering off or restarting, through crowbar_join. And we should not break that.
> 
yes, thanks, (note-to-self: I need to learn to communicate well)
> Also, for the record, removing `/etc/crowbarrc` unfortunately doesn't help much if `/etc/crowbar.install.key` is still there (which I guess is still true nowadays).
yes, its still there

cmurphy:
[**@vuntz**](https://github.com/vuntz) commented on [Mar 8, 2019, 11:21 AM GMT+1](https://github.com/crowbar/crowbar-core/pull/1802#issuecomment-470879214 "2019-03-08T10:21:16Z - Replied by Github Reply Comments"):
> I think @sjamgade's point is that the node is reporting to crowbar when it's is powering off or restarting, through crowbar_join. And we should not break that.

Ah okay, I think I understand.

> 
> Also, for the record, removing `/etc/crowbarrc` unfortunately doesn't help much if `/etc/crowbar.install.key` is still there (which I guess is still true nowadays).

The bug report is only about the regression introduced by https://github.com/crowbar/crowbar-core/pull/1305 so I intentionally only focused on that. Do you know when/whether it's safe to remove `/etc/crowbar.install.key`?

vuntz:
> The bug report is only about the regression introduced by #1305 so I intentionally only focused on that. 

Makes sense.

> Do you know when/whether it's safe to remove /etc/crowbar.install.key?

It's the same issue in the end: we only need this to allow crowbar_join to report a new state for the node to the crowbar server. There's no alternative; I agree it should be done, though.

cmurphy:
[**@sjamgade**](https://github.com/sjamgade) commented on [Mar 8, 2019, 11:01 AM GMT+1](https://github.com/crowbar/crowbar-core/pull/1802#issuecomment-470873152 "2019-03-08T10:01:16Z - Replied by Github Reply Comments"):
> there are some `Power off/on/restart` states too. 
> crowbarctl will not be able to inform server when a node is rebooting
> 
> edit: unless they are unauthticated transitions

Should we be accounting for all these states then? https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/node.rb#L287-L302

sjamgade:
I wonder if you want to run the chef-client (when a `reboot` is fired for the node) only to create a crowbarrc file. 

Since we rely entirely on apache to do auth, IMHO this issue can be only solved by limiting a special user for transitions by including a new `Location` in vhosts/crowbar.conf 
```
    <Location />
        AuthType Digest
        AuthName "SUSE OpenStack Cloud - Node Transitions "
        AuthDigestProvider file
        AuthUserFile /opt/dell/crowbar_framework/htdigest-transition
        Require valid-user
    </Location>
```

and include a new user in `htdigest-transition` with the new realm (AuthName) and install the new username password in crowbarrc.

cmurphy:
The digest user still has complete authorization on the whole crowbar server, doesn't it? We don't have any kind of access control, so creating a special user is the same as reusing the crowbar or machine-install user, or am I missing something?

sjamgade:
>     <Location /> 

` <Location crowbar/crowbar/1.0/transition/default/> `
 I missed this

cmurphy:
[**@sjamgade**](https://github.com/sjamgade) commented on [Mar 8, 2019, 3:15 PM GMT+1](https://github.com/crowbar/crowbar-core/pull/1802#issuecomment-470942367 "2019-03-08T14:15:08Z - Replied by Github Reply Comments"):
> >     <Location /> 
> 
> ` <Location crowbar/crowbar/1.0/transition/default/> `
>  I missed this

This is a good idea but if the problem is that if a node is allowed to access the "transition" command, they can access it for all nodes, including the crowbar node, not just their own. So I don't think that provides enough protection.

cmurphy:
Trying out @sjamgade 's idea

cmurphy:
@jsuchome @sjamgade @vuntz this is ready to re-review

chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -257,6 +257,21 @@
...
     mode "0640"
   end
+
+  client_users = users.dup
+  node["crowbar"]["client-users"].each do |k, h|
+    next if h["disabled"]
+    # Fix passwords into digests.
+    h["digest"] = Digest::MD5.hexdigest("#{k}:#{realm}:#{h["password"]}") if h["digest"].nil?
+    client_users[k] = h
+  end
```
jsuchome: Why do you need to create `client_users`? Can't you just pass same `users` argument to the template when creating htdigest-clients?

chef/data_bags/crowbar/migrate/crowbar/301_add_client_password.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
```
jsuchome: Maybe we do not plan to backport it, but to be sure, make this work for upgrades and check for the presence of the keys...

chef/cookbooks/crowbar/recipes/default.rb
cmurphy: The `users` variable doesn't contain the client user, and if it did that would mean the client machines could access `/`. So we need a separate variable that adds the client user as well as the other users to this new htdigest file.

chef/data_bags/crowbar/migrate/crowbar/301_add_client_password.rb
cmurphy: Good point, this will have to be backported to stable/5.0-pike

chef/cookbooks/crowbar/recipes/default.rb
jsuchome: Ah, now I know why I was confused, it's that `client_users = users.dup`. Is there a reason for that? Would it be better to initialize it as empty hash?

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -350,7 +350,14 @@
 address = crowbar_node["crowbar"]["network"]["admin"]["address"]
 protocol = crowbar_node["crowbar"]["apache"]["ssl"] ? "https" : "http"
 server = "#{protocol}://#{address}"
-password = crowbar_node["crowbar"]["users"]["crowbar"]["password"]
+is_admin = CrowbarHelper.is_admin?(node)
+if is_admin
+  username = "crowbar"
+  password = crowbar_node["crowbar"]["users"][username]["password"]
+else
+  username = crowbar_node["crowbar"]["client-users"].keys[0]
```
jsuchome: `...keys.first`?

chef/cookbooks/crowbar/recipes/default.rb
cmurphy: If it was initialized to an empty hash then we would still need to add the crowbar and machine-install user to it, as those users also need to use the transition API. Better to just copy the existing one.

chef/data_bags/crowbar/migrate/crowbar/301_add_client_password.rb
cmurphy: Done

chef/cookbooks/provisioner/recipes/base.rb
cmurphy: Done

chef/cookbooks/crowbar/recipes/default.rb
jsuchome: Thanks for explanation

cmurphy:
@sjamgade can you re-approve the latest update?

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -350,7 +350,14 @@
 address = crowbar_node["crowbar"]["network"]["admin"]["address"]
 protocol = crowbar_node["crowbar"]["apache"]["ssl"] ? "https" : "http"
 server = "#{protocol}://#{address}"
-password = crowbar_node["crowbar"]["users"]["crowbar"]["password"]
+is_admin = CrowbarHelper.is_admin?(node)
+if is_admin
+  username = "crowbar"
+  password = crowbar_node["crowbar"]["users"][username]["password"]
```
vuntz: Not really something new (as the bug was already there): we should check that there's a crowbar user, and if not, take another user. Nothing prevents the cloud admin to change usernames...

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -359,10 +366,18 @@
...
     password: password,
     verify_ssl: verify_ssl
   )
   owner "root"
   group "root"
   mode "0o600"
 end
+
+# Remove installer key, otherwise all root users could use it to talk to the crowbar server
+unless is_admin || ["hardware-installing", "installed", "readying"].include?(node[:state])
```
vuntz: I haven't checked, but why do we still need the file in these states? Is it because of some scripts that run (during autoyast/crowbar_register)? Would it be possible to do a followup (no need to block things here) to remove the need for this?

chef/data_bags/crowbar/template-crowbar.schema
```diff
@@ -23,6 +23,15 @@
                 }
               }
             },
+            "client-users": { "type": "map", "required": true, "mapping": {
```
vuntz: I think we generally avoid `-` in the attribute names (because it prevents a number of magic things in ruby).

Btw, any reason that it's a list? Do we anticipate we may need more than one user?

cmurphy:
> do we still need the machine-install user? Or could we replace it with this new client user?

The machine-install user is used here:

https://github.com/crowbar/crowbar-core/blob/master/bin/gather_logs.sh

to do `crowbarctl node list` and `crowbarctl proposal list` so I think it's not safe to remove it completely since it is doing things other than transitioning nodes.

> more importantly, the PXE files (/srv/tftpboot/discovery/x86_64/bios/pxelinux.cfg/default and maybe also all the node-specific ones) will still reference credentials that can do everything. So we need to fix that too. Could be done in a later step, but until this is done, things are not secure yet.

I'll try to see how we can possibly remove this from the client nodes. I worry that the way we've constructed this might make it impossible, as we've always assumed that it's safe to have these keys on the client nodes (I'm surprised we're deciding to worry about it just now).

chef/cookbooks/provisioner/recipes/base.rb
cmurphy: crowbar_join needs the key in place during these states. I rather doubt it would be possible to undo that.

chef/data_bags/crowbar/template-crowbar.schema
cmurphy: > I think we generally avoid - in the attribute names 

Will fix

> Btw, any reason that it's a list?

Just consistency with the other user list.

vuntz:
> I'm surprised we're deciding to worry about it just now

It's probably not the highest priority, I agree. There are very likely some bugs with higher priority.

cmurphy:
@jsuchome is the machine-install user needed for upgrades?

I can remove the /etc/crowbar.install.key and the credentials from the grub config, but the logs still contain all of the machine-install users' credentials, which still work for all crowbar operations. I will try to see how we can disable the machine-install user from the crowbar node itself after all the nodes have joined, but will nodes be able to upgrade after that?

jsuchome:
crowbar_join is executed from each node once it has been upgraded and rebooted. Now I do not know exactly if it machine-install user. I think it uses credentials from `/etc/crowbarrc` ...

jsuchome:
You might actually try adding this PR to https://ci.suse.de/view/Cloud/view/Cloud9/job/cloud-mkcloud9-job-upgrade-disruptive-x86_64. 
The job is red (fails during initial chef-client run) but we should be able to find out if this change causes some new failures...

cmurphy:
[**@jsuchome**](https://github.com/jsuchome) commented on [Mar 25, 2019, 4:57 PM GMT+1](https://github.com/crowbar/crowbar-core/pull/1802#issuecomment-476261090 "2019-03-25T15:57:13Z - Replied by Github Reply Comments"):
> crowbar_join is executed from each node once it has been upgraded and rebooted. Now I do not know exactly if it needs machine-install user. I think it uses credentials from `/etc/crowbarrc` ...

Okay, I see crowbar_join only executes `crowbarctl node transition` and `crowbarctl node status` so I think if the crowbar client user has permissions to those two APIs then that will be sufficient, if crowbar_join is the only thing that needs to work for upgrades.

cmurphy:
Actually, it wouldn't just be an issue with upgrading but also with adding new nodes. I don't think we can disable the machine-install user server-side if we expect the operator to add new nodes. And I don't think it's feasible to remove the machine-install credentials from all the logs on the client nodes.

cmurphy:
I think I've come up with a plan that works. This is what changed:

* fixed the client user to just be singular and not a list in the barclamp data bag template
* added /nodes/status as an allowed endpoint so that crowbar_join will work
* once nodes are deployed, change the machine-install user's password by changing /etc/crowbar.install.key on the crowbar node and changing the value in the proposal (via a gross hack)

We'll see how the mkcloud jobs do. I tested that it doesn't break things too much by doing a node reinstall and it works perfectly, getting the new machine-install key. I haven't been able to get mkcloud to do a setuplonelynode/crowbar_register for me so I haven't fully tested that part.

cmurphy:
mkcloud needs https://github.com/SUSE-Cloud/automation/pull/3293

cmurphy:
I think this is ready. @jsuchome @sjamgade @vuntz @rhafer @dirkmueller can you review/re-review?

vuntz:
I'm not clear why we still need the machine-install user? Or rather, why don't we move the machine-install user as a client user? Quickly looking around (and arguably I could have missed something), it seems that this user does not require access to other APIs than the ones that are restricted.

cmurphy:
The machine-install user credentials are used to set the CROWBAR_KEY environment variable which is used for every `crowbar` command call. It would be very destructive to disable that user completely.

vuntz:
@cmurphy point taken. But I guess the direction I was taking is that I feel that we do not need a privileged user outside of the admin server: even for provisioning, we could do with the restricted access. If I'm correct, there's no need in rotating the password for machine-install.

If we can't restrict the machine-install account, then we should change the provisioning process to use the new restricted account (that's changes in the provisioner cookbook as well as in `updates/control_lib.sh`)

vuntz:
For the record, I'm playing with that right now. The main problem is that the discovery process requires access to the machines controller, which we want to be restricted (it can be used to get access to attributes on a node, including secrets, and to reboot machines).

vuntz:
@cmurphy this is still WIP, but I've added a few more commits to https://github.com/vuntz/crowbar-core/commits/rm-crowbarrc

I also believe we should have a controller in the rails app that will deal with the restricted API calls, to simplify things a bit (clearer that we should not leak anything there, and also clearer what API calls are allowed).

vuntz:
I pushed another commit to my branch showing what I have in mind. This seems to work fine, but there are still things to do:

  - need to make `valid_transition_states.include?` usable from this controller (the transition fails because of this)
  - need to implement support for that in the client side
  - need to change crowbar_register, `updates/control_lib.sh` and crowbar_join to use the new calls

Is it useful, or is it heading in a wrong direction?

vuntz:
I've split my work:

* https://github.com/vuntz/crowbar-core/commits/rm-crowbarrc can possibly be merged in this PR (restarting some test to confirm it's all fine)
* https://github.com/vuntz/crowbar-core/commits/restricted-provisioning is my proposal for a next step; as said in the previous comment, it's not completely functional there, especially because the client side support for the new APIs needs to be written. But some curl tests were promising.

I wouldn't block this PR on the next step.

cmurphy:
I think it's the right direction if it works. Verifying your first branch locally first, then will incorporate it in this PR.

susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

vuntz:
@cmurphy for the next step... https://github.com/vuntz/crowbar-core/commits/restricted-provisioning together with https://github.com/vuntz/crowbar-client/commits/restricted-provisioning seems to be working now. I could only do some manual testing as patching crowbar-client for the discovery process is not fun (crowbar-client is in the sleshammer image), but if you could take a look, that'd be great.

Feel free to create the PRs already too if you feel it's not insane -- I'm calling it a day for now.

chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -242,6 +242,34 @@
...
+      group "root"
+      mode "0o600"
+    end
+    execute "update crowbar proposal with new machine-install password" do
+      command "crowbarctl proposal edit crowbar default \
+               --data \"$(crowbarctl  proposal show crowbar default --json | \
+               /opt/dell/bin/json-edit \
+                 --attribute attributes.crowbar.users.machine-install.password \
+                 --value #{secret} - )\""
+    end
```
rhafer: Hm, this means that the new password will only ever appear in the crowbar-config-default role when the user explicitly applies the crowbar "proposal". I guess that's what you're referring to with "gross hack" ? :)

I think this doesn't break anythink. It might just confuse users looking at the role with knife role show. And it will cause the orange exclamation mark to appear for the crowbar barclamp  all the time ("The proposal has been changed since last deployment") until somebody clicks apply.  But I think it's fine to address this in a subsequent change (if possible at all)

cmurphy:
Closing in favor of https://github.com/crowbar/crowbar-client/pull/187 and https://github.com/crowbar/crowbar-core/pull/1817

PR #1804, opened 2019-03-05
dns: allow using dns-server as designate target

sjamgade:
Designate needs an rndc key to CRUD zones on the real bind9 master
server, and updates to bind9 can only be sent via a key that is
mentioned in its(bind9) config.

So the key is supposed to be shared across the nodes, (admin and
designate-worker) therefore adding it to the proposal. one could
rsync the key between the nodes but this opens up differnt
types of race conditions/syncronization problems.

- added rndc key field to dns proposal
- a default key is pregnerated (512 bits)
- the designate barclamp will install this key on desginate-worker
nodes as these nodes need authoritative command on the dns-server to
create zones
-  syntax for the controls clause[1]

1 - http://www.zytrax.com/books/dns/ch7/controls.html

### DIFF
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -390,6 +390,9 @@ end
 
 ### FIXME Change to "any" once IPv6 support has been implemented
 admin_addr6 = "none"
+if node[:dns][:enable_designate] && !node[:dns][:master]
+  node[:dns][:forwarders].push master_ip
+end
 
 # Rewrite our default configuration file
 template "/etc/bind/named.conf" do
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -400,7 +400,7 @@ template "/etc/bind/named.conf" do
             allow_transfer: allow_transfer,
             ipaddress: admin_addr,
             ip6address: admin_addr6,
-            enable_designate: node[:dns][:enable_designate]
+            enable_designate: node[:dns][:enable_designate] && node[:dns][:master]
            )
   notifies :restart, "service[bind9]", :immediately
 end
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -294,7 +294,8 @@ when "suse"
 end
 
 # We would like to bind service only to ip address from admin network
-admin_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+admin_network = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin")
+admin_addr = admin_network.address
 
 # Load up our default zones.  These never change.
 if node[:dns][:master]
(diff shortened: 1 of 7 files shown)


sjamgade:
this is ready to be reviewed

sjamgade:
I am not really sure about the last commit, because we do have `nameserver=dns-master` on all nodes, so the zone "should" be resolvable.  But I could not get it to work, glad if someone could shed some light.

sjamgade:
> Looks ok to me, just one question: this grants full RNDC access to Designate...is that the only way it can update DNS, or is it possible to use AXFR - which can enabled on a per-zone basis?
> 
> I'm just asking out of curiosity, though. I'd be happy with a RNDC based implementation, even if AXFR was possible. We can always implement a choice between the two later.

if you look at the [bind9 implementation](http://git.openstack.org/cgit/openstack/designate/tree/designate/backend/impl_bind9.py#n72) it expects an rndc control, explicitly sets everyone slave.  And I am not sure I know/understand the alternative way you mentioned.

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -373,6 +390,9 @@ def make_zone(zone)
 
 ### FIXME Change to "any" once IPv6 support has been implemented
 admin_addr6 = "none"
+if node[:dns][:enable_designate] && !node[:dns][:master]
+    node[:dns][:forwarders].push master_ip
```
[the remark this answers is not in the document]
dirkmueller: @sjamgade could you fix this one?

crowbar_framework/app/models/dns_service.rb
```diff
@@ -49,6 +49,11 @@ def validate_proposal_after_save proposal
       validation_error("At least one nameserver or one node with the dns-server role must be specified.")
     end
 
+    if proposal["attributes"]["dns"]["enable_designate"]
+      rndckey = proposal["attributes"]["dns"]["designate_rndc_key"]
+      validation_error("designate_rndc_key cannot be empty with designate enabled") if rndckey.empty?
```
dirkmueller: I wonder if we want to just generate one? that would basically be a migration for the dns proposal..

sjamgade: generated one during the migration. but I would still like to keep the check

sjamgade:
> Looks okay to me. I'm curious if we need to change something more for the designate integration though, like for example making the dns server listen to the public network

I was able to resolve a designate-zone's SOA from one of the vms. 

`dnsmasq_dns_servers` config from dhcp_agent.ini contains `node[:dns][:forwarders].join(",")`  which is list of ip address for admin network

chef/data_bags/crowbar/migrate/dns/301_add_rndckey.rb
```diff
@@ -0,0 +1,12 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  service = ServiceObject.new "fake-logger"
+  deployment["designate_rndc_key"] = service.random_password(64)
```
dirkmueller: this needs to be something like:

```
unless @@dns_designate_rndc_key
   @@dns_designate_rndc_key = service.random_password(64)
end
deployment["designate_rndc_key"] =  @@dns_designate_rndc_key

to ensure that we don't accidentally change the rndc key when applying a proposal.

sjamgade: whats a `@@` ?

PR #1808, opened 2019-03-13
Fixup uid/gids if they don't match (SCRD-8014)

JanZerebecki:
To fix the ones on disk from rpm packages by reinstalling the package in
question with --force.

The rpm package scripts in question just pick a free id. It can not be
ensured that creating the user/groups with specific ids is done before
the packages are installed as crowbar_register may be run on an existing
system that has one or more of the packages already installed. Also
there is the update case, though less likely by now as this code already
existed in the last release.

Note that this does not fix up any on disk uid/gids for files that may
have been created by crowbar (though I don't know any that are) or
manually.

### DIFF
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -99,8 +99,10 @@ add_group() {
   if [[ -z "$gid" ]]; then
     groupadd --system --gid $2 $1
   elif [[ "$gid" != "$2" ]]; then
-    echo "Group $1 does not have gid $2."
-    exit 1
+    groupmod -g $2 $1
+    if [[ -n "$3" ]] && rpm -q $3 >/dev/null; then
+      zypper --non-interactive install --force $3
+    fi
   fi
 }
 
@@ -110,8 +112,15 @@ add_user() {
   if [[ -z "$uid" ]]; then
     useradd --system --shell /sbin/nologin -d / --gid $2 --uid $2 --groups $3 $1
   elif [[ "$uid" != $2 ]] || [[ "$gid" != $2 ]]; then
-    echo "User $1 does not have uid/gid $2."
-    exit 1
+    if [[ -n "$gid" ]]; then
+      echo "Group $1 for the same username doesn't exist. Please clean up manually."
+      exit 1
+    fi
+    usermod -u $2 $1
+    groupmod -g $2 $1
+    if [[ -n "$4" ]] && rpm -q $4 >/dev/null; then
+      zypper --non-interactive install --force $4
+    fi
   fi
 }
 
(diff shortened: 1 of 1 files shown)


chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -110,8 +112,15 @@ add_user() {
   if [[ -z "$uid" ]]; then
     useradd --system --shell /sbin/nologin -d / --gid $2 --uid $2 --groups $3 $1
```
cmurphy: With `--groups $3` this will create a user with group `openstack-glance` which is incorrect

JanZerebecki: I don't follow. According to the man page `useradd --groups` doesn't create groups, only adds the user also to that group. For the glance case `$3` is `glance` and only `$4` is `openstack-glance`. Did I miss anything?

cmurphy: You're right, I was looking at the add_group on line 196 and not at add_user. Sorry for the confusion.

cmurphy:
This will not only reinstall a package if it was there but install it fresh if it wasn't there, are you sure that's the right behavior. I don't think all nodes necessarily need to have openstack-glance, openstack-nova, and qemu installed on them, right?

Do you know what the real reason is for forcing the uid/gids instead of letting the package choose them? Does that reason still apply?

JanZerebecki:
>  but install it fresh if it wasn't there

I'll need to fix that.

> Do you know what the real reason is for forcing the uid/gids instead of letting the package choose them?

According to commit history the reason is when multiple nodes share an NFS mount. E.g. for sharing glance images. Also I heard certain types of testing are done by using NFS, I don't remember what exactly.
With NFSv4 and some setup this could still work with different ids: https://unix.stackexchange.com/questions/438939/how-to-get-nfsv4-idmap-working-with-sec-sys/464950#464950 (I checked and the v4 nfsidmap package is available in SLES 12 SP2, 3 and 4. A reason this might not work is if the server in use doesn't support NFSv4.)

Do you have any indication that we can rely on the later
 or don't need this type of NFS sharing feature at all?

cmurphy:
> According to commit history the reason is when multiple nodes share an NFS mount. E.g. for sharing glance images

So maybe we only need to do this for the glance user? A conflict with a preexisting glance user is probably much less likely than a conflict with a preexisting kvm user.

> Do you have any indication that we can rely on the later
or don't need this type of NFS sharing feature at all?

I have no information about that.

JanZerebecki:
>  but install it fresh if it wasn't there

fixed.

> So maybe we only need to do this for the glance user?

And cinder for the same reason.

> A conflict with a preexisting glance user is probably much less likely than a conflict with a preexisting kvm user.

Yes, maybe we can drop the kvm/qemu ones. At least the reason give in https://github.com/crowbar/crowbar-core/commit/9bb9eaa9948f7653bfd6aa47c75b58dd48c5eb3c is not relevant anymore. But that might be the riskier move (as we are close to the release). Your call.

cmurphy:
> Yes, maybe we can drop the kvm/qemu ones. At least the reason give in 9bb9eaa is not relevant anymore. But that might be the riskier move (as we are close to the release). Your call.

Maybe something we can revisit later then.

PR #1811, opened 2019-03-22
Octavia: Definition of octavia network

ilausuch:
Octavia requires a management network to comunicate with amphora VMs


ilausuch:
Ok, let's do that. I am going to close this PR

PR #1812, opened 2019-03-27
Fix PostgreSQL 10 support (SCRD-7126)

rsalevsky:
PostgreSQL 10 isn't officially supported in rails 4.X, see
https://help.heroku.com/WKJ027JH/rails-error-after-upgrading-to-postgres-10.
As a workaround we can monkeypatch active record code.


rsalevsky:
I wont fix the hound issues as it is a direct backport from Activerecord.

PR #1813, opened 2019-03-27
barclamp: Fix setting MTU on networks using a bridge

vuntz:
When add_bridge is set to true, we set the MTU on the bridge interface,
but nothing sets it on the slave interface, so it doesn't work.

So whenever we set the MTU on a bridge, also set it on the slave
interface.


toabctl:
Backport for stable/5.0 is https://github.com/crowbar/crowbar-core/pull/1845

toabctl:
Backport for stable/4.0 is https://github.com/crowbar/crowbar-core/pull/1846

PR #1815, opened 2019-03-27
Use same line-height for all buttons (SCRD-8582)

eapodaca:
Button groups can contain inputs and anchor tags that should be styled similar. Removing this over-ride because they are consistent with out it. This is especially true on barclamp edit pages.

### DIFF
--- a/crowbar_framework/app/assets/stylesheets/content/content.scss
+++ b/crowbar_framework/app/assets/stylesheets/content/content.scss
@@ -132,10 +132,6 @@ h3.modal-title {
   }
 }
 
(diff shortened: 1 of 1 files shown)


skazi0:
@eapodaca please extend the commit message with some explanation/example of what this change fixes

jeremy-moffitt:
@skazi0 I saw Ethan updated the message, not sure if this is sufficient or needs to be revisited again, please let us know, thanks!

skazi0:
@jeremy-moffitt looks good except the spelling mistakes :o)

PR #1817, opened 2019-04-03
Add some API to allow non-privileged user to be used for discovery/crowbar_join/crowbar_register (bsc#1117080)

vuntz:
The goal here is to remove the use of the `machine-install` user (who has access to all API, and can therefore do destructive operations) from non-trusted clients. This is really about crowbar_join, crowbar_register and the whole discovery/allocation process.

To that end, a new set of "restricted" APIs is introduced so that we fully understand what we allow access to.

This also depends on a crowbar-client change (and CI is expected to fail until that change lands): https://github.com/crowbar/crowbar-client/pull/187

The "Restricted" name is probably bad... but nothing better came to my mind yesterday.

Note that it builds on top of (and therefore contains) https://github.com/crowbar/crowbar-core/pull/1802


susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

cmurphy:
This is missing the part where the actual machine-install user is removed from the htdigest file. Once that is done then the whole password change hack won't be needed at all.

I'm trying to figure out how to test this properly because I think there will be corner cases here that are hard to find without testing, however I can't get the sleshammer image to build yet: https://build.suse.de/project/monitor/home:comurphy:branches:Devel:Cloud:9:Staging?arch_x86_64=1&defaults=0&failed=1&repo_SLE_12_SP4=1

I can't think of a better name than "restricted" either.

vuntz:
> This is missing the part where the actual machine-install user is removed from the htdigest file. Once that is done then the whole password change hack won't be needed at all.

I was thinking about that earlier today; I think we also need to change the deprecated crowbar CLI to read /etc/crowbarrc (you mentioned that earlier), and see if we can remove /etc/crowbar.install.key from the admin server (with this PR, it's already gone on other nodes).

vuntz:
> however I can't get the sleshammer image to build yet:

Yeah, I'm building it on my machine now to work around this.

vuntz:
I clearly didn't take into account the horrible upload time... While it's still uploading: I can confirm that `crowbar_register` works fine with the patches. Which also means that `crowbar_join` is fine too. That really leaves the discovery/allocation process.

vuntz:
There was a bug (forgot to change `--json` to `--plain` in one file after a change); after fixing this, I can go through the allocation process just fine.

cmurphy:
I was able to get the sleshammer image built locally and deployed but there seems to still be an issue even with the latest version of this branch, the nodes never become ready and just go into an infinite reboot loop.

vuntz:
@cmurphy does your sleshammer image contain a patched crowbar-client? You can check that by looking at the apache logs for the crowbar vhost, and look at the requests being made there.

vuntz:
(I'm assuming you fetched the latest from the branch, including the fix I made a couple of hours ago)

cmurphy:
It was my mistake, the sleshammer image was fine but I didn't have the update repos set up correctly and so the deployed image didn't get the right version of crowbar-client installed. After I got it working it runs smoothly, power functions and reinstall functions work fine.

I think we should completely revert the part of my change that changes the machine-install user's password, I sometimes see race conditions where the key is changed on disk but CROWBAR_KEY is already set in the environment. We don't have to completely remove the machine-install user from the crowbar admin node yet because there's no trace of that user on the client nodes any more.

I think it would also be good to squash some of these commits before merging.

Otherwise LGTM, thanks for jumping in and helping with this.

vuntz:
> I think we should completely revert the part of my change that changes the machine-install user's password

Done.

> I think it would also be good to squash some of these commits before merging.

I've cleaned up commits that are not in https://github.com/crowbar/crowbar-core/pull/1802; the commits from that PR are good from my perspective, and don't need any cleanup.

vuntz:
The CI will not be green until we have a new release of crowbar-client in the product, with https://github.com/crowbar/crowbar-client/pull/187

vuntz:
> Might it possible somehow to restrict request authenticated as "client_user" to just do stuff for the node from which they're calling?

I was thinking about that as a next step. It's not trivial actually, because of the discovery bit which by definition comes from an unknown node. Couldn't find a good solution.

The good news, however, is that yes, you can get chef-client to change behavior (or even not run, due to the missing client key). But this should not impact what is running on the node in the end.

vuntz:
Actually, maybe we can enforce the restricted API to go through a specific flow for transitions.

cmurphy:
[**@vuntz**](https://github.com/vuntz) commented on [Apr 4, 2019, 8:43 AM PDT](https://github.com/crowbar/crowbar-core/pull/1817#issuecomment-479951665 "2019-04-04T15:43:29Z - Replied by Github Reply Comments"):
> Actually, maybe we can enforce the restricted API to go through a specific flow for transitions.

I think we have a deadline to complete this by Monday...so I would suggest calling this good enough and continuing to optimize later?

chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -242,34 +242,6 @@
...
-      mode "0o600"
-    end
-    execute "update crowbar proposal with new machine-install password" do
-      command "crowbarctl proposal edit crowbar default \
-               --data \"$(crowbarctl  proposal show crowbar default --json | \
-               /opt/dell/bin/json-edit \
-                 --attribute attributes.crowbar.users.machine-install.password \
-                 --value #{secret} - )\""
-    end
-  end
```
cmurphy: This commit could be squashed into 004169812f8b2a49354710b080e9a431d1a444ef, no need to add this and then undo it in the same PR.

cmurphy:
The PR and/or commits should also reference the bsc #.

vuntz:
> > Actually, maybe we can enforce the restricted API to go through a specific flow for transitions.
> 
> I think we have a deadline to complete this by Monday...so I would suggest calling this good enough and continuing to optimize later?

Agreed.

> The PR and/or commits should also reference the bsc #.

Commits already do, adding it to the PR now.

chef/cookbooks/crowbar/recipes/default.rb
vuntz: Yeah, I thought we'd merge your PR first and then consider only the new commits. Let me clean this up further, then.

vuntz:
> Actually, maybe we can enforce the restricted API to go through a specific flow for transitions.

Since I had some bandwidth, I went ahead and did that. I also tested it.

@rhafer with that, I don't see how the API can be abused anymore. What do you think?

Note that I didn't enforce the flow for non-restricted transitions, as I think it's valid for the admin to override the state of a node through a transition.

rhafer:
> with that, I don't see how the API can be abused anymore. What do you think?

@vuntz Yeah. That seems to cover it.

vuntz:
> I guess it would require at least the crowbarctl change to be merged first and an updated sleshammer image being build and included in the D:C:9:S images, right?

Yes, we need https://github.com/crowbar/crowbar-client/pull/187 to be merged and in D:C:9 so that we get a new sleshammer. And then the CI should be working. @rsalevsky will look at the crowbar-client bits today.

rsalevsky:
Needs rebase.

vuntz:
rebased

PR #1818, opened 2019-04-03
[4.0] network: run wicked ifdown for interface cleanup (bsc#1063535)

stefannica:
#1441 backport

When cleaning up vlan interfaces, simply running 'vconfig rem'
doesn't have any effect for interfaces that are already managed
by wicked. Only running 'wicked ifdown' properly cleans up these
interfaces.

Instead of patching only the vlan interfaces in the nic.rb library,
all interfaces are brought down with 'wicked ifdown' in the network
barclamp when removed.

(cherry picked from commit 8f15749f2834f3c490358a59f44278d79078996f)


stefannica:
already backported in #1831

PR #1819, opened 2019-04-04
Remove dubious uses of machine-install user

vuntz:
Both gather_logs.sh and transition.sh seem to use it, for no good reason.

### DIFF
--- a/bin/gather_logs.sh
+++ b/bin/gather_logs.sh
@@ -16,10 +16,6 @@
 # limitations under the License.
 #
 
-if [[ -f /etc/crowbar.install.key ]]; then
-    export CROWBAR_KEY=$(cat /etc/crowbar.install.key)
-    export CROWBAR_PASS="$(sed -e 's/^machine-install://' <<< $CROWBAR_KEY)"
-fi
 mkdir -p /tmp/crowbar-logs
 tarname="${1-$(date '+%Y%m%d-%H%M%S')}"
 targetdir="/opt/dell/crowbar_framework/public/export"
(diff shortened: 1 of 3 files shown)


bin/gather_logs.sh
```diff
@@ -48,11 +44,24 @@ sort_by_last() {
 	-o 'UserKnownHostsFile /dev/null')
     logs=(/var/log /etc)
     logs+=(/var/chef/cache /var/cache/chef /opt/dell/crowbar_framework/db)
-    crowbarctl node list -U machine-install -P $CROWBAR_PASS --no-verify-ssl
 
-    for to_get in proposals roles; do
-        crowbarctl $to_get proposal list crowbar -U machine-install -P $CROWBAR_PASS --no-verify-ssl
+    mkdir nodes
+    for node in $(crowbarctl node list --no-verify-ssl --no-meta --plain); do
+        crowbarctl node show $node --no-verify-ssl --json > nodes/$node.json
```
cmurphy: If I understand correctly, this will be collected by the `find` on line 73 now?

cmurphy:
The crowbarrc file is created by chef, is it guaranteed that it will exist before gather_logs.sh and transition.sh need to use it?

bin/gather_logs.sh
vuntz: Correct. Anything that is not a file will not end up in the tarball generated by that script and is therefore useless.

(we don't really recommend using that script since we have supportconfig, so that's why noone ever noticed)

vuntz:
> The crowbarrc file is created by chef, is it guaranteed that it will exist before gather_logs.sh and transition.sh need to use it?

These scripts are not run before the install process is complete; during the install process, one of the chef-client run will create /etc/crowbarrc. So from my perspective, yes, it is guaranteed.

PR #1820, opened 2019-04-04
bin: Read /etc/crowbarrc instead of /etc/crowbar.install.key

vuntz:
We also stop using the CROWBAR_KEY environment variable and use
CROWBAR_USERNAME and CROWBAR_PASSWORD, like crowbarctl.

This depends on https://github.com/crowbar/crowbar/pull/2379 (which creates /etc/crowbarrc during install-chef-suse, and also removes the use of the crowbar CLI in that script).

This removes a good part of the use of the machine-install user.

### DIFF
--- a/bin/barclamp_lib.rb
+++ b/bin/barclamp_lib.rb
@@ -19,6 +19,7 @@ require "rubygems"
 require "net/http"
 require "net/http/digest_auth"
 require "uri"
+require "inifile"
 require "json"
 require "getoptlong"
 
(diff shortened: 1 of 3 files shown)


bin/barclamp_lib.rb
```diff
@@ -529,8 +536,9 @@ def parse_extra_opt(opt, arg)
 
 def check_user_password
   if @username.nil? or @password.nil?
-    STDERR.puts "CROWBAR_KEY not set, will not be able to authenticate!"
-    STDERR.puts "Please set CROWBAR_KEY or use -U and -P"
+    STDERR.puts "Incomplete credentials, will not be able to authenticate!"
+    STDERR.puts "Please a crowbarrc configuration file, " \
```
jsuchome: I feel the verb is missing here

vuntz: Thanks, as agreed, I'll fix in the next PR (to avoid retriggering CI here and move on the other PR needed on GMC1).

PR #1821, opened 2019-04-06
Only use /etc/crowbarrc for storing crowbar admin credentials (SCRD-7537)

vuntz:
Storing the admin credentials for crowbar in the crowbar proposal still enables a user on a non-admin node to do some destructive operations (reinstalling a node), as the non-admin node has access to chef. Instead, we now only rely on /etc/crowbarrc for the credentials, which crowbar will not modify.

This is related to bsc#1117080, SCRD-7537, CVE-2018-17954.

This needs to be merged at the same time as these other changes:
 - https://github.com/crowbar/crowbar/pull/2380
 - https://github.com/yast/yast-crowbar/pull/63
 - https://github.com/SUSE-Cloud/doc-cloud/pull/897

~~Note: this contains both https://github.com/crowbar/crowbar-core/pull/1817 and https://github.com/crowbar/crowbar-core/pull/1820.~~


susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please rebase to get rid of the merge commits in this PR</td>
    </tr>
  </tbody>
</table>

</p>

vuntz:
While I triggered suse/mkcloud again, the last failure (https://ci.suse.de/job/openstack-mkcloud/164580/console) was one tempest test failure, unrelated to this change and CI in https://github.com/crowbar/crowbar/pull/2380 shows that it's actually fine.

chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -231,7 +231,15 @@
...
 end
 
-unless node["crowbar"].nil? or node["crowbar"]["users"].nil? or node["crowbar"]["realm"].nil?
+if !node["crowbar"].nil? && !node["crowbar"]["realm"].nil?
+  # After installation of a gem, we have a new path for the new gem, so we
+  # need to reset the paths if we can't load the gem
+  begin
+    require "inifile"
+  rescue LoadError
+    Gem.clear_paths
```
jsuchome: Pardon my ruby ignorance, but does this mean that after `Gem.clear_paths`, `inifile` can be succesfuly used later in this block?

chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -244,11 +252,19 @@
...
-  node["crowbar"]["users"].each do |k,h|
-    next if h["disabled"]
-    # Fix passwords into digests.
-    h["digest"] = Digest::MD5.hexdigest("#{k}:#{realm}:#{h["password"]}") if h["digest"].nil?
-    users[k] = h
+
+  begin
+    crowbarrc = IniFile.load("/etc/crowbarrc") || {}
+  rescue IniFile::Error
+    Chef::Log.warn("Could not parse config file /etc/crowbarrc")
```
jsuchome: Any reason why this is not a fatal error?

chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -231,7 +231,15 @@
   action :create
 end
 
-unless node["crowbar"].nil? or node["crowbar"]["users"].nil? or node["crowbar"]["realm"].nil?
+if !node["crowbar"].nil? && !node["crowbar"]["realm"].nil?
```
jsuchome: Could this be written as
`if node["crowbar"] && node["crowbar"]["realm"]` ?

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -346,31 +346,71 @@
...
+    crowbarrc_config = crowbarrc["default"]
+
+    if server != crowbarrc_config["server"]
+      crowbarrc_config["server"] = server
+      Chef::Log.info("Will update \"server\" option in /etc/crowbarrc to \"#{server}\"")
+      do_save = true
+    end
+
+    crowbarrc_verify_ssl = crowbarrc_config["verify_ssl"].nil? ||
+      ![false, 0, "0", "f", "F", "false", "FALSE"].include?(crowbarrc_config["verify_ssl"])
```
jsuchome: WOW

chef/cookbooks/crowbar/recipes/default.rb
vuntz: Yes

chef/cookbooks/provisioner/recipes/base.rb
vuntz: Yeah, that's the same bits we have in crowbar-client ;-)

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -346,31 +346,71 @@
...
+
+    if protocol == "http" && !crowbarrc_config["verify_ssl"].nil?
```
jsuchome: I would prefer `crowbarrc_config.key?("verify_ssl")` to get rid of double negatives, but that's just matter of taste

chef/cookbooks/crowbar/recipes/default.rb
vuntz: If this happens, we simply don't change the htdigest bits and let everything go. The reason this matters is that having chef-client fail on the admin server is pretty bad -- it would stop the discovery process from working for instance (since the dhcp config wouldn't get updated anymore).

chef/cookbooks/crowbar/recipes/default.rb
vuntz: I guess it can, yes. I didn't want to move away from the `unless` syntax to keep changes minimal but hound made me do this ;-)

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -346,31 +346,71 @@
...
+      crowbarrc_config.delete("verify_ssl")
+      Chef::Log.info("Will remove \"verify_ssl\" option in /etc/crowbarrc")
+      do_save = true
+    elsif protocol == "https" && verify_ssl != crowbarrc_verify_ssl
+      crowbarrc_config["verify_ssl"] = verify_ssl ? 1 : 0
+      Chef::Log.info("Will update \"verify_ssl\" option in /etc/crowbarrc to " \
+          "\"#{crowbarrc_config["verify_ssl"]}\"")
```
jsuchome: Is there a reason why you choose to use 0/1 as default values instead of just the boolean value of `verify_ssl`?

vuntz: That's what we did in the past (see the old code that is just moved)

chef/cookbooks/provisioner/recipes/base.rb
vuntz: Happy to fix this in a followup PR ;-)

chef/cookbooks/crowbar/recipes/default.rb
jsuchome: OK, so if you do followup PR, could you put this explanation into a comment?

chef/cookbooks/crowbar/recipes/default.rb
rhafer: Hm, I don't quite get that reasoning :). The current state is rather ugly.

vuntz:
I'll do a followup in the morning with the cleanups

PR #1822, opened 2019-04-08
Fix cloud-mkcloud9-job-backup-restore (SCRD-7126)

chajain:
Added code to fix the ownership of admin node log file to crowbar:crowbar after restore since it run chef-client as root and creates the log file with ownership root:root

### DIFF
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -222,6 +222,13 @@ module Crowbar
           "-L",
           "#{ENV["CROWBAR_LOG_DIR"]}/chef-client/#{Node.admin_node.name}.log"
         )
+        Rails.logger.info("Updating admin node log file ownership")
+        system(
+          "sudo",
+          "chown",
+          "crowbar:",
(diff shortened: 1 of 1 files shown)


crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -222,6 +222,13 @@ def restore_chef
           "-L",
           "#{ENV["CROWBAR_LOG_DIR"]}/chef-client/#{Node.admin_node.name}.log"
         )
+        Rails.logger.info("Updating admin node log file ownership")
+        system(
+          "sudo",
+          "chown",
+          "crowbar:crowbar",
```
dirkmueller: NIT: "crowbar:" should be enough (use default group)

vuntz:
https://ci.suse.de/job/openstack-mkcloud/164806/console failed, but the problem is that on restore, we don't have the patched package. Looks like a bug in mkcloud that doesn't add the repo for the patched package during `onhost_reset_admin`...

vuntz:
Submitted https://github.com/SUSE-Cloud/automation/pull/3331

vuntz:
> Submitted SUSE-Cloud/automation#3331

With that automation patch and the PR here, the restore step is passing. See https://ci.suse.de/job/openstack-mkcloud/164906/

PR #1824, opened 2019-04-09
Fix backup restore gate job

chajain:
Fixed the ownership of crowbar admin node log file

### DIFF
--- a/crowbar_framework/lib/crowbar/backup/restore.rb
+++ b/crowbar_framework/lib/crowbar/backup/restore.rb
@@ -226,7 +226,7 @@ module Crowbar
         system(
           "sudo",
           "chown",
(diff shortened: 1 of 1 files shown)


crowbar_framework/lib/crowbar/backup/restore.rb
```diff
@@ -226,7 +226,7 @@ def restore_chef
         system(
           "sudo",
           "chown",
-          "crowbar:crowbar",
+          "crowbar:",
```
nicolasbock: Why a trialling `:`? Isn't it enough to set the user?

nicolasbock:
I presume you are working on #1822 instead? Can we close this one then?

PR #1825, opened 2019-04-09
Update some styles for readability (SCRD-8787)

eapodaca:
* Updates navbar styles to closer match CLM Day2
* Updates default button styles
* Updates table headers
* Updates info alert colors

### DIFF
--- a/crowbar_framework/vendor/assets/stylesheets/_eos-branding.scss
+++ b/crowbar_framework/vendor/assets/stylesheets/_eos-branding.scss
@@ -63,7 +63,7 @@ $font-family-serif:       Georgia, "Times New Roman", Times, serif !default;
 $font-family-monospace:   Menlo, Monaco, Consolas, "Courier New", monospace !default;
 $font-family-base:        $font-family-sans-serif !default;
 
-$font-size-base:          13px;
+$font-size-base:          14px;
 $font-size-large:         ceil(($font-size-base * 1.25)) !default; // ~18px
 $font-size-small:         ceil(($font-size-base * .85)) !default; // ~12px
 
@@ -75,7 +75,7 @@ $font-size-h5:            $font-size-base !default;
 $font-size-h6:            ceil(($font-size-base * .85)) !default; // ~12px
 
 //** Unit-less `line-height` for use in components like buttons.
-$line-height-base:        1.1;
+$line-height-base:        1.428571429;
 //** Computed "line-height" (`font-size` * `line-height`) for use with `margin`, `padding`, etc.
 $line-height-computed:    floor(($font-size-base * $line-height-base)); // ~20px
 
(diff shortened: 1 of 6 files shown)


eapodaca:
These changes can be seen on our [crowbar instance](http://suserose.us.to:1338).

jeremy-moffitt:
scrd-8787-0
attached a screenshot of the current L&F with this change

skazi0:
Is it just me or all texts seem a bit "up-shifted"? This is mostly visible on menu highlights.

crowbar_framework/vendor/assets/stylesheets/_eos-branding.scss
```diff
@@ -646,9 +646,9 @@ $alert-success-bg:            $state-success-bg !default;
 $alert-success-text:          $state-success-text !default;
 $alert-success-border:        $state-success-border !default;
 
-$alert-info-bg:               $state-info-bg !default;
-$alert-info-text:             $state-info-text !default;
-$alert-info-border:           $state-info-border !default;
+$alert-info-bg:               #d1ecf1;
+$alert-info-text:             #0c5460;
+$alert-info-border:           #bee5eb;
```
vuntz: Mostly a question for @rsalevsky... these values may actually be better in our branding patch, not in the upstream theme. But we kind of lost the branding branch, it seems (except for the patch in the package). How do we deal with that now?

jeremy-moffitt:
> Is it just me or all texts seem a bit "up-shifted"? This is mostly visible on menu highlights.

I see that too, usually thats a line-height thing, but tweaking it in my debugger isnt generating the impact I'd expect :-/

jeremy-moffitt:
scrd-8787-1
weird height issue on the menus is fixed... think there's a slight upward drift on the buttons, though its far less noticeable

crowbar_framework/vendor/assets/stylesheets/_eos-branding.scss
eapodaca: That patch file can be applied to a local branch using `git am` and regenerated with `git format-patch`. Even still because of the way the scss is being built applying it in a patch there does not make sense.

jeremy-moffitt:
scrd-8787-2
button text spacing fixed now as well

jeremy-moffitt:
I think the latest version looks pretty good, asked Sorin to take a look as well (I can't flag him here, so I assume he's not part of the crowbar org), and sent him the url for the local crowbar instance

dotNomad:
Found some things to either consider or fix for this change.

* The SUSE Green on white, and on the panel-default heading, is outside of the
  contras requirementst. We may want to consider using `#686565` for h1, h2,
  and h3.modal-title.
* The `panel-default > .panel-heading` background-color is not a eos color. The
  closest color we have is `#FFECB5`. Unfortunately its a quite a bit darker.
* Currently body is using `font-size: 13px`, but 14px is the recommended in the
  eos design system.
* Should the `bottom-line` be using the primary green for its background?
  This would change `#1fad7a` -> `#009C51`. `#1fad7a` is not a eos color
  which is why I recommend this change.
* Buttons `.btn btn-default` don't follow any specific guideline from the eos
  design system. It seems like either the default, primary, or outline-primary
  can be followed to fix this. It changes their colors, how :hover and :active
  are handled, but also touches the padding and line-height.
* Similar to the above `btn btn-default btn-block` needs to follow a specific
  button guideline.
* Similar to the above `btn btn-danger` which can be seen on Utilities > Backup
  & Restore.
* The Browse button under Utilities > Backup & Restore > Click Upload Backup
  Image is a bit compressed due to the height on the form-control div.

en3sis:
Hello everyone,

I was having a look and agree with what @dotNomad already mentioned, those changes would be a great start.

> Should the bottom-line be using the primary green for its background?
This would change #1fad7a -> #009C51. #1fad7a is not a eos color
which is why I recommend this change.

A background color with the #00C081 instead of gradient would be great, we're trying to remove the usage of gradients across products. Also the link color  (SUSE® OpenStack Cloud 9 Provided by SUSE®) should be blue #008ACF

> The panel-default > .panel-heading background-color is not a eos color. The
closest color we have is #E0DFDF. Unfortunately its a quite a bit darker.

Since we have not defined it, there's no need to use any specific color. The default bootstrap color should work and keep it consistent with the rest of the products (also as you mentioned, all headers should use #686565)

For buttons situation please check https://www.eosdesignsystem.com/buttons. 
- Under Barclamps, I think it's safe to have all of them as Default buttons.
- Once you edit a Proposal, would be great to have only one row of buttons, maybe the bottom one and remove the one on top (I know there's layout limitation).  About the color, Save can be `primary` the rest can stay default.

eapodaca:
> The SUSE Green on white, and on the panel-default heading, is outside of the
contrast requirementst. We may want to consider using #686565 for h1, h2,
and h3.modal-title.

Got it.

> The panel-default > .panel-heading background-color is not a eos color. The
closest color we have is #E0DFDF. Unfortunately its a quite a bit darker.

that is an [eos colors](https://gitlab.com/SUSE-UIUX/eos/blob/master/assets/stylesheets/base/variables/branding.scss#L82). The `bc-gray-50` and `bc-gray-80` do not appear on the website but are in their colors file provided.

> Buttons .btn btn-default don't follow any specific guideline from the eos
design system. It seems like either the default, primary, or outline-primary
can be followed to fix this. It changes their colors, how :hover and :active
are handled, but also touches the padding and line-height.

I think we are getting a little out of scope on this. At this late stage I don't think we should be making drastic changes like this. I've adjusted the buttons a tiny bit, let me know what you think.

> Under Barclamps, I think it's safe to have all of them as Default buttons. Once you edit a Proposal, would be great to have only one row of buttons, maybe the bottom one and remove the one on top (I know there's layout limitation). About the color, Save can be primary the rest can stay default.

Got it.

> Under Utilities > Exported files the delete icon should be gray 

Got it. All icons using correct colors.


Pushed and updated the UI.

eapodaca:
> Should the bottom-line be using the primary green for its background?
This would change #1fad7a -> #009C51. #1fad7a is not a eos color
which is why I recommend this change.

These are specified in a patch file in an IBS package. Will submit a change to that to update this.

> Should the bottom-line be using the primary green for its background?
This would change #1fad7a -> #009C51. #1fad7a is not a eos color
which is why I recommend this change.

Same as above

> Buttons .btn btn-default don't follow any specific guideline from the eos ....

I've added the same styles we use in the horizon theme to address this.

dotNomad:
With the latest changes addressing the feedback its looking good! Following the EOS Design System closer. :+1:

eapodaca:
Updated links at the bottom use default page font size and colors.

crowbar_framework/vendor/assets/stylesheets/_patches.scss
```diff
@@ -13,19 +13,203 @@
...
+        color: $btn-danger-color;
+        background-color: $btn-danger-bg;
+        border-color: $btn-danger-border;
+      }
+    }
+  }
+}
+
+.btn {
+  padding: 7px 12px 5px 12px;
```
dotNomad: This stick out a bit. The EOS Design System uses `padding: 8px 16px;` for buttons, but changing this seems to make the text move a bit above center of the button.

jeremy-moffitt: We tried having even top/bottom padding, and it definitely floated the text up :-/  can take another swing at it, but I think in this case having the text at the right height should take precedence over the styleguide adherence... Do you think the buttons should be made wider? I'm not sure we tried 16px

jeremy-moffitt:
the latest version of the button padding causes some of the buttons to be mis-aligned vertically

eapodaca:
Ok, I think we've got the buttons wrangled the and bottom bar smoothed out.

vuntz:
This is not fully consistent yet. We remove the buttons at the top of the barclamp pages, but then for the "nodes"->"bulk edit" page, we don't do that.

The SES settings (in Utilities) is still using a group for buttons. Should we remove that?

vuntz:
Also, in the "Utilies"->"Exported Items" page, I guess we want a "Delete" button instead of a trash icon?

vuntz:
I'd still recommend merging and having a followup for the next steps, though.

crowbar_framework/vendor/assets/stylesheets/_eos-branding.scss
vuntz: @eapodaca sorry that I was not clear. My point is more than the whole EOS branding would actually not be part of the upstream branding, but instead in the SUSE-specific branding (which in the past was done in a different branch) -- and we still have a small patch for the SUSE-specific branding. So the situation is kind of strange, and should probably be resolved eventually... Either we stop having a SUSE-specific theme, or we move the EOS bits to the SUSE-specific theme.

PR #1826, opened 2019-04-09
Update in_process led animation (SCRD-8786)

eapodaca:
* Updates led animations to be more apparent and easy to see.
* Updates all led animations to use css animations so they blink at the same rate.

### DIFF
--- a/crowbar_framework/app/assets/javascripts/application.js
+++ b/crowbar_framework/app/assets/javascripts/application.js
@@ -255,13 +255,6 @@ jQuery(document).ready(function($) {
   $('#nodelist').nodeList();
   $('input[type=password]').hideShowPassword();
 
-  setInterval(
-    function() {
-      $('.led.failed, .led.pending, .led.waiting, led.red').toggleClass('blink');
-    },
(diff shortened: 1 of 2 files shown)


eapodaca:
I don't understand why hound is complaining.

skazi0:
> I don't understand why hound is complaining.

Because it's a dog :o)

eapodaca:
The thing is... all that worked before for one of my other PRs ¯\\\_(ツ)_/¯

skazi0:
@eapodaca I merged the hound fix. Please (force)push something here to retrigger hound.

jeremy-moffitt:
the error in the ha test does not look related to this change
```
sshrun TIMEFORMAT="timing for mkcloud call 'onadmin_testsetup' real=%R user=%U system=%S" ; time onadmin_testsetup failed! (safelyret=102) Aborting.
```

dirkmueller:
Would it not make sense to have the 'inprogress' blink differently? Not that I care much though, but how do we indicate that the visible state is not final?

PR #1827, opened 2019-04-09
Remove arrows (SCRD-8737)

eapodaca:
Update Codemirror to overive attribute set by bootstrap.


eapodaca:
Change is deployed on our [crowbar instance](http://suserose.us.to:1338).

PR #1830, opened 2019-04-10
crowbar: Also add access to /restricted/ in SSL vhost

vuntz:
This became visible in the CI where SSL is used.


vuntz:
If anyone wants to see the patch in action: running on the vm2 deployment right now.

PR #1833, opened 2019-04-11
[5.0] Designate enablement

dirkmueller:


sjamgade:
iiuc neutron integration is missing in master too, could we please have it master first ?

dirkmueller:
@sjamgade how is neutron integration (which is a topic for the git repository called crowbar-openstack) related to this one (crowbar-core) ? these changes are prerequisite for the crowbar-openstack PR, which still needs to be done.

sjamgade:
LGTM

PR #1836, opened 2019-04-12
Fix 500 error on node edit (SCRD-8787)

eapodaca:


jeremy-moffitt:
please add the SCRD number to the commit message

PR #1796, opened 2019-02-11
provisioner: enabled tuned profiles

Itxaka:
Enable the tuned profiles virtual-host or throughput-performance based
on the current role of the node to get better sysctl settings

PR #1838, opened 2019-04-15
Revert "provisioner: enabled tuned profiles"

Itxaka:
Should remember that we are between GMC1 and 2....

Reverts crowbar/crowbar-core#1796


vuntz:
The HA failure is unrelated, and we agreed to revert the patch that unexpectedly went in before GMC2.

@Itxaka I assume you'll submit the patch again?

Itxaka:
@vuntz I will

PR #1840, opened 2019-04-16
provisioner: enabled tuned profiles

Itxaka:
Enable the tuned profiles virtual-host or throughput-performance based
on the current role of the node to get better sysctl settings


Itxaka:
[Saner sysctl defaults for compute nodes and database nodes](https://trello.com/c/hSbEqGyx/336-saner-sysctl-defaults-for-compute-nodes-and-database-nodes)

PR #1841, opened 2019-04-16
crowbar: Move crowbarrc mgmt into crowbar cookbook (SCRD-8330)

cmurphy:
In the case of a restoration from backup, /etc/crowbarrc will already
have been modified by the last installation. During the restore, the
install-suse-chef.sh script runs chef on the admin node, but only with a
select set of roles which at certain points does not include the
provisioner role. The result is that from crowbar's point of view at
that point in time this is a new installation and the proposal values,
including whether to use SSL, are set to default values. However, the
crowbarrc file may already have been modified to use SSL. In this case,
trying to use crowbarctl will fail because the crowbar server is not yet
configured for SSL.

This patch addresses the problem by moving the management of the
crowbarrc file into the crowbar cookbook for the admin node. This way,
if only the crowbar role is included, it can still properly manage the
crowbarrc file so that the server and verify_ssl parameters are correct
for the stage it is running in.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/default.rb
+++ b/chef/cookbooks/crowbar/recipes/default.rb
@@ -253,7 +253,7 @@ else
   chef_solr_data = "/var/cache/chef/solr/data"
 end
 
-if node["crowbar"] && node["crowbar"]["realm"]
+if node["crowbar"]
   # After installation of a gem, we have a new path for the new gem, so we
   # need to reset the paths if we can't load the gem
   begin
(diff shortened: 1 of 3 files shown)


cmurphy:
It's failing on the issue this should be fixing. I can see that addupdaterepo is being called so I'm not sure yet why it's not working.

nicolasbock:
You could sprinkle a few log statements into your code to see whether it's executed and why it's not doing what you think it should be doing.

cmurphy:
I've tested this a few time and I'm now confident in it, please review.

cmurphy:
This is clearly affecting the CI in some strange way.

cmurphy:
I think it needs this patch: https://github.com/SUSE-Cloud/automation/pull/3359

nicolasbock:
Good point. Too bad we can't stack PR's :(

cmurphy:
The backup-restore-ssl job is now green \o/

vuntz:
As HA is not relevant, let me merge this.

PR #1847, opened 2019-05-08
network: Don't set datapath-ids on ovs-bridges anymore

rhafer:
Neutron is now taking care of setting a unique datapath-id. So we don't
need this anymore. See:

https://review.opendev.org/#/c/587244/


rhafer:
> do we have the fix already in GM?

Yes. It's part of GM.

The pike backport seems to have landed in the Update Channels as well.

PR #1848, opened 2019-05-13
Ignore CVE-2019-11068 during Travis (SOC-9262)

rsalevsky:
SLES 12 only ships Ruby 2.1.9 but nokogiri 1.10.3 requires ruby >=
2.3.0. As we can't update to this version we will ignore the CVE during
travis but we will add a patch to the current SLES 12 nokogiri to fix
the CVE.


rsalevsky:
Backports are https://github.com/crowbar/crowbar-core/pull/1850 and https://github.com/crowbar/crowbar-core/pull/1849.

PR #1851, opened 2019-05-13
bind: Disable listening on IPv6 addresses for now

dirkmueller:
The newer bind version in SLE12 SP4 enables listening on IPv6
by default, however we can't deal with that yet. disable it
so that we can bootstrap crowbar.

(cherry picked from commit 997efa72169dc175d83ef1cdce13bd04283f88cc)


nicolasbock:
In the meantime I found what `draft` means. Another exciting new feature from GitHub :wink:

PR #1853, opened 2019-05-16
network: reduce risk of creating network loops

rhafer:
This adds some additional checks to the network recipe that should help reducing the risk of creating network loops when interfaces need to be reconfigured. This mainly done by

1. Erroring out when the conduit mapping for the node turns out to be broken in a way the would require putting two interfaces into the same bridge.
2. Properly tearing down existing configurations when the conduit mapping change in a way the would move interfaces from one bridge to anther.
3. Erroring out when the interface mapping for a bond changes in an unexpected manner.

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -131,14 +131,15 @@ def kill_nic_files(nic)
     end
   when "suse"
     # SuSE also has lots of small files, but in slightly different locations.
-    if ::File.exist?("/etc/sysconfig/network/ifcfg-#{nic.name}")
-      ::File.delete("/etc/sysconfig/network/ifcfg-#{nic.name}")
-    end
-    if ::File.exist?("/etc/sysconfig/network/ifroute-#{nic.name}")
-      ::File.delete("/etc/sysconfig/network/ifroute-#{nic.name}")
-    end
-    if ::File.exist?("/etc/wicked/scripts/#{nic.name}-pre-up")
-      ::File.delete("/etc/wicked/scripts/#{nic.name}-pre-up")
+    [
+      "/etc/sysconfig/network/ifcfg-#{nic.name}",
+      "/etc/sysconfig/network/ifroute-#{nic.name}",
+      "/etc/wicked/scripts/#{nic.name}-pre-up"
+    ].each do |nicfile|
+      if ::File.exist?(nicfile)
+        ::File.delete(nicfile)
+        Chef::Log.info("Deleting #{nicfile}")
+      end
     end
   end
 end
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -315,6 +315,45 @@ sorted_networks.each do |network|
   if network.add_ovs_bridge
     bridge = network.bridge_name || "br-#{network.name}"
 
+    # Check if our current base interface (bond, vlan or physical) is already
+    # attached to another OVS bridge. Take care about that appropriately.
+    if our_iface.ovs_master && our_iface.ovs_master.name != bridge
+      old_ovs = our_iface.ovs_master
+
+      # We are already attached to an ovs bridge, check whether that attachment
+      # was already verified (or created) during this chef-client run, by looking
+      # through the list of interfaces that we already touched ('ifs'). If it is,
+      # find out which crowbar network requires that attachement and error out
+      # with a meaningful error message.
+      if ifs.key? old_ovs.name
+        other_net = sorted_networks.find do |n|
+          n.bridge_name == old_ovs.name
+        end
+        other_net ||= sorted_networks.find do |n|
+          old_ovs.name == "br-#{n.name}"
+        end
+        raise "The current conduit mapping for network '#{network.name}' requires " \
+          "interface '#{our_iface.name}' to be attached to ovs-brige '#{bridge}'. " \
+          "But network '#{other_net.name}' which was already set up, requires " \
+          "the same interface to be attached to ovs-bridge '#{old_ovs.name}'. " \
+          "Refusing to reconfigure. There seems to be a conflict in the conduit" \
+          "mappings for the networks '#{other_net.name}' and '#{network.name}'."
+      end
+
+      # We didn't create the attachement during the current run of chef-client.
+      # Tear down the attachement (and bridge configuration) to be able to attach
+      # the current interface to the new bridge.
+      Chef::Log.warn("Current conduit mapping for network '#{network.name}' requires " \
+                     "ovs-brige '#{bridge}' to be attached to device '#{our_iface.name}' " \
+                     "but the device is already attached to '#{old_ovs.name}'.")
+      Chef::Log.warn("Unplugging '#{our_iface.name}' from '#{old_ovs.name}' and " \
+                     "taking the bridge down for reconfiguration to reduce the " \
+                     "risk of creating a network loop.")
+      old_ovs.unplug(our_iface.name)
+      ::Kernel.system("wicked ifdown #{old_ovs.name}")
+      kill_nic_files old_ovs
+    end
+
     # This flag is used later to enable wicked-nanny (on SUSE platforms)
     ovs_bridge_created = true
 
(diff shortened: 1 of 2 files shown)


susecloudcommitpolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

aojea:
This SGTM but I let others approve, my Ruby skills are not good 😓

rhafer:
stable/4.0 backport: https://github.com/crowbar/crowbar-core/pull/1899
stable/5.0 backport: https://github.com/crowbar/crowbar-core/pull/1898

rhafer:
Rebased, addressed remaining hound issues and re-formatted commit messages to make travis happy

chef/cookbooks/network/recipes/default.rb
```diff
@@ -314,9 +334,76 @@ def kill_nic(nic)
...
+    if Nic.exists?(bridge) && Nic.ovs_bridge?(bridge)
+      ovs = Nic.new bridge
+      ovs_slaves = ovs.slaves
+
+      # Find out if the ovs bridge for the current network is already attached to
+      # any other physical, vlan or bond interface than what the current configuration
+      # requires. If it is, tear down the bridge and slave devices for reconfig to
+      # reduce possible risk creating network loops.
+      non_virtual_slaves = ovs_slaves.find_all do |slave|
+        next if slave.name == our_iface.name
```
skazi0: Wouldn't it be simpler like this: `slave.name != our_iface.name && Nic.base_interface? slave` ? This `next` is a bit confusing inside a predicate.

chef/cookbooks/network/recipes/default.rb
```diff
@@ -314,9 +334,76 @@ def kill_nic(nic)
...
+      # was already verified (or created) during this chef-client run, by looking
+      # through the list of interfaces that we already touched ('ifs'). If it is,
+      # find out which crowbar network requires that attachement and error out
+      # with a meaningful error message.
+      if ifs.key? old_ovs.name
+        other_net = sorted_networks.find do |n|
+          n.bridge_name == old_ovs.name
+        end
+        other_net ||= sorted_networks.find do |n|
+          n.name == old_ovs.name.sub(/\A#{Regexp.escape("br-")}/, "")
```
skazi0: Some explanation of this 'br-XXX' vs 'XXX' would be helpful.

chef/cookbooks/network/recipes/default.rb
```diff
@@ -314,9 +334,76 @@ def kill_nic(nic)
...
+          "But network '#{other_net.name}' which was already set up, requires " \
+          "the same interface to be attached to ovs-bridge '#{old_ovs.name}'. " \
+          "Refusing to reconfigure. There seems to be a conflict in the conduit" \
+          "mappings for the networks '#{other_net.name}' and '#{network.name}'."
+      end
+
+      # We didn't create the attachement during the current run of chef-client.
+      # Tear down the attachement (and bridge configuration) to be able to attach
+      # the current interface to the new bridge.
+      Chef::Log.warn("Current conduit mapping for network '#{network.name}' requires " \
```
jgrassler: Not sure whether I like this. This would take the other bridge down more or less silently (with just a log message), wouldn't it, potentially breaking functionality that was there, previously. Shouldn't we fail barclamp application and require the user to indicate that the action just performed was indeed fine (by deleting a state file or something like that) manually?

chef/cookbooks/network/recipes/default.rb
```diff
@@ -314,9 +334,76 @@ def kill_nic(nic)
...
+      # Find out if the ovs bridge for the current network is already attached to
+      # any other physical, vlan or bond interface than what the current configuration
+      # requires. If it is, tear down the bridge and slave devices for reconfig to
+      # reduce possible risk creating network loops.
+      non_virtual_slaves = ovs_slaves.find_all do |slave|
+        next if slave.name == our_iface.name
+        Nic.base_interface? slave
+      end
+      unless non_virtual_slaves.nil? || non_virtual_slaves.empty?
+        Chef::Log.warn("Current conduit mapping for network '#{network.name}' requires " \
```
jgrassler: Same as above. The user may miss that warning and thus wonder why that bridge ceased working. With a failed barclamp that persists this warning (and requires some sort of manual action to retroactively sign off on this bridge removal) that couldn't happen.

chef/cookbooks/network/recipes/default.rb
rhafer: @jgrassler Actually the block above this (line 347 to 360) is trying to make sure that this change is intentional an won't break anything. If the old bridge (the one we're tearing down here) is needed by any of the  enabled crowbar networks on this node the chef run will fail with a the pretty verbose usable error message above, latest when trying to recreate that bridge (the timing of that depends on the order in which the networks are processed, but it will fail at some point).

chef/cookbooks/network/recipes/default.rb
rhafer: All this code it just there to keep the bridge working (it will be created further down) after some change in either the network proposal, the node HW  or role assignment of the node was done.

chef/cookbooks/network/recipes/default.rb
rhafer: I'll do that. Though I assumed is was pretty obvious, that the bridge's name is derived from the network name in crowbar.

chef/cookbooks/network/recipes/default.rb
rhafer: Hm, still I am still unsure which I like better, but I can adapt to your suggestion.

chef/cookbooks/network/recipes/default.rb
skazi0: OK, I misread the code above. I thought that the bridge could be named XXX or br-XXX. Still I think this code could be done simpler:
a. `"br-#{n.name}" == old_ovs.name`
b. `n.name == old_ovs.name.sub(/\Abr-/, "")` I don't think we need to escape `-` outside of character range

chef/cookbooks/network/recipes/default.rb
rhafer: Done

chef/cookbooks/network/recipes/default.rb
rhafer: Done

chef/cookbooks/network/recipes/default.rb
```diff
@@ -314,9 +334,75 @@ def kill_nic(nic)
...
+    if Nic.exists?(bridge) && Nic.ovs_bridge?(bridge)
+      ovs = Nic.new bridge
+      ovs_slaves = ovs.slaves
+
+      # Find out if the ovs bridge for the current network is already attached to
+      # any other physical, vlan or bond interface than what the current configuration
+      # requires. If it is, tear down the bridge and slave devices for reconfig to
+      # reduce possible risk creating network loops.
+      non_virtual_slaves = ovs_slaves.find_all do |slave|
+        slave.name != our_iface.name && Nic.base_interface? slave
```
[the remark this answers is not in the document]
skazi0: brackets?

chef/cookbooks/network/recipes/default.rb
```diff
@@ -314,9 +334,75 @@ def kill_nic(nic)
...
+      # was already verified (or created) during this chef-client run, by looking
+      # through the list of interfaces that we already touched ('ifs'). If it is,
+      # find out which crowbar network requires that attachement and error out
+      # with a meaningful error message.
+      if ifs.key? old_ovs.name
+        other_net = sorted_networks.find do |n|
+          n.bridge_name == old_ovs.name
+        end
+        other_net ||= sorted_networks.find do |n|
+          n.name == "br-#{old_ovs.name}"
```
skazi0: shouldn't this be the other way around? `n.name` is a network name and `old_ovs.name` is a bridge name and bridge=br-network

rhafer: Argh yeah.

PR #1855, opened 2019-05-20
crowbar: Modify regex of group names to forbid dots (SOC-8011)

ilausuch:
Group names cannot contain dots because it conflicts with
the url route

/nodes/groups/1.0/:id/:group(.:format)

Now JS doesn't allow dots in the group names

More info at https://jira.suse.com/browse/SOC-8011


ilausuch:
I think there is not a bug, but there is this card https://jira.suse.com/browse/SOC-8011

ilausuch:
Done

JanZerebecki (approved):

PR #1858, opened 2019-06-12
keystone: delete old Member role before upgrade

jgrassler:
Until Cloud 8, the default keystone role for simple membership
in a tenant was named "Member". It got renamed to "member" as
of Cloud 9. Since Keystone ignores case, Crowbar will attempt
to create a "member" role and fail since "Member" exists
already. Deleting the old "Member" role first fixes this
problem.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-delete-Member-role-before-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-delete-Member-role-before-upgrade.sh.erb
@@ -1,0 +1,47 @@
+#!/bin/bash
+#
+# This script deletes the Member role as part of the upgrade from Cloud 8 to
+# Cloud 9. This is necessary because the "Member" role was renamed to "member"
+# in Cloud 9 and the attempt to create the "member" role will fail if "Member"
+# exists already (Keystone is case-insensitive).
+#
+# The script should be run only on the controller (first controller/cluster
+# founder in an HA cloud).
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+UPGRADEDIR=/var/lib/crowbar/upgrade
+STATEFILE=${UPGRADEDIR}/crowbar-delete-Member-role-before-upgrade-ok
+
+mkdir -p "`dirname "$LOGFILE"`"
+exec >>"$LOGFILE" 2>&1
+
+log()
+{
+    set +x
+    echo "[$(date --iso-8601=ns)] [$$] $@"
+    set -x
+}
+
+log "Executing $BASH_SOURCE"
+
+source /root/.openrc
+set -x
+
+mkdir -p $UPGRADEDIR
+
+if [[ -f $STATEFILE ]] ; then
+    log "Member role was already deleted"
+    exit 0
+fi
+
+if openstack role list | grep -q ' Member '; then
+  openstack role delete Member
+  ret=$?
+  if [ $ret != 0 ]; then
+    log "Failed to delete Member role."
+    exit $ret
+  fi
+fi
+
+touch $STATEFILE
+log "$BASH_SOURCE is finished."
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -443,6 +443,8 @@ module Api
         upgrade_nodes = ::Node.find("state:crowbar_upgrade")
         cinder_node = nil
         nova_node = nil
+        keystone_node = nil
+
         upgrade_nodes.each do |node|
           node.set_pre_upgrade_attribute
           if node.roles.include?("cinder-controller") &&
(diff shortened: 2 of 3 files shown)


chef/cookbooks/crowbar/templates/default/crowbar-delete-Member-role-before-upgrade.sh.erb
```diff
@@ -0,0 +1,47 @@
...
+
+if [[ -f $STATEFILE ]] ; then
+    log "Member role was already deleted"
+    exit 0
+fi
+
+if openstack role list | grep -q ' Member '; then
+  openstack role delete Member
+  ret=$?
+  if [ $ret != 0 ]; then
```
jsuchome: That's a nitpick, but this time can we do something like `if ! openstack role delete Member` ? (my syntax might be wrong)

jsuchome: Oh, but that we would miss the exit value we want to return ...

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -455,6 +457,32 @@ def services
                   node["pacemaker"]["founder"] == node[:fqdn])
             nova_node = node
           end
+          if node.roles.include?("keystone-server") &&
```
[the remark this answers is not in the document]
jgrassler: What do we do about this one? All the code around it does the naughty thing, for what it's worth...

chef/cookbooks/crowbar/templates/default/crowbar-delete-Member-role-before-upgrade.sh.erb
```diff
@@ -0,0 +1,47 @@
...
+
+mkdir -p $UPGRADEDIR
+
+if [[ -f $STATEFILE ]] ; then
+    log "Member role was already deleted"
+    exit 0
+fi
+
+if openstack role list | grep -q ' Member '; then
+  openstack role delete Member
```
cmurphy: This has the side effect that all users with the Member role assignment on any project now lose that role assignment and won't be able to authenticate. Operators will have to recreate those role assignments manually and potentially update all of their policy files if they had been using 'Member' in their policies. Is this upgrade meant to be so disruptive?

chef/cookbooks/crowbar/templates/default/crowbar-delete-Member-role-before-upgrade.sh.erb
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+#
+# This script deletes the Member role as part of the upgrade from Cloud 8 to
+# Cloud 9. This is necessary because the "Member" role was renamed to "member"
+# in Cloud 9 and the attempt to create the "member" role will fail if "Member"
+# exists already (Keystone is case-insensitive).
```
cmurphy: Slight nitpick here, it's not keystone that is case-insensitive but the mysql backend. If it was up to keystone, we wouldn't have this problem.

chef/cookbooks/crowbar/templates/default/crowbar-delete-Member-role-before-upgrade.sh.erb
jgrassler: Oh...missed that problem. No the upgrade definitely isn't meant to be that disruptive. Then we're in a bit of a pickle. I see two potential solutions off the top of my head:

1. We could retain the old `Member` role and modify Crowbar to be case insensitive. Not too fond of that one...

2. We could extend `crowbar-delete-Member-role-before-upgrade.sh.erb` to save role assignments, create the `member` role right away and restore these role assignments. That's the approach I'd prefer.

Thoughts?

jsuchome: For part 2, do you have idea how complicated could it be? Would you plan to automatically modify all possible policy files? Can't we miss something? Is it right to adapt what users set manually?

chef/cookbooks/crowbar/templates/default/crowbar-delete-Member-role-before-upgrade.sh.erb
guangyee: Why are we creating the member role in Crowbar? It is being created by 'keystone-manage bootstrap'. See

https://github.com/openstack/keystone/blob/stable/rocky/keystone/cmd/bootstrap.py#L112

Are we not using bootstrap?

chef/cookbooks/crowbar/templates/default/crowbar-delete-Member-role-before-upgrade.sh.erb
guangyee: I would go for 1 since we only support mysql backend now. Otherwise, I'd recommend doing SQL injection to rename 'Member' to 'member' so we can retain the role ID.

guangyee: How about this for option 1? https://github.com/crowbar/crowbar-openstack/pull/2153

jgrassler:
@guangyee has come up with a more elegant solution for this: case-insensitive role name lookup. That's a lot more elegant than deleting and recreating the role, so closing this one.

PR #1861, opened 2019-06-12
travis: pin sexp_processor to 4.12.0

sjamgade:
we use brakeman during our travis run. It has a dependency
sexp_processor (deep buried). This gem got a new commit[1] which is not
compatible with brakeman.

The best case would be to fix brakeman and backport it to v2.6.3 (also
pinned) which had a last commit in 2014.  So pinning the dependency is
the easier fix

1 - https://github.com/seattlerb/sexp_processor/commit/ce284487f057203360c41b14d2b25f8c5453fbb9

### DIFF
--- a/crowbar_framework/Gemfile
+++ b/crowbar_framework/Gemfile
@@ -65,6 +65,7 @@ gem "activeresource", "~> 4.0.0",
 
 unless ENV["PACKAGING"] && ENV["PACKAGING"] == "yes"
   group :development, :test do
+    gem "sexp_processor", "<= 4.12.0" # remove when moving to newer brakeman
     gem "brakeman", "~> 2.6.3"
     gem "rspec-rails", "~> 3.3.0"
     gem "byebug", "~> 8.2.2"


skazi0:
@sjamgade there is already a fix in upstream brakeman (https://github.com/presidentbeef/brakeman/commit/a84f6661ee1a3b353d6367744e44042b6d56f52d) but I agree that pinning is the way to go here since we have very old brakeman pinned already.

crowbar_framework/Gemfile
```diff
@@ -45,6 +45,7 @@ gem "rails-observers", "<= 0.1.2"
 gem "ohai", "~> 6.24.2"
 gem "chef", "~> 10.32.2"
 gem "net-http-digest_auth", "~> 1.4"
+gem "sexp_processor", "<= 4.12.0"
```
JanZerebecki: This is only needed by brakman, right? Please move it down there, so it is only installed under the same condition. Also please add a comment, to remove it, once a newer brakman is in use.

sjamgade: currently only brakeman is failing, So I hope its only brakeman. 
Moved and added comment.

PR #1862, opened 2019-06-13
Allow restricted API access during upgrade

jgrassler:
The upgrade uses the restricted API to transition nodes from
state "readying" to state "ready" in the course of the
upgrade. Since RestrictedController inherits the update()
method from ApplicationController this is not possible due to
ApplicationController disallowing all POST requests during
API. This commit fixes the problem by giving the restricted
controller an update() method of its own which does allow
them.

Setting to WIP for now - currently testing. I will clear the flag once my test cloud has (hopefully) finished upgrading.

### DIFF
--- a/crowbar_framework/app/controllers/restricted_controller.rb
+++ b/crowbar_framework/app/controllers/restricted_controller.rb
@@ -133,4 +133,15 @@ class RestrictedController < ApplicationController
   def load_node
     @node = Node.find_node_by_name_or_alias(params[:id])
   end
+
+  def upgrade
+    respond_to do |format|
+      format.any(:json, :jsonapi) do
+        return
+      end
+      format.html do
+        render "errors/during_upgrade", status: :service_unavailable
+      end
+    end
+  end
 end


jsuchome:
Would it make sense to link the PR/commit that actually introduced this new behavior, or are there too many of them?

jgrassler:
There were several, but this one introduced the behaviour I'm fixing (it added the new controller for the restricted API): https://github.com/crowbar/crowbar-core/commit/d613e71e3e739d53320c3cbd6ca8711d8c1522f1

jsuchome:
@jgrassler, could you please include the commit reference in your commit message?

Also, was the original solution ported to SOC8? If so, upgrade patch would be required for 7-8 path.... but I do not see any backport label so I assume it wasn't. Just to make sure, let's ask @vuntz

jgrassler:
Ok, I added the reference now.

crowbar_framework/app/controllers/restricted_controller.rb
```diff
@@ -133,4 +133,15 @@ def load_node_or_render_not_found
   def load_node
     @node = Node.find_node_by_name_or_alias(params[:id])
   end
+
+  def upgrade
```
jsuchome: You're mentioning `update` in the commit message

jgrassler: Indeed, and twice no less...fixed now. Thanks!

jgrassler:
Ok, local testing successful, clearing the WIP flag. Ready to merge as far as I'm concerned.

PR #1863, opened 2019-06-13
crowbar: Save sync_mark attributes in databag

skazi0:
sync_mark provider from crowbar-ha is using sync_mark attributes from
crowbar-config databag so they need to be transferred from proposal to
this databag.


skazi0:
@AbelNavarro if it is causing problems we could add caching like in some other cases (invalidated with every chef-client run).

PR #1872, opened 2019-06-28
crowbar: Give more time for reboot for physical hardware reboots

dirkmueller:
Some more enterprisy physical servers take a little bit more time
to reboot than just the 5-6 minutes. Wait at least 10 minutes
to be compatible with those vendors.


toabctl:
@dirkmueller travis is complaining about your commit msg...

PR #1873, opened 2019-07-01
bind9: Fix spelling error in template

dirkmueller:


nicolasbock:
The gitlinter is complaining that the title needs a bug reference.

PR #1876, opened 2019-07-04
monasca: remove Java persister (SOC-9482)

jgrassler:
Until Cloud 8 we used the Java based Monasca Persister. As of
Cloud 9 we use the Python based Monasca Persister
implementation. Since the two packages conflict, we need to
uninstall the Java based persister when preparing for the
Cloud 8 to Cloud 9 upgrade.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -105,6 +105,15 @@ template "/usr/sbin/crowbar-delete-cinder-services-before-upgrade.sh" do
   only_if { cinder_controller && (!use_ha || is_cluster_founder) }
 end
 
+template "/usr/sbin/crowbar-monasca-cleanups.sh" do
+  source "crowbar-monasca-cleanups.sh.erb"
+  mode "0775"
+  owner "root"
+  group "root"
+  action :create
+  only_if { roles.include?("monasca-server") }
+end
+
 controller_nodes = search(:node, "run_list_map:nova-controller").map { |n| n["hostname"] }
 compute_nodes = search(:node, "run_list_map:nova-compute-*").map { |n| n["hostname"] }
 
--- a/chef/cookbooks/crowbar/templates/default/crowbar-monasca-cleanups.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-monasca-cleanups.sh.erb
@@ -1,0 +1,72 @@
+#!/bin/bash
+#
+# This script performs various Monasca cleanups required before
+# the Cloud 8 to Cloud 9 upgrade.
+#
+# The script should be run only on the monasca-server node.
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+UPGRADEDIR=/var/lib/crowbar/upgrade
+STATEFILE=${UPGRADEDIR}/crowbar-monasca-cleanups-ok
+
+mkdir -p "`dirname "$LOGFILE"`"
+exec >>"$LOGFILE" 2>&1
+
+log()
+{
+    set +x
+    echo "[$(date --iso-8601=ns)] [$$] $@"
+    set -x
+}
+
+install_new_persister()
+{
+    # Do nothing if java persister is not installed
+    rpm -q openstack-monasca-persister-java || return
+
+    zypper --non-interactive remove openstack-monasca-persister-java
+    local ret=$?
+
+    if [ $ret != 0 ]; then
+      log "Failed to remove Java based Monasca persister."
+      exit $ret
+    fi
+
+    zypper --non-interactive install openstack-monasca-persister
+    local ret=$?
+
+    if [ $ret != 0 ]; then
+      log "Failed to install Python based Monasca persister."
+      exit $ret
+    fi
+
+    # This is required because persister deinstallation will perform a userdel on
+    # the monasca-persister user which may then get recreated with a different user
+    # ID by the Python persister package.
+    chown monasca-persister /var/log/monasca-persister/monasca-persister.log
+}
+
+
+log "Executing $BASH_SOURCE"
+
+set -x
+
+mkdir -p $UPGRADEDIR
+
+if [[ -f $STATEFILE ]] ; then
+    log "Monasca cleanup already ran."
+    exit 0
+fi
+
+# Replace Java persister by Python persister
+
+install_new_persister
+
+# Clear old VHost definitions. Without this, apache2 will fail to start due to
+# duplicate VHost definitions for monasca{-log,}-api
+rm -vf /etc/apache2/vhosts.d/wsgi-monasca-*
+
+###
+
+touch $STATEFILE
+log "$BASH_SOURCE is finished."
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -443,6 +443,8 @@ module Api
         upgrade_nodes = ::Node.find("state:crowbar_upgrade")
         cinder_node = nil
         nova_node = nil
+        monasca_node = nil
+
         upgrade_nodes.each do |node|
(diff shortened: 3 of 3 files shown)


jgrassler:
Do not merge, yet: the script isn't getting deployed to the Monasca node properly. Checking why...

jgrassler:
Ok, problem fixed: I deployed the script to controller rather than the monasca-server node accidentally. Tested and works now.

jgrassler:
Setting back to WIP: there's a whole bunch of things that will need removing (such as stale vhost configurations), so I'll turn this into a generic Monasca cleanup script that removes all of these things in one fell swoop.

jgrassler:
Ok, this one should do the trick. Building a new cloud to test. Will update on Monday morning...

chef/cookbooks/crowbar/templates/default/crowbar-monasca-cleanups.sh.erb
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+#
+# This script performs various Monasca cleanups required before
+# the Cloud 8 to Cloud 9 upgrade.
+#
+# The script should be run only on the monasca-server node.
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+UPGRADEDIR=/var/lib/crowbar/upgrade
+STATEFILE=${UPGRADEDIR}/crowbar-remove-java-persister-ok
```
toabctl: nit: `crowbar-remove-monasca-java-persister-ok` to know what this is about.

chef/cookbooks/crowbar/templates/default/crowbar-monasca-cleanups.sh.erb
```diff
@@ -0,0 +1,72 @@
...
+    set -x
+}
+
+install_new_persister()
+{
+    # Do nothing if java persister is not installed
+    rpm -q openstack-monasca-persister-java || return
+
+    zypper --non-interactive remove openstack-monasca-persister-java
+    ret=$?
```
toabctl: `local ret=...`

chef/cookbooks/crowbar/templates/default/crowbar-monasca-cleanups.sh.erb
```diff
@@ -0,0 +1,72 @@
...
+
+    if [ $ret != 0 ]; then
+      log "Failed to remove Java based Monasca persister."
+      exit $ret
+    fi
+
+    zypper --non-interactive install openstack-monasca-persister
+    ret=$?
```
toabctl: `local ret ....`

chef/cookbooks/crowbar/templates/default/crowbar-monasca-cleanups.sh.erb
```diff
@@ -0,0 +1,72 @@
...
+
+    if [ $ret != 0 ]; then
+      log "Failed to install Python based Monasca persister."
+      exit $ret
+    fi
+
+    # This is required because persister deinstallation will perform a userdel on
+    # the monasca-persister user which may then get recreated with a different user
+    # ID by the Python persister package.
+    chown monasca-persister /var/log/monasca-persister/monasca-persister.log
```
toabctl: what about the group

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -105,6 +105,15 @@
   only_if { cinder_controller && (!use_ha || is_cluster_founder) }
 end
 
+template "/usr/sbin/crowbar-monasca-cleanups.sh" do
```
toabctl: `/usr/local/sbin`

chef/cookbooks/crowbar/templates/default/crowbar-monasca-cleanups.sh.erb
jgrassler: Done (renamed to crowbar-monasca-cleanups-ok since there's more than just the persister now).

chef/cookbooks/crowbar/templates/default/crowbar-monasca-cleanups.sh.erb
jgrassler: Done.

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
jgrassler: Are you sure about this? I know it's wrong, viewed strictly from a FHS point of view, but there is a considerable amount of precedent here in that same file. If we were to use `/usr/local/` here, we should refactor everything which will greatly inflate this pull request. If you really want me to do that, please confirm.

chef/cookbooks/crowbar/templates/default/crowbar-monasca-cleanups.sh.erb
jgrassler: Done.

chef/cookbooks/crowbar/templates/default/crowbar-monasca-cleanups.sh.erb
```diff
@@ -0,0 +1,72 @@
...
+
+    zypper --non-interactive install openstack-monasca-persister
+    local ret=$?
+
+    if [ $ret != 0 ]; then
+      log "Failed to install Python based Monasca persister."
+      exit $ret
+    fi
+
+    # This is required because persister deinstallation will perform a userdel on
```
dirkmueller: I wonder if that isn't more like a bug.. can we just remove that code that deletes the user on postun? doesn't seem to make sense to me.

chef/cookbooks/crowbar/templates/default/crowbar-monasca-cleanups.sh.erb
```diff
@@ -0,0 +1,72 @@
...
+    exit 0
+fi
+
+# Replace Java persister by Python persister
+
+install_new_persister
+
+# Clear old VHost definitions. Without this, apache2 will fail to start due to
+# duplicate VHost definitions for monasca{-log,}-api
+rm /etc/apache2/vhosts.d/wsgi-monasca-*
```
dirkmueller: maybe use rm -vf to make it idempotent and visible what it does?

dirkmueller: alternatively move it inside the install_new_persister function so that it is executed only once

dirkmueller:
@jgrassler why are we doing this so complicated? the first barclamp apply chef run could just deinstall the package and install the right one..

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
dirkmueller: yeah, it is consistent with all the other scripts (that are also in the wrong location)

chef/cookbooks/crowbar/templates/default/crowbar-monasca-cleanups.sh.erb
jgrassler: Done. Didn't put it into the function, though - this one doesn't have anything to do with the persister but with the switch from monasca-installer to the native Crowbar implementation (they use different file names for the VHost definitions).

chef/cookbooks/crowbar/templates/default/crowbar-monasca-cleanups.sh.erb
jgrassler: We can, but we'd have to do that in the Cloud 8 package, which is the one doing the deleting.

jgrassler:
> @jgrassler why are we doing this so complicated? the first barclamp apply chef run could just deinstall the package and install the right one..

We try to keep upgrade related code in barclamps to a minimum. It usually needs to be conditional and if it needs to be removed again later, it will be easier to find in a standard location, such as these upgrade scripts.

PR #1877, opened 2019-07-05
ipv6: Add a wrap_ip helper to NetworkHelper

s-t-e-v-e-n-k:
To make it easier to write out IP addresses which may be v6, introduce a
wrap_ip method to NetworkHelper. Make use of it in the provisioner
barclamp when writing out the crowbar rc file.

Co-Authored-By: Matthew Oliver <matt@oliver.net.au>


nicolasbock:
I restarted the HA CI job.

PR #1881, opened 2019-07-22
Cleanup clone_stateless_services leftovers (SOC-9842)

skazi0:
Since clone_stateless_services attribute was dropped in
crowbar-ha/1cf82d3ff3142a700ac0e2c55c20944bf364e868 and implicitly
set to false, we don't need some parts of code which were active
when it was set to true.

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -340,14 +340,11 @@ class CrowbarService < ServiceObject
       prop.save
     end
 
-    # Change pacemaker setting to not use pacemaker for stateless services.
-    # Also, turn off possible DRBD enablement.
+    # Turn off possible DRBD enablement.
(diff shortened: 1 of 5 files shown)


crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -340,14 +340,11 @@ def prepare_nodes_for_os_upgrade
       prop.save
     end
 
-    # Change pacemaker setting to not use pacemaker for stateless services.
-    # Also, turn off possible DRBD enablement.
+    # Turn off possible DRBD enablement.
```
jsuchome: I wonder if we should remove all this DRBD related code soon.

skazi0: The attributes are still there so I left this part intact.

skazi0:
This was tested together with https://github.com/crowbar/crowbar-openstack/pull/2174 and both passes went fine over the keystone part (failed later for various reasons).

PR #1883, opened 2019-07-24
upgrade: dump Monasca, Grafana databases (SOC-9772)

jgrassler:
Up until SUSE OpenStack Cloud 8, Monasca ran its own MariaDB
instance on the monasca-server node which hosted Monasca's
metrics database and Grafana's configuration database. As of
Cloud 9 we use the OpenStack cloud's main MariaDB instance for
both of these, hence we need to create a dump of the the data
from both in order to later migrate them to the main MariaDB
instance.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -48,6 +48,9 @@ end
 
 new_alias = repo_alias.gsub("SP2", "SP3").gsub(node[:platform_version], target_platform_version)
 
+monasca_node = search(:node, "run_list_map:monasca-server").first
+monasca_enabled = !monasca_node.nil?
+
 template "/usr/sbin/crowbar-prepare-repositories.sh" do
   source "crowbar-prepare-repositories.sh.erb"
   mode "0755"
@@ -90,7 +93,10 @@ template "/usr/sbin/crowbar-shutdown-services-before-upgrade.sh" do
   variables(
     use_ha: use_ha || remote_node,
     cluster_founder: is_cluster_founder,
-    nova_controller: roles.include?("nova-controller")
+    nova_controller: roles.include?("nova-controller"),
+    monasca_server: roles.include?("monasca-server"),
+    monasca_enabled: monasca_enabled,
+    horizon_node: roles.include?("horizon-server")
   )
 end
 
@@ -317,4 +323,65 @@ template "/usr/sbin/crowbar-heat-migrations-after-upgrade.sh" do
   owner "root"
   group "root"
   only_if { roles.include?("heat-server") }
+end
+
+if monasca_enabled
+  monasca_node_fqdn = monasca_node[:fqdn]
+
+  if !monasca_node[:monasca].key?(:db_monapi)
+    # Pre proposal migrations: monasca proposal data structure looks the way it
+    # looks in Cloud 8.
+    metrics_db_user = "monapi" # hardwired in Cloud 8
+    metrics_db_password = monasca_node[:monasca][:master][:database_monapi_password]
+    metrics_db_name = "mon" # hardwired in Cloud 8
+    grafana_db_user = "grafana" # hardwired in Cloud 8
+    grafana_db_password = monasca_node[:monasca][:master][:database_grafana_password]
+    grafana_db_name = "grafana" # hardwired in Cloud 8
+  else
+    # Post proposal migrations: monasca proposal data structure looks the way
+    # it looks in Cloud 9. We need this if the "services" upgrade step is
+    # interrupted after proposals have been migrated to their Cloud 9 schema
+    # and resumed with the new proposal data structure.
+    metrics_db_user = monasca_node[:monasca][:db_monapi][:user]
+    metrics_db_password = monasca_node[:monasca][:db_monapi][:password]
+    metrics_db_name = monasca_node[:monasca][:db_monapi][:database]
+    grafana_db_user = monasca_node[:monasca][:db_grafana][:user]
+    grafana_db_password = monasca_node[:monasca][:db_grafana][:password]
+    grafana_db_name = monasca_node[:monasca][:db_grafana][:database]
+  end
+
+  stop_db = false
+
+  if roles.include?("horizon-server")
+    db_user = grafana_db_user
+    db_password = grafana_db_password
+    db_type = "grafana"
+    db_name = grafana_db_name
+  elsif roles.include?("monasca-server")
+    db_user = metrics_db_user
+    db_password = metrics_db_password
+    db_type = "metrics"
+    db_name = metrics_db_name
+    stop_db = true
+  end
+
+  template "/usr/sbin/crowbar-dump-monasca-db.sh" do
+    source "crowbar-dump-monasca-db.sh.erb"
+    mode "0755"
+    owner "root"
+    group "root"
+    action :create
+    only_if do
+      (roles.include?("horizon-server") || roles.include?("monasca-server")) &&
+        (!use_ha || is_cluster_founder)
+    end
+    variables(
+      db_user: db_user,
+      db_password: db_password,
+      db_host: monasca_node_fqdn,
+      db_type: db_type,
+      db_name: db_name,
+      stop_db: stop_db
+    )
+  end
 end
--- a/chef/cookbooks/crowbar/templates/default/crowbar-dump-monasca-db.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-dump-monasca-db.sh.erb
@@ -1,0 +1,56 @@
+#!/bin/bash
+#
+# This script will backup the Monasca and grafana databases
+# from the dedicated MariaDB instance on the monasca-server
+# node (where it resided up to Cloud 8) for later migration to
+# the OpenStack cloud's shared mariadb instance.
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+UPGRADEDIR=/var/lib/crowbar/upgrade
+STATEFILE=${UPGRADEDIR}/crowbar-dump-monasca-db-ok
+DBNAME=<%= @db_name %>
+
+DUMPFILE=${UPGRADEDIR}/monasca-${DBNAME}-database.dump.gz
+
+DB_USER="<%= @db_user %>"
+DB_PASSWORD="<%= @db_password %>"
+DB_HOST="<%= @db_host %>"
+DB_NAME="<%= @db_name %>"
+
+STOP_DB="<%= @stop_db %>"
+
+mkdir -p "$UPGRADEDIR"
+mkdir -p "`dirname "$LOGFILE"`"
+exec >>"$LOGFILE" 2>&1
+
+log()
+{
+    set +x
+    echo "[$(date --iso-8601=ns)] [$$] $@"
+    set -x
+}
+
+exitok()
+{
+touch $STATEFILE
+log "$BASH_SOURCE is finished."
+}
+
+log "Executing $BASH_SOURCE"
+
+set -x
+
+if [[ -f $STATEFILE ]] ; then
+    exitok
+fi
+
+set -e
+
+# Connect to the monasca-server node from the node where this script is running
+# so we can store the dump right where we will need it later. That may either
+# be the first database node (where the Grafana DB dump will go) or the
+# monasca-server node (where the Metrics DB dump will go).
+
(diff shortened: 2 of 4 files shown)


chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -309,3 +310,45 @@
   group "root"
   only_if { roles.include?("heat-server") }
 end
+
+monasca_node = search(:node, "run_list_map:monasca-server").first
+monasca_enabled = ! monasca_node.nil?
```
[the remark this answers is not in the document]
skazi0: Too much bash? I like hound's version better.

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -309,3 +310,45 @@
...
+
+if monasca_enabled
+  if ! monasca_node[:monasca].key?(:db_monapi)
+    # Pre proposal migrations: monasca proposal data structure looks the way it
+    # looks in Cloud 8.
+    metrics_db_user = "monapi" # hardwired in Cloud 8
+    metrics_db_password = monasca_node[:monasca][:master][:database_monapi_password]
+    grafana_db_user = "grafana" # hardwired in Cloud 8
+    grafana_db_password = monasca_node[:monasca][:master][:database_grafana_password]
+    monasca_node_fqdn = monasca_node[:fqdn]
```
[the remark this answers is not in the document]
skazi0: Is this by mistake or is this really the same value?

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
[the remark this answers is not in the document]
jgrassler: No arguing about that. There's no arguing with Hound anyway :-)

jgrassler:
No need to comment on Hound comments. I'll address these...that's why I set this to WIP (Hound will always find something, no matter how careful one is...).

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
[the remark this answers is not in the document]
skazi0: Well... we could argue about arguing with hound :o)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -455,6 +456,32 @@ def services
                   node["pacemaker"]["founder"] == node[:fqdn])
             nova_node = node
           end
+          if node.roles.include?("database-server") &&
```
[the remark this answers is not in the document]
jgrassler: There is no nested conditional here, just a multi condition conditional.

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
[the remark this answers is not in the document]
jgrassler: It's really the same one, yes.

chef/cookbooks/crowbar/templates/default/crowbar-dump-monasca-db.sh.erb
```diff
@@ -0,0 +1,65 @@
...
+fi
+
+if [ "$MONASCA_ENABLED" = "false" ]; then
+    log "Monasca barclamp not enabled, skipping Monasca database backup."
+    exit 0
+fi
+
+set -e
+
+# Connect to the monasca-server node from the first controller node where this
```
skazi0: "first controller node" should rather be "first database node" to better cover cases with multiple clusters.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -455,6 +456,32 @@ def services
...
+              (!node.roles.include?("pacemaker-cluster-member") ||
+                  node["pacemaker"]["founder"] == node[:fqdn])
+            database_node = node
+          end
+        end
+
+        begin
+          unless database_node.nil?
+            database_node.wait_for_script_to_finish(
```
skazi0: I think this should be called before this "stopping" loop. If monasca is running on a node which is before the database-server in `upgrade_nodes` list then the monasca database will already be stopped when we try to dump it from database node.

chef/cookbooks/crowbar/templates/default/crowbar-dump-monasca-db.sh.erb
```diff
@@ -0,0 +1,65 @@
...
+
+mkdir -p $UPGRADEDIR
+rm -f $UPGRADEDIR/crowbar-dump-monasca-db
+
+if [[ -f $STATEFILE ]] ; then
+    log "Monasca database backup is already done."
+    exit 0
+fi
+
+if [ "$MONASCA_ENABLED" = "false" ]; then
```
skazi0: This script is generated only when `monasca_enabled==true`. I don't see any scenario where this part would be triggered.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -455,6 +456,32 @@ def services
...
+              "/usr/sbin/crowbar-dump-monasca-db.sh",
```
skazi0: This script is generated only when `monasca_enabled==true`. There should be some guard code to avoid trying to execute non-existing script.

chef/cookbooks/crowbar/templates/default/crowbar-dump-monasca-db.sh.erb
jgrassler: Changed to "machine where this script is running" because it will either run on the cluster founder for dumping the Grafana database or on the monasca-server node for dumping the Monasca database (for they will be restore from these machines, respectively, later).

crowbar_framework/app/models/api/upgrade.rb
jgrassler: Not a problem anymore: I moved back to dumping from the monasca-server node (since that's where we'll restore from later - the restore happens in a recipe running there as well, so it makes sense to put have the dump available on that machine).

chef/cookbooks/crowbar/templates/default/crowbar-dump-monasca-db.sh.erb
jgrassler: True. Removed.

crowbar_framework/app/models/api/upgrade.rb
jgrassler: Done.

jgrassler:
This is untested and unfinished, as you can see from the plethora of Hound comments. Feel free to take a look and comment already, but be aware that I'll at least fix the Hound comments (and any breakage I may yet discover during testing...). To make a diff available I created a dedicated commit. I'll squash once this is closer to being done...

jgrassler:
See also: https://github.com/crowbar/crowbar-openstack/pull/2177

This is where I'm currently working on the restore side of things.

jgrassler:
Testbuild failed because https://github.com/crowbar/crowbar-core/pull/1876 only merged recently and hasn't made it into the crowbar-core package, yet.

chef/cookbooks/crowbar/templates/default/crowbar-dump-monasca-db.sh.erb
```diff
@@ -0,0 +1,55 @@
...
+fi
+
+set -e
+
+# Connect to the monasca-server node from the node where this script is running
+# so we can store the dump right where we will need it later. That may either
+# be the first database node (where the Grafana DB dump will go) or the
+# monasca-server node (where the Metrics DB dump will go).
+
+mysqldump -h ${DB_HOST} -u ${DB_USER} "-p${DB_PASSWORD}" --all-databases | gzip > $DUMPFILE
```
skazi0: Are you sure you want to dump all databases (into one file)? This would include things like `mysql`, `performance_schema` and `information_schema` databases. You could get into problems when restoring these dumps into a shared server.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -481,6 +487,32 @@ def services
           return
         end
 
+        begin
+          unless monasca_node.nil? || database_node.nil?
+            database_node.wait_for_script_to_finish(
+              "/usr/sbin/crowbar-dump-monasca-db.sh",
+              timeouts[:dump_shutdown_monasca_db]
```
skazi0: the timeout name is misleading now after you moved the shutdown code out of the dump script.

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -481,6 +487,32 @@ def services
...
+            )
+            Rails.logger.info("Dump and shutdown of Monasca database sucessful.")
+          end
+        rescue StandardError => e
+          ::Crowbar::UpgradeStatus.new.end_step(
+            false,
+            services: {
+              data: "Problem while dumping/shutting down monasca databases: " + e.message,
```
skazi0: again, no shutdown here.

chef/cookbooks/crowbar/templates/default/crowbar-dump-monasca-db.sh.erb
jgrassler: Oh. I'd thought using an unprivileged user prevented that from happening, but I forgot about `information_schema`. Fixed now.

crowbar_framework/app/models/api/upgrade.rb
jgrassler: Indeed...fixed.

crowbar_framework/app/models/api/upgrade.rb
jgrassler: Fixed as well.

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -318,3 +319,67 @@
...
+    metrics_db_password = monasca_node[:monasca][:db_monapi][:password]
+    metrics_db_name = monasca_node[:monasca][:db_monapi][:database]
+    grafana_db_user = monasca_node[:monasca][:db_grafana][:user]
+    grafana_db_password = monasca_node[:monasca][:db_grafana][:password]
+    grafana_db_name = monasca_node[:monasca][:db_grafana][:database]
+  end
+
+  stop_db = false
+
+  if roles.include?("database-server")
```
skazi0: I think this should also refer to horizon not to database.

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -318,3 +319,67 @@
...
+  end
+
+  template "/usr/sbin/crowbar-dump-monasca-db.sh" do
+    source "crowbar-dump-monasca-db.sh.erb"
+    mode "0755"
+    owner "root"
+    group "root"
+    action :create
+    only_if do
+      (roles.include?("database-server") || roles.include?("monasca-server")) &&
```
skazi0: same here `s/database/horizon/'

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -481,6 +487,34 @@ def services
           return
         end
 
+        begin
+          unless horizon_node.nil?
```
skazi0: This should not be executed if `monasca_node.nil?` also the error message in `rescue` refers to `monasca_node.name` and `horizon_node.name` which could be undefined if those nodes are not found.

jgrassler: Ok, fixed. Not rebasing so you can look at the diff (back into rebase-once-all-is-tested mode :-))

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -488,12 +488,26 @@ def services
...
               "/usr/sbin/crowbar-dump-monasca-db.sh",
               timeouts[:dump_grafana_db]
             )
           end
+        rescue StandardError => e
+          ::Crowbar::UpgradeStatus.new.end_step(
+            false,
+            services: {
+              data: "Problem while dumping Grafana database: " + e.message,
+              help: "Check /var/log/crowbar/production.log at admin server, " \
```
skazi0: "at" admin server but "on" horizon node :o)

jgrassler: Fixed (and in a couple of other places, too. Let's break the bad grammar convention :-))

jgrassler:
Note: no code changes in the force push - only commits squashed.

PR #1884, opened 2019-07-26
travis: Whitelist CVE-2015-3448 (SOC-9911)

skazi0:
CVE-2015-3448 is not fixed in rest-client for a long time. As we are
using custom version of the gem, fixing it would require considerable
effort. Whitelisting it in travis config should unblock the gating runs.

Additional commit changes gitlint parameters for easier backporting.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,8 @@ matrix:
       script:
       - pip install gitlint
       - wget https://raw.githubusercontent.com/SUSE-Cloud/automation/master/scripts/jenkins/gitlint.ini
-      - gitlint --commits master..HEAD -C gitlint.ini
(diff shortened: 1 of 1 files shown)


jgrassler:
Looks good, thanks! Please submit the same for the other repositories (i.e. https://github.com/crowbar/crowbar-openstack and friends) as well.

skazi0:
@jgrassler I don't think others need it... at least crowbar-openstack works fine for my PRs.

jgrassler:
> @jgrassler I don't think others need it... at least crowbar-openstack works fine for my PRs.

Indeed, it doesn't appear to be affected - the crowbar-core PR I looked at when I wrote this failed the commit message Regex. Sorry about that.

PR #1887, opened 2019-07-31
Public ips for dns nodes when designate integration is in use (SOC-9635)

matthewoliver:
When crowbar's DNS is set to intergrate with designate we need the DNS
servers to be listening on the public network so tennent users of
designate can access the zones they create via the API.

If designate integration is enabled, this patch allocates a public ip
for each DNS node. Each node's bind9 is also setup to listen on both the
public and admin ips.

### DIFF
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -293,7 +293,8 @@ when "suse"
   end
 end
 
-# We would like to bind service only to ip address from admin network
+# We would like to bind service only to ip address from admin network unless enable_designate is
+# enabled. In which case bind both the admin _and_ public.
 admin_network = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin")
 admin_addr = admin_network.address
 
@@ -390,8 +391,18 @@ end
 
 ### FIXME Change to "any" once IPv6 support has been implemented
 admin_addr6 = "none"
+public_addr6 = "none"
 if node[:dns][:enable_designate] && !node[:dns][:master]
   node[:dns][:forwarders].push master_ip
+end
+
+ipaddresses = [admin_addr]
+ip6addresses = [admin_addr6]
+if node[:dns][:enable_designate]
+  public_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "public").address
+  public_addr = nil if admin_addr == public_addr
+  ipaddresses << public_addr unless public_addr.nil?
+  ip6addresses << public_addr6 unless public_addr6 == "none"
 end
 
 # Rewrite our default configuration file
@@ -402,8 +413,8 @@ template "/etc/bind/named.conf" do
   group bindgroup
   variables(forwarders: node[:dns][:forwarders],
             allow_transfer: allow_transfer,
-            ipaddress: admin_addr,
-            ip6address: admin_addr6,
+            ipaddresses: ipaddresses,
+            ip6addresses: ip6addresses,
             enable_designate: node[:dns][:enable_designate] && node[:dns][:master]
            )
   notifies :restart, "service[bind9]", :immediately
--- a/chef/cookbooks/bind9/templates/default/named.conf.erb
+++ b/chef/cookbooks/bind9/templates/default/named.conf.erb
@@ -39,8 +39,8 @@ options {
         };
 <% end -%>
         auth-nxdomain no;    # conform to RFC1035
-        listen-on { <%= @ipaddress %>; };
-        listen-on-v6 { <%= @ip6address %>; };
+        listen-on { <%= @ipaddresses.join("; ") %>; };
+        listen-on-v6 { <%= @ip6addresses.join("; ") %>; };
         minimal-responses yes;
         allow-new-zones yes;
 };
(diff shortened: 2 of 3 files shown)


matthewoliver:
Just pushing up a first version to get it off my machine. Thus the WIP label. Still need to test this code. Just ran out of time today.

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -404,7 +407,10 @@ def make_zone(zone)
             allow_transfer: allow_transfer,
             ipaddress: admin_addr,
```
dirkmueller: simply change ipaddress here to public_addr if enable_designate is added? then you don't need to do the if/else inside the template?

chef/cookbooks/bind9/templates/default/named.conf.erb
```diff
@@ -49,6 +54,6 @@ include "/etc/bind/named.conf.default-zones";
 include "/etc/bind/named.conf.crowbar";
 include "/etc/bind/named.conf.local";
 include "/etc/named.d/rndc-access.conf";
-<% if @enable_designate -%>
+<% if @enable_designate and @dns_master -%>
```
dirkmueller: is this an intentional change?

chef/cookbooks/bind9/recipes/default.rb
matthewoliver: That's good for designate, but this is the DNS infrastructure for the crowbar deployed cloud. If we make named/DNS _only_ listen on the public network then any node in the cloud that isn't on the public network can't access DNS. So unless we also go and make sure _every_ node can route to the DNS public ips then there will be nodes in the cloud that can't talk to DNS. Which I think is more problems then an if in a template.

ATM every node in the cloud, all compute, controllers, storage nodes listen on the admin network. Which means they can talk to the crowbar DNS infra, but not all listen on the Public network.

that's my 2 cents anyway.

chef/cookbooks/bind9/templates/default/named.conf.erb
matthewoliver: Yeah, this was intentional. If you go look at the changed line 407 of 'chef/cookbooks/bind9/recipes/default.rb' you'll see that confusingly `@enable_designate` wan't simply the value of `node[:dns][:enable_designate]`. It's:

`enable_designate: node[:dns][:enable_designate] && node[:dns][:master]`

In the old version of the template this is OK. Because when it's the DNS master and enable_designate we want to incloud the rndc key. But that's not the case anymore. We now want to listen on the public ip if designate is enabled **and** if designate is enabled and the node also happens to be the DNS_master then we want to drop the rndc.

So I separated the `&&` and passed them in separately so we could do the right thing. But the and on line 57.

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -296,6 +296,8 @@ def make_zone(zone)
 # We would like to bind service only to ip address from admin network
```
s-t-e-v-e-n-k: This comment is now misleading

chef/cookbooks/bind9/templates/default/named.conf.erb
```diff
@@ -39,8 +39,13 @@ options {
         };
 <% end -%>
         auth-nxdomain no;    # conform to RFC1035
+<% if @enable_designate and !@publicaddress.nil? -%>
+        listen-on { <%= @ipaddress %>; <%= @publicaddress %>; };
+        listen-on-v6 { <%= @ip6address %>; <%= @public6address%>; };
+<% else -%>
```
s-t-e-v-e-n-k: I wonder if we want to switch to an array and then you can just ipaddrs.join('; ') and you can stop passing enable_designate into the template, and the smarts remain in the recipe.

chef/cookbooks/bind9/recipes/default.rb
matthewoliver: oh good spot. Fixed.

chef/cookbooks/bind9/templates/default/named.conf.erb
matthewoliver: Great idea!! That simplifies it greatly. And then the old dns_master + enable_designate logic remains untouched. You sir are a genius!

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -293,9 +293,12 @@ def make_zone(zone)
   end
 end
 
-# We would like to bind service only to ip address from admin network
+# We would like to bind service only to ip address from admin network unless enable_designate is
+# enabled. In which case bind both the admin _and_ public.
 admin_network = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin")
 admin_addr = admin_network.address
+public_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "public").address
```
dirkmueller: why do we put this here and not inside the if below (line 403ff) ? it is only used once. I find it pretty cluttery to handle this generically here if its only for the designate case..

matthewoliver:
Rebased to include the tavis fix and addressed dirk's inline comment

PR #1888, opened 2019-07-31
network: override sysctl netfilter disable parameter (SOC-6229)

sandonovsuse:
SLES12 comes with sysctl net.bridge.bridge-nf-call-iptables
parameter set to zero, thus disabling netfilter on bridges on each
sysctl --system execution.

This patch overrides net.bridge.bridge-nf-call-iptables = 0 found in
sysctl.d/50-default.conf by creating a file 51-bridge-netfilter.conf,
setting the netfilter parameter to 1. (SOC-6229)


sandonovsuse:
Agreed, although I was trying to avoid possibly long discussion on how the change might affect products other than cloud - given that the bug is already in customer production. Personally, I would rather have this reach the MU quickly, and chase the aaa_base change in peace, even though it's double the effort. Also, happy to close the PR if you prefer otherwise?

rhafer:
>  Also, happy to close the PR if you prefer otherwise?

No. It's fine. I was just wondering if we're going to drop the ball on the aaa_base change (again).

sandonovsuse:
>@@sandonovsuse sandonovsuse force-pushed the sandonovsuse:override-sysctl-netfilter-disable-parameter-SOC-6229 branch from 8cefcff to 54e8405 

Commit title changed to match convention. Required by ci/cd

s-t-e-v-e-n-k:
That has been merged to master, rebase.

PR #1890, opened 2019-08-06
upgrade: Re-run upgrade on non-compute nodes (SOC-10072)

skazi0:
Previous search for not-upgraded non-compute nodes was just checking for `crowbar_upgrade` state. In case where crowbar_join wait loop would time out but the script would eventually finish, state would be changed to `ready` but the `node_upgrade_state` would still be `upgrading`. 
Old version was skipping such nodes when retrying the upgrade and they were left in half-upgraded state with no easy way to correct things.
    
The new version of non-compute nodes lookup follows the pattern present in other parts of upgrade code: find some subset of nodes, then exclude those which are upgraded, then sort to move half-upgraded nodes to front and (re)start the upgrade for this list.

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -1372,7 +1372,15 @@ module Api
       # This has to be done before compute nodes upgrade, so the live migration
       # can use fully upgraded cinder stack.
       def upgrade_non_compute_nodes
-        ::Node.find("state:crowbar_upgrade AND NOT run_list_map:nova-compute-*").each do |node|
+        non_compute_nodes = ::Node.find("NOT run_list_map:nova-compute-*")
+
+        # this should also remove all controllers and crowbar node from the list
+        non_compute_nodes.reject!(&:upgraded?)
+
+        # if we started upgrade of some node before, let's continue with it
+        non_compute_nodes.sort! { |n| n.upgrading? ? -1 : 1 }
+
+        non_compute_nodes.each do |node|
           upgrade_one_node node.name
         end
         save_nodes_state([], "", "")
--- a/crowbar_framework/spec/models/api/upgrade_spec.rb
+++ b/crowbar_framework/spec/models/api/upgrade_spec.rb
@@ -551,7 +551,7 @@ describe Api::Upgrade do
       # upgrade_non_compute_nodes:
       allow(Node).to(
         receive(:find).with(
-          "state:crowbar_upgrade AND NOT run_list_map:nova-compute-*"
+          "NOT run_list_map:nova-compute-*"
         ).and_return([ceph, testing])
       )
       allow_any_instance_of(Api::Node).to receive(:save_node_state).with(


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1372,7 +1372,13 @@ def upgrade_one_node(name)
       # This has to be done before compute nodes upgrade, so the live migration
       # can use fully upgraded cinder stack.
       def upgrade_non_compute_nodes
-        ::Node.find("state:crowbar_upgrade AND NOT run_list_map:nova-compute-*").each do |node|
+        # node_upgrade_state:upgrading is there to catch nodes which timed out waiting in
+        # join_and_chef and never got to save_node_state("upgraded") but are actually already
+        # upgraded.
```
jsuchome: I do not like this. If there's such node that is already upgraded by some step did not set it correctly, we should fix it in that step (`join_and_chef` here)

By extending the search here we may be asking for new problems. For example what if a node was in `upgrading` state but never really finished the upgrade? With this check, it will be ignored in next run (which could normally fix it)

skazi0: How would such node be ignored by this condition? It is wider than before so any node which was caught before will be caught now plus some more which were skipped before.

To have this fixed in `join_and_chef` we would need to do one of these:
1. wait for `crowbar_join` forever or very long time. As we're not killing the process, if it ever loops forever then we'll end up with a locked state anyway. Timeout here just makes things worse IMO.
2. kill `crowbar_join` after timeout. risky stuff.
3. (locally) modify `crowbar_join` to not move the node out of `crowbar_upgrade` state and do this transition inside the upgrade code.
4. set node state to `upgraded` inside `crowbar_join` call. risky as it would skip some keystone migration code and re-enabling of network agents.
5. other ideas?

jsuchome: You are right about condition, I somehow thought it's a condition to exclude the nodes. 
I'm not sure though what's the right approach now. I still feel like you are trying to work around the problem instead of facing it, I would rather see something different...

jsuchome: Just one thing that could be minor change of this idea for the better - make sure that the node in "upgrading" state comes as first, that way we should prevent situation with multiple nodes in such state. 
I think we actually have such search somewhere when upgrading compute or controller nodes.

skazi0: Good point! The nodes should be sorted to pick the one with `upgrading` first.

skazi0: The existing code you mentioned is here https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/api/upgrade.rb#L1093 and the logic around there is actually closer to what I added than the existing code for non-computes (i.e. checking `upgrading` not `crowbar_upgrade` state). I will change this part to follow the same pattern.

skazi0:
force-pushed with fixed tests

PR #1896, opened 2019-08-13
upgrade: remove nova-consoleauth service entries on upgrade (SOC-10164)

dirkmueller:
nova-consoleauth is no longer used in Cloud 9, so we should
be removing the service entries related to it.


skazi0:
@dirkmueller is consoleauth package and service removed somewhere? All I found is https://github.com/crowbar/crowbar-openstack/pull/1979 which just covers new deployments.

PR #1900, opened 2019-08-15
Allow enabling bpdu-forwarding on OVS bridges

zzaimeche:
This change allows users to enable BPDU forwarding on OVS bridges
that are used for Neutron provider networks (br-fixed, br-floating).
Enabling it will allow the physical network to detect and stop loops
if these bridges are accidentally misconfigured. It is turned off
by default.

More context in:

  https://bugzilla.suse.com/show_bug.cgi?id=1133804

### DIFF
--- a/chef/cookbooks/barclamp/libraries/nic.rb
+++ b/chef/cookbooks/barclamp/libraries/nic.rb
@@ -843,6 +843,10 @@ class ::Nic
       ::Kernel.system("ovs-vsctl add-port #{@nic} #{slave}")
     end
 
+    def ovs_forward_bpdu(forward)
+      ::Kernel.system("ovs-vsctl set Bridge #{@nic} other_config:forward-bpdu=#{forward}")
+    end
+
     def self.create(nic, slaves = [])
       Chef::Log.info("Creating new OVS bridge #{nic}")
       if self.exists?(nic)
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -448,6 +448,12 @@ sorted_networks.each do |network|
     our_iface = br
     net_ifs << our_iface.name
   end
+
+  unless network.ovs_forward_bpdu.nil?
+    Chef::Log.info("OVS BPDU forwarding set to #{network.ovs_forward_bpdu}.")
+    ifs[our_iface.name]["ovs_forward_bpdu"] = network.ovs_forward_bpdu
+  end
+
   if network.mtu
     Chef::Log.info("Using mtu #{network.mtu} for #{network.name} network on #{our_iface.name}")
     ifs[our_iface.name]["mtu"] = network.mtu
@@ -682,7 +688,10 @@ when "suse"
       end
 
       pre_up_script = "/etc/wicked/scripts/#{nic.name}-pre-up"
+      ovs_forward_bpdu = ifs[nic.name]["ovs_forward_bpdu"] || false
       is_admin_nwk = if_mapping.key?("admin") && if_mapping["admin"].include?(nic.name)
+
+      nic.ovs_forward_bpdu(ovs_forward_bpdu)
 
       template pre_up_script do
         owner "root"
@@ -691,6 +700,7 @@ when "suse"
         source "ovs-pre-up.sh.erb"
         variables(
           bridgename: nic.name,
+          ovs_forward_bpdu: ovs_forward_bpdu,
           is_admin_nwk: is_admin_nwk
         )
       end
--- a/chef/cookbooks/network/templates/default/ovs-pre-up.sh.erb
+++ b/chef/cookbooks/network/templates/default/ovs-pre-up.sh.erb
@@ -1,6 +1,7 @@
 #! /bin/bash
 
 ovs-vsctl br-exists <%= @bridgename %> || exit 0
+ovs-vsctl set bridge <%= @bridgename %> other-config:forward-bpdu=<%= @ovs_forward_bpdu %>
 <%
   # remove the "secure" fail-mode for bridges that share an interface
   # with the "admin" network, otherwise the admin network will be offline
--- a/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+++ b/chef/cookbooks/barclamp/libraries/barclamp_library.rb
@@ -92,6 +92,7 @@ module BarclampLibrary
         attr_reader :vlan, :use_vlan
         attr_reader :add_bridge, :add_ovs_bridge, :bridge_name
         attr_reader :conduit
+        attr_reader :ovs_forward_bpdu
 
         def initialize(node, net, data)
           @node = node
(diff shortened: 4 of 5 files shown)


zzaimeche:
> The pre-up script will only run at the next boot, so should chef execute the new ovs-vsctl command too?

Ah, probably; I feel like the user would expect any changes they make to apply immediately upon applying the barclamp. Not sure if there's a preferred/standard way for crowbar to do this; will take a look.

rhafer:
> Ah, probably; I feel like the user would expect any changes they make to apply immediately upon applying the barclamp. Not sure if there's a preferred/standard way for crowbar to do this; will take a look.

The preferred way is to apply the changes immediately. In this case you'd probably just enhance the `Nic::OvsBridge` class in chef/cookbooks/barclamp/libraries/nic.rb with a method(s) to get/set the bpdu-forwarding flag on a bridge and all that from with chef/cookbooks/network/recipes/default.rb

chef/cookbooks/barclamp/libraries/nic.rb
```diff
@@ -820,6 +820,10 @@ def replug(slave)
       ::Kernel.system("ovs-vsctl add-port #{@nic} #{slave}")
     end
 
+    def ovs_forward_bpdu(forward)
+      ::Kernel.system("ovs-vsctl set Bridge #{@nic} other-config:forward-bpdu=#{forward}")
```
rhafer: I am not entirely sure how ovs work internally. But wouldn't it be better to actually check the current value before writing it?
E.g. call `ovs-vsctl get Bridge #{@nic} other-config:forward-bpdu` first and only run the set command if an update is needed?

chef/data_bags/crowbar/template-network.json
```diff
@@ -184,6 +184,7 @@
           "add_bridge": false,
           "add_ovs_bridge": false,
           "bridge_name": "br-fixed",
+	  "ovs_forward_bpdu": false,
```
rhafer: There seem to be an indentation issue here.

Also I am wondering, since we marked the attribute optional. Should we really actually set a default here? Wouldn't it be better to rely on what ever ovs defaults to? Which means you also wouldn't need to do a schema migration. But in that case `Nic::OvsBridge.ovs_forward_bpdu(forward)` would need to learn about a third possible value `nil` which would mean, do whatever OVS does by default. @djoreilly  Any opinion on this?

zzaimeche: argh, will fix the indentation! (that's what happens when you paste from the file on your crowbar box at the end of the day...)  I think the field is just blank by default (at least, if you do `ovs-vsctl list Bridge` you won't see anything at all in the other_config section until you specify keys and values)-- but I don't know ovs really, just reporting what I see.

djoreilly: > There seem to be an indentation issue here.
> 
> Also I am wondering, since we marked the attribute optional. Should we really actually set a default here? Wouldn't it be better to rely on what ever ovs defaults to? Which means you also wouldn't need to do a schema migration. But in that case `Nic::OvsBridge.ovs_forward_bpdu(forward)` would need to learn about a third possible value `nil` which would mean, do whatever OVS does by default. @djoreilly Any opinion on this?

yes, that seems cleaner. When OVS does not find `other_config:forward-bpdu` defined it uses false.

chef/cookbooks/barclamp/libraries/nic.rb
djoreilly: s/other-config/other_config/ - `other-config` actually works too, but `other_config` is more in keeping with the convention.

nicolasbock:
The linter is complaining about a missing Jira reference in the commit message subject. Could you add one?

chef/data_bags/crowbar/template-network.json
zzaimeche: I'm not sure why this wouldn't need a schema migration? Since the value hasn't been set from the barclamp so far, and aiui the user would be setting the value in the raw view of the barclamp. I may be misunderstanding something.

chef/cookbooks/network/templates/default/ovs-pre-up.sh.erb
```diff
@@ -1,6 +1,7 @@
 #! /bin/bash
 
 ovs-vsctl br-exists <%= @bridgename %> || exit 0
+ovs-vsctl set bridge <%= @bridgename %> other-config:forward-bpdu=<%= @ovs_forward_bpdu %>
 <%
```
djoreilly: Just found out that setting this here will have no effect if neutron has https://github.com/openstack/neutron/commit/1f8378e0ac4b8c3fc4670144e6efc51940d796ad because the ovs agent overwrites other_config when it starts.

chef/data_bags/crowbar/template-network.json
rhafer: You defined the attribute to be optional in the schema `{ "type": "bool", "required": false },` , which IMO is correct as we don't what every network to carry that attribute. But it means the validator doesn't require the attribute to be present. So you need to prepare the rest of your code to cope with proposals where the attribute is not set at all (even for the networks where you set a default, as the user could just delete it and the validator would not complain).

My proposal would be to not set a default anywhere and just document the new attribute in the deployment Guide `http://docserv.nue.suse.com/documents/SUSE_OpenStack_Cloud_9/suse-openstack-cloud-crowbar-deployment/single-html/#sec-deploy-network-json-networks` as I don't expect many setups to fiddle around with this setting.

When there is a network that doesn't have the `ovs_forward_bpdu` set to anything `BarclampLibrary::Barclamp::Inventory::Network.ovs_forward_bpdu` would return `nil` (I think). In that case the network cookbook just IMO just clear any existing value from the OVS database (and use what ever default OVS is using)

Only if the attribute is explicitly set in the proposal the cookbook should set it to either true of false.

chef/cookbooks/network/templates/default/ovs-pre-up.sh.erb
rhafer: > Just found out that setting this here will have no effect if neutron has [openstack/neutron@1f8378e](https://github.com/openstack/neutron/commit/1f8378e0ac4b8c3fc4670144e6efc51940d796ad) because the ovs agent overwrites other_config when it starts.

@djoreilly Did you test this with something newer than newton? I thought the above problem is newton specific (and since yesterday fixed in our packages)

djoreilly: I had only tested on newton. I have just tested pike and there is no problem.

chef/data_bags/crowbar/template-network.json
zzaimeche: Ah, okay, gotcha. :) Thanks.

chef/cookbooks/network/templates/default/ovs-pre-up.sh.erb
dirkmueller: this problem should be fixed though

chef/cookbooks/network/recipes/default.rb
```diff
@@ -448,6 +448,12 @@ def kill_nic(nic)
     our_iface = br
     net_ifs << our_iface.name
   end
+
+  if network.ovs_forward_bpdu
```
dirkmueller: this looks wrong to me. you don't want to see if it is "true" but if it is "defined". because setting it to false should also be done (even if its the code default, it might not have been the setting in the currently installed configuration)

zzaimeche: oh, good point, thanks

PR #1901, opened 2019-08-19
allow user to ask for FQDN as public hostname (SOC-9616)

guangyee:
Currently CrowbarHelper.get_host_for_public_url() indiscriminately
return the IP instead of FQDN in the situation where the
node's public_name is not set and SSL is not being used.
This behavior was introduced in

https://github.com/crowbar/barclamp-glance/commit/22d3c190f1edc6f4d278982eb58076acdde16505

to address certain situations where the crowbar generated hostnames are not
resolveable externally. For SSL, we need the FQDN or hostname for
certificate validation. However, for WebSSO, certain identity providers
(i.e. Google) requires FQDN to match the authorized domain, regardless of
whether SSL is used.

In theory, we could change the behavior of get_host_for_public_url() to
always return the FQDN instead of IP as we *should be* supporting
external name resolution by now, but doing so may risk breaking
backward compatibility. Therefore, this patch introduce an option to allow
users to ask for FQDN regardless whether SSL is used without breaking
backward compatibility.

### DIFF
--- a/chef/cookbooks/utils/libraries/helpers.rb
+++ b/chef/cookbooks/utils/libraries/helpers.rb
@@ -33,7 +33,8 @@ module CrowbarHelper
     end
   end
 
-  def self.get_host_for_public_url(node, use_ssl, use_cluster = false)
+  def self.get_host_for_public_url(node, use_ssl, use_cluster = false,
+                                   want_fqdn = false)
     if use_cluster && defined?(CrowbarPacemakerHelper)
       # loose dependency on the pacemaker cookbook
       cluster_vhostname = CrowbarPacemakerHelper.cluster_vhostname(node)
@@ -49,14 +50,15 @@ module CrowbarHelper
     end
 
     # For the public endpoint, we prefer the public name. If not set, then we
-    # use the IP address except for SSL, where we always prefer a hostname
-    # (for certificate validation).
+    # use hostname when approprate (i.e. for SSL certificate validation or
+    # WebSSO identification), otherwise we use the IP address.
     if public_name.nil? || public_name.empty?
-      if use_ssl
-        public_name = public_fqdn
-      else
-        public_name = public_ip
-      end
+      public_name =
+        if use_ssl || want_fqdn
+          public_fqdn
+        else
+          public_ip
+        end
     end
 
     public_name


toabctl:
@rsalevsky do we plan to fix travis? Currently failing because of CVE-2019-5477 (see https://travis-ci.org/crowbar/crowbar-core/jobs/574040609#L1570 )

rhafer:
> can you please explain this a bit more? if you set the public name on the cluster or the host then that one should be picked?

Yes, and AFAICS this PR doesn't change that behavior. It only changes what is returned when no public name is set.

> I am not necessarily adding another option, but it seems incorrect to me to do that only for keystone/websso then. we should generalize the solution (for example by simply always returning the hostname rather than ip addresses).

IIRC we specifically introduced the behavior to return IP addresses for the endpoints by default, to be able to work when no public name is set and the crowbar generated hostnames are not resolveable externally (see: https://github.com/crowbar/barclamp-glance/commit/22d3c190f1edc6f4d278982eb58076acdde16505 for some very old history). Not sure if that is really a good reason still (or if it ever was :) ). But we might break existing setups if we set fqdn for the public endpoints everywhere now. That's why I hinted in https://github.com/crowbar/crowbar-openstack/pull/2188#discussion_r315249080 at making this optional.

guangyee:
> > can you please explain this a bit more? if you set the public name on the cluster or the host then that one should be picked?
> 
> Yes, and AFAICS this PR doesn't change that behavior. It only changes what is returned when no public name is set.
> 
> > I am not necessarily adding another option, but it seems incorrect to me to do that only for keystone/websso then. we should generalize the solution (for example by simply always returning the hostname rather than ip addresses).
> 
> IIRC we specifically introduced the behavior to return IP addresses for the endpoints by default, to be able to work when no public name is set and the crowbar generated hostnames are not resolveable externally (see: [crowbar/barclamp-glance@22d3c19](https://github.com/crowbar/barclamp-glance/commit/22d3c190f1edc6f4d278982eb58076acdde16505) for some very old history). Not sure if that is really a good reason still (or if it ever was :) ). But we might break existing setups if we set fqdn for the public endpoints everywhere now. That's why I hinted in [crowbar/crowbar-openstack#2188 (comment)](https://github.com/crowbar/crowbar-openstack/pull/2188#discussion_r315249080) at making this optional.

Probably won't break as we are already returning the FQDN if SSL is enabled. That is unless we do a bunch of other hacks to DNS if SSL is enabled.

dirkmueller:
> Probably won't break as we are already returning the FQDN if SSL is enabled. That is unless we do a bunch of other hacks to DNS if SSL is enabled.

right, which is why I was wondering why we need to make it extra complicated. But I do accept and agree with ralfs comment that this is on a maintenance branch so we should be careful  with changing default behavior in a maintenance update. so I'll no longer object. would still be nice to capture this in the commit message rather than in a comment on the PR..

guangyee:
@rhafer thanks for the reminder. I've updated the commit msg.

PR #1903, opened 2019-08-20
ohai: Hardcode ruby version for package installation (SOC-10010)

rsalevsky:
Sometimes there is a race condition and ohai didn't collect the ruby
version. to_f evalutes then the version to 0.0 and zypper fails to
install the rubygem `ruby0.0-rubygem-cstruct' not found in package
names`.


rsalevsky:
All other packages already use hardcoded ruby versions.

PR #1906, opened 2019-08-23
IPV6 - Expose methods in chef NetworkHelper to crowbar_framework (SOC-6397)

guglo:
Barclamp horizon need to handle ipv6. NetworkHelper in chef/cookbooks/network/libraries/helpers.rb has methods that can be used to deal with ipv6 in barclamp horizon. 
We assume the addresses used in barclamp horizon are ip addresses. 

This commit includes:
1) Because the methods in Chef NetworkHelper are not exposed to crowbar_framework, added the path of chef/cookbooks/network/libraries/helpers.rb in crowbar_framework/config/application.rb so they can be called from crowbar_framework. 
 2) Added boolean ipv6 method to check if the address is in ipv6 form.

### DIFF
--- a/chef/cookbooks/network/libraries/helpers.rb
+++ b/chef/cookbooks/network/libraries/helpers.rb
@@ -7,4 +7,13 @@ module NetworkHelper
       address.to_s
     end
   end
+
+  def self.ipv6(address)
+    require "ipaddr"
+    if IPAddr.new(address).ipv6?
+      true
+    else
+      false
+    end
+  end
 end
--- a/crowbar_framework/config/application.rb
+++ b/crowbar_framework/config/application.rb
@@ -22,6 +22,7 @@ module Crowbar
     # Explicitely eager load /lib/crowbar/lock so we can use SharedNonBlocking
     # with threading without hitting circular dependencies
     config.eager_load_paths += Dir["#{config.root}/lib/crowbar/lock"]
+    config.eager_load_paths += Dir["#{config.root}/../chef/cookbooks/network/libraries"]
 
     config.autoload_paths += %W(
       #{config.root}/lib


crowbar_framework/app/helpers/barclamp/network_helper.rb
```diff
@@ -17,5 +17,13 @@
 
 module Barclamp
   module NetworkHelper
+    def self.wrap_ip(address)
```
nicolasbock: This duplicates the code in https://github.com/crowbar/crowbar-core/blob/204289d652f4a6c974c458595fdc8308c986b40f/chef/cookbooks/network/libraries/helpers.rb#L2. I would prefer if we had this code only once. @matthewoliver , when @guglo and I tried to use your code in `horizon_service.rb` we couldn't convince crowbar to use it. I am guessing it's a namespace issue or a missing include somewhere. Could you comment?

guglo: I experimented more today, however I am newbee and not ruby savvy so I could not find a better way. Need help :)

matthewoliver: Looking closer at rails apps. We might be able to add something to `crowbar_framework/config/application.rb`. Like:

`config.eager_load_paths += <path into the chef/barclamp ruby code>`

Which will apparently extend the "scope" outside of the app/ directory. Once my current deploy is done I can have a play.

matthewoliver: Ok. So I've just been playing with hard coding paths.. because I want to know if it'll work.
I started with:
  `config.eager_load_paths += Dir["/opt/dell/chef"]`
But that died with not knowing a Node object was. There is a bunch of chef and other ruby code that may get in the way and cause issues. So instead I was more specific and tried:
`config.eager_load_paths += Dir["/opt/dell/chef/cookbooks/network/libraries"]`
I.e: http://pastebin.nue.suse.com/18645

And restarting crowbar (and crowbar-jobs) seemed to be fine with it. This includes attempting to use the NetworkHelper in `crowbar_framework/app/models/horizon_service.rb` with:
`node.crowbar["crowbar"]["links"]["OpenStack Dashboard (public)"] = "#{protocol}://#{NetworkHelper.wrap_ip(public_server_ip)}/"`

So from the surface it "looks" like it works.Looks kinda ugly having a static path. So it would be nice if we could find a way of making it more portable. 

But have run out of time to do any more playing today.

guglo: @matthewoliver  Thanks for helping, tried config.eager_load_paths += Dir["#{config.root}/../chef/cookbooks/network/libraries"] 
it seems to be ok. Will update and remove the duplication.

matthewoliver:
Seems people are happy with our way of extending crowbars ruby on rails scope with eager_load. Because I was involved in researching it I shouldn't be one to approve.

But because we have 2 approvals, I _can_ press the merge, this way crowbar-openstack's 2194 PR's CI, which is dependent on this, will pass.

PR #1907, opened 2019-08-27
upgrade: Add more prechecks for 8->9 (SOC-9868)

skazi0:
Additional prechecks were added to detect problematic scenarios before
upgrade is started. These cover:
- Aodh deployment
- Trove deployment
- "Legacy" Ceilometer (without Monasca) deployment

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -278,6 +278,18 @@ module Api
           end
           ret[:controller_roles] = { node: node.name, roles: wrong_roles } if wrong_roles.any?
         end
+        # Make sure Aodh is not deployed
+        if Proposal.where(barclamp: "aodh").any?
+          ret[:aodh_proposal] = true
+        end
+        # Make sure Trove is not deployed
+        if Proposal.where(barclamp: "trove").any?
+          ret[:trove_proposal] = true
+        end
+        # Make sure Ceilometer is deployed with Monasca or not at all
+        if ::Node.find("roles:ceilometer-server").any? && ::Node.find("roles:monasca-server").none?
+          ret[:legacy_ceilometer] = true
+        end
         ret
       end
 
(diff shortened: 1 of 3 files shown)


skazi0:
Zrzut ekranu z 2019-08-27 17-00-49

crowbar_framework/app/models/api/crowbar.rb
```diff
@@ -278,6 +278,18 @@ def deployment_check
...
+        if Proposal.where(barclamp: "aodh").any?
+          ret[:aodh_proposal] = true
+        end
+        # Make sure Trove is not deployed
+        if Proposal.where(barclamp: "trove").any?
+          ret[:trove_proposal] = true
+        end
+        # Make sure Ceilometer is deployed with Monasca or not at all
+        if Proposal.where(barclamp: "ceilometer").any?
+          ret[:legacy_ceilometer] = Proposal.where(barclamp: "monasca").none?
```
jsuchome: Does this make sure that the  monasca is actually active?

skazi0: Nope. This is just a rough check for some legacy deployments.
@jgrassler do you think we need more or checking the existence of proposal is enough?

jsuchome: I don't know the proposal values for now, but it should not be difficult to check for active proposals ... I guess we are doing it for some other check in the same file

jgrassler: Hmm. We could indeed check if the `monasca-server` role is indeed applied, since that's what other barclamps use. It doesn't what (working or broken) state Monasca is in since the Monasca barclamp is going to rectify that later...as long as there's a node with the `monasca-server` role applied. That would be cleaner, for that eliminates a possible (though probably uncommon) failure mode during the upgrade.

skazi0: I changed the ceilometer/monasca check to look at the nodes not proposals. For Aodh and Trove I keep it on proposals since we want a negative check there.

skazi0:
I changed condition for Ceilometer/Monasca to set `ret[:legacy_ceilometer]` only when it's `true`. Other parts of code rely on `ret` being empty for passing check.

JanZerebecki:
This merged something that broke existing spec tests. It added new find calls that these tests aren't expecting.

skazi0:
@JanZerebecki uh, sorry. the fix: https://github.com/crowbar/crowbar-core/pull/1923

PR #1908, opened 2019-08-28
dns: Always set master_ip

s-t-e-v-e-n-k:
Currently, the DNS barclamp will only set master_ip and slave_ips if
there is more than one node. However, this is causing an issue with the
designate barclamp which now relies on it being set all the time, so set
for the single node case too.

### DIFF
--- a/crowbar_framework/app/models/dns_service.rb
+++ b/crowbar_framework/app/models/dns_service.rb
@@ -123,6 +123,7 @@ class DnsService < ServiceObject
       node.set[:dns] = {} if node[:dns].nil?
       unless node[:dns][:master]
         node.set[:dns][:master] = true
+        node.set[:dns][:master_ip] = node.get_network_by_type("admin")["address"]
         node.save
       end
     elsif nodes.length > 1


matthewoliver:
Makes sense to me. Always adding it shouldn't hurt anything but will wait to for CI to confirm.

If it did we could add an:
`if role.default_attributes[:dns][:enable_designate]`

But I think just setting it all the time is easier to grok.. as if there is one.. it should be the master :)

crowbar_framework/app/models/dns_service.rb
```diff
@@ -123,6 +123,7 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
       node.set[:dns] = {} if node[:dns].nil?
       unless node[:dns][:master]
         node.set[:dns][:master] = true
+        node.set[:dns][:master_ip] = node.get_network_by_type("admin")["address"]
```
dirkmueller: this is for non-master nodes to look up the ip of the master ip. should we set it to the same node actually? why not simply fix the designate barclamp to just not look for master_ip if it is master?

s-t-e-v-e-n-k:
Okay, I'll fix the Designate barclamp instead

PR #1912, opened 2019-08-30
Allow designate rndc for all nodes (SOC-10339)

ritesh216:
With designate enabled, all Bind servers act as slaves for
designate created zones. These zones are created via rndc addzone
from designate server(s).

Adding permissions and key for allowing rndc access.

### DIFF
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -355,7 +355,7 @@ template "/etc/bind/named.conf.crowbar" do
   notifies :reload, "service[bind9]"
 end
 
-if node[:dns][:enable_designate] && node[:dns][:master]
+if node[:dns][:enable_designate]
   template "/etc/named.d/designate-rndc-access.conf" do
     source "designate-rndc-access.conf.erb"
     mode 0o640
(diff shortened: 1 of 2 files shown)


sjamgade:
I have a few questions, (this dns stuff is still very complicated for me) 
- why localhost ? asaik mdns is not listening on localhost, so it will never send requests via "localhost" to add zone.
- we use bind slaves as forwarders (forward only), so all requests are forwarded to bind master (forwarders clause in named.conf), do we no longer have a dns-master ?
- is the change of master_ip to admin_ip necessary ?

ritesh216:
Hi Sumit,

With the approach we are taking, the Bind9 servers become slaves to Designate-mdns service(s) and zones created/updated via/in Openstack will be rndc'ed to the slaves. For this we need the changes.

Answers to your questions: 
    why localhost ? asaik mdns is not listening on localhost, so it will never send requests via "localhost" to add zone.
-> I was covering for the case in which we have Bind9 server and Designate-mdns running on the same node. I will undo this change and see if anything fails. Better to be less permissive. 
    we use bind slaves as forwarders (forward only), so all requests are forwarded to bind master (forwarders clause in named.conf), do we no longer have a dns-master ?
-> afaik, master slave concept is not black and white; with Designate integration, for Zones created in Openstack, Designate-mdns is master and Bind9s are all slaves.
    is the change of master_ip to admin_ip necessary ?
-> Yes, as we are permitting rndc to all Bind9s.

PR #1916, opened 2019-09-05
upgrade: Upgrade compute services early (SOC-10273)

skazi0:
As RPC 4.x support was dropped with https://review.opendev.org/#/c/543580/
Rocky controller can't talk to Pike compute anymore. We need to upgrade
nova compute services before the controllers so that everything is running
on new RPC when we get to live-migrate the VMs during upgrade.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
@@ -90,7 +90,7 @@ log "Upgrading packages needed for nova-compute node"
 
 zypper --non-interactive up openstack-nova-compute
 
-log "Restarting services for Pike"
+log "Restarting services for Rocky"
 
 <% unless @remote_node %>
 systemctl restart <%= @neutron_agent %>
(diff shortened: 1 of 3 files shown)


chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -86,6 +86,10 @@ log "No HA setup found..."
 
 <% if @compute_node %>
 
+log "Upgrading packages needed for nova-compute node"
+
+zypper --non-interactive up openstack-nova-compute
+
 log "Restarting services for Pike"
```
dirkmueller: can you fix the "Pike" comment?

jsuchome:
So is this at the same place we had it during 6-7 upgrade?

skazi0:
@jsuchome I have not confirmed it yet but it seems to be exactly the same situation as we had before.

chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
skazi0: done

PR #1918, opened 2019-09-09
[5.0] Add some API to allow non-privileged user to be used for discovery/crowbar_join/crowbar_register (bsc#1117080) 

skazi0:
port of #1817


cmurphy:
mkcloud failures look potentially related?

skazi0:
@cmurphy looks like we need this https://github.com/crowbar/crowbar-client/pull/191 followed by gem release (and later maintenance updates for crowbar-client and sleshammer).

PR #1923, opened 2019-09-11
upgrade: Fix pre-checks tests (SOC-9868)

skazi0:


nicolasbock:
I restated the HA test

JanZerebecki:
HA fails because of zypper mirror problems, tracked in SOC-10328.

PR #1935, opened 2019-09-24
Designate: Don't add the admin node to the public network (SOC-10658)

matthewoliver:
The DNS barclamp fails to apply when designate is enabled. This is
because in that case we try and add the admin node to the public
network.
In our CI this fails as the admin node only has 1 NIC and the admin
network has a lower router preference then public, so the default route
is killed.

Due to the topology supported by designate and potential security
concerns it's better that we never connect the admin node.

There is a crowbar-openstack PR to filter the admin node out in the
designate barclamp. This is the core side.

### DIFF
--- a/crowbar_framework/app/models/dns_service.rb
+++ b/crowbar_framework/app/models/dns_service.rb
@@ -113,14 +113,19 @@ class DnsService < ServiceObject
     return if all_nodes.empty?
 
     tnodes = role.override_attributes["dns"]["elements"]["dns-server"]
-    # If designate is enabled, we need each DNS node to be attached to the public network.
+    # If designate is enabled, we need each DNS node to be attached to the
+    # public network, except for the admin node
     net_svc = NetworkService.new @logger
+    nodes = []
     tnodes.each do |node|
-      if role.default_attributes[:dns][:enable_designate]
+      n = ::Node.find_by_name(node)
+      if role.default_attributes[:dns][:enable_designate] && !n.admin?
         net_svc.allocate_ip "default", "public", "host", node
+        # Reload the node
+        n = ::Node.find_by_name(node)
       end
+      nodes << n
     end
-    nodes = tnodes.map { |n| Node.find_by_name(n) }
 
     if nodes.length == 1
       # remember that this node will stick as master node, in case we add some


crowbar_framework/app/models/dns_service.rb
```diff
@@ -116,7 +116,8 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
     # If designate is enabled, we need each DNS node to be attached to the public network.
     net_svc = NetworkService.new @logger
     tnodes.each do |node|
-      if role.default_attributes[:dns][:enable_designate]
+      n = ::Node.find_by_name(node)
+      if role.default_attributes[:dns][:enable_designate] && !n.admin?
         net_svc.allocate_ip "default", "public", "host", node
       end
     end
```
s-t-e-v-e-n-k:
```
    # If designate is enabled, we need each DNS node to be attached to the
    # public network, except for the admin node
    net_svc = NetworkService.new @logger
    nodes = []
    tnodes.each do |node|
      n = ::Node.find_by_name(node)
      nodes << n
      if role.default_attributes[:dns][:enable_designate] && !n.admin?
        net_svc.allocate_ip "default", "public", "host", node
      end
    end
```

s-t-e-v-e-n-k: Fixes the comment, and stops looping over the list twice.

matthewoliver: The problem here, and it's confirm after testing. Is what allocate_ip does is it takes a node name, loads it up, adds a new network IP then saves it again. However back here we have a node object loaded that is stale, doesn't have the network set in it. So when it is saved it removes what allocate_ip has just done.
When we get to the network barclamp, it's suppose to see that there is a new network and IP in the node so doesn't the actual connection magic. But because the stale version has overwritten it. It never happens.

So I like this refactor, but we either need to ignore it or:
```
   # If designate is enabled, we need each DNS node to be attached to the
    # public network, except for the admin node
    net_svc = NetworkService.new @logger
    nodes = []
    tnodes.each do |node|
      n = ::Node.find_by_name(node)
      if role.default_attributes[:dns][:enable_designate] && !n.admin?
        net_svc.allocate_ip "default", "public", "host", node
        # reload the node
        n = ::Node.find_by_name(node)
      end
     nodes << n
    end

```

matthewoliver: Cool the run hasn't finished yet. But I do see both nodes connected to the public network now.. so looks like my above snippet what reloads the node id allocate_ip is run works.

matthewoliver: pushed up Steve's refactor + the reload node addition.

PR #1936, opened 2019-09-25
Revert "[upgrade] Use block-migration when needed"

jgrassler:
This reverts commit 5c2fece34bdbd51c1700d883dac955ca7b6ff468.
Specifying block migration explicitly is no longer needed as
of Nova API version 2.25 (Mitaka and newer). From this API
version onward Nova will automatically select the appropriate
migration mode.

Setting "do not merge yet" because we are currently testing this on a QA cloud for both the Cloud 7 -> Cloud 8 and Cloud 8 -> Cloud 9 upgrade. I'll update once that's through.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -130,19 +130,12 @@ template "/usr/sbin/crowbar-delete-unknown-nova-services.sh" do
   only_if { roles.include?("nova-controller") }
 end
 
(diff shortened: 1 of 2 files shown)


jgrassler:
Tested and works on the 7 -> 8 upgrade. Merging for Cloud 9 and backporting because we couldn't test for Cloud 9, yet due to a bit of an RPC API incompatibility issue we still need to fix (SOC-10761). If there _are_ problems on the Cloud 9 side we'll fix them in the course of getting the live migration upgrade working for Cloud 9.

PR #1940, opened 2019-10-01
[5.0] Designate: Don't add the admin node to the public network (SOC-10658)

matthewoliver:
The DNS barclamp fails to apply when designate is enabled. This is
because in that case we try and add the admin node to the public
network.
In our CI this fails as the admin node only has 1 NIC and the admin
network has a lower router preference then public, so the default route
is killed.

Due to the topology supported by designate and potential security
concerns it's better that we never connect the admin node.

There is a crowbar-openstack PR to filter the admin node out in the
designate barclamp. This is the core side.

(cherry picked from commit 5a5c6a94544c2226cb04914054d87bad88ae6a6f)

### DIFF
--- a/crowbar_framework/app/models/dns_service.rb
+++ b/crowbar_framework/app/models/dns_service.rb
@@ -113,14 +113,19 @@ class DnsService < ServiceObject
     return if all_nodes.empty?
 
     tnodes = role.override_attributes["dns"]["elements"]["dns-server"]
-    # If designate is enabled, we need each DNS node to be attached to the public network.
+    # If designate is enabled, we need each DNS node to be attached to the
+    # public network, except for the admin node
     net_svc = NetworkService.new @logger
+    nodes = []
     tnodes.each do |node|
-      if role.default_attributes[:dns][:enable_designate]
+      n = ::Node.find_by_name(node)
+      if role.default_attributes[:dns][:enable_designate] && !n.admin?
         net_svc.allocate_ip "default", "public", "host", node
+        # Reload the node
+        n = ::Node.find_by_name(node)
       end
+      nodes << n
     end
-    nodes = tnodes.map { |n| Node.find_by_name(n) }
 
     if nodes.length == 1
       # remember that this node will stick as master node, in case we add some


matthewoliver:
> @here where are we actually checking and notifying the user that you must select at least one (ideally more than one) non-admin dns-server node role for designate to work? I would expect at least here if enable_designate is true to fail validation and return a nice error instead.

@dirkmueller The problem here though is the DNS barclamp is deployed with just the admin node during instcrowbar, also when we only have 1 node, the admin node. Which would mean if enable_designate is enabled instcrowbar will fail. Breaking CI/CD.

We either leave it to the users to deal with having extra nodes in DNS if they want to deploy designate. Change the CI/CD to setup an extra node when designate is enabled and known by crowbar for instcrowbar (seems complicated). Or in mkcloud run instcrowbar with enable_designate to false, then once that's finished modify the DNS proposal to enable it so the check would work when redeploying the DNS barclamp during the proposal step.

I just went with the less evasive to the rest of crowbar approach.

matthewoliver:
I guess form the CI/CD side, we could not set the `enable_designate` in the DNS proposal until after installcrowbar. Ie part of the proposal step.
That would mean an additional patch to automation. But then we may also need more documentation on to users about not enabling `enable-designate` until the crowbar admin node has been deployed, even if you know you'll want to enable designate in the future.

Carl has already added documentation about the DNS barclamp needing at least 1 non admin DNS node in the DNS barclamp if you want to deploy designate. So currently as it stands there is a social solution in place.

ritesh216:
imho, Designate can still be deployed with no non-admin node playing dns-server role. Just that, the interesting aspect of Designate, namely, "Integration with an external DNS service" is not achieved; as we have an empty default Bind9 pool. Note, we can still assign dns-domain and dns-name to Ports and FIPs and dns-domain to Networks. So I am ok with this patch. The doc change also helps Customers.

ritesh216:
While I think we are right, in theory, to apply Designate barclamp, even when no non-admin node is dns-server; looks like the Designate barclamp fails in that case.
It fails at:
[2019-10-02T16:02:35.922Z]      86>>   variables(rndc_key: dns_all.first[:dns][:designate_rndc_key])

with no method error ('undefined method `[]' for nil:NilClass'). Found by Designate-CI story the Jenny is working on.

dirkmueller:
@matthewoliver it just means we can't eanble the designate integration until there is a 2nd node imho, to avoid breaking CI/CD. the whole point of CI/CD is to test what a customer would test, so the deployment needs to be realistic.

I think the validation still needs to be done but we don't have to block this patch on it.

matthewoliver:
Ritesh did add logging a warning if there where no DNS barclamp servers in the designate barclamp (after the filtering out of the admin node) in https://github.com/crowbar/crowbar-openstack/pull/2281

So maybe we can land this one.

PR #1943, opened 2019-10-15
upgrade: Reload repo config in repochecks (SOC-10718)

skazi0:
When /etc/crowbar/repos.yml was modified manually, nodes repochecks
sometimes returned old content which could be confusing.
Forcefully reloading the configs each time the repochecks are run
solves the problem.


dirkmueller:
Hopefully performance impact is minimal

skazi0:
@dirkmueller it is done only once when prechecks are executed s

PR #1945, opened 2019-10-21
[5.0] Ignore Cloud repository when checking for required ones

jsuchome:
Cloud repo should not be necessary to have, so if user does not have it, upgrade checks should not complain.

### DIFF
--- a/crowbar_framework/app/models/api/node.rb
+++ b/crowbar_framework/app/models/api/node.rb
@@ -257,9 +257,10 @@ module Api
                                                                    architecture)
                   ::Openstack::Upgrade.enable_repos_for_feature(feature, Rails.logger)
                 end
-                available, repolist = ::Crowbar::Repository.provided_and_enabled_with_repolist(
-                  feature, platform, architecture
-                )
+                available, repolist =
+                  ::Crowbar::Repository.provided_and_enabled_with_repolist_for_upgrade(
+                    feature, platform, architecture
+                  )
                 ret[addon]["available"] &&= available
                 ret[addon]["errors"].deep_merge!(repolist.deep_stringify_keys)
               end
--- a/crowbar_framework/lib/crowbar/repository.rb
+++ b/crowbar_framework/lib/crowbar/repository.rb
@@ -137,6 +137,12 @@ module Crowbar
         provided_with_enabled(feature, platform, arch, true)
       end
 
+      # For upgrade related checks we need that Cloud repo is not needed
+      # so we explicitly exclude it from the list. See also comment in feature_repository_map
+      def provided_and_enabled_with_repolist_for_upgrade(feature, platform = nil, arch = nil)
+        provided_with_enabled(feature, platform, arch, true, true)
+      end
+
       def provided?(feature, platform = nil, arch = nil)
         provided_with_enabled(feature, platform, arch, false).first
       end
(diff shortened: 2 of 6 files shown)


crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -256,6 +256,14 @@ def provided_with_enabled(feature,
             unless repos[:inactive][r.arch.to_sym].include?(r.name)
               repos[:inactive][r.arch.to_sym].push(r.name)
             end
+
+            next if r.required == "mandatory"
+              repos[:optional] ||= {}
```
skazi0: the indentations look somehow wrong here

jsuchome: True

crowbar_framework/app/models/api/node.rb
```diff
@@ -257,9 +257,27 @@ def repocheck(options = {})
...
                 end
-                available, repolist = ::Crowbar::Repository.provided_and_enabled_with_repolist(
+                _, repolist = ::Crowbar::Repository.provided_and_enabled_with_repolist(
                   feature, platform, architecture
                 )
+                # Filter the result and remove the repositories that are missing but not mandatory
+                repolist.each do |category, repomap|
+                  if [:missing, :inactive].include? category
+                    repomap.each do |arch, repos|
+                      repos.reject! { |repo| repolist[:optional][arch].include?(repo) }
```
skazi0: 1. might this be `repos -= repolist[:optional][arch]` ?
2. looks like `repolist[:optional]` might not be defined if all repos under given feature are "mandatory"

skazi0: for #2 we need something like `next unless repolist.include? :optional` before the reject! line

jsuchome: Ah true. Or I can make the lib function always return at least empty list for optional

jsuchome:
@dirkmueller @vuntz Hi, as this is change (about repo priorities) of the behavior that used to be same for a while, could you take a look if there's not a potential to break things?

(I'm trying to limit the change to upgrade cause though, but putting most of the changes into the upgrade code and not the repository lib)

vuntz:
Wouldn't that break badly if the HA repos (which are optional) are missing, while HA itself is used?

crowbar_framework/lib/crowbar/repository.rb
```diff
@@ -256,6 +259,11 @@ def provided_with_enabled(feature,
             unless repos[:inactive][r.arch.to_sym].include?(r.name)
               repos[:inactive][r.arch.to_sym].push(r.name)
             end
+
+            next if r.required == "mandatory"
+            unless repos[:optional][r.arch.to_sym].include?(r.name)
+              repos[:optional][r.arch.to_sym].push(r.name)
+            end
```
vuntz: This block is after `next if r.active?`, which is strange. It looks like we only put inactive optional repos in that variable.

jsuchome:
> Wouldn't that break badly if the HA repos (which are optional) are missing, while HA itself is used?

Good question. I _think_ we have a specific check for HA repos but I may be wrong, I'll need a closer look

crowbar_framework/lib/crowbar/repository.rb
jsuchome: Yes, that's true. If they are active we do not need any more info, and if they are inactive then there's a problem unless we know they are optional.
(It just seemed nicer to me to test for "optional" then "not mandatory", but in the end it does not matter)

jsuchome:
So you are right. HA is marked as optional so when we check for HA and it is missing, repocheck would not fail with this patch :-(

PR #1946, opened 2019-10-24
[5.0] upgrade: Add a precheck for XEN compute nodes presence (SOC-10495)

jsuchome:
XEN compute nodes are no longer supported in SOC9 so they need
to be removed before the upgrade from SOC8 to SOC9 starts.


jsuchome:
The experimental commit for ignoring CVE does not have the right title... I'm still not sure if we are allowed to do it

skazi0:
@jsuchome remember to add crowbar-ui translation for the new precheck

PR #1948, opened 2019-10-28
upgrade: cold start nova before live migration (SOC-10761)

jgrassler:
In order to prevent any and all RPC API incompatibilities
through lingering nova-compute processes, this commit stops
all Nova services (both on controllers and compute nodes)
after the early nova-compute update performed at the end of
the controller upgrade. Subsequently, these services are
restarted in a staggered manner to keep them from DDoSing
RabbitMQ on larger clouds. This ensures all Nova services use
the new Nova RPC API (5.0), thus allowing the live migrations
that are part of non-disruptively upgrading the compute nodes,
to take place.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -81,8 +81,9 @@ end
 
 # Find out now if we have HA setup and pass that info to the script
 use_ha = roles.include? "pacemaker-cluster-member"
-remote_node = roles.include? "pacemaker-remote"
+is_remote_node = roles.include? "pacemaker-remote"
 is_cluster_founder = use_ha && node["pacemaker"]["founder"] == node[:fqdn]
+remote_nodes = CrowbarPacemakerHelper.remote_nodes(node)
 
 template "/usr/sbin/crowbar-shutdown-services-before-upgrade.sh" do
   source "crowbar-shutdown-services-before-upgrade.sh.erb"
@@ -91,7 +92,7 @@ template "/usr/sbin/crowbar-shutdown-services-before-upgrade.sh" do
   group "root"
   action :create
   variables(
-    use_ha: use_ha || remote_node,
+    use_ha: use_ha || is_remote_node,
     cluster_founder: is_cluster_founder,
     nova_controller: roles.include?("nova-controller"),
     monasca_server: roles.include?("monasca-server"),
@@ -177,7 +178,7 @@ template "/usr/sbin/crowbar-pre-upgrade.sh" do
   variables(
     use_ha: use_ha,
     compute_node: compute_node,
-    remote_node: remote_node,
+    is_remote_node: is_remote_node,
     swift_storage: swift_storage,
     cinder_volume: cinder_volume,
     neutron_agent: neutron_agent,
@@ -297,7 +298,24 @@ template "/usr/sbin/crowbar-reload-nova-after-upgrade.sh" do
   owner "root"
   group "root"
   variables(
-    nova_controller: roles.include?("nova-controller")
+    nova_controller: roles.include?("nova-controller"),
+    remotes_present: remote_nodes.any?,
+    is_remote_node: is_remote_node,
+    is_cluster_founder: is_cluster_founder
+  )
+  only_if { nova_node }
+end
+
+template "/usr/sbin/crowbar-stop-nova-services.sh" do
+  source "crowbar-stop-nova-services.sh.erb"
+  mode "0775"
+  owner "root"
+  group "root"
+  variables(
+    nova_controller: roles.include?("nova-controller"),
+    remotes_present: remote_nodes.any?,
+    is_remote_node: is_remote_node,
+    is_cluster_founder: is_cluster_founder
   )
   only_if { nova_node }
 end
--- a/chef/cookbooks/crowbar/templates/default/crowbar-reload-nova-after-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-reload-nova-after-upgrade.sh.erb
@@ -27,21 +27,43 @@ if [[ -f $UPGRADEDIR/crowbar-reload-nova-after-upgrade-ok ]] ; then
     exit 0
 fi
 
+# Check the python-nova package to make sure it is at the latest version with
+# no more recent versions in any zypper repositories (this package can be used
+# as a proxy for making sure all Nova service packages are up to date since
+# they are version locked to python-nova):
+
+if zypper lu -a | grep -w python-nova; then
+  echo "There are available update candidates for python-nova. This means that "
+  echo "the nova service packages on this node are not at the latest possible "
+  echo "version. Please update them and restart the upgrade."
+
+  exit 1
+fi
+
 <% if @nova_controller %>
 for service in conductor scheduler novncproxy serialproxy api; do
     fullname="openstack-nova-$service"
-    if systemctl --quiet is-active $fullname 2>/dev/null ; then
-        systemctl restart $fullname
+    if systemctl cat $fullname &>/dev/null; then
+      systemctl start $fullname
     fi
 done
-# Remove and unmanage openstack-nova-consoleauth
-systemctl disable openstack-nova-consoleauth
-systemctl stop openstack-nova-consoleauth
-systemctl kill openstack-nova-consoleauth
-rpm -e openstack-nova-consoleauth
+
+# Apache was also stopped by crowbar-stop-nova-services.sh
+systemctl start apache2
+
+  <% if @remotes_present and @is_cluster_founder %>
+# Start nova-compute on remote nodes:
+for remote in $(crm resource list | awk '$1 ~ /^remote-/  {print $1}'); do
+  crm resource start $remote
+done
+  <% end %>
 
 <% else %>
-systemctl restart openstack-nova-compute
+  # Only start on non-remote compute nodes - service on remote nodes will be started by
+  # the controller side code above.
+  <% unless @is_remote_node %>
+systemctl start openstack-nova-compute
+  <% end %>
 <% end %>
 
 touch $UPGRADEDIR/crowbar-reload-nova-after-upgrade-ok
--- a/chef/cookbooks/crowbar/templates/default/crowbar-stop-nova-services.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-stop-nova-services.sh.erb
@@ -1,0 +1,88 @@
+#!/bin/bash
+#
+# After the upgrade of all compute services on all nodes is finished, it's
+# necessary to stop all nova services so that they start using the latest RPC
+# API version. All services must be stopped without exception: if any service
+# still uses the old RPC API version, it may cause other services to pick that
+# RPC API version as well.
+
+LOGFILE=/var/log/crowbar/node-upgrade.log
+UPGRADEDIR=/var/lib/crowbar/upgrade
+SCRIPTNAME=stop-nova-services
+mkdir -p "`dirname "$LOGFILE"`"
+exec >>"$LOGFILE" 2>&1
+
+log()
+{
+    set +x
+    echo "[$(date --iso-8601=ns)] [$$] $@"
+    set -x
+}
+
+log "Executing $BASH_SOURCE"
+
+set -x
+
+if [[ -f $UPGRADEDIR/crowbar-${SCRIPTNAME}-ok ]] ; then
+    log "Stop nova script was already successfully executed"
+    exit 0
+fi
+
+<% if @nova_controller %>
+for service in conductor scheduler novncproxy serialproxy api; do
+    fullname="openstack-nova-$service"
+    if systemctl --quiet is-active $fullname 2>/dev/null ; then
+        systemctl stop $fullname
+    fi
+done
+
+# This is needed to prevent the Nova Placement API from causing RPC version
+# trouble.
+if systemctl --quiet is-active apache2 2>/dev/null ; then
+    systemctl stop apache2
+fi
+
+  <% if @remotes_present and @is_cluster_founder %>
+# Stop nova-compute on remote nodes:
+for remote in $(crm resource list | awk '$1 ~ /^remote-/  {print $1}'); do
+  crm resource stop $remote
+done
+  <% end %>
+
+# Remove and unmanage openstack-nova-consoleauth
+systemctl disable openstack-nova-consoleauth
+systemctl stop openstack-nova-consoleauth
+systemctl kill openstack-nova-consoleauth
+rpm -e openstack-nova-consoleauth
+
+<% else %>
+  <% unless @is_remote_node %>
+# Only stop on non-remote compute nodes - service on remote nodes will be stopped by
+# the controller side code above.
+systemctl stop openstack-nova-compute
+  <% end %>
+<% end %>
+
+# This check can run on both remote and non-remote nodes: the `crm resource
+# stop` for the remote resources will have been run before this script runs on
+# the compute node it was intended for. This check needs to fail on the compute
+# node itself if it is going to fail. That way it will pinpoint the problematic
+# node.
+
+for wait in 10 20 40 0; do
+  if ps -U nova u | grep -e apache -e python; then
+    echo "Warning: there are still Nova processes running. These processes must be stopped before continuing the upgrade."
+    if [ $wait -eq 0 ]; then
+      echo "Error: Waited for 70s and Nova processes are still active."
+      echo "Please stop these processes manually and restart the upgrade's nodes step."
+      exit 1
+    fi
+    echo "Sleeping for $wait seconds until rechecking."
+    sleep $wait
+  else
+    break
+  fi
+done
+
+touch $UPGRADEDIR/crowbar-${SCRIPTNAME}-ok
+log "$BASH_SOURCE is finished."
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -817,6 +817,8 @@ module Api
           upgrade_controller_clusters
           upgrade_non_compute_nodes
           prepare_all_compute_nodes
+          # Stop nova services on all nodes to prevent any and all RPC API trouble
+          stop_nova_services
           reload_nova_services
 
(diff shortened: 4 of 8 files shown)


chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -297,7 +298,24 @@
   owner "root"
   group "root"
   variables(
-    nova_controller: roles.include?("nova-controller")
+    nova_controller: roles.include?("nova-controller"),
+    remotes_present: !remote_nodes.empty?,
+    remote_node: remote_node,
```
jsuchome: Just a nit ... for the code clarity it might be better to rename `remote_node` to `is_remote_node`...

chef/cookbooks/crowbar/templates/default/crowbar-reload-nova-after-upgrade.sh.erb
```diff
@@ -27,21 +27,41 @@ if [[ -f $UPGRADEDIR/crowbar-reload-nova-after-upgrade-ok ]] ; then
     exit 0
 fi
 
+# Check the python-nova package to make sure it is at the latest version with
+# no more recent versions in any zypper repositories (this package can be used
+# as a proxy for making sure all Nova service packages are up to date since
+# they are version locked to python-nova):
```
jsuchome: How could this happen? Aren't we responsible for updating the package? Repositories should be set up and the package was updated in different step, right?

chef/cookbooks/crowbar/templates/default/crowbar-reload-nova-after-upgrade.sh.erb
```diff
@@ -27,21 +27,41 @@ if [[ -f $UPGRADEDIR/crowbar-reload-nova-after-upgrade-ok ]] ; then
...
-    if systemctl --quiet is-active $fullname 2>/dev/null ; then
-        systemctl restart $fullname
-    fi
+    systemctl start $fullname
+done
+
+# Apache was also stopped by crowbar-stop-nova-services.sh
+systemctl start apache2
+
+  <% if @remotes_present and @is_cluster_founder %>
```
jsuchome: Indentation?

jsuchome: Ah, this is part of `if @nova_controller` block...

chef/cookbooks/crowbar/templates/default/crowbar-reload-nova-after-upgrade.sh.erb
```diff
@@ -27,21 +27,41 @@ if [[ -f $UPGRADEDIR/crowbar-reload-nova-after-upgrade-ok ]] ; then
...
+# Start nova-compute on remote nodes:
+for remote in $(crm resource list | awk '$1 ~ /^remote-/  {print $1}'); do
```
jsuchome: Why not use `remote_nodes` varaible?

chef/cookbooks/crowbar/templates/default/crowbar-reload-nova-after-upgrade.sh.erb
```diff
@@ -27,21 +27,41 @@ if [[ -f $UPGRADEDIR/crowbar-reload-nova-after-upgrade-ok ]] ; then
...
+  crm resource start $remote
 done
-# Remove and unmanage openstack-nova-consoleauth
-systemctl disable openstack-nova-consoleauth
-systemctl stop openstack-nova-consoleauth
-systemctl kill openstack-nova-consoleauth
-rpm -e openstack-nova-consoleauth
```
jsuchome: I assume this can be deleted as the service was already removed in SOC8, right?

jsuchome: Ah, no, you've moved it to the other script.

chef/cookbooks/crowbar/templates/default/crowbar-stop-nova-services.sh.erb
```diff
@@ -0,0 +1,77 @@
...
+systemctl stop openstack-nova-compute
+  <% end %>
+<% end %>
+
+# This check can run on both remote and non-remote nodes: the `crm resource
+# stop` for the remote resources will have been run before this script runs on
+# the compute node it was intended for. This check needs to fail on the compute
+# node itself if it is going to fail. That way it will pinpoint the problematic
+# node.
+if ps auxw | grep ^nova | grep python; then
```
jsuchome: Hm, are we sure systemctl stop already killed every such process? Shouldn't there be some short wait loop?

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -297,7 +298,24 @@
   owner "root"
   group "root"
   variables(
-    nova_controller: roles.include?("nova-controller")
+    nova_controller: roles.include?("nova-controller"),
+    remotes_present: !remote_nodes.empty?,
```
jsuchome: What about `remote_nodes.any?`

chef/cookbooks/crowbar/templates/default/crowbar-stop-nova-services.sh.erb
jsuchome: Yep, my test just failed with `'Error: there are still nova processes running. Stop these processes before continuing the upgrade.'`
I logged in, checked manually and no nova process is there. I guess we need to give it some time..

chef/cookbooks/crowbar/templates/default/crowbar-reload-nova-after-upgrade.sh.erb
jgrassler: We are, but the `zypper --non-interactive up openstack-nova-compute` we run for the early `nova-compute` upgrade we require for RPC API compatibility is not trustworthy. For instance, it exited `0` when we'd forgotten to set `solver.allowVendorChange` earlier, thus causing the installation to fail with a prompt asking about vendor change. I don't know how `zypper` behaves with other problems, such as a NIC flap preventing access to the package repository, but I'd rather play it safe and check up on it.

jsuchome: OK, then. It does not hurt to have another check I guess

chef/cookbooks/crowbar/templates/default/crowbar-reload-nova-after-upgrade.sh.erb
jgrassler: Because I'd be building the resource name from the host name in there and a prepended `remote-`. I'm not comfortable assuming that this is always the correct resource name. For instance, what happens if there's a FQDN in the list and a short host name in pacemaker (or vice versa)? I can rely on whatever matches that regex in the pacemaker resource list being a remote resource since pacemaker is the canonical source of resource names.

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
jgrassler: Done.

chef/cookbooks/crowbar/templates/default/crowbar-stop-nova-services.sh.erb
jgrassler: Ok, I added a retry loop now.

chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
jgrassler: Done.

chef/cookbooks/crowbar/templates/default/crowbar-stop-nova-services.sh.erb
```diff
@@ -0,0 +1,88 @@
...
+<% end %>
+
+# This check can run on both remote and non-remote nodes: the `crm resource
+# stop` for the remote resources will have been run before this script runs on
+# the compute node it was intended for. This check needs to fail on the compute
+# node itself if it is going to fail. That way it will pinpoint the problematic
+# node.
+
+for wait in 10 20 40 0; do
+  if ps -U nova u | grep nova-compute; then
```
jsuchome: It's not just for compute services, right?

chef/cookbooks/crowbar/templates/default/crowbar-reload-nova-after-upgrade.sh.erb
jsuchome: OK

chef/cookbooks/crowbar/templates/default/crowbar-stop-nova-services.sh.erb
```diff
@@ -0,0 +1,88 @@
...
+<% end %>
+
+# This check can run on both remote and non-remote nodes: the `crm resource
+# stop` for the remote resources will have been run before this script runs on
+# the compute node it was intended for. This check needs to fail on the compute
+# node itself if it is going to fail. That way it will pinpoint the problematic
+# node.
+
+for wait in 10 20 40 0; do
+  if [ $(ps -U nova u | wc -l) -gt 1 ]; then
```
jsuchome: This is actually still wrong:

```
# ps -U nova u 
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
nova     28717  0.0  0.1  36912  4656 ?        Ss   14:24   0:00 /usr/lib/systemd/systemd --user
nova     28720  0.0  0.0 166136  1980 ?        S    14:24   0:00 (sd-pam)
```

jgrassler: Sigh. This is a veritable Rubik's Cube :-)

Modified again...and that looks like it does the trick on both controller...

```
root@d00-25-90-e5-04-54:~ # ps -U nova u | grep -e apache -e python
nova      3811  0.0  0.5 307760 95552 ?        Ss   Oct30   0:17 /usr/bin/python /usr/bin/nova-novncproxy
nova      8240  0.0  0.6 593152 111972 ?       Sl   Oct29   0:28 (wsgi:nova-placement-ap -DSYSCONFIG -C PidFile /var/run/httpd.pid -C Include /etc/apache2/sysconfig.d//loadmodule.conf -C Include /etc/apache2/sysconfig.d//global.conf -f /etc/apache2/httpd.conf -c Include /etc/apache2/sysconfig.d//include.conf -DSYSTEMD -DFOREGROUND -k start
nova      8241  0.0  0.6 593144 110904 ?       Sl   Oct29   0:27 (wsgi:nova-placement-ap -DSYSCONFIG -C PidFile /var/run/httpd.pid -C Include /etc/apache2/sysconfig.d//loadmodule.conf -C Include /etc/apache2/sysconfig.d//global.conf -f /etc/apache2/httpd.conf -c Include /etc/apache2/sysconfig.d//include.conf -DSYSTEMD -DFOREGROUND -k start
nova      8242  0.0  0.6 658704 111116 ?       Sl   Oct29   0:28 (wsgi:nova-placement-ap -DSYSCONFIG -C PidFile /var/run/httpd.pid -C Include /etc/apache2/sysconfig.d//loadmodule.conf -C Include /etc/apache2/sysconfig.d//global.conf -f /etc/apache2/httpd.conf -c Include /etc/apache2/sysconfig.d//include.conf -DSYSTEMD -DFOREGROUND -k start
nova      8291  1.1  0.6 274660 100344 ?       Ss   Oct29  27:12 /usr/bin/python /usr/bin/nova-conductor
nova      8517  0.0  0.6 288568 113976 ?       Ss   Oct29   2:02 /usr/bin/python /usr/bin/nova-scheduler
nova      8624  1.1  0.7 311244 130692 ?       Ss   Oct29  27:26 /usr/bin/python /usr/bin/nova-api
nova      8807  0.0  0.6 284080 109700 ?       Ss   Oct29   1:41 /usr/bin/python /usr/bin/nova-consoleauth
nova     24523  0.0  0.8 323888 137768 ?       S    Oct30   0:00 /usr/bin/python /usr/bin/nova-api
nova     24527  0.0  0.8 322960 136960 ?       S    Oct30   0:00 /usr/bin/python /usr/bin/nova-api
nova     24528  0.0  0.8 323924 137784 ?       S    Oct30   0:01 /usr/bin/python /usr/bin/nova-api
nova     24529  0.0  0.8 324360 137868 ?       S    Oct30   0:00 /usr/bin/python /usr/bin/nova-api
nova     24530  0.0  0.8 318868 131488 ?       S    Oct30   0:00 /usr/bin/python /usr/bin/nova-api
nova     24531  0.0  0.8 318868 131488 ?       S    Oct30   0:00 /usr/bin/python /usr/bin/nova-api
nova     24532  0.0  0.8 318868 131488 ?       S    Oct30   0:00 /usr/bin/python /usr/bin/nova-api
nova     24533  0.0  0.8 318868 131488 ?       S    Oct30   0:00 /usr/bin/python /usr/bin/nova-api
nova     24577  0.2  0.7 297400 117364 ?       S    Oct30   2:07 /usr/bin/python /usr/bin/nova-conductor
nova     24578  0.2  0.7 297476 117592 ?       S    Oct30   2:08 /usr/bin/python /usr/bin/nova-conductor
nova     24585  0.2  0.7 296896 116796 ?       S    Oct30   2:05 /usr/bin/python /usr/bin/nova-conductor
nova     24586  0.2  0.7 297032 116952 ?       S    Oct30   2:09 /usr/bin/python /usr/bin/nova-conductor
root@d00-25-90-e5-04-54:~ #
```

...and compute node:

```
root@d00-25-90-e5-0c-fc:~ # ps -U nova u | grep -e apache -e python
nova     13415  0.2  0.7 1961284 130156 ?      Ssl  Oct30   2:30 /usr/bin/python /usr/bin/nova-compute
nova     13667  0.2  0.7 1961524 129976 ?      Ssl  Oct30   2:28 /usr/bin/python /usr/bin/nova-compute
root@d00-25-90-e5-0c-fc:~ #
```

chef/cookbooks/crowbar/templates/default/crowbar-stop-nova-services.sh.erb
```diff
@@ -0,0 +1,88 @@
...
+
+# Remove and unmanage openstack-nova-consoleauth
+systemctl disable openstack-nova-consoleauth
+systemctl stop openstack-nova-consoleauth
+systemctl kill openstack-nova-consoleauth
+rpm -e openstack-nova-consoleauth
+
+<% else %>
+  # Only stop on non-remote compute nodes - service on remote nodes will be stopped by
+  # the controller side code above.
```
jsuchome: Could you move the comment into the `unless` block?

This way it is rendered with the comment only on remote nodes and is not clear what does the comment refer to. So it's better to not show it in the script (while it is perfectly good in template)

jsuchome:
will need a rebase after https://github.com/crowbar/crowbar-core/pull/1950 is merged

jsuchome:
Please rebase

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -817,6 +817,8 @@ def do_controllers_substep(substep)
           upgrade_controller_clusters
           upgrade_non_compute_nodes
           prepare_all_compute_nodes
+          # Stop nova services on all nodes to prevent any and all RPC API trouble
+          stop_nova_services
           reload_nova_services
```
skazi0: this `reload_nova_services` call doesn't make much sense here now, does it?

jsuchome: It starts the services. So it still makes sense, though the naming might be bit confusing

jgrassler: I didn't change the name on purpose - didn't want to risk breaking something due to a dangling reference to the old name somewhere...

jgrassler:
> Please rebase

Done.

crowbar_framework/app/models/api/upgrade.rb
skazi0: ah, so the idea here is that we stop everything and then start it all again without anything in between, correct? if so then I missed the point before.

jgrassler: Yes.

jgrassler:
This needed another rebase now due to landing (the merge conflict only affected the cleanup line for the "run-nova-online-migrations" script).

chef/cookbooks/crowbar/templates/default/crowbar-reload-nova-after-upgrade.sh.erb
```diff
@@ -27,21 +27,41 @@ if [[ -f $UPGRADEDIR/crowbar-reload-nova-after-upgrade-ok ]] ; then
...
+  exit 1
+fi
+
 <% if @nova_controller %>
 for service in conductor scheduler novncproxy serialproxy api; do
     fullname="openstack-nova-$service"
-    if systemctl --quiet is-active $fullname 2>/dev/null ; then
-        systemctl restart $fullname
-    fi
+    systemctl start $fullname
```
skazi0: why did you remove the `is-active` check? do we always have all of these services installed? if something is missing then you should not try to start it.

skazi0: `if systemctl cat $fullname &>/dev/null; then ...` seems to do the trick.

crowbar_framework/lib/crowbar/upgrade_timeouts.rb
```diff
@@ -38,6 +38,7 @@ def values
         delete_nova_services: @timeouts_config[:delete_nova_services] || 300,
         wait_until_compute_started: @timeouts_config[:wait_until_compute_started] || 60,
         reload_nova_services: @timeouts_config[:reload_nova_services] || 120,
+        stop_nova_services: @timeouts_config[:stop_nova_services] || 120,
```
skazi0: Add this to `configs/upgrade_timeouts.yml` as well. This file is packaged as an example for the users.

chef/cookbooks/crowbar/templates/default/crowbar-reload-nova-after-upgrade.sh.erb
jgrassler: Done.

crowbar_framework/lib/crowbar/upgrade_timeouts.rb
jgrassler: Done.

PR #1949, opened 2019-10-30
[4.0] Allow pacemaker remotes for upgrade (SOC-10133)

jgrassler:
This commit allows clouds with Pacemaker remote nodes to be
upgraded. To this end it changes the upgrade precheck to
include remote nodes in the list of valid elements.

Please do not merge this, yet. We can only merge it once all other changes required for SOC-10133 have landed and we have verified the upgrade works.

### DIFF
--- a/crowbar_framework/app/models/api/crowbar.rb
+++ b/crowbar_framework/app/models/api/crowbar.rb
@@ -100,7 +100,8 @@ module Api
         # Not using Node.all here as we only need node names
         nodes = ChefObject.fetch_nodes_from_cdb.first.map(&:name)
         clusters = ServiceObject.available_clusters.keys
-        valid_elements = nodes + clusters
+        remotes = ServiceObject.available_remotes.keys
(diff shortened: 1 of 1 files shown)


jsuchome:
Actually, this does not need a forward port exactly, as we forgot to port https://github.com/crowbar/crowbar-core/pull/1684 into 5.0.
But I think we should port the old PR together with the fix that is present here.

jsuchome:
Needs rebase after merging https://github.com/crowbar/crowbar-core/pull/1953

jsuchome:
... and another rebase after https://github.com/crowbar/crowbar-core/pull/1972

jgrassler:
Needs https://github.com/crowbar/crowbar-core/pull/1972 for travis to pass.

jgrassler:
Rebase done. I removed the needs-forward-port label as well since we've got no equivalent of this precheck in Cloud 8 anyway.

PR #1959, opened 2019-11-08
Allow enabling bpdu-forwarding on OVS bridges (SOC-9172)

zzaimeche:
This change allows users to enable BPDU forwarding on OVS bridges
that are used for Neutron provider networks (br-fixed, br-floating).
Enabling it will allow the physical network to detect and stop loops
if these bridges are accidentally misconfigured. It is turned off
by default.

More context in:

  https://bugzilla.suse.com/show_bug.cgi?id=1133804

(backported from: 273312c0c8be73d050bec2ed912958c74c7652b7 )


JanZerebecki:
This branch should now only receive important fixes. This seems not to be one. Please reopen if you think otherwise.

PR #1961, opened 2019-11-14
[5.0] upgrade: Make the time before next upgrade configurable

jsuchome:
crowbar upgrade tool returns the information about the finished
upgrade for some time even after everything has been done.
For the cases where user would like to continue with the upgrade
to the next product version ASAP, we can make the required time
configurable.


skazi0:
...and add some ref to the commit title

jsuchome:
Yeah, I need to create Jira card...

skazi0:
tested and working :o)

PR #1962, opened 2019-11-15
[upgrade] Remove existing upgrade directories from nodes (SOC-10956)

jsuchome:
If such dir exists, it would contain indications of upgrade scripts' runs
from previous upgrade.
If they'd stay, script wouldn't be executed in current upgrade.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -92,6 +92,13 @@ when "crowbar_upgrade"
     end
   end
 
+  # Remove existing upgrade directory. If it exists, it would contain
+  # indications of upgrade scripts runs from previous upgrade.
+  # If they'd stay, script wouldn't be executed in current upgrade
+  directory "/var/lib/crowbar/upgrade" do
+    action :delete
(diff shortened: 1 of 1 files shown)


chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -92,6 +92,13 @@
     end
   end
 
+  # Remove existing upgrade directory. If it exists, it would contain
+  # indications of upgrade scripts runs from previous upgrade.
+  # If they'd stay, script wouldn't be executed in current upgrade
+  directory "/var/lib/crowbar/upgrade" do
```
skazi0: where is this directory created?

jsuchome: It's created directly from the upgrade bash scripts, see e.g. https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/crowbar/templates/default/crowbar-chef-upgraded.sh.erb#L21

jsuchome:
This actually needs to go to 5.0 branch

jsuchome:
Not needed for Cloud9

PR #1967, opened 2019-11-18
[5.0] upgrade: Fixes for prepare step

skazi0:
port of #1549


jsuchome:
Actually, can you drop the drbd parts? There should be no trace of them in SOC8 so no need to take care of it during upgrade.

(But we can also do it in another PR)

PR #1970, opened 2019-11-20
[5.0] Disable automatic repo services and installation repo (bsc#1152007)

skazi0:
When nodes are installed from SMT, they can have some automatic repo
services configured. These conflict with crowbar repo management and
should be disabled during upgrade (or earlier).

In addition this change disables configuration of install media repo
as it is not needed for upgraded nodes (just for new ones).

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-prepare-repositories.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-prepare-repositories.sh.erb
@@ -25,6 +25,9 @@ if [[ -f $UPGRADEDIR/crowbar-prepare-repositories-ok ]] ; then
     exit 0
 fi
 
+log "Disabling Repository Index Services (RIS)..."
+zypper services --show-enabled-only | tr -d ' ' | awk -F "|" '{ if ($7=="ris") print $1}' | xargs -i zypper modifyservice --disable {}
+
 
 log "Removing old repositories..."
 rm -f /etc/zypp/repos.d/*.repo
--- a/chef/cookbooks/crowbar/templates/default/crowbar-prepare-repositories.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-prepare-repositories.sh.erb
@@ -40,7 +40,6 @@ zypper --non-interactive addrepo -f <%= attrs[:url] %> <%= name %>
 zypper --non-interactive addrepo <%= attrs[:url] %> <%= name %>
     <% end -%>
 <% end %>
-zypper --non-interactive addrepo <%= @new_base_repo %> <%= @new_alias %>
 
 touch $UPGRADEDIR/crowbar-prepare-repositories-ok
 log "$BASH_SOURCE is finished."
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -30,23 +30,9 @@ new_repos = Provisioner::Repositories.get_repos(
   target_platform, target_platform_version, arch
 )
 
-# Find out the location of the base system repository
 provisioner_config = Barclamp::Config.load("core", "provisioner")
 
-web_path = "#{provisioner_config['root_url']}/#{node[:platform]}-#{node[:platform_version]}/#{arch}"
-old_install_url = "#{web_path}/install"
-
 web_path = "#{provisioner_config['root_url']}/#{node[:target_platform]}/#{arch}"
-new_install_url = "#{web_path}/install"
-
-# try to create an alias for new base repo from the original base repo
-repo_alias = "SLES12-SP2-12.2-0"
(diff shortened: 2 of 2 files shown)


jsuchome:
Installation repo is not SMT specific, could you make it a separate commit?

jsuchome:
BTW, how does `zypper services --show-enabled-only` look like in your setup?

skazi0:
@jsuchome currently like this:
```
# | Alias                                  | Name                                   | Enabled | GPG Check | Refresh | Type  
--+----------------------------------------+----------------------------------------+---------+-----------+---------+-------
2 | PTF                                    | PTF                                    | Yes     | (r ) Yes  | No      | rpm-md
3 | SLE12-SP3-HA-Pool                      | SLE12-SP3-HA-Pool                      | Yes     | (r ) Yes  | No      | rpm-md
4 | SLE12-SP3-HA-Updates                   | SLE12-SP3-HA-Updates                   | Yes     | (r ) Yes  | Yes     | rpm-md
5 | SLES12-SP3-Pool                        | SLES12-SP3-Pool                        | Yes     | (r ) Yes  | No      | rpm-md
6 | SLES12-SP3-Updates                     | SLES12-SP3-Updates                     | Yes     | (r ) Yes  | Yes     | rpm-md
7 | SUSE-OpenStack-Cloud-Crowbar-8-Pool    | SUSE-OpenStack-Cloud-Crowbar-8-Pool    | Yes     | (r ) Yes  | No      | rpm-md
8 | SUSE-OpenStack-Cloud-Crowbar-8-Updates | SUSE-OpenStack-Cloud-Crowbar-8-Updates | Yes     | (r ) Yes  | Yes     | rpm-md

**Comment by jsuchome** [comment]
I mean with SMT (where's the 'ris' you are grepping for?). Also, can't the grep show some false positives (could ' ris ' be part of an alias or full name)?

**Comment by skazi0** [comment]
Now I have this (on a different node):
```
# | Alias                         | Name                          | Enabled | GPG Check | Refresh | Type
--+-------------------------------+-------------------------------+---------+-----------+---------+-----
1 | SMT-http_smt-scc_nue_suse_com | SMT-http_smt-scc_nue_suse_com | Yes     | ----      | Yes     | ris 
```
I don't think spaces are allowed in aliases or names, are they?

**Comment by skazi0** [comment]
btw, I changed `grep` to `egrep` because in polish:
```
# | Alias                         | Nazwa                         | Włączono | Sprawdzaj GPG | Odśwież | Typ
--+-------------------------------+-------------------------------+----------+---------------+---------+----
1 | SMT-http_smt-scc_nue_suse_com | SMT-http_smt-scc_nue_suse_com | Tak      | ----          | Tak     | ris

chef/cookbooks/crowbar/templates/default/crowbar-prepare-repositories.sh.erb
```diff
@@ -25,6 +25,8 @@ if [[ -f $UPGRADEDIR/crowbar-prepare-repositories-ok ]] ; then
     exit 0
 fi
 
+log "Disabling RIS services..."
+zypper services --show-enabled-only | egrep ' ris( |$)' | cut -d' ' -f1 | xargs -i zypper modifyservice --disable {}
```
jsuchome: I'm still not sure about the grepping. You're probably right. We can be more precise with checking the value of the type column I guess.

(Not supported by zypper, bash magic required)

skazi0: I was surprised that zypper doesn't have `--type` option to apply changes based on service type while it has `--media-type` option which filters by url scheme. I don't have any better idea for filtering out those services. We could try `--xmlout` but without proper xpath utils it would probably be even more hcky than regular table grep.

skazi0: oh, wait, we have `xml_grep`... let me try that.

skazi0:
I split the PR into two commits and changed parsing of `zypper services` output.

chef/cookbooks/crowbar/templates/default/crowbar-prepare-repositories.sh.erb
```diff
@@ -37,7 +40,7 @@ zypper --non-interactive addrepo -f <%= attrs[:url] %> <%= name %>
 zypper --non-interactive addrepo <%= attrs[:url] %> <%= name %>
     <% end -%>
 <% end %>
-zypper --non-interactive addrepo <%= @new_base_repo %> <%= @new_alias %>
+#zypper --non-interactive addrepo <%= @new_base_repo %> <%= @new_alias %>
```
jsuchome: Could you just remove it? And also creation of `@new_base_repo` and `@new_alias` in the recipe..

skazi0: done but not tested yet

skazi0:
I tested the master version of this patch and it seems to work fine. I will include this version in my next full 7-8-9 run.

PR #1973, opened 2019-11-26
[5.0] upgrade: Remove DRBD specific code from the upgrade

jsuchome:
DRBD support was dropped in SOC8, so upgrade from this version
does not need to handle it.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
+++ b/chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
@@ -60,11 +60,6 @@ when "crowbar_upgrade"
 
   ha = node["run_list_map"].key? "pacemaker-cluster-member"
 
-  service "drbd" do
-    action :disable
-    only_if { ha && node["drbd"] && node["drbd"]["rsc"] && node["drbd"]["rsc"].any? }
-  end
-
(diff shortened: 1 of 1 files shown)


chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -158,11 +158,6 @@
 
 swift_storage = roles.include? "swift-storage"
 
-rabbitmq_server = search(:node, "run_list_map:rabbitmq-server").first
-# original DRBD setup is switched to cluster one in prepare_nodes_for_os_upgrade
-rabbitmq_cluster = rabbitmq_server[:rabbitmq][:ha][:enabled] &&
```
skazi0: this variable is still passed to the template belo

jsuchome: True.

jsuchome:
Actually the method `prepare_nodes_for_os_upgrade` is executed from `services` step, which is already using SOC9 code. So that part of change belongs to master branch

jsuchome:
See https://github.com/crowbar/crowbar-core/pull/1974 for the next part

PR #1976, opened 2019-12-04
Make sure the crowbar migrations are OK

jsuchome:
Migrations are executed at the end of admin upgrade step and if there was a failure,
there's a big chance services step would fail
(because correct chef templates might not be rendered)

So do the extra check right at the start of services step. Not during
admin upgrade step, beacuse that one is not repeatable after the packages installations.

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -434,11 +434,38 @@ module Api
         false
       end
 
+      # Make sure the crowbar migrations are OK
+      #
+      # They were executed at the end of admin upgrade step and if there was a failure,
+      # there's a big chance services step would fail (because chef templates might not be rendered)
+      # We have to do the check here at the start of 'services' step and not during admin upgrade,
+      # because admin upgrade step is not repeatable.
+      def check_schema_migrations
+        schema_migrate = run_cmd(
+          "cd /opt/dell/crowbar_framework/; " \
+          "sudo -u crowbar RAILS_ENV=production bin/rake crowbar:schema_migrate_prod"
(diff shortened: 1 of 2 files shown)


skazi0:
So the check for failed migrations is done by trying to (re)run the migrations? Have you checked how it behaves if some migrations failed before?

jsuchome:
@skazi0 I do not have broken migrations to test now, but why would it be a problem?

Repeating same called that has (presumably) failed before should not yield different result.

(I remember doing the manual call when we really hit such problem before)

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -416,10 +416,38 @@ def check_for_running_instances
...
+          "sudo -u crowbar RAILS_ENV=production bin/rake crowbar:schema_migrate_prod"
+        )
+        return true if schema_migrate[:exit_code].zero?
+        msg = "There was an error during crowbar schema migrations."
+        Rails.logger.error msg
+        ::Crowbar::UpgradeStatus.new.end_step(
+          false,
+          services: {
+            data: msg,
+            help: "Check the admin server upgrade log /var/log/crowbar/admin-server-upgrade.log " \
```
jgrassler: Might be a good idea to mention `/var/log/crowbar/component_install.log` as well since that will have the error output from barclamp migrations.

jsuchome: That's a good point. I'll change it, I'd need to rebase the PR anyway...

jsuchome:
Rebased and added note about `/var/log/crowbar/component_install.log`

PR #1977, opened 2019-12-05
upgrade: Do not stop and reload nova services in normal mode (SOC-10995)

jsuchome:
In normal mode, all compute nodes are full upgraded in parallel
and rebooted after that. So we do not need to specifically stop
and reload nova services because that's done after the reboot automatically.

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -1025,7 +1025,9 @@ module Api
       # After the controller upgrade we need to stop Nova services on all nodes
       # to ensure that no old services remain running and cause restarted Nova
       # services to autonegotiate the old RPC API version.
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -1026,6 +1026,8 @@ def finalize_node_upgrade(node)
       # to ensure that no old services remain running and cause restarted Nova
       # services to autonegotiate the old RPC API version.
       def stop_nova_services
+        # Only related to non-disruptive mode where nodes will be fully rebooted
```
skazi0: This comment is not very clear to me... aren't nodes rebooted independent of the upgrade mode?

jsuchome: Oh, I did not read it after myself

PR #1978, opened 2019-12-05
upgrade: Make cluster health check at the start of services step 

jsuchome:
If the cluster is unhealthy, it could lead to errors during upgrade.
At this point all services are still running so it should be possible to
fix the problems.

I'm reusing existing health check functions here we are normally using for prechecks ... but why not?

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -416,10 +416,29 @@ module Api
         )
       end
 
+      # Do another pacemaker cluster health check at the start of services step
+      # If the cluster is unhealthy, it could lead to errors during upgrade.
+      # At this point all services are still running so it should be possible to
+      # fix the problems.
+      def cluster_health_check
+        return true unless Api::Crowbar.addons.include?("ha")
+
+        clusters_health = Api::Pacemaker.health_report
+        return true if clusters_health.empty?
+
+        Rails.logger.error "Pacemaker cluster is not healthy."
+        ::Crowbar::UpgradeStatus.new.end_step(
+          false,
+          clusters_health_report_errors(clusters_health)
+        )
+        false
+      end
+
       #
       # service shutdown
       #
       def services
+        return unless cluster_health_check
         begin
           # prepare the scripts for various actions necessary for the upgrade
           service_object = CrowbarService.new


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -416,10 +416,37 @@ def check_for_running_instances
...
+
+        msg = "Pacemaker cluster is not healthy. Fix the problem before repeating the upgrade step."
+        help_messages = []
+        clusters_health_report_errors(clusters_health).each do |problem, desc|
+          help_messages.push desc[:help]
+        end
+        Rails.logger.error msg
+        ::Crowbar::UpgradeStatus.new.end_step(
+          false,
+          services: {
```
skazi0: I'm not sure this is correct. There could be multiple errors from the cluster health check and we will loose the `data` fields for these while merging the `help` fields together. Can we merge the original structure with `services: {}` error? The clients should be able to handle this correctly.

jsuchome: Hmm... OK, do you find this better? (Ignore the fake data values)
```
| steps.services.status                                     | failed                                                                                                                                                                     |
| steps.services.errors.clusters_health_crm_failures.data   | crm_failures.values                                                                                                                                                        |
| steps.services.errors.clusters_health_crm_failures.help   | Please check that all cluster resources are online or refer to the SUSE High Availability documentation for possible troubleshooting.                                      |
| steps.services.errors.clusters_health_failed_actions.data | failed_actions.values,                                                                                                                                                     |
| steps.services.errors.clusters_health_failed_actions.help | Please clean the resource history and re−check the current state with "crm_resource -C" or refer to the SUSE High Availability documentation for possible troubleshooting. |
| steps.services.errors.clusters_health_unready_nodes.data  | unready_nodes.values                                                                                                                                                       |
| steps.services.errors.clusters_health_unready_nodes.help  | Some pacemaker nodes are in maintenance or standby node. Make sure all nodes are online and ready.  
```

jsuchome: Updated

jsuchome:
services-cluster-health

@skazi0 do we need to update some labels in crowbar-ui?

skazi0:
@jsuchome https://github.com/crowbar/crowbar-ui/pull/193

PR #1979, opened 2019-12-06
crowbar: add crowbar-pacemaker dependency (SOC-10986)

jgrassler:
Without this dependency, barclamp application will fail with a

  uninitialized constant Chef::Recipe::CrowbarPacemakerHelper

for any node that does not have other barclamps with this
dependency declaration applied. In practice, this problem only
becomes apparent for nodes that have no roles from crowbar-openstack
applied since all Openstack barclamps declare a dependency on
crowbar-pacemaker.

### DIFF
--- a/chef/cookbooks/crowbar/metadata.json
+++ b/chef/cookbooks/crowbar/metadata.json
@@ -9,6 +9,7 @@
       "apache2": [],
       "barclamp": [],
       "bluepill": [],
+      "crowbar-pacemaker": [],
       "utils": []
     },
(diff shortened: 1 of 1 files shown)


skazi0:
@jgrassler can you add more specific example of scenario where this would fail? What kind of node are we talking about here (i.e. what roles)?

jgrassler:
> @jgrassler can you add more specific example of scenario where this would fail? What kind of node are we talking about here (i.e. what roles)?

Done.

PR #1980, opened 2019-12-09
Ignore CVE-2019-16770

jgrassler:
The fix for our rubygem-puma package is currently being tested
and should make it into a maintenance update soon. Meanwhile,
we can add this CVE to the ignore list since Travis consumes
puma-2.16.0 from Ruby Gems which will never have the
backported patch we've got in our package.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ matrix:
        - bundle exec rake spec brakeman:run
        # ignore rest-client issues, chef 10 requires that
        - bin/bundle exec bundle-audit update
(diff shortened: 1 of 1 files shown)


jsuchome:
I assume you have to mention bnc# (or Jira ticket) in the commit message

skazi0:
> I assume you have to mention bnc# (or Jira ticket) in the commit message

If you don't have one you can also use "noref"

jgrassler:
I even have a ticket number for this...sorry, forgot about that one.

PR #1985, opened 2019-12-12
upgrade: Make a check for SLES product version (SOC-3089)

jsuchome:
If some packages are incorrectly upgaded during admin server upgrade,
it could blow to our faces later. Better to make some explicit check
if things went well when it's still time and check for the installed
SLES product version.

It is not necessary to check for Cloud product version because another
check for the state of crowbar migrations should cover it.

Do the extra check right at the start of services step. Not during
admin upgrade step, beacuse that one is not repeatable after the reboot.

### DIFF
--- a/crowbar_framework/app/models/api/upgrade.rb
+++ b/crowbar_framework/app/models/api/upgrade.rb
@@ -460,12 +460,39 @@ module Api
         false
       end
 
+      # Check the version of installed SLES to verify admin upgrade went well.
+      # We have to do the check here at the start of 'services' step and not during admin upgrade,
+      # because admin upgrade step is not repeatable.
+      def check_product_version
+        zypper_stream = Hash.from_xml(
+          `sudo /usr/bin/zypper-retry --xmlout products -i`
+        )["stream"]
+        products = zypper_stream["product_list"]["product"]
+        return true if products.any? do |product|
+          product["name"] == "SLES" && product["version"] == "12.4"
+        end
+        Rails.logger.error "Incorrect SLES version present"
+        ::Crowbar::UpgradeStatus.new.end_step(
+          false,
+          services: {
+            data: "It appears that incorrect version of SLES product is installed.",
+            help: "Check the admin server upgrade log /var/log/crowbar/admin-server-upgrade.log " \
+              "for possible hints of what went wrong during admin server upgrade. " \
+              "For fixing the problem, it might be needed to check the state of admin server " \
+              "repositories, upgrade the admin server manually (using 'zypper dup' command) and " \
+              "rebooting it. Once done, repeat this upgrade step."
+          }
+        )
+        false
+      end
+
       #
       # service shutdown
       #
       def services
         return unless cluster_health_check
         return unless check_schema_migrations
+        return unless check_product_version
         begin
           # prepare the scripts for various actions necessary for the upgrade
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -460,12 +460,41 @@ def check_schema_migrations
...
+      # We have to do the check here at the start of 'services' step and not during admin upgrade,
+      # because admin upgrade step is not repeatable.
+      def check_product_version
+        # "zypper products -i" should look like: (shortened from real output)
+        # S  | Repository         | Internal Name   | Name                                | Version
+        # ---+--------------------+-----------------+-------------------------------------+--------
+        # i  | Cloud              | suse-openstack..| SUSE OpenStack Cloud Crowbar 9      | 9-0
+        # i+ | SLES12-SP4-Updates | SLES            | SUSE Linux Enterprise Server 12 SP4 | 12.4-0
+        cmd = run_cmd(
+          "LANG=C zypper products -i | grep SLES | tr -d ' ' | cut -d '|' -f 5"
```
skazi0: why not reuse product info extraction logic from `adminrepocheck`?

jsuchome: Hm, I'm not sure if it's easily extractable this time

jsuchome: I can parse the xml output as well. But I do not see an easy option to reuse existing method.
I guess I should do the `zypper-retry` call, that's a good one

skazi0: ok, if you don't want to do the extra checks for "system locked" or "prompts" then it all boils down to around two lines of xml processing code to be copied... imo still better than the bash magic :o)

jsuchome: Yes, I will try that.
(Although I would see bash magic as not as bad compared to xml parsing...)

skazi0: If the xml parsing was to be done in bash then I'd probably go with grep on text but in ruby with `Hash.from_xml()` at hand I think it's better to use that.

jsuchome:
@skazi0 what about this one?

crowbar_framework/app/models/api/upgrade.rb
```diff
@@ -460,12 +460,39 @@ def check_schema_migrations
...
 
+      # Check the version of installed SLES to verify admin upgrade went well.
+      # We have to do the check here at the start of 'services' step and not during admin upgrade,
+      # because admin upgrade step is not repeatable.
+      def check_product_version
+        zypper_stream = Hash.from_xml(
+          `sudo /usr/bin/zypper-retry --xmlout products -i`
+        )["stream"]
+        products = zypper_stream["product_list"]["product"]
+        return true if products.detect do |product|
```
skazi0: 1. it will die if it doesn't find anything as `detect` will return `nil` and you can't call `any?` on `nil`
2. you don't need `detect` at all. you just want to check presence of "any" item matching your predicate so something like:
```
return true if products.any? { |p| p["name"] == "SLES" && p["version"] == "12.4" }
```
should do the trick.

jsuchome: True.

PR #1986, opened 2019-12-18
apache2: Restart after enabling SSL flag (SOC-11029)

skazi0:
When SSL flag is enabled in apache's sysconfig file it is translated
into a -DSSL command argument that will be used the next time apache
is (re)started. When apache is reloaded, it is just signaled with
SIGUSR1 but not completely restarted.
Without a restart, the main process doesn't get the -DSSL option and
this causes all configs with <IfDefine SSL> to be skipped.
Additional restart in the mod_ssl recipe after enabling SSL flag is
required to handle this properly.

### DIFF
--- a/chef/cookbooks/apache2/recipes/mod_ssl.rb
+++ b/chef/cookbooks/apache2/recipes/mod_ssl.rb
@@ -32,6 +32,12 @@ end
 if node[:platform_family] == "suse"
   execute "/usr/sbin/a2enflag SSL" do
     command "/usr/sbin/a2enflag SSL"
+    # apache needs to be hard-restarted or -DSSL will not be added to the main process
+    # this would result in some config files with <IfDefine SSL> not being picked up by
+    # following reloads
+    notifies :restart, resources(service: "apache2"), :immediately
+    not_if "grep '^[[:space:]]*APACHE_SERVER_FLAGS=' /etc/sysconfig/apache2 |"\
+           "sed -r 's/[\"=]|$/ /g' | grep -q ' SSL '"
   end
   apache_module "version"
 end


chef/cookbooks/apache2/recipes/mod_ssl.rb
```diff
@@ -32,6 +32,12 @@
 if node[:platform_family] == "suse"
   execute "/usr/sbin/a2enflag SSL" do
     command "/usr/sbin/a2enflag SSL"
+    # apache needs to be hard-restarted or -DSSL will not be added to the main process
+    # this would result in some config files with <IfDefine SSL> not being picked up by
+    # following reloads
+    notifies :restart, resources(service: "apache2"), :immediately
+    not_if "grep '^[[:space:]]*APACHE_SERVER_FLAGS=' /etc/sysconfig/apache2 |"\
+           "sed -r 's/[\"=]|$/ /g' | grep -q ' SSL '"
```
jsuchome: Why `sed`? Aren't 2 greps enough?

skazi0: This is my test input:
```
APACHE_SERVER_FLAGS="XX"
 APACHE_SERVER_FLAGS="SSL"
APACHE_SERVER_FLAGS="aa SSL"
 APACHE_SERVER_FLAGS="SSL ss"
APACHE_SERVER_FLAGS="aa SSL aa"
APACHE_SERVER_FLAGS="NOSSL"
APACHE_SERVER_FLAGS=SSL

Comment (skazi0) [reply chef/cookbooks/apache2/recipes/mod_ssl.rb]: Actually a perfect solution would be to add another command for checking if flag is currently set. We have `a2enflag/a2disflag`. If there was something like `a2testflag` then we wouldn't need to worry about those seds and greps :o)

Comment (jsuchome) [reply chef/cookbooks/apache2/recipes/mod_ssl.rb]: I still do not see why any of your test input data needs that `sed` command.

(And I agree that it would be best if we would not need to grep+sed at all)

Comment (skazi0) [reply chef/cookbooks/apache2/recipes/mod_ssl.rb]: OK, it is possible to do with even one grep but the regexp would be way more complicated... or maybe I'm missing something. Do you have some example for your idea?

Comment (jsuchome) [reply chef/cookbooks/apache2/recipes/mod_ssl.rb]: I'm probably just missing something. Which of your cases requires the sed command between 2 grep calls?

Comment (skazi0) [reply chef/cookbooks/apache2/recipes/mod_ssl.rb]: with perl regexp syntax this could be simplified as:
```
grep -qP '^\s*APACHE_SERVER_FLAGS=.*\bSSL\b.*' /etc/sysconfig/apache2

jgrassler: > OK, it is possible to do with even one grep but the regexp would be way more complicated

Let's not go there...the regex is complicated enough as it is. The way it looks right now it's still reasonably readable, which is a better thing to optimize for than maximum efficiency in this case.

skazi0: The sed is there to simplify " X " matching on the ends of string.
The `sysconf_addword` used by `a2enflag` makes a clever (but not very secure) trick and sources /etc/sysconfig/apache2 file. This way we get real `APACHE_SERVER_FLAGS` variable with all quotes removed so we can test it simply like that: `echo " $APACHE_SERVER_FLAGS " | grep " SSL "` (they actually use different kind of shell magic there but I hope you know what I mean).

skazi0:
Commit message updated.
The regexp/grep/sed is up to discussion. Now that I wrote it, I think I like the "perl" way the most :o)

chef/cookbooks/apache2/recipes/mod_ssl.rb
jgrassler: By all means, leave the regex in the state it is. You tested it and it works, so no need to change it to something that may break in subtle, new ways (and that will be less readable).

PR #1941, opened 2019-10-02
Revert "Use block-migration when needed" (SOC-10133)

jgrassler:
This reverts commit 5c2fece34bdbd51c1700d883dac955ca7b6ff468.
Specifying block migration explicitly is no longer needed as
of Nova API version 2.25 (Mitaka and newer). From this API
version onward Nova will automatically select the appropriate
migration mode.

(cherry picked from commit f6d4e4d8a0b82b870005434f61ba81e67a4723c3)

This is a backport of https://github.com/crowbar/crowbar-core/pull/1936. It's already tested and works against the Cloud 7 -> Cloud 8 upgrade (see Jira issue for details).

PR #1987, opened 2019-12-20
Revert "Use block-migration when needed" (SOC-10133)

sandonovsuse:
Backport for crowbar/crowbar-core#1941


jsuchome:
Why would we need this in `stable/4.0` branch? That means 6-7 upgrade...

sandonovsuse:
> Why would we need this in `stable/4.0` branch? That means 6-7 upgrade...

Good point. Was tidying up customer issues, sorry for the noise. Closing PR

PR #1988, opened 2020-01-02
Upgrade neutron agent together with nova-compute package (SOC-11031)

jsuchome:
In some cases we start losing network connectivity to VM's when
all nova-compute services are upgraded and live-migration is being
performed. Upgrading neutron agents on the compute nodes seems to help.

Just a note: this is consistent with the code in 4.0 branch, where we also upgraded nova-compute on all nodes during the upgrade...

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
@@ -97,7 +97,7 @@ if [ $ret != 0 ] ; then
     exit $ret
 fi
 
-log "Restarting services for Rocky"
+zypper --non-interactive up <%= @neutron_agent %>
 
 <% unless @is_remote_node %>
 systemctl restart <%= @neutron_agent %>


chef/cookbooks/crowbar/templates/default/crowbar-pre-upgrade.sh.erb
```diff
@@ -97,7 +97,7 @@ if [ $ret != 0 ] ; then
     exit $ret
 fi
 
-log "Restarting services for Rocky"
```
skazi0: why did you remove this log line?

jsuchome: Because it's no longer true (no Rocky when upgrading to SOC9) ... and anyway it does not have much sense, the script actions are logged already

jsuchome: OK, I might be wrong about Rocky and SOC9. But I still think the line does not bring any value (esp. when services are not _always_ restarted, there are further checks, see below)

jsuchome:
It seems it has proved to be useful, removing wip flag.
@jgrassler, chance to veto?

skazi0:
Tested in the scale lab and it seems to solve our conn-loss problem.

PR #1994, opened 2020-01-13
Do not log an error for a case that is correct

jsuchome:


nicolasbock:
you could add a `(trivial)` to the commit message subject.

jsuchome:
@nicolasbock thanks for the tip

PR #2000, opened 2020-01-23
network: start OVS before wickedd (SOC-11067)

stefannica:
According to the official documentation [1], the OVS services need
to start long before the network (wicked) service, to be able to
interact with the service during boot. However, both OVS and wicked
are multi-service applications and the way that OVS and wicked systemd
service files are being set up, using a `Before=network.service` entry
in the OVS systemd unit file is not enough to guarantee that OVS
services are fully started before wicked services start.

The network.service systemd unit is an alias for wicked.service, which
is just a front-end unit for the multiple wickedd-* services that are
part of the wicked application. These wickedd-* services are also configured
to start `Before=wicked.service`, which means that OVS services and
wicked services usually end up being started in parallel. This is
the source of a number of heisenbugs that only manifest during
reboot and intermittently impact the gating and periodic CI jobs.

NOTE: this service startup dependency cannot be modeled as an OVS
patch, because wicked is a SUSE specific product. It also cannot be
modeled as a wicked patch because there is no direct dependency
between wicked and OVS. It needs to be modeled by SOC, because
it is in the context of SOC where both applications need to be
running side by side. I did file an upstream ticket though, just in
case this can be solved in the wicked systemd files: https://github.com/openSUSE/wicked/issues/822

[1] https://en.opensuse.org/Portal:Wicked/OpenvSwitch

### DIFF
--- a/chef/cookbooks/network/recipes/default.rb
+++ b/chef/cookbooks/network/recipes/default.rb
@@ -53,6 +53,27 @@ if node[:network][:needs_openvswitch]
     end
     s.run_action :disable
   end
+
+  directory "/etc/systemd/system/openvswitch.service.d" do
+    owner "root"
+    group "root"
+    mode 0o755
+    action :create
+  end
+
+  template "/etc/systemd/system/openvswitch.service.d/99-wickedd.conf" do
+    source "openvswitch.service-wickedd.conf.erb"
+    owner "root"
+    group "root"
+    mode 0o644
+  end
+
+  bash "reload systemd for openvswitch.service extension" do
+    code "systemctl daemon-reload"
+    action :nothing
+    subscribes :run, "template[/etc/systemd/system/openvswitch.service.d/99-wickedd.conf]", :immediate
+  end
+
 end
 
 require "fileutils"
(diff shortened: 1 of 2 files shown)


stefannica:
Testing with: https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar/detail/openstack-crowbar/741/pipeline

JanZerebecki:
> NOTE: this service startup dependency cannot be modeled as an OVS
patch, because wicked is a SUSE specific product.

This is not true because: 1) The target of a Before does not need to be started or installed, it matters only and is only an ordering requirement if both declaring unit and target unit are started in the same transaction. 2) There is a general [network-pre.target](https://www.freedesktop.org/software/systemd/man/systemd.special.html#network-pre.target). 3) This patch reads as if wicked provides a service that OVS needs instead of merely needing to be done before the network is started like a firewall. systemd is probably able to more correctly model this. 4) OVS already does this interaction with wicked, so the above is no argument to not fix the existing interaction.

JanZerebecki:
> 2) There is a general network-pre.target.

I was wrong here. This can't be used as OVS already needs to be After this, to prevent it from creating interfaces before it.

PR #2003, opened 2020-02-03
upgrade: Run DHCP evacuation (SOC-11046)

skazi0:
DHCP agent handles networks and ports in a way similar to L3 agent's
routers. The assignment of these to the network nodes need to be
properly orchestrated to keep things under control in upgrade scenario
where we need to move network agents out from node before upgrading
to make the process non-disruptive.

pre-req: https://github.com/SUSE-Cloud/cookbook-openstack-network/pull/38

### DIFF
--- a/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
+++ b/chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
@@ -205,6 +205,14 @@ template "/usr/sbin/crowbar-router-migration.sh" do
   action :create
 end
 
+template "/usr/sbin/crowbar-dhcp-migration.sh" do
+  source "crowbar-dhcp-migration.sh.erb"
+  mode "0755"
+  owner "root"
+  group "root"
+  action :create
+end
+
 template "/usr/sbin/crowbar-set-network-agents-state.sh" do
   source "crowbar-set-network-agents-state.sh.erb"
   mode "0755"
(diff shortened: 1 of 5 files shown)


chef/cookbooks/crowbar/recipes/prepare-upgrade-scripts.rb
```diff
@@ -205,6 +205,14 @@
   action :create
 end
 
+template "/usr/sbin/crowbar-dhcp-migration.sh" do
```
jsuchome: It might be better to just put this file on the controller nodes and not everywhere (but I know reouter scripts are also everywhere...)

jsuchome:
@skazi0 please update also `delete_upgrade_scripts` method https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/api/upgrade.rb#L1193

skazi0:
> @skazi0 please update also `delete_upgrade_scripts` method [/crowbar_framework/app/models/api/upgrade.rb@`master`#L1193](https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/api/upgrade.rb#L1193) [](/crowbar/crowbar-core/blob/HEAD@{2020-02-03T10:45:40Z}/crowbar_framework/app/models/api/upgrade.rb#L1193)

I was looking for this place... but I was grepping for `crowbar-router-migration...` :o)

PR #1799, opened 2019-02-19
ses: Configuration upload & SES cookbook

skazi0:
Added Crowbar UI part for uploading and viewing SES settings based on
yaml file exported from SES cluster.
Settings are saved in a data bag for future use.

This change is a base for: https://github.com/crowbar/crowbar-openstack/pull/2035
and alternative to: https://github.com/crowbar/crowbar-openstack/pull/1977

TODO:
- [ ] validation of uploaded yaml (beyond yaml syntax check). Currently everything is stored in the data bag "as is".
- [ ] CLI interface to upload yaml configs (good for CI)
- [ ] add some service restart logic after the ceph config/keyring files are modified (?)

Some screenshots:
Menu entry:
zrzut ekranu z 2019-02-19 22-46-19

Empty settings view:
zrzut ekranu z 2019-02-19 22-44-28

Upload modal (similar to backup upload):
zrzut ekranu z 2019-02-19 22-44-11

Part of uploaded settings:
zrzut ekranu z 2019-02-19 22-43-50

PR #2006, opened 2020-02-20
[5.0] ses: Configuration upload & SES cookbook + other fixes

skazi0:
port of #1799 and #2009 and #2010


jsuchome:
Could you add some validation when importing SES configuration so the UI does not break when I try to add a faulty one?

skazi0:
instead of validation (which would not be trivial) I added this: https://github.com/crowbar/crowbar-core/pull/2010

PR #2010, opened 2020-03-03
ses: Make SES UI safe for unknown options (trivial)

skazi0:
When uploaded yaml file is invalid or contains unrecognized parts the UI
would crash and block user from re-uploading a correct file.
With this change the unrecognized parts will not get translated but UI
would be fully operational.


skazi0:
before:
Zrzut ekranu z 2020-03-03 09-17-05

after:
Zrzut ekranu z 2020-03-03 09-15-53

PR #2012, opened 2020-03-05
network: check for duplicate ip addr (SOC-11102)

sandonovsuse:
Crowbar doesn't check for duplicate IP addresses in the network prior to
adding the address to the interface. This patch adds this functionality,
with chef run failing if duplicate IP address detected.


sandonovsuse:
Closing due to travis failure

PR #2018, opened 2020-03-20
upgrade: Fixes for neutron-evacuate-lbaasv2-agent execution (SOC-11171)

jsuchome:
use_crm option is there for the cases when lbaas agents are handled
by pacemaker resources. So the script puts the node into the maintenance
mode for the time it is restarting the agent.

However since SOC8 we do not use pacameker for lbaas agents, so there
is no need for maintenance mode and no need to execute the script with
use_crm option.


jsuchome:
It actually does not look important enough for the backport....

PR #2019, opened 2020-03-20
Ignore latest CVEs in the CI, update default OS version used by CI

jsuchome:


jsuchome:
Thanks to @skazi0 for the help with the 3rd commit!

PR #2021, opened 2020-03-24
[5.0] upgrade: Remove the assignement of crowbar-upgrade role (SOC-11166)

jsuchome:
At the end of upgrade, remove elements of crowbar-upgrade from
crowbar barclamp. Otherwise if user makes changes to crowbar proposal
(or even just reapplies it for some reason), nodes would be moved
into crowbar-upgrade state again.

(cherry picked from commit c164fc1c1dfe3493d755f2597bac1a8939935c18)

port of https://github.com/crowbar/crowbar-core/pull/2016


jsuchome:
Requires https://github.com/crowbar/crowbar-core/pull/2020

PR #1998, opened 2020-01-15
Add lb-mgmt-net to network.json (SOC-10904)

nicolasbock:
This change adds a network to `network.json` for configuring a
provider network for Octavia management network traffic.

Signed-off-by: Nicolas Bock <nicolas.bock@suse.com>

PR #2023, opened 2020-03-31
Revert: Add lb-mgmt-net to network.json (SOC-10904) 

stefannica:
The octavia network.json network is not mandatory (customers
may choose to deploy their own provider network). If we add
it to network.json, the network will always be detected and
used by Octavia.

Instead, we need to document this as an optional network
that customers may choose to add to their deployment.

Side-note: this is also a way of resolving the circular dependency
between Octavia (the consumer of this octavia network) and Neutron
(the provider that needs to configure the octavia network only if the
Octavia barclamp is deployed).

This reverts PR #1998

### DIFF
--- a/chef/data_bags/crowbar/template-network.json
+++ b/chef/data_bags/crowbar/template-network.json
@@ -179,7 +179,7 @@
         },
         "nova_fixed": {
           "conduit": "intf1",
-          "vlan": 700,
+          "vlan": 500,
           "use_vlan": true,
           "add_bridge": false,
           "add_ovs_bridge": false,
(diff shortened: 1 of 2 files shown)


guangyee:
I thought the management network is already an internal network, which is completely segmented off from public network. For CLM, both OCTAVIA-MGMT-NET and management (controller) network are internal.

stefannica:
> I thought the management network is already an internal network, which is completely segmented off from public network. For CLM, both OCTAVIA-MGMT-NET and management (controller) network are internal.

For CLM, OCTAVIA-MGMT-NET is indeed internal, in the sense that it is isolated from the public network. However, it is not internal in the sense that it needs routing (customer provided routing) to be able to communicate with the management network. 

With this approach, the OCTAVIA-MGMT-NET (the Crowbar version of it) is completely isolated, in that it doesn't even need to reach the management network. All Octavia components communicate directly over this network. Furthermore, it requires no customer provided routing.

stefannica:
> > I thought the management network is already an internal network, which is completely segmented off from public network. For CLM, both OCTAVIA-MGMT-NET and management (controller) network are internal.
> 
> For CLM, OCTAVIA-MGMT-NET is indeed internal, in the sense that it is isolated from the public network. However, it is not internal in the sense that it needs routing (customer provided routing) to be able to communicate with the management network.
> 
> With this approach, the OCTAVIA-MGMT-NET (the Crowbar version of it) is completely isolated, in that it doesn't even need to reach the management network. All Octavia components communicate directly over this network. Furthermore, it requires no customer provided routing.

On the second thought, we might be able to achieve the same with an internal network without a gateway, which will bring Crowbar closer to what CLM is using. I'll run some tests to confirm.

guangyee:
We need to revert the doc changes as well. https://github.com/SUSE-Cloud/doc-cloud/pull/1268

PR #2028, opened 2020-04-23
upgrade: disable zypper process check temporarily (SOC-11203)

jgrassler:
This commit temporarily sets zypper's psCheckAccessDeleted
(introduced by bsc#945169) setting to 'no' before package
upgrade on nodes. This check launches an lsof command that may
hang indefinitely to check for running processes using files
that were changed by a package upgrade. In our case, this
check is entirely unneccessary since the nodes will be
rebooted after package upgrade anyway. Hence we can safely
turn it off, thus saving the operator the trouble of killing
these lsof processes manually on each node that experiences
the problem.

### DIFF
--- a/chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.sh.erb
+++ b/chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.sh.erb
@@ -76,12 +76,19 @@ initiate_node_upgrade()
       sed -i s/django.utils.log.NullHandler/logging.NullHandler/g $horizon_custom_config
     fi
 
+    # SOC-11203: record old setting and disable lsof check for running
+    # processes since it may hang indefinitely
+    zypper --non-interactive install crudini
+    zypper_upgrade_conf=/tmp/zypper.conf.upgrade
+    cp /etc/zypp/zypper.conf $zypper_upgrade_conf
+    crudini --set $zypper_upgrade_conf commit psCheckAccessDeleted no
+
     # Upgrade the distribution non-interactively
     log "Executing zypper ref"
-    zypper --no-color --releasever <%= @target_platform_version %> ref -f
+    zypper --config $zypper_upgrade_conf --no-color --releasever <%= @target_platform_version %> ref -f
 
     log "Executing zypper dist-upgrade"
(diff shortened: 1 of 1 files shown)


jsuchome:
I don't know ... it's a zypper bug after all and it's more likely it's gonna be fixed before we can release this patch as MU...

jgrassler:
Added one more small change (tested, too): the {else} branch for deleting the `psCheckAccessDeleted` in case it was entirely unset before.

Regarding the Zypper bug: it's already closed. They fixed it by introducing this setting to deal with environments where the problem occurs (it doesn't occur everywhere)...

chef/cookbooks/crowbar/templates/default/crowbar-upgrade-os.sh.erb
```diff
@@ -76,13 +76,28 @@ initiate_node_upgrade()
       sed -i s/django.utils.log.NullHandler/logging.NullHandler/g $horizon_custom_config
     fi
 
+    # SOC-11203: record old setting and disable lsof check for running
+    # processes since it may hang indefinitely
+
+    zypper --non-interactive install crudini
+    check_access_old=$(crudini --get /etc/zypp/zypper.conf commit psCheckAccessDeleted || true)
```
jsuchome, line 83: What's the reason for `|| true` ? We're not checking the return value, and if the key is not there, it won't change anything...

jgrassler: I thought we had a `set -e` in there...but I just checked and we don't. So theoretically it's redundant. Actually I'm currently pondering another overhaul in the light of Jacek's "hopefully it won't die before we restore the old version" comment: a backup file instead would preserve the state we might lose and render that and a lot more crudini mongering redundant. So if you'd like to see it gone I'd remove it in the course of that.

jgrassler:
Ok, this is the second, improved iteration with temporary `zypper.conf`. Testing currently in progress (`crowbarctl upgrade nodes all` running). Will update on how that went once it's done...

jgrassler:
Upgrade went off without a hitch - ready to merge as far as I'm concerned.

PR #2029, opened 2020-04-30
network: check for duplicate ip addr (SOC-11102)

sandonovsuse:
Crowbar doesn't check for duplicate IP addresses in the network prior to
adding the address to the interface. This patch adds this functionality,
with chef run failing if duplicate IP address detected.

### DIFF
--- a/chef/cookbooks/barclamp/libraries/nic.rb
+++ b/chef/cookbooks/barclamp/libraries/nic.rb
@@ -43,6 +43,11 @@ class ::Nic
     ::Kernel.system("ethtool", *args)
   end
 
+  # Helper for running arping commands.
+  def run_arping(arg)
+    ::Kernel.system("arping #{arg}")
+  end
+
   # Return an unsorted array of all visible nics on the system.
   # This will skip virtual nics that OVS creates.
   def self.__nics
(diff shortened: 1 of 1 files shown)


jsuchome:
Why is it a problem?

sandonovsuse:
> Why is it a problem?

A use  case where IP addresses allocated in crowbar during deployment for any of the network ranges are taken at some later point in time by some other devices external to cloud, i.e. external ceph deployment residing in network range allocated to crowbar storage network. Now deploying a new node might end up in setting a duplicate IP address in storage subnet due to chef not checking for duplicate address prior to setting it. This commit only checks for duplicate and if one is detected it fails the chef run until the conflict is resolved by admin.

skazi0:
Is this even a supported scenario? Shouldn't external nodes use range which doesn't overlap the one in Crowbar?

sandonovsuse:
Yes, I don't think it is a supported scenario, although it may happen years after deployment when it's too late to change crowbar ranges (bsc#1163563). This patch would fail the chef-run only if such a scenario took place (even though it is unsupported), otherwise it should act the same way on a supported deployment scenario

sandonovsuse:
Rebased

JanZerebecki:
The ticket was closed as "Won't do". Please reopen if you think this is useful.

PR #2030, opened 2020-05-27
upgrade: Do not stop pacemaker managed apache service (SOC-11298)

jsuchome:
apache service should not be stopped by usual upgrade scripts,
because we need it so the keystone API can function as long as possible.

Recently we found a corner case where apache is stopped because
it is handled by Pacemaker (setting that is non-default for a long time)
and most of pacemaker resources are stopped and deleted at certain point.

Solution is to exclude apache resource from this deletion command.


skazi0:
.... and a bunch of new CVEs :o(

PR #2036, opened 2020-07-15
crowbar: Also add access to /restricted/ in SSL vhost (SOC-11352)

flaviodsr:
The `/restricted/` path is missing on the SSL vhost configuration.

(cherry picked from commit 5f7b8c4e685a4e3de497714af6d58aca7dc30fba)


skazi0:
@jgrassler remember to add it to soc7 version as well

PR #2042, opened 2020-09-14
provisioner: check for client_user (SOC-11389)

jgrassler:
This commit adds a check for the presence of the client_user
attribute in the Crowbar barclamp's data bag. This attribute
will only become available in chef once chef-client has run on
the Crowbar admin node or the Crowbar data bag has been
applied. Until that has happened, the code using it would
error out on all other nodes.

Note that this is _not_ a backport from Cloud 9 since the code for writing crowbarrc has changed considerable between Cloud 8 and Cloud 9. I'll create a separate pull request for Cloud 9 once I've come up with and tested a patch that fits the Cloud 9 code base.


skazi0 (approved):
lgtm

jsuchome:
Could you add a link to the change which added _client_user_ attribute?

jsuchome (dismissed):
It looks correct

gosipyan (approved):

gosipyan (approved):

skazi0 (approved):

PR #2043, opened 2020-09-14
provisioner: check for client_user (SOC-11389)

jgrassler:
This commit adds a check for the presence of the client_user
attribute in the Crowbar barclamp's data bag. This attribute
will only become available in chef once chef-client has run on
the Crowbar admin node or the Crowbar data bag has been
applied. Until that has happened, the code using it would
error out on all other nodes.

Note that this pull request is a Cloud 9 specific solution to the same problem and separate from https://github.com/crowbar/crowbar-core/pull/2042 (i.e. https://github.com/crowbar/crowbar-core/pull/2042 is not a backport). Setting the _do not merge yet_ label for now - I still need to test this against a Cloud 9 deployment.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/base.rb
+++ b/chef/cookbooks/provisioner/recipes/base.rb
@@ -357,8 +357,17 @@ package "ruby2.1-rubygem-crowbar-client"
 
 unless is_admin
   # On non-admin nodes, setup /etc/crowbarrc with the restricted client
-  username = crowbar_node["crowbar"]["client_user"]["username"]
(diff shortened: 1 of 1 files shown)


skazi0:
@jgrassler I don't think we need this in SOC9. The restricted API was released for SOC9 from the beginning so there should be no case where we have update which adds this feature. For SOC8-9 upgrade we orchestrate chef-client runs explicitly so we should not hit this problem either.

jgrassler:
Hmm, indeed. It was merged on April 8th 2019. I thought we'd only added it a bit after release. Looks like I misremembered then.

